PR #27, opened 2013-10-17
Sets Default to Maximum Number of HT Cores - 1

4616:
Current concurrency default of 3 cores fails to make use of additional processing power available on many machines. My edit sets the initial default to the total number of cores available, minus one to keep the user's computer running smoothly.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -21,11 +21,16 @@ def setup_config_file(config_file):
   ''' Config file setup, adds defaults if not existing '''
   config = SafeConfigParser()
   config.read(config_file)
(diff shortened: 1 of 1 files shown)


Stefano80:
Should we close it? @glinscott expressed its opinion against it and I can follow its argument. If nobody argues for it, I will close it soon.

ppigazzini:
I trust the use of HT cores (useful in some scenarios), but I'm against the use of HT cores by default, so OK to me to close this PR.

Stefano80:
So, closing.

PR #32, opened 2014-01-11
Cpu survey

jromang:
Surveys CPU load while running fishtest and stops it for 1 minute if it gets too high.


alg4032:
"Please don't use your computer while running fishtest." should better be extended to ".. or use a lower concurrency value". I think most fishtest users don't use their computers exclusively for fishtest.

Stefano80:
Is this still up-to-data, otherwise I will close it.

Stefano80:
Looks inactive, closing.

PR #43, opened 2014-07-05
Fix mako templates on Pyramid 1.5

hwiechers:
As of Pyramid 1.5, Mako templates are no longer configured by default.
See http://docs.pylonsproject.org/projects/pyramid/en/1.5-branch/whatsnew-1.5.html

This probably breaks things on Pyramid <1.5 so you should keep it around for when you update.

### DIFF
--- a/fishtest/fishtest/__init__.py
+++ b/fishtest/fishtest/__init__.py
@@ -15,6 +15,8 @@ def main(global_config, **settings):
                         session_factory=session_factory,
                         root_factory='fishtest.models.RootFactory')
 
+  config.include('pyramid_mako')
+
   # Authentication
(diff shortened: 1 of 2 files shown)


Stefano80:
Any opinion about this?

ppigazzini:
@Stefano80 please keep this PR opened. It should be committed before making a fresh installation of the official server. In the [server installation script](https://github.com/glinscott/fishtest/wiki/Installing-a-development-fishtest-server-on-Ubuntu) ATM I'm using a diff patch.

tomtor:
Fixed in

https://github.com/glinscott/fishtest/pull/153

PR #49, opened 2015-01-05
Initial commit: regression tests page

kamyarkaviani:
Added graphs for fishtest and Jens Lehmann regression testing data

TODO: organize data properly, standing table, table of possible regressions


glinscott:
Wow, thanks!  Looks great, I will try it now :).

PR #52, opened 2015-03-02
Detect if threads have been specified manually (for scalability testing)

cuddlestmonkey:
and do not set global value in -each clause to cute chess if so. Keep the global value to correctly compute concurrency.

This is suggested to allow Lucas to perform Elo scalability testing in the framework.

For normal use, nothing changes (no need to change howtos etc)

In order to test a version against itself with different threads, set the normal threads parameter to the *higher* value and then add the usual Threads=N option to the custom UCI options for each.

See Lucas' post here:

https://groups.google.com/d/msg/fishcooking/_aLnMjMH4ec/jbKXgJrpcq0J

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -421,6 +421,10 @@ def run_games(worker_info, password, remote, run, task_id):
     games_to_play = games_remaining
     pgnout = ['-pgnout', 'results.pgn']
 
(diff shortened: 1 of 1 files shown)


glinscott:
Looks good!  Thanks, merged.

PR #53, opened 2015-03-03
Update worker version (API) 

cuddlestmonkey:
as otherwise new worker will not be rolled out to existing clients and new thread option handling won't work.
Thanks to Alfred for the spot.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -142,7 +142,7 @@ def request_version(request):
   token = authenticate(request)
   if 'error' in token: return json.dumps(token)
 
-  return json.dumps({'version': 55})
+  return json.dumps({'version': 56})
 
 @view_config(route_name='api_request_spsa', renderer='string')
 def request_spsa(request):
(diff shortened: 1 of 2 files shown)


joergoster:
Your new threads-option cannot work if put into the -each section of the cutechess command line options ...

cuddlestmonkey:
But the option put into the -each section is only filled in if the engine-specific options do *not* contain "Threads". Please look at how the original patch works. If there is a problem with the indentation (I know this is critical in Python) please point it out.

cuddlestmonkey:
Here is the relevant bit:

threads_cmd=[]
if not any("Threads" in s for s in new_options + base_options):
  threads_cmd = ['option.Threads=%d' % (threads)]

threads_cmd is initialised to an empty list. It is then set to "option.Threads=whatever" if the manually specified options (new options and base options) *don't* contain the word "Threads".

Otherwise threads_cmd stays empty and its inclusion in -each is a NOOP.

joergoster:
Thx for the explanation. I am far from being an expert in python.
I just realized that if "Threads=x" is specified, it is sent via 'new_options' and 'base_options' to the engines. Sorry for the noise!

cuddlestmonkey:
No problem, I'm no expert either! I had to cheat and look up the 'in s for s in' trick on stack overflow :)

PR #54, opened 2015-03-24
Add link to official stockfish to menu

hxim:
I miss direct link to official-stockfish/Stockfish github in main menu.


glinscott:
Thanks!

PR #56, opened 2015-04-03
Throughput and internal priority. Better balancing between tests.

zamar:
All tests are run in parallel and ordered by constantly changing internal priority rather than being processed one by one.

This has several advantages:
- No long running test should get stuck at the end of the queue. Instead these are progressing slowly.
- After a test has been approved in fishtest, the execution starts immediately, so a developer can immediately get feedback if the test is promising or not.

python is not my strongest language, so please review carefully :-)

I'm still running some internal tests in my local fishtest instance, but I'm submitting this for early review to get feedback...

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -1,6 +1,7 @@
 import copy
 import os
 import random
+import time
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


glinscott:
Looks great!

PR #57, opened 2015-04-04
Detect the new nodestime option and use timemargin to reduce noise

cuddlestmonkey:
The new 'nodes time' option should allow virtual machines and more loaded hosts in general to take part in the framework. But local testing revealed that time losses are still possible if there is a local spike in CPU load.

So update the worker such that if 'nodes time' is being used, the 'timemargin' option is passed to cutechess. This allows the engines (which are not really using the clock at all) to overstep the time limit by up to 10 seconds (a huge amount, and one that should stop any time losses at all)

Since the SF instances are still playing without the knowledge they have this reserve, it won't distort the results.

The next refinement I guess would be for workers to be able to indicate that they are only good for 'nodestime' jobs, but that is rather more complicated.

In the meantime this should eliminate time losses and thus noise from SPSA runs.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -425,6 +425,12 @@ def run_games(worker_info, password, remote, run, task_id):
   if not any("Threads" in s for s in new_options + base_options):
     threads_cmd = ['option.Threads=%d' % (threads)]
 
+  # If nodestime is being used, give engines extra grace time to 
+  # make time losses virtually impossible
+  nodestime_cmd=[]
+  if any ("nodestime" in s for s in new_options + base_options):
+    nodestime_cmd = ['timemargin=10000']
+
   while games_remaining > 0:
     # Run cutechess-cli binary
     cmd = [ cutechess, '-repeat', '-rounds', str(games_to_play), '-tournament', 'gauntlet'] + pgnout + \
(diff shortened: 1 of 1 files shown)


mcostalba:
Nice :-)

I didn't know of this cutechess option....

cuddlestmonkey:
Yes.. at first I forked cutechess expecting to add a similar option but found it already there, so just a worker patch is required. Of course, if this is pulled then #53 needs pulling too, so that the worker nodes update themselves to the latest games.py etc.

glinscott:
Great, this looks nice.  Originally the intent was that worker updates would be seamless.  However, since a lot of people are using the precompiled windows version, whenever the worker is updated, it breaks all of them until they download a new version.  Not an optimal situation for sure.  That's why I'm hesitant to update the worker version unless absolutely necessary.

cuddlestmonkey:
Ah, the precompiled version is actually compiled, rather than scripts with a local python copy, correct? So the code in worker.py and updater.py can't work?

```python
    req = requests.post(remote + '/api/request_version', data=json.dumps(payload), headers={'Content-type': 'application/json'})
    req = json.loads(req.text)

    if 'version' not in req:
      print 'Incorrect username/password'
      time.sleep(5)
      sys.exit(1)

      print 'Updating worker version to %d' % (req['version'])
      update()
```

glinscott:
@cuddlestmonkey, it's not precompiled, but it is packed into one exe, while the worker expects to be able to overwrite it's own files.  The end result is the same, if I update a worker verison and enforce it, most of the windows machines will go offline for a while.

cuddlestmonkey:
Ah, that's a shame. Unfortunately the only parameters you can inject into cutechess at the moment are the UCI ones for the engines, so there isn't a trick for adding timemargin manually - again, unless the worker is updated.

Which makes me think - is it worthwhile adding support for custom cutechess (as opposed to UCI) options into the worker? At least then if you went through the pain of pushing a worker update, it would open up a whole bag of possible tweaks that could be driven from the server in future.

cuddlestmonkey:
Related point - if 'nodestime' is detected, then ideally should we not scale the incoming TC at all? That way all workers, regardless of clock speed (as long as their real NPS is not less than the nodestime value) will use the same number of total nodes per game and you get a more even contribution.

cuddlestmonkey:
Cheers Gary, let the tuning be unconfined!

PR #58, opened 2015-04-18
PGO

lucasart:
We've (re)learnt recently that the compiler can really optimize differently
with PGO, making use of profiling information collected at runtime. Whether a
patch is a clear speedup or a clear speed regression can be radically different
with and without PGO. Often fishtest is being used to verify non functional
changes that are supposed to speedup the engine (or simplifications that are
supposed not to slow it down). To be able to rely on fishtest for that, it must
use PGO.


glinscott:
Thanks, yes, this makes sense.

PR #59, opened 2015-04-18
Scale increment

lucasart:
Without that we cannot test time management patches correctly, because each
worker uses a different increment/base ratio.

This problem is also known to cause chi-square to fail for tcs where the
increment is not small respective to the base, like 5+0.05 for example
(for the same reason).

Now with C++11 std::chrono, SF uses a sub millisecond precision on every
platform. The only constraint is that the Timer thread only checks for
time every 5 msec, but that's still 10x less than our 50msec increment,
so no risk to go below that.


lucasart:
@glinscott: can you have a look?

glinscott:
Sorry, missed this one in my email!  Looks good.

PR #62, opened 2015-08-09
Change default SPRT bounds

lucasart:
- Reduce resolution from 6 to 5 bayes elo. Detect >= 2 elo patches more
reliably, and reject regressive to almost useless patches more strictly
(less lucky runs).
- No average cost increase. even a slight decrease in the most common cases.

Supporting evidence provided + approved by Joona:
https://groups.google.com/forum/?fromgroups=#!topic/fishcooking/Vai7kJKlwaM

Also remove OwnBook while at it. SF does not support book, which must be handled
by the GUI.


glinscott:
Thanks for doing the analysis!

PR #67, opened 2015-10-12
Test with increment by default

lucasart:
Use a standard `100:1` ratio, proportional to `5'+3"`, more or less in line with typical rating lists.
Joona is ok with it, so it's up to you:
https://groups.google.com/forum/?fromgroups=#!topic/fishcooking/e4aApDQKlCY

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -113,7 +113,7 @@ Cowardice,150,0,200,10,0.0020"""})['raw_params']}</textarea>
   <div class="control-group">
     <label class="control-label">Time Control:</label>
     <div class="controls">
-      <input name="tc" value="${args.get('tc', '15+0.05')}">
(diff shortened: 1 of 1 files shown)


jhellis3:
This is a good change.  The minimum TC still won't be enough to allow a fixed move overhead solution to the time loss issue however. From the TCEC logs, we would need to set move overhead to at least 100ms to be on the safe side (and probably more is better at that TC).  So at 100 move overhead vs 100 increment we are already in trouble for testing, but some machines may even scale that down.   We will also need to include a notice with Stockfish that it does not function properly with time control increments less than or equal to the move overhead setting, and that end users will need to adjust this accordingly.

glinscott:
Agreed, merged!

Stefano80:
Why nothing changed on the test submission page?

jhellis3:
Probably requires server restart?

glinscott:
Yes, should be there now.

PR #69, opened 2015-10-17
Increase the time waiting for a new game result.

joergoster:
This should increase the maximum allowed game duration
to 30 minutes, and allows tests with longer time controls.
Also bump up worker version.

This change worked recently for CoffeeOne.
I propose to run a test of about 1,000 games at 5min+3sec.

I chose a limit of 30 minutes as I think this is a good compromise
between the maximum game duration, and the time an unattended PC
must wait until the hanging process is being killed.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -14,7 +14,7 @@ from optparse import OptionParser
 from games import run_games
 from updater import update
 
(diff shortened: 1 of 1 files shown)


glinscott:
Thanks!!

PR #70, opened 2015-10-21
Feed a random seed to cutechess.

wfenchel:
Cutechess presently uses the current time, in seconds, to seed its RNG.
This makes it very easy for various machines in fishtest to choose the
same random seed on the same workload, particularly when a large number
of machines join a machine pool for a specific test.

The fix here is to pass a random integer to cutechess via the "-srand"
option.  A platform-independent seed is chosen using os.urandom(),
which raises an exception if a randomness source is unavailable.

Cutechess only uses 32 bits of random state (a "quint32"), and the
parser for -srand requires an unsigned integer (although the
Mersenne::initialize routine just takes an int), so the routine here
generates an unsigned integer.

Issue originally discovered by VoyagerOne at
https://groups.google.com/forum/#!topic/fishcooking/CW7Vlwur8CQ

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -13,6 +13,7 @@ import threading
 import time
 import traceback
 import platform
+import struct
 import zipfile
 from base64 import b64decode
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


glinscott:
Thanks!

PR #72, opened 2015-11-07
Rescale throughput with threads

lucasart:
The purpose of throughput was to increase parallelism and ensure equal distribution of resources between tests. So there is no reason to allocate 7x more resources to a 7 thread test, and throughput should be rescaled accordingly.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -253,7 +253,7 @@ class RunDb:
     # Formula: - second_since_epoch - played_and_allocated_tasks * 3600 * 1000 / games_throughput
     # With default value 'throughput = 1000', this means that the priority is unchanged as long as we play at rate '1000 games / hour'.
     if (run['args']['throughput'] != None and run['args']['throughput'] != 0):
-      run['args']['internal_priority'] = - time.mktime(run['start_time'].timetuple()) - task_id * 3600 * 1000 / run['args']['throughput']
+      run['args']['internal_priority'] = - time.mktime(run['start_time'].timetuple()) - task_id * 3600 * 1000 * run['args']['threads'] / run['args']['throughput']
     self.runs.save(run)
 
     return {'run': run, 'task_id': task_id}


wfenchel:
On the other hand, the present behavior ensures that many-core machines give higher priority to many-thread tests.

This preference may reduce overall load (defined as the sum of all waiting times) on fishtest because the single-core machines can not process many-thread tests.  For example, it happened a few days back that fishtest had no 1-thread tests, only 7-thread; if the 7-thread tests were interrupted to run a 1-thread test, it would have increased the total wait across all jobs.

Of course, there are many different ways to define fairness, and your point stands that the current setup is unfair from one perspective.  But this is a delicate issue since, going back to the preceding example, machines with more than 4 cores are still a minority.

lucasart:
@glinscott @zamar: please have a look.

zamar:
We want all tests to finish in reasonable time... After this change some 23 thread tests would practically never finish.

Also throughput is easily understood as "games per hour". After this change, it would be much less clear.

So my first thought is that this change is probably not a good idea.

lucasart:
@wfenchel: what you want is that 7+ machines only run 7+ threads tests (where available). That would correctly be achieved by raising priority of such tests, rather than a dysfunctional throughput.

@zamar : "games per hour" is not a constant metric: without a number of threads and a time control, it does not really mean anything.

So, for equal priority (based on throughput behaviour):
* you think that a 23 thread test should use 23 times more resources than a 1 thread test ?
* you think that a 23 thread test at 60+0.4 should use 121x more resources than a 1 thread test ?

I thought the idea of throughput was to increase parallelism and ensure equal allocation of resources between tests...

wfenchel:
@lucasart, I have to say that now that a few days of clogged resources have gone by, I'm seeing that the situation is indeed not ideal.

I still don't think scaling by number threads is ideal, as it suggests a 7thread test will take 7x as long on average.  (Or did I misunderstand how throughput weights things?)

Unfortunately I don't currently have a proposal I feel is adequate.

lucasart:
To ensure that all tests are given equal testing bandwidth, we must rescale throughput by threads **and** tc.

the argument that LTC should have more bandwidth than STC doesn't hold, since people abusively use direct LTC, speculative LTC, LTC for SPSA tuning, etc.

how about calculate a correct scaling factor based on tc and threads, and take the **square root** of it ? Example 23 thread test in 60+0.4 is 121x costlier than 1 thread test in 10+0.1

lucasart:
… so we.rescale by `sqrt(121)=11`. will still be using 11x more bandwidth than a standard 1 thread STC test. I think that is enough.

wfenchel:
@lucasart,

Indeed, I was considering similar formulas.  I think it has potential, though the specific choice of formula is tricky.

Based on @zamar's suggestion of changing the TC, I put a proposal in fishcooking about codifying TC and th= settings for SMP patches.  Part of this is based on an anecdotal experience over the past two days; I was trying some tweaks to SMP formulas and had trouble going from 3 to 7 threads.

Perhaps a combination of your patch here and @zamar's test submission policies can resolve this issue (the two approaches seem complementary).

lucasart:
What I propose is completely orthogonal to choice of tc and threads. It's about fair allocation of testing bandwidth. Joona's proposal to reduce tc for SMP tests is also a good idea, but it's completely independent.

wfenchel:
Sorry I wasn't clear.  I agree they are fairly orthogonal.  I only meant to say that perhaps both are necessary.

lucasart:
@wfenchel: agree.

Stefano80:
I think lucas idea is good. For instance if you think in terms of ELO gained pro computing hour, there is really no reason for allocating more resources for SMP tests. It would even suggest to allocate less than proportional resoureces to those tests.

glinscott:
The core idea of lowering the resource usage of 23 core tests makes sense.  I'll defer to @zamar on this one for the final call though.

mstembera:
What is clearly wrong is that high core machines get allocated to single threaded tests while multi threaded tests are waiting in the queue as is happening right now.    Later many low core machines go idle when only multi threaded tests remain.

zamar:
I think it should be fine. Let's give it a try.

zamar:
@glinscott: Please commit when you have time.

lucasart:
I put back SMP tests at throughput 1000 now.

Eventually we will need to rescale with tc as well, if we want to create the correct incentive and make people use 5+0.05/20+0.2 for SMP tests

zamar:
OK. Let's see how it goes. We might get complaints that all SMP tests are
stuck at the end of the queue.

PR #75, opened 2015-12-10
Change LTC to 60+0.6

zamar:
100+1 is just taking too long to be practical.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -195,7 +195,7 @@ $(function() {
   $('#slow_test').click(function() {
     $('input[name=sprt_elo0]').val('0');
     $('input[name=sprt_elo1]').val('5');
-    $('input[name=tc]').val('100+1');
+    $('input[name=tc]').val('60+0.6');
   });
 });
 </script>


mstembera:
I completely agree.  We already spend most of our time waiting for results.  Thank you.

lucasart:
That's much better: 1.5x longer than the old standard 60+0.05, and 6x longer than STC. So it's still pretty long, but not unreasonably so.

VoyagerOne:
should I stop hpf2...or let it continue?

mcostalba:
In case hpf2 passes at 60+0.6 it is a confirmation that TC is not enough
(because it was going to fail at 100+1).

We can also act as we don't care or assume it does not exists, but actually
from 60 to 100 it makes a difference and we will see it soon.

IMO this patch does not add anything above original LTC apart from games
that last longer.

So it would be even better to return to the original 60+0.05, but if we
want to do some serious action for TC scaling than I am sorry for people
using fishtest as their quick slot machine, but we have to do something
better.

Of course we can also decide that we just want to have fun with fishtest
and not caring about longer TC, but this patch is totally useless, much
better to revert to original TC: or you do something or you don't.

This is a political patch, not a technical one.

zamar:
Marco: In case hpf2 passes at 60+0.6 it is a confirmation that TC is not
enough (because it was going to fail at 100+1).

No, it's not. It's just a nature of stochastical tests.

Marco: We can also act as we don't care or assume it does not exists, but
actually from 60 to 100 it makes a difference and we will see it soon.

There is no way to reliably observe the impact.

Marco: IMO this patch does not add anything above original LTC apart from
games that last longer.

It's a compromise. Where did you get the idea that only base time matters
and increment does not? Lucas proved long time ago that using increment
results in higher quality games. So using 60+0.6 over 60+0.05 is a major
boost. The former plays at least 100 elo stronger chess than the latter.

Marco: This is a political patch, not a technical one.

It's a practical patch. The framework was becoming overloaded with 100+1
tests running forever and hogging up all the resources.

If we reach the point where we consistently have ~350 cpu cores running on
the framework, then we can switch back

zamar:
In the future, I think that the correct course of action is to raise the
time control bit by bit - if we can afford it.

joergoster:
@zamar @mcostalba  This seems to be an acceptable compromise, for now.

lucasart:
@mcostalba: Regarding your hpf2 "argument", there is no sign that it baheves any differently between 60+0.6 and 100+1
http://tests.stockfishchess.org/tests/view/566a2c950ebc592d552a3faf
http://tests.stockfishchess.org/tests/view/5667294a0ebc592d552a3ecc
It even failed harder at 60+0.6, but that does not prove anything of course. The difference between the two is simply noise.

mcostalba:
@lucasart if passing at one TC and failing at another does not prove
anything I wonder what could prove that a patch does not scale?

lucasart:
@mcostalba: please read before answering. patch failed 60+0.6, even harder than it would have failed 100+1. so what is your point ?

VoyagerOne:
Just my 2 cents:

One of the best things we did recently was add increments to TC.  To my knowledge, since the birth of fishtest, both LTC, STC was played without increments. (i.e. 0.05).  I feel that all tuning/patches will favor MG... it will be a priority for an engine to gain an advantage at MG than at EG.  The reason for this is that playing accurate EG will require more time as each move needs to be near perfect, one small blunder it will be lost.  Whereas, MG is more tactical, giving the engine a high chance of scoring a big score jump... where it can afford to make sub optimal moves at EG, and still win.  So it would make sense for all those patches to be bias towards MG.

Having increments will give the engine higher chance to win EG's...when the score is nearly equal, as time will tend to be more uniform, it will train the engine to consider to play positional.

So I believe increments is highly important.  More so than base.

VoyagerOne:
One observation during TCEC.
SF spent most of it's time in MG, whereas K spent most of it's time at EG.

If this time bias is the same during tuning.  It makes logical sense that K's tuning will favor EG, whereas SF tuning will favor MG.

PR #77, opened 2016-01-11
fix count of finished runs

wfenchel:
I believe this corrects the number of finished tests (and pages of them) listed on user pages (which I believe occurs when the number in the runs db exceeds one page).


glinscott:
Apologies, I apparently was not getting notifications for pull requests here!

glinscott:
Thanks for the fix :)

PR #78, opened 2016-01-18
Correct default hash size

lucasart:
STC: 4MB
LTC: 64MB

These values were calibrated using fishtest:

STC: 2MB is too low, 4MB is enough
http://tests.stockfishchess.org/tests/view/569b14d50ebc594a0871d925
http://tests.stockfishchess.org/tests/view/569adc340ebc594a0871d910

LTC: 32MB is too low, 64MB is enough
http://tests.stockfishchess.org/tests/view/569b5c090ebc594a0871d937
http://tests.stockfishchess.org/tests/view/569c6d980ebc594a0871d958

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -26,7 +26,7 @@
   <div class="control-group">
     <label class="control-label">Test options:</label>
     <div class="controls">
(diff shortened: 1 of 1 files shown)


Yery:
Just a question: what is the advantage of running fishtest with lower hash (what is there to gain?)

glinscott:
It's more representative of hash pressure in real matches.  Thanks for running the tests Lucas.

PR #82, opened 2016-05-22
increase spsa iter count only with completed games

wfenchel:
Prior to this patch, spsa iteration count is increased when spsa work is
requested, not only when it is completed.  The spsa iteration count is
important to convergence as it affects the step size.  This has been
a problem recently as there have been many disconnecting clients; e.g.,
checking the most recently completed spsa run
http://tests.stockfishchess.org/tests/view/573fc6890ebc59301a354f99
the final iteration count is 60837 instead of 20000.  See also related fishcooking thread
https://groups.google.com/forum/#!topic/fishcooking/Np13OpEO1_w

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -367,14 +367,11 @@ class RunDb:
 
     spsa = run['args']['spsa']
 
-    # Increment the iter counter
-    spsa['iter'] += 1 
-    self.runs.save(run)
-
     # Generate the next set of tuning parameters
+    iter_local = spsa['iter'] + 1 #assume at least one completed, and avoid division by zero
     for param in spsa['params']:
-      a = param['a'] / (spsa['A'] + spsa['iter']) ** spsa['alpha']
-      c = param['c'] / spsa['iter'] ** spsa['gamma']
+      a = param['a'] / (spsa['A'] + iter_local) ** spsa['alpha']
+      c = param['c'] / iter_local ** spsa['gamma']
       R = a / c ** 2
       flip = 1 if bool(random.getrandbits(1)) else -1
       result['w_params'].append({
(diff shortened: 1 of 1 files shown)


glinscott:
Thanks!

This seems like it could be a bit problematic, in that we will send the same parameters down to a bunch of workers until we get their results back.  The current situation isn't great either though.  Hm...

wfenchel:
Hey Gary,

Since each work packet will randomly sample "flip" for each parameter, different workers will get distinct workloads even for small parameter sets with high probability.  Thus I'd argue the only real distinction is that "a" and "c" are not decreasing as smoothly as with the current approach, particular when a clump of machines make a request at once.

I chose this fix as it is such a small change.  Alternatives seem to be:

* Current behavior, but adjusted to decrement spsa['iter'] if a machine doesn't respond in a while.  Downsides: spsa['iter'] (and things depending on it) may now move non-monotonically, and also the extra bookkeeping is a pain.
* Use further randomness when choosing the work.  There is some justification for this, but I avoided changing things up too much without thorough testing.

(I discussed some of this and my other rationale in fishcooking.)

What do you think?

joergoster:
@glinscott This seems to be problematic only at the start of a tuning run. Once running, there should be a balance of incoming results and new parameter requests.
Alternatively, you can block the user causing the problems.

glinscott:
Good point, merged.

joergoster:
@glinscott Don't you need to bump the worker version to 60 to make the changes work?

glinscott:
Nope, this is all a server side change.  Server restarts every six hours to pick up new changes.

joergoster:
@glinscott OK. I will wait until tomorrow then. Thank you Gary!

wfenchel:
Thanks Gary!

I agree it's not perfect; I just wanted a minimal change while addressing the issue which is alarming people.

wfenchel:
@glinscott  @joergoster 

So I can only conclude from
http://tests.stockfishchess.org/tests/view/57450c740ebc59301a35509a
that something is horribly wrong; iter should always be half as much as num_games now
T_T

I can only imagine two possibilities:
* the server reset did not happen, those 2002 iterations are from the old behavior.
* one of the lines I deleted self.runs.save(run), was crucial to some sort of background initialization or bookkeeping which I overlooked due to not fully comprehending the how pyramids and the rest of the machinery behind fishtest works....

comments?

joergoster:
My first guess is the server reset did not happen by then. I will resubmit.

joergoster:
@wfenchel Looking at other places in the code, I guess you need to re-insert
```
self.runs.save(run)
```
right before we return the result.
But someone more knowledgeable to python should confirm!

wfenchel:
@joergoster , I removed the line since 'run' is no longer modified in any way throughout the function.

Definitely though we should wait for gary to comment and/or check the logs.

PR #83, opened 2016-05-28
sanity check number games versus number spsa games

wfenchel:
This patch adds some server-side sanity checking on reported spsa games in rundb.py::update_task().
Specifically, it disallows two behaviors:
* spsa games have just completed, but this task has no games completed.
* spsa games have just completed, but this task has no _new_ games completed.

The sanity checks are not perfect, in the sense that ideally we'd make sure 'spsa' and 'stats', as reported, are in sync, but this would add complexity since 'stats' totals games across the lifetime of a task, whereas the counters in 'spsa' reset with each set of parameters.  As such, my feeling is that if a worker is so broken that such a test is needed, we have other problems on our hands?

Anyway, the above tests should at least prevent the present anomalous behavior as assessed in the following forum post:
https://groups.google.com/d/msg/fishcooking/Np13OpEO1_w/LnQjsLooDwAJ

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -268,8 +268,13 @@ class RunDb:
       return {'task_alive': False}
 
     # Guard against incorrect results
-    num_games = stats['wins'] + stats['losses'] + stats['draws']
-    if 'stats' in task and num_games < task['stats']['wins'] + task['stats']['losses'] + task['stats']['draws']:
+    count_games = lambda d: d['wins'] + d['losses'] + d['draws']
+    num_games = count_games(stats)
+    old_num_games = count_games(task['stats']) if 'stats' in task else num_games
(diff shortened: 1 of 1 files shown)


wfenchel:
Perhaps I should explain one other thing I went back-and-forth on when making this patch.

Say that two consecutive remote api calls are made which state 5 spsa games finished, and 5 games finished (according to 'stats'); that is, two consecutive calls are made with no new games.  The second check I added will tell the worker to piss off, which it does whenever "spsa games completed is positive and number games completed does not exceed old number games completed".  The current worker implementation never makes such consecutive api calls, but technically we're adding some extra constraints to how the API is invoked.  I didn't feel great about this, but decided to add this second test to try harder to rule out the weirdness we were seeing in recent spsa runs...

glinscott:
Thanks!  Let's hope this helps fix the problem.

wfenchel:
Thanks Gary!

PR #84, opened 2016-06-12
Spsa boundary tweak

wfenchel:
I was watching some recent tunings by Fisherman, and saw spsa reverse its trend after its bounds were changed:
http://tests.stockfishchess.org/tests/view/575a7d0f0ebc59029919b385
http://tests.stockfishchess.org/tests/view/575ae6ee0ebc59029919b39d

There's a chance that the current behavior at boundaries is to blame (though I'm not sure), so I created the present pull request.  It contains two changes: one attempting to make spsa more stable at boundaries, and one to make it more stable in general by replacing the round() with a randomized rounding that in expectation gives a smoother function.

It's possible that neither of these changes are improvements, so I've added options to the new run page to enable or disable them (they are disabled by default).  I can post on the mailing list so people know to try them; if they like them, maybe they can become default, or if they dislike them, we can remove them.

Note that the first two commits in this pull request have a more detailed explanation of their behavior.  (The third third commit fixes some typos.)

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -363,8 +363,11 @@ class RunDb:
       value = min(max(param['theta'] + increment, param['min']), param['max'])
     else: #clipping == 'careful':
       inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
-      inc_sgn = 0 if increment == 0 else increment / abs(increment)
-      value = param['theta'] + inc_sgn * inc
+      if inc > 0:
+          inc_sgn = 0 if increment == 0 else increment / abs(increment)
+          value = param['theta'] + inc_sgn * inc
+      else: #revert to old behavior to bounce off boundary
+          value = min(max(param['theta'] + increment, param['min']), param['max'])
 
     #'deterministic' rounding calls round() inside the worker.
     #'randomized' says 4.p should be 5 with probability p, 4 with probability 1-p,
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -1,6 +1,7 @@
 import copy
 import os
 import random
+import math
 import time
 from datetime import datetime
 from bson.objectid import ObjectId
@@ -355,13 +356,27 @@ class RunDb:
     self.runs.save(run)
     return True
 
-  def spsa_param_clip(param, increment, clipping):
+  def spsa_param_clip_round(param, increment, clipping, rounding):
+    value = 0.0
+
     if clipping == 'old':
-      return min(max(param['theta'] + increment, param['min']), param['max'])
+      value = min(max(param['theta'] + increment, param['min']), param['max'])
     else: #clipping == 'careful':
       inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
       inc_sgn = 0 if increment == 0 else increment / abs(increment)
-      return param['theta'] + inc_sgn * inc
+      value = param['theta'] + inc_sgn * inc
+
+    #'deterministic' rounding calls round() inside the worker.
+    #'randomized' says 4.p should be 5 with probability p, 4 with probability 1-p,
+    #  and is continuous (albeit after expectation) unlike round().
+    if rounding == 'randomized':
+        fl = math.floor(value) #greatest integer <= value, thus works for negative.
+        if random.uniform(0,1) < value - fl:
+            value = fl + 1
+        else:
+            value = fl
+
+    return value
 
   def request_spsa(self, run_id, task_id):
     run = self.get_run(run_id)
@@ -381,6 +396,8 @@ class RunDb:
     spsa = run['args']['spsa']
     if 'clipping' not in spsa:
         spsa['clipping'] = 'old'
+    if 'rounding' not in spsa:
+        spsa['rounding'] = 'deterministic'
 
     # Generate the next set of tuning parameters
     iter_local = spsa['iter'] + 1 #assume at least one completed, and avoid division by zero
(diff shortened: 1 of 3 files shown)


mcostalba:
You have replaced a proved math formula with a random one without any
rationale apart from "maybe" with the option to disable them if people
don't like(!). It is SPSA, there is nothing to like or dislike: it is like
his. I think that changing SPSA, where we have some serious math, should be
supported by (1) an understanding of what you are doing (2) a strong
rationale/proof of the math involved.

This is up to Gary, but I really fail to make any sense out of this.

wfenchel:
@mcostalba,

Did you read my commit messages?  They are more detailed.  In particular, I've brought us _closer_ to the settings where SPSA has guarantees, not farther.  SPSA on fishtest or even when running on a single core for stockfish is _far_ from any of the settings where SPSA has mathematical guarantees, so that is why I said "maybe".  Also, all of my changes are guarded behind options at test creation time and not default.

* regarding the rounding: I've taken a discontinuous objective function and made it lipschitz continuous.  Stochastic optimization methods like spsa have guarantees for objective functions with a good deal of continuity.

* regarding the treatment of boundary conditions: the current handling (the min/max) is heuristic and breaks the convergence proofs, even in the simplified settings where SPSA has guarantees.  The boundary handling I added is also heuristic, but at least preserves some of the properties needed for SPSA to make sense (the symmetric sampling), and is closer to boundary methods which are principled (e.g., barrier methods, which Spall discusses in his papers, but can be found elsewhere).  The more principled methods require another parameter so I wanted this simpler approach first.

wfenchel:
@mcostalba ,  let me give a little more mathematical detail:

* regarding the randomized rounding.  In the current rounding (using (round())), notice that for any integer parameters x, the expected game value (randomization over opening books) is constant for any x+delta where delta's coordinates are within (-0.5, 0.5).  Then there is suddenly a jump once we get a 0.5.  These jumps, meaning the discontinuities in the minimization problem, are what I'm trying to fix.  If you look at the Spall or any other SPSA papers, they definitely can not handle these sorts of discontinuities.  

* regarding the boundary behavior.  SPSA works when the perturbation distribution has a specific form; one option, as with the fishtest implementation when far from boundaries, is to sample from {-1,+1} for each coordinate, scaled by the step size.  Near a boundary though, we only shrink one of these, but not both, giving an asymmetric distribution which breaks the proofs.  What I've done is shrunk both coordinates so we have a symmetric distribution again.  This is motivated by more recent perturbation analyses that give some flexibility in the perturbation distribution, see for instance http://arxiv.org/abs/1312.2139 .  What is still heuristic in my modification is that I also shrink the step sizes near the boundary, so that we never land exactly on the boundary.  This is similar to a barrier method, which is a standard suggestion (see for instance spall's paper http://www.jhuapl.edu/spsa/PDF-SPSA/Rezatat_Contrained_SPSA.PDF ), but I've removed a tuning parameter, which is simiar to a standard approach of reducing the weight on the barrier over time.

wfenchel:
@glinscott , any comments?

Even if you and marco disagree with my implementation, can we at least agree on a technical statement of the issues, and perhaps someone can find a solution that everyone likes?

wfenchel:
@glinscott , @mcostalba , @zamar 

I'll try one last time to get some attention to this pull and discussion.  While reading the spsa code in fishtest, I noticed three things that move the implementation away from the standard setting of SPSA.  Of course, since we are operating far from the standard setting of SPSA (e.g., due to convexity), it's unclear that addressing these will give improvements, which is why my pull request did not make any of the changes default, but instead introduced options to try them out.  But even so I'd just like to raise awareness, and even if this pull request is not merged, maybe it motivates some other discussions and attempts.

Anyway:

1. The use of round() to convert parameters to integers makes the objective function discontinuous.  There are many ways to fix; the present pull request uses randomized rounding, which in expectation gives a lipschitz continuous function.  This is discussed above.  Another idea is to force the objective to have higher orders of smoothness.

2. Treatment of boundaries/constraints.  The present implementation only clips one side of the distribution, meaning it becomes asymmetric; meanwhile, though there are many spsa variants with many distributions, they are always symmetric.  This patch suggests merely adjusting the distribution to be symmetric while changing nothing else.  The are of course more aggressive possibilities, for instance as motivated by the Duchi et al paper discussed above (for instance, permanently shrinking the sampling ball in order to keep a small scaling sequence as suggested there).

3. Resampling of perturbations.  SPSA wants a fresh perturbation for each trial (or pair of trials), whereas the fishtest implementation reuses the perturbation according to the concurrency of the worker.  I did not address this in the present pull request, was going to do so next.  There are many ways to proceed here.

In Marco's post above, he claimed I am proceeding without a mathematical understanding of SPSA.  I hope I have laid that criticism to rest with my detailed posts above.

mcostalba:
Sorry for my comment. This code is up to Gary.

glinscott:
Sorry, don't take my lack of response as not agreeing with the change.  I just don't have much time to review changes to fishtest these days.  Will take a much deeper look over the next few days, thanks!

wfenchel:
@mcostalba no problem, sorry for my wall of text!

@glinscott no rush

Stefano80:
I think the symmetric handling of boundaries has really potential to be an substantial improvement. It should remove bizarre dependencies on the perturbation sizes.

wfenchel:
@glinscott Hi gary, but any chance you can take a look at and discuss this over the next two weeks or so?  During that time I'll have a number of chances to discuss here and make any changes you desire, but after that I'll be a little overloaded.  Thanks either way!

glinscott:
Sorry for the super long wait!  It's been way too busy here.  This looks great!  It's deployed on fishtest now.

wfenchel:
@glinscott thanks gary!  I will write some comments on fishcooking about the new options, so that people can document their use so they can be either documented or removed based on their performance.

PR #86, opened 2016-07-14
Switch master and the patch in master diff

theo77186:
Before this patch, in unapproved test run, the master diff link is broken because it makes Github attempting to compare the master against the patch, which gives normally nothing. Instead, it should compare the patch against the master to see the diffs.


theo77186:
Retracted my PR. It was only a very weird case for this: https://github.com/official-stockfish/Stockfish/compare/master...1d09ee7

PR #88, opened 2016-07-30
Dont build twice for spsa

Stefano80:
Dont build twice if spsa tuning. In the command sent to cutechess-cli use cmd=stockfish both for stockfish and base.

When spsa tuning there is no reason to build twice the binaries, this also introduces a further noise source due to the varying speed of the PGO builds.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -407,6 +407,8 @@ def run_games(worker_info, password, remote, run, task_id):
 
   # Verify signatures are correct
   base_nps = verify_signature(new_engine, run['args']['new_signature'], remote, result, games_concurrency * threads)
+  if not spsa_tuning:
+    verify_signature(base_engine, run['args']['base_signature'], remote, result, games_concurrency * threads)
 
   # Benchmark to adjust cpu scaling
   scaled_tc, tc_limit = adjust_tc(run['args']['tc'], base_nps, int(worker_info['concurrency']))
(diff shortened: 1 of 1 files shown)


Stefano80:
Hi,

I dont know how to test fishtest locally, so I dont have a way to check that my changes actually achieves what I want to.

Could you give feedback both on the PR and on how to possibly test it locally?

ppigazzini:
Sorry if this is already solved.
You can install a development fishtest server, set a spsa test and have some workers doing the works

https://github.com/glinscott/fishtest/wiki/Installing-a-development-server-on-Ubuntu

snicolet:
> [building twice the binaries] also introduces a further noise source due to the varying speed of the PGO builds.

+1. PGO builds are a small plague on fishtest :-)

Stefano80:
Hi,

my branch is obviously broken, it blocked twice Luca's test

http://tests.stockfishchess.org/tests/view/57e4593a0ebc59763f358d09

(I forgot to get back locally to master).

Can anybody spot the mistake?

ianfab:
@Stefano80 I just had a quick look at it, and if I am not mistaken, the only issue with your PR is that you do not skip the [verification of bench for the "base" version](https://github.com/glinscott/fishtest/pull/88/files#diff-e8d8184dfbe43a62177e9eb695449cd2R411) for SPSA runs, which perhaps fails, since you skip its compilation.

Stefano80:
Ok, I keep it open until I find some time to fix this.

Stefano80:
This is not anymore needed since pgo builds are reproducible now.

PR #89, opened 2016-09-07
Don't use profile-build in fishtest

mcostalba:
Don't use profile-build because it is not deterministic. In particular the same source, comiled twice on the same machine can yield to a couple od binaries that have diferent speed.

This is a big no-no for fishtest becuase introduces a random bias that could affect testing result.

See https://groups.google.com/forum/?fromgroups=#!searchin/fishcooking/profiling%7Csort:date/fishcooking/4_ausUwMXP0/eebU5xZZFAAJ

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -43,11 +43,11 @@ HTTP_TIMEOUT = 5.0
 FISHCOOKING_URL = 'https://github.com/mcostalba/FishCooking'
 ARCH = 'ARCH=x86-64-modern' if is_64bit() else 'ARCH=x86-32'
 EXE_SUFFIX = ''
-MAKE_CMD = 'make profile-build COMP=gcc ' + ARCH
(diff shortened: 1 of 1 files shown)


Stefano80:
I see one point against this: we release PGOs, so we should also test PGOs. And I would increase TC by 15% as this what we gain with PGO

ppigazzini:
I just posted new tests in the forum: I don't view a statistical difference between PGO and NO-PGO builds. 
Anyway I attached the raw data for people with statistical knowledge stronger that mine.

glinscott:
@mcostalba please merge if you think this is still required -- sorry, I had somehow missed the message.  You should have merge privileges on the fishtest repo.

mcostalba:
@glinscott no more needed now that we do PGO with fixed depth. I forgot this one open, closing it now...

PR #95, opened 2016-11-19
Move routes into a list of tuples

madnight:
No functional change.


madnight:
I think its better to write the routes into a list of tuples than explicitly calling config.add_route 28 times

Stefano80:
Why do you think is better? I think is better to use an appropriate abstraction level, so that if we change the implementation of routes, we do have to make changes only in one place.

PR #99, opened 2017-01-09
Restore old waiting time of 5 minutes to check for new tasks.

joergoster:
30 minutes is not really necessary and was mistakenly increased by me.
Bump up worker version to 60 to force updating.


glinscott:
Let's batch all the worker updates into one big PR.

Stefano80:
@ppigazzini : this is ok for me, should we merge?

Stefano80:
Combined into a single PR for the worker updates https://github.com/glinscott/fishtest/pull/128

PR #101, opened 2017-02-27
Replace hard-coded value by chunk size

ianfab:
The internal priority formula implicitly uses the chunk size, so make this dependency explicit by replacing the hard-coded value.


glinscott:
Thanks!

PR #103, opened 2017-03-19
Replace "mingw32-make" with "make"

ppigazzini:
Replace the custom "mingw32-make" command with the standard "make" command to build the applications with a Windows worker.
In this way fishtest will work out of the box with all the MinGW-w64 distributions (MSYS provides also the "make" command that works fine).


glinscott:
Should add this to the worker upgrade PR.

Stefano80:
@ppigazzini : how do we proceed here?

Stefano80:
Included in https://github.com/glinscott/fishtest/pull/128

PR #108, opened 2017-04-18
Fix OSError raised by Popen.kill() with a terminated PID

ppigazzini:
With MSYS2 python, using Popen.kill() with a terminated PID raises an OSError
https://bugs.python.org/issue6973
Drop the Popen.kill() call, and use Popen.communicate(), that wait for process to terminate, to end the subprocess used to load the CPU.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -88,8 +88,7 @@ def verify_signature(engine, signature, remote, payload, concurrency):
 
   finally:
     if concurrency > 1:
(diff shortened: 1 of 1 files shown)


glinscott:
Thanks.  We should batch all these PRs together into a worker upgrade patch though.

Stefano80:
Included in https://github.com/glinscott/fishtest/pull/128

PR #112, opened 2017-05-28
Add "Show/Hide" button for the running machines table

ppigazzini:
When the number of running machines is high, reaching the tables of running and finished tests requires some serious scrolling.
Add a "Show/Hide" button for the running machines table, the page is loaded with the running machines table hidden.


Bernton:
Merge plis 😀

Stefano80:
@ppigazzini : maybe another good idea would be to invert the order: on the top pending tests, then results, then machines.

ppigazzini:
@Stefano80 in your way the table of machines will be buried, changing the problem: developers are interested in tests, CPU contributors are interested in machines status.
The "Show/Hide" button was also a simple solution: the code was there to shows/hides the table of pending tests.

glinscott:
Thanks!  Sorry for the huge delay...

PR #113, opened 2017-06-05
Sort active tests

ianfab:
Idea by Stehane Nicolet. Sorting criteria:
1) SPRT jobs sorted by decreasing LLR, with higher LLR at the top
2) Fix number of games jobs, with larger number of games at the top
3) SPSA tuning jobs, with larger number of games at the top


glinscott:
Thanks!

PR #115, opened 2017-07-09
Protect elo formula from math domain error

ianfab:
If the Elo difference in a fixed number of games test is huge, the calculation of the statistical error in `get_elo` will fail for `mu_max >= 1`. I encountered this error on my fishtest instance for chess variants (http://35.161.250.236:6543/tests), but it is not very likely to occur in the main fishtest instance, since Elo differences usually are much smaller there.


Stefano80:
Looks good. If nobody objects, will merge soon.

ppigazzini:
OK also for me.

PR #116, opened 2017-07-09
Do not change bounds when switching time control

ianfab:
Since SPRT bounds are not directly related to the choice of time control, automatically resetting the bounds when changing time control might be surprising for the user.


Stefano80:
Looks good to me, I will merge if @ppigazzini does not object.

ppigazzini:
ok for me

Stefano80:
Merged and server updated.

PR #117, opened 2017-07-09
Sort pending jobs by internal priority

ianfab:
...instead of by starting time. This should result in a more intuitive sorting of the pending tests, since internal priority determines which test will be picked first.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -788,7 +788,7 @@ def tests(request):
 
     runs[state].append(run)
 
-  runs['pending'].sort(reverse=True, key=lambda run: (-run['args']['priority'], run['start_time']))
+  runs['pending'].sort(key=lambda run: (run['args']['priority'], run['args']['internal_priority']))
   runs['active'].sort(reverse=True, key=lambda run: ('sprt' in run['args'], run['results_info']['llr'] if 'llr' in run['results_info'] else 0,
                                                      'spsa' not in run['args'], run['results']['wins'] + run['results']['draws'] + run['results']['losses']))
 


Stefano80:
I am pretty neutral about this. Opinions?

Stefano80:
Thinking about it again, I feat that people will start tweaking throughput to get to the top of the list. Does this make sense or is just lack of trust from my side?

ianfab:
It could be, but at least there still will not be immediate feedback after changing throughput, because internal priority is only updated when a test is picked up. Usually that "feature" is annoying, because you sometimes have to temporarily increase priority to force an update of the internal priority, but in this case it might be useful.

It is already possible to increase throughput (and even priority) for quite some time and we haven't really seen much misuse I think, so I would not except that it suddenly changes just because it is more transparent to the user, but I can not exclude that of course. One way to restrict misuse would be to only allow approvers to set priorites above 0 and throughputs above the default.

Stefano80:
Ok, I see. Indeed, it is disturbing that the internal priority is only changed when a test is picked up. Would it be difficult to also change the logics and let the internal priority updated on a throughput change? If we add this additional feature, than I would be willing to try it out (and monitor abuse). @ppigazzini : thoughts on this?

ppigazzini:
I like the patch and I'm confident in a responsible behavior of the users.

ianfab:
@Stefano80 That should be doable. The only question is how to determine the task id for the [internal priority calculation](https://github.com/glinscott/fishtest/blob/6cd02d9a219cbcfa74f2cd8aa29236ae642b25e9/fishtest/fishtest/rundb.py#L250-L261). When we update throughput we can not use the same logic since we do not update a task then. We could count the number of tasks that are not pending, or find the task with the highest index that is not pending, or the task with the lowest index that is pending (and subtract 1). In principle I think all three should give the same result, but I am not entirely sure.

Stefano80:
Can't we just extract out a "recalculate_internal_priority" function and use it also [this line](https://github.com/glinscott/fishtest/blob/6cd02d9a219cbcfa74f2cd8aa29236ae642b25e9/fishtest/fishtest/views.py#L388)?

Or am I missing something fundamental?

ianfab:
Yes, that is the idea. However, at the position I linked to, `task_id` is calculated with the help of the information that the active task with the highest index has just been updated, which will not necessarily be the case when throughput is modified. Therefore, we need a different way to find this `task_id`. For consistency it would then of course be good to include this into the internal priority calculation method and get rid of the dependency from `last_updated`. I think any of the three methods I mentioned above should work, we just need to be careful to also handle the special cases when there is no active and/or no pending task. I think the simplest solution (and also consistent with how it works currently) would be something like:
```
task_id = 0
for task in run['tasks']:
  if task.get('pending') is True and task.get('active') is False:
    break
  task_id += 1
```

Stefano80:
Ok, I will look into it in some more detail the next days. By the way, I think you could write

`if task.get('pending') and not task.get('active'):`

which seems simpler to me...

Stefano80:
Hi Fabian, will you have a go at my proposal? If yes, let us keep this open, otherwise, please rebase and I will happily merge.

ianfab:
Hi Stefano. I would like to try that, but probably not in the next weeks, so for now I rebased the PR. The recalculation of the internal priority could also be added indepently.

PR #118, opened 2017-07-18
Update iteration limit when modifying SPSA run

ianfab:
Increasing the number of games during an SPSA run (e.g., from 10000 to 20000) can result in something like:
9806/5000 iterations
20000/20000 games played

Update the number of iterations when modifying the number of games so that the results are more consistent:
9806/10000 iterations
20000/20000 games played

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -378,6 +378,8 @@ def tests_modify(request):
 
     run['finished'] = False
     run['args']['num_games'] = num_games
+    if 'spsa' in run['args']:
+        run['args']['spsa']['num_iter'] = int(num_games / 2)
     run['args']['priority'] = int(request.POST['priority'])
     run['args']['throughput'] = int(request.POST['throughput'])
     request.rundb.runs.save(run)


Rocky640:
I think that it is not advisable, by design of the SPSA algorithm, to change the number of games in a middle of an spsa run.

The right thing to do is to forbid any change there.

ianfab:
The tuning most probably is more efficient if the number of games is not changed, since the parameters have been optimized under the assumption of a fixed number of games, but in contrast to other tests, we at least do not have to care about biasing the results. Increasing the number of games is similar to starting a second run with the final parameter values of the first run, but I think it is more convenient to just extend the run when you see that parameters are still changing close to the end. Forbidding changes is also an option, but I think it is not really necessary, since there basically is no risk in allowing it, and it might save resources compared to restarting an SPSA run with a larger number of games.

joergoster:
@ianfab No good idea. 
From the wiki:

> Note: Do not modify the number of games in an SPSA test while it's running. This breaks the algorithm.

ianfab:
Well, this PR does **not** introduce the option of changing the number of games, since the code already exlicitly allows it for tuning sessions (https://github.com/glinscott/fishtest/blob/2b9e27f9e49a8fa84b36aeb85f6d01e85d651a60/fishtest/fishtest/views.py#L370), but it just tries to fix the behavior in case this happens. It should either be fixed or disabled, but the current code is inconsistent there, IMO. (As far as I understand the comment and the code, that it "breaks the algorithm" only means that the mathematical proofs no longer apply, but it might still give good results.)

joergoster:
@ianfab Thanks for explaining. In this case I'm with @Rocky640 and SPSA runs should also be excluded from modifying number of games.

alwey:
@Stefano80 : 

> We need some data here.

I also have a modied SPSA tester. It has been derived from Joona's perl code and later ported to python. It actually is a SPSA tester's tester. It uses dummy games and thus is much faster than using fishtest. 

The first thing to do when testing SPSA is to define the number N of iterations (and thus games) of the test. The (probable) optimal parameters alpha, gamma and A depend on this definition of N.

Usually it is not advisable to change N within a test. It does not entirely invalidate the results but it probably will not use the resources as efficiently as a correctly designed test.

Stefano80:
Do you have a repository for that? How does this simulator work? Can you set a Elo dependent on the parameter? In this case, we could just test whether changing the number of games in average changes the Elo gain per game.

alwey:
There is no repository yet, I froze the code about two years ago. 

It is an SPSA elo-simulator. There is a target elo-function (using something like Joona's simulator function in the perl code or an other functional) which depends on several parameters which have to be tuned. The optimum elo strength is zero. The starting values for the parameters, the parameter ranges and the SPSA tuning parameters like gamma, alpha and A can be specified by some init-files. One of these files is equivalent and yet (almost) compatible with the original format for variables in Joona's spsa.pl.

While simulating, the program outputs data into some log files. One (not any) of these is the equivalent of the log generated by spsa.pl. My goal has been to make chess engine tuning more efficient/faster.

Stefano80:
This sounds very cool, why don't you just create a repository?

alwey:
This is not ready for the public eye. It is half-baked "research code" and I will have to get reacquainted with the status reached. 

What I have learned is that you cannot force the success of a SPSA run but encourage it given the right conditions. Under the same starting conditions and same code, every run will have an individual outcome. If you have great elo potential, e.g. improving a new variant for multivariant stockfish you will have the chance to save many iterations with an aggressive setup, i.e. lower iteration limit N, higher values for gamma and alpha, lower value A. Possibly a second stage with more conservative parameters may then enhance the result. BUT: There is no guarantee for this. 

I have also seen many simulations where the best elo already was achieved after e.g. a third of the planned number of iterations and then improved no more or even showed a systematical drift to the worse (depending on the modelled setup).

Basically more tuning variables need more iterations, and interdependent parameters need much more iterations than sets with "orthogonal" parameters.

So there is no standard recipe for SPSA. One size (e.g. number of iterations N) does not fit all problems.

Stefano80:
Hi alwey, none is going to complain about half baked research code, given that the alternative is no code at all. For instance, I myself have exactly this simulator in mind dice years, but it never did to the top of my personal queue. 

If you upload, there is a fair chance that I will join the project of turning this into something more stable.

For the rest, it is clear to be me that every run will have a different outcome wrt to Elo, but sampling many SPSA runs should give you an idea of the distribution of Elo gains

alwey:
@Stefano80 : I see you have been thinking the subject. OK, then let me clean up a bit and get this into a status describable by words.

> For the rest, it is clear to be me that every run will have a different outcome wrt to Elo, but sampling many SPSA runs should give you an idea of the distribution of Elo gains.

This is what the python script does. It runs a number of simulations, and collects the final elo results. It outputs average and median of the total elos of (of the respective last iteration of each simulation).

I also tried to exploit the "special case of chess" and detect phases of possible regressions (meaning iterations with probable negative progress in elo) from the collected iteration data, using only the _obtainable_ knowledge. This is hard.

The ultimate goal has been to develop a "washing machine" for chess tuning, a multi-stage tuner. Select a program, fill in some parameters, start. Lean back and watch the bull's eye.


So now I am motivated to restart this.

PR #122, opened 2017-08-28
Show change of throughput in action history

ianfab:
The action history does not show a meaningful message when the throughput is changed (`modify:`). With this PR, `'throughput changed from %s to %s' % (before, after)` is added to the message.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -135,6 +135,11 @@ def actions(request):
       if before != after:
         item['description'].append('games changed from %s to %s' % (before, after))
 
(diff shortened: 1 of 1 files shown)


Stefano80:
A question: what happen if more than one property are changed at once?

ianfab:
@Stefano80 The actions are comma separated, see https://github.com/glinscott/fishtest/pull/122/files#diff-7f1499b600078591351c4d45d5f27a95R143 (or http://35.161.250.236:6543/actions for examples).

Stefano80:
It's ok for me, will merge if no objections arise in the next days (this has bothered me too for quite some time)

Issue #129, opened 2018-01-17
Slow/Fast switch


Rocky640:
A common usage scenario is to run a STC test, and if it is green, 
a) user will reschedule the test
b) Click the Slow button
c) Submit

However step b) always overwrite the "sprt elo0" and "sprt elo1" to [0, 5]
which is unexpected behaviour if there were already some values
(especially if former STC test was a [0,4] or [3,1])

I think it would make more sense to move the Slow/Fast buttons at the right of the time control,
and have those two buttons only change the time control, and nothing else.

ianfab:
@Rocky640 You are lucky: #116 was merged yesterday. The position can also be changed if that is preferred. I think that the hash should still automatically be changed, not only the time control.

Rocky640:
Changing the hash is indeed a good thing for "SPRT" tests or "numgames" tests.

https://github.com/glinscott/fishtest/wiki/Creating-my-first-test
For SPSA tests, I would hide the slow/fast button since the time control would depend if the user goes with the "nodestime" method or not.
Also, for SPSA, most people use hash=128 for both STC and LTC

ianfab:
I just tried out the changes and the change of the position of the buttons indeed makes usage more intuitive to me. However, hiding the buttons for SPSA can be annoying in some cases. E.g., I clicked on LTC, then switched to SPSA, and then wanted to change back to STC, but the buttons had disappeared, so I had to change it manually or switch back to SPRT first. Maybe not a very common case, but might still be a reason not to hide them.

Rocky640:
OK.

Stefano80:
May I close this rocky?

Rocky640:
Sure, and congrats to you @ppigazzini and @ianfab for your work

Stefano80:
Thanks!

PR #130, opened 2018-01-17
Move Fast/Slow buttons to time control field (#129)

ianfab:
It seems more intuitive to have them next to the field they change.


Stefano80:
Merged and restarted the server.

PR #132, opened 2018-01-21
Add buttons for choosing SPRT bounds

ianfab:
Add buttons to automatically set the SPRT bounds with one click for convenience. The naming of the buttons is "std" [0,5], "tweak" [0,4], "simp" [-3,1] to keep them small, but can of course be changed if desired. Similar buttons can be found on http://35.161.250.236:6543/tests/run.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -70,12 +70,24 @@
     </div>
   </div>
   <div class="control-group stop_rule sprt">
+    <label class="control-label">SPRT bounds:</label>
+    <div class="controls">
+      <select name="bounds">
+        <option value="standard">Standard [0,5]</option>
+        <option value="tweak">Parameter Tweak [0,4]</option>
+        <option value="regression">Non-regression [-3,1]</option>
+        <option value="simplification">Simplification [-3,1]</option>
+        <option value="custom">Custom bounds...</option>
+      </select>
+    </div>
+  </div>
+  <div class="control-group stop_rule sprt custom_bounds">
     <label class="control-label">SPRT Elo0:</label>
     <div class="controls">
       <input name="sprt_elo0" value="${args.get('sprt', {'elo0': 0})['elo0']}">
     </div>
   </div>
-  <div class="control-group stop_rule sprt">
+  <div class="control-group stop_rule sprt custom_bounds">
     <label class="control-label">SPRT Elo1:</label>
     <div class="controls">
       <input name="sprt_elo1" value="${args.get('sprt', {'elo1': 5})['elo1']}">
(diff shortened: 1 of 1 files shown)


Stefano80:
At the moment I don't have access to your server, so I can't say how does it look like.

In principle I would say ok. What about writing out the names? standard, parameter tweak, simplification?

ianfab:
@Stefano80 You only need to [sign up](http://35.161.250.236:6543/signup) (email does not have to be valid, it is not used for anything), then you can see the test submission page.

Writing out the names would make the buttons quite big and the line very long in total, so short names should be preferable.

Stefano80:
The server seems to be down.

Stefano80:
Would a drop down menu do it?

Stefano80:
Or, alternatively, just use the sprt bounds as a text.

ianfab:
The server is running and everything seems to be working fine, at least I can access it without problems and others were also contributing tests and workers in the last hours.

Do you mean a drop down menu as a replacement for the free text input of the bounds? It depends a bit on whether we want to allow non-standard bounds or not. Replacing the text by the bounds or adding the bounds behind the short names would be an option. Any label that is short enough to not look ugly but is still understandable to the user is fine with me.

Stefano80:
I was thinking of a drop down menu on the side (like where the buttons are now on your server) allowing to select bounds, which you can then change manually.

I would like to avoid shortcuts, as we do not use any in the fishtest page as for now. "simp" for instance is quite cryptic, I find.

ianfab:
The drop down menu sounds like a good option, that would keep it compact and flexible, with the minor downside of requiring one more click. I'll maybe give that a try.

Stefano80:
Let me know then!

ianfab:
@Stefano80 I had a quick look and saw that I can reuse existing code (since I am not very experienced in JavaScript), so I gave it a try immediately. I did not like the style when the select box was at the right, so I moved it to above the SPRT bounds. You can now check it out here: http://35.161.250.236:6543/tests/run. If you like it, I can adjust the PR or open a new one, of course with adjusted bounds compared to my instance.

Stefano80:
Hi, I like it as it is now, this is an elegant solution. Can we let this open some time and maybe ask in the forum whether people like it? By the way, what happens if I use some custom bounds and then reschedule? Would it keep the custom bounds or will it change them to the selected drop down item? I think the desired behaviour would be to keep the custom bounds.

ianfab:
I have not tried it yet, but from the code I read that the bounds are only changed if the selection in the drop down menu is changed. So if you reschedule a [-3,1] test, elo0 and elo1 would be -3 and 1, respectively, i.e., keeping the custom bounds, but the drop down menu would show the default "Standard [0,5]". When you click on submit, the drop down is completely ignored and only the values from elo0 and elo1 are read.

Sure, feedback from other developers/users would be very welcome (not necessarily only about this PR).

snicolet:
It is already nice as it is.

The only strange thing is that it is possible to have at the same time a menu with "Standard" and bounds of [-3..1] (say), which may be a bit confusing.

Would it be a lot of work to create something like this:

Menu =
   Standard [0,5]
   Parameter Tweak [0,4]
   Non-regression [-3,1]
   Simplification [-3,1]
   Custom bounds...

With the following behaviour:
   
A) Selecting any of the items of the menu changes the bounds (except "Custom bounds...", which doesn't change any bound at all)

B) After the user (or the website) sets any text in the bounds, if the bounds are not [0,5] or [0,4] or [-3,1] then the menu changes automatically to "Custom bounds..."

C) [Optional] if the bounds are manually set to [0,5] or [0,4] (but not [-3,1]), then the menu changes back to "Standard" or "Parameter Tweak"

Thanks a lot for taking the website maintenance, of course! :-)
Stéphane

ianfab:
@snicolet Thanks for the feedback. I think a similar but simpler behavior could be achieved by keeping it as it is now, but hiding the fields for the custom bounds if a different option than "custom bounds" is selected. This way you would not see any inconsistent or redundant data in the form. If you select "custom bounds", then the initial values of the custom bounds would be the values of the option that was last selected. What do you think?

snicolet:
Sounds perfect :-)

ianfab:
Ok, I then only need to find a solution for the case of a rescheduled test. Only the bounds, but not the selected option is currently stored, so the selection would either have to be determined from the bounds (possibly tricky) or one could simply auto-select "custom bounds" and fill in the saved bounds. If there are no objections against the latter solution, I will try to implement that.

Stefano80:
Hi, I found the solution on your page just perfect, including the logics for rescheduling. Thanks!

Could you squash, rebase and open a PR?

Stefano80:
I mean updating this PR, to have the history all in one place.

ianfab:
Ok, thanks, I'll update the PR then. I just need the list of possible options. Should I take @snicolet's list?

Stefano80:
Yes take that.

ianfab:
I updated the PR accordingly.

PR #135, opened 2018-01-27
Make book_depth = 0 disable the opening book

xoto10:
Currently, setting Book Depth to 0 breaks the worker, which loops with an error.
To allow testing from the start position, i.e. no starting book, this change makes Book Depth 0 miss out all of the book arguments in the cutechess command.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -142,7 +142,7 @@ def request_version(request):
   token = authenticate(request)
   if 'error' in token: return json.dumps(token)
 
-  return json.dumps({'version': 60})
+  return json.dumps({'version': 61})
 
 @view_config(route_name='api_request_spsa', renderer='string')
 def request_spsa(request):
--- a/worker/games.py
+++ b/worker/games.py
@@ -415,7 +415,9 @@ def run_games(worker_info, password, remote, run, task_id):
   # Handle book or pgn file
   pgn_cmd = []
   book_cmd = []
-  if book.endswith('.pgn') or book.endswith('.epd'):
+  if book_depth <= 0:
+    pass
+  elif book.endswith('.pgn') or book.endswith('.epd'):
     plies = 2 * int(book_depth)
     pgn_cmd = ['-openings', 'file=%s' % (book), 'format=%s' % (book[-3:]), 'order=random', 'plies=%d' % (plies)]
   else:
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -14,7 +14,7 @@ from optparse import OptionParser
 from games import run_games
 from updater import update
 
-WORKER_VERSION = 60
+WORKER_VERSION = 61
 ALIVE = True
 
 HTTP_TIMEOUT = 5.0


ppigazzini:
I read the forum posts. Please verify if an up to date version of cutechess-cli solve the problem (I'm not an expert w/ cutechess-cli options).
In that case I can open a PR on Marco's FishCooking Setup branch to update cutechess-cli

A couple of weeks ago I built a new cutechess-cli for multi variant fishtest, now it's thoroughly tested.
Here the links
https://github.com/ianfab/FishCooking/blob/setup/cutechess-cli-win.zip

xoto10:
No, doesn't help.
If I'm looking at the right place line 188 in https://github.com/cutechess/cutechess/blob/master/projects/cli/src/main.cpp is the where cutechess throws the error.

Stefano80:
Otherwise, I think this is good.

alwey:
There is a workaround without need of a fishtest update:

cutechess-cli accepts empty pgn-files, e.g. _empty.pgn_. One can use a reasonable value for plies, this will make no difference. The result is a match with all games starting from the default FEN position (startpos of chess).

Example command-line in bash (for illustration):
`cutechess-cli-0.6.0 -engine cmd=stockfish-4 proto=uci -engine cmd=stockfish-8 proto=uci -each tc=10+.1 -openings file=empty.pgn plies=4   -games 2 -rounds 3 -recover -repeat -pgnout match.pgn`

xoto10:
What are you thinking, also disable the book if depth < 0? Seems reasonable.
I have created a new branch startpos2 if you want to use that: https://github.com/xoto10/fishtest/commit/ea85b1f3f4f1fb16d6cbb20766a7aae4005b0705

xoto10:
@alwey If we're creating a pgn it's easy enough to create a startpos.pgn which contains the start position. I forget what my reasoning was, but I thought it was better/easier/some_other_thing to modify fishtest than create a new pgn. Interesting to know that an empty pgn is possible, I hadn't thought of that.

alwey:
@xoto: The method of using an empty game in startpos.pgn (only PGN tags header and an asterisk in move text section) is quite equivalent to the empty file scenario /wrt PGN output.

When using a FEN setup you will get an additional setup tag` [FEN "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1"]`.

Nevertheless I think your code is a valuable improvement of fishtest.

alwey:
@ppigazzini : The new cutechess-cli behaves like the old 0.6.0 version /wrt plies and bookdepth: only positive integers are acceptable.

Stefano80:
Can you update this PR with the <= case? I will merge.

xoto10:
Done.

Stefano80:
Can you please squash all commits into one?

Stefano80:
And I think we also have to update the worker's version

xoto10:
OK, I didn't know if I could squash the commits, but I've looked up how to do it now. Not sure what you mean about the worker's version?

ianfab:
@xoto10 When something is changed in the code for the workers, they will not automatically get that update. By increasing the worker version on the server side, the workers will notice that they have an outdated worker version (which is hard-coded in the worker code) and run an update.

xoto10:
Ah, I see! Have updated the version number as well ...

ianfab:
The [version on the server side](https://github.com/glinscott/fishtest/blob/3e0617cfa9ad0768be183df1387e18ed157bfb42/fishtest/fishtest/api.py#L145) needs to be updated to the same value.

Stefano80:
Thanks, I will merge as soon as possible. @ppigazzini : when updating the worker we also have to do some additional work server-side, right? Can we document it somewhere?

ppigazzini:
@Stefano80 no additional work server side is required when updating the worker. In the week end I will document my works on the server.

Stefano80:
Merged, server restarted.

PR #136, opened 2018-01-27
Add option for disabling auto-purge per test

ianfab:
This is a proposal to add an option for disabling auto-purge on a per-test basis. This might be useful if for certain tests it is expected that there will be a big statistical fluctuation, e.g., when using no/small opening book, when testing patches that give different speed-ups depending on the architecture, for time management tests, etc. By disabling auto-purge for such tests, one can avoid to bias the result and/or wasting too much resources by reapeatedly purging results.

Test submission page looks like this: http://35.161.250.236:6543/tests/run
An intentionally problematic test (in terms of fluctuations) where auto-purge was disabled: http://35.161.250.236:6543/tests/view/5a6474bb6e23db7297acc507

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -58,6 +58,7 @@ class RunDb:
               spsa=None,
               username=None,
               tests_repo=None,
+              auto_purge=True,
               throughput=1000,
               priority=0):
     if start_time == None:
@@ -83,6 +84,7 @@ class RunDb:
       'new_signature': new_signature,
       'username': username,
       'tests_repo': tests_repo,
+      'auto_purge': auto_purge,
       'throughput': throughput,
       'priority': priority,
       'internal_priority': - time.mktime(start_time.timetuple()),
--- a/fishtest/fishtest/templates/base.mak
+++ b/fishtest/fishtest/templates/base.mak
@@ -15,6 +15,10 @@
       padding:1px;
       line-height:13px
     }
+    .checkbox.inline {
+      margin-left: 20px;
+      margin-right: 0;
+    }
   </style>
 
   ${self.head()}
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -165,6 +165,12 @@ Cowardice,150,0,200,10,0.0020"""})['raw_params']}</textarea>
     </div>
   </div>
   <div class="control-group">
+    <label class="control-label">Advanced:</label>
+    <div class="controls checkbox inline">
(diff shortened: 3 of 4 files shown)


Stefano80:
Hi Fabian,

Some remarks.

1) in principle yes, sounds like a good idea
2) however, from an another point of view: we are adding even more code and we don't have a real test environment (yet)

I would tend to merge, I would like to wait for ppigazzini opinion.

If, in the meantime, you feel like diverting your time and energy to some healthy coffee pruning activities, be assured that you will win my gratitude.

ianfab:
> some healthy coffee pruning activities

If by that you mean code simplifications/refactoring: Yes, that would be nice to have and I think the fishtest code needs some refactoring, but I currently do not really have the time. My PRs mostly are byproducts of when I need and implement a feature for my fishtest instance, and since I then anyway have already done the work, I propose to reuse the code in official fishtest in case it is considered useful. There might be different priorities here, so no problem if it is preferred to not add the feature or postpone the decision. I guess it should be clarified whether the feature would be used before adding it.

Stefano80:
So, I assume @ppigazzini agrees with the patch. If no objection arises, will merge over the weekend.

ppigazzini:
I think that we should update the ["Creating my first test" wiki page](https://github.com/glinscott/fishtest/wiki/Creating-my-first-test) to set the rules on when and why to use this new feature in order to avoid abuses.

Stefano80:
Right.

locutus2:
I think this option makes sense (with default auto purge on). A later purge is later manually possible.

Vizvezdenec:
This is for sure a good thing when you tweak TM, for example.
Since TM tweaks can provide time losses -> leading to high residuals -> leading to autopurge... Which will only corrupt results.

Stefano80:
Merged, server and wiki updated

https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#advanced-options

ianfab:
Thanks. Further possible options for the advanced section would be:
- adjudication on/off
- swap colors on/off (if turned off, first engine always plays the side that is to move in an opening book position, no rematch with reversed colors)
- ponder on/off

Is any of these (or something not on this list) requested?

Stefano80:
I think adjudication would be very time saving for devs, this would be a welcome improvement.

Ponder I don't know, I don't think anybody cares, but maybe people would start to care.

Revert colour I also am skeptical, I see no reason for this.

ianfab:
Ok, thanks, I agree.

snicolet:
Thanks for the work you are doing for the the site, everybody :-)

And adding @xoto10 too...

PR #149, opened 2018-03-05
Rough fix for empty book problem

Stefano80:
Add a cleanup function to the worker, such that obvious problems with opening books are fixed.

Still to do: add some check for signature or hash.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -93,13 +93,29 @@ def verify_signature(engine, signature, remote, payload, concurrency):
 
   return bench_nps
 
+def cleanup(item, testing_dir):
+  """Remove corrupted opening books"""
+  target_path = os.path.join(testing_dir, item)
+  stat_info = os.stat(target_path)
+  if not os.path.exists(os.path.join(testing_dir, book)):
+    return True
+  elif statinfo.st_size == 0:
+    os.remove(target_path)
+    return False
+  else:
+    return True
+
 def setup(item, testing_dir):
   """Download item from FishCooking to testing_dir"""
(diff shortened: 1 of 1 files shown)


Stefano80:
@ppigazzini @ianfab could you give it a look? I think I covered all cases, but it would be good to have a second opinion and maybe have it checked on the dev or multivariant server. In the meanwhile, I am trying to build a local test case.

Stefano80:
Ok, now in the folder `./test` you can find a very simple python script which try to emulate some possible cases. I would really appreciate if you would try to let it run locally and report any problems.

Enough for today.

Stefano80:
Now that the waters have calmed, could @ppigazzini @noobpwnftw @Rocky640 give a check on this? I have a small test scripts, I would build on to go in the direction of having continuous integration for fishtest.

This is a fix for [this issue](Rocky640 ).

noobpwnftw:
The fix makes sense and I've tested it by deleting the book files and restart my workers, I do not see any corrupted books now.

Stefano80:
Ok, thanks. I will give this time until Tuesday then I will merge if no one complains. I am still not yet certain of whether I want to merge the test too, but I think is a basic starting point to create a CI framework here, so I am more for merging, although it is a bit temporary.

Stefano80:
I have to skip this, I had some issues at home and I am leaving for vacation and business. Will be back in ten days.

Stefano80:
I removed the test, this is ready to merge.

Stefano80:
@ppigazzini : I dont feel like merging my own patches. If you think this is still needed, you can merge it now.

PR #150, opened 2018-03-05
Wait longer when satisfying HTTP requests

Stefano80:
This is an attempt to make the problem described [here](https://groups.google.com/forum/?fromgroups=#!topic/fishcooking/kA28W9fyCGQ) less damaging.

Just increase the timeout in the HTTP request from 5 to 10 secs and the number of repeating attempts from 5 to 10.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -142,7 +142,7 @@ def request_version(request):
   token = authenticate(request)
   if 'error' in token: return json.dumps(token)
 
-  return json.dumps({'version': 61})
+  return json.dumps({'version': 63})
 
 @view_config(route_name='api_request_spsa', renderer='string')
 def request_spsa(request):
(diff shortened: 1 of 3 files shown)


Stefano80:
I have put the worker version to 63, so if you want, you can merge both PRs one after the other (this one second).

xoto10:
How about increasing this sleep: https://github.com/glinscott/fishtest/blob/712250c57b2ea4e22523364a92ac20d4a7ddef1c/worker/worker.py#L73
If we get into a state where say 100 workers have no work, there are going to be about 10 a second asking the server for a task - that sounds like a recipe for keeping the server overloaded!
Maybe increase it to 60s ? (120s?) Then if the server is struggling the workers will slow down to give it some breathing space.
The same applies at line 82, if there are no tasks available and lots of workers waiting, we don't want them swamping the server with requests, I would raise this to 60s or 120s as well. 

Wow, imagine if there are 10 workers doing tests and 240 not doing anything at the moment ... that means 24 requests a second for more work! That index I suggested should help the server cope with this but I think increasing both of these to 120s would be good.

Stefano80:
Closing as https://github.com/glinscott/fishtest/pull/156#pullrequestreview-102763310 solves two problems at once.

PR #153, opened 2018-03-07
Add caching of data for /tests view

tomtor:
This is a first quick hack.

It adds the caching of the DATA for the /tests view (5s timeout). 

The HTML template conversion itself is still executed, not clear for me how to do caching for that and if it is really needed. 

If you can get this to work in YOUR test environment then adding caching to other critical spots is easy.
I tested this by adding some random data and verifying that it remained constant on refreshes for 5 seconds.

### DIFF
--- a/fishtest/fishtest/__init__.py
+++ b/fishtest/fishtest/__init__.py
@@ -15,6 +15,9 @@ def main(global_config, **settings):
                         session_factory=session_factory,
                         root_factory='fishtest.models.RootFactory')
 
+  config.include('pyramid_mako')
+  config.include('pyramid_beaker')
+
   # Authentication
   with open(os.path.expanduser('~/fishtest.secret'), 'r') as f:
     secret = f.read()
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -15,6 +15,8 @@ from pyramid.view import view_config, forbidden_view_config
 from pyramid.httpexceptions import HTTPFound, HTTPBadRequest
 
 import stat_util
+import pyramid_beaker
+from beaker.cache import cache_region
 
 @view_config(route_name='home', renderer='mainpage.mak')
 def mainpage(request):
@@ -749,7 +751,12 @@ def post_result(run):
 
 @view_config(route_name='tests', renderer='tests.mak')
 @view_config(route_name='tests_user', renderer='tests.mak')
+def tests_cached(request):
+    return tests(request)
+
+@cache_region('short_term', 'tests_view')
 def tests(request):
+  
   username = request.matchdict.get('username', '')
   success_only = len(request.params.get('success_only', '')) > 0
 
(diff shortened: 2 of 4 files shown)


Stefano80:
I am not an expert on this: you said you tried locally. Do we expect some possible problems when merging on the server?

tomtor:
@Stefano80 You might have to add new Python libraries to the server, that might be difficult to roll back if there is an issue with installing these. If the libraries are (already) available then it is a piece of cake.

It all depends on how similar your testing environment is to the production server? Does it even exist? 
You should consider migrating to Docker for ease of testing and deploying.

Stefano80:
Hi Tom thank you for your comments. I think I cannot be of any help here, I hope @ppigazzini has further information...

Stefano80:
Ah, now I understand your question. We have a test environment, which has some differences to the productive one, the exact ones being known to Pasquale.

It must be said that he and I just stepped in as maintainers, after it became clear that Gary has no time left for this project.

The idea of migrating to Docker is of course tempting, but I personally are not able to start this. But if you could formulate a plan for that, I would help the best I can.

tomtor:
Ok. The most important for now is the ability to test changes which go beyond changing a few lines of Python code.

Adding a front end webserver like nginx is one of those changes which need thorough testing. A webserver would also enable https. Now all passwords are transmitted as clear text.

snicolet:
@tomtor would you be interested to help @Stefano80 and @ppigazzini as fishtest maintainers? I think both @glinscott and @mcostalba can give you the credentials if necessary.

ppigazzini:
@tomtor the official server has very outdated packages (2013), and some packages (e.g. MongoDB) was installed by hand.
IMO the best thing will be to switch to a new server. I installed a [test server](http://94.198.98.239) using [this script](https://gist.github.com/ppigazzini/8d3608910616eba855f649dd57dbe0f9), that replicates the official server.

tomtor:
@snicolet I think @ppigazzini has done an excellent job with the test server and install script. I could help if needed (e.g. with pull requests) but I do not think that I need the server credentials do to that.

Stefano80:
@tomtor : but maybe you want to think about it. You are contributing since years to the project, you obviously have quite some skill, and we need everybody on board.

Stefano80:
@tomtor, @ppigazzini  : is this ready to be merged?

tomtor:
@Stefano80 @ppigazzini I think that this is not urgent and the efficiency gain is small. I see no reason to merge this now. Other PRs don't need new Python libraries and gain more.

Stefano80:
Ok, then maybe you can close this for now and reopen when it's ready.

tomtor:
Closing for now because https://github.com/glinscott/fishtest/pull/157 should give a sufficient boost.

In addition we can add caching to a proxy http server which we need for https anyway.

PR #154, opened 2018-03-07
add index on tasks.pending to speed up query of new tasks

xoto10:
This index should speed up the finding of a new task when a worker has finished the previous one. Gives only a marginal gain on my home test, but I only have a handful of tests in the database at the moment. Should give a useful gain (possibly a very large gain) on a large db.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -30,6 +30,10 @@ class RunDb:
 
   def build_indices(self):
     self.runs.ensure_index([('finished', ASCENDING), ('last_updated', DESCENDING)])
+    self.runs.ensure_index([('tasks.pending', ASCENDING)],
+                           partialFilterExpression = { 'tasks': { '$elemMatch': {'pending': True} } } )
+    self.runs.ensure_index([('tasks.active', ASCENDING)],
+                           partialFilterExpression = { 'tasks': { '$elemMatch': {'active': True} } } )
 
   def generate_tasks(self, num_games):
     tasks = []
--- a/fishtest/utils/index.py
+++ b/fishtest/utils/index.py
@@ -1,0 +1,16 @@
+#!/usr/bin/python
+import os, sys
+
+# For tasks
+sys.path.append(os.path.expanduser('~/fishtest/fishtest'))
+from fishtest.rundb import RunDb
+
+def create_indices():
+  rundb = RunDb()
+  rundb.build_indices()
+
+def main():
+  create_indices()
+
+if __name__ == '__main__':
+  main()


Stefano80:
Ok, thanks, it looks like a non-invasive change. I am for merging, if @ppigazzini agrees.

xoto10:
I have a probable refinement to make, so don't do it just yet. There is an
extra detail I can set on the index to make it much smaller, which will
make it more effective. I can push that tomorrow. (or possibly later
tonight if you really want it. Out for an evening class now.)

Stefano80:
@xoto10, take your time. I will try to dive in this mongoDB stuff.

xoto10:
More importantly, I think my current code is wrong! Updated index looking good so far though ...

xoto10:
OK. Updated code now committed. It creates an index on tasks.pending, but only indexes the rows where tasks.pending is true. This keeps the index small and quick to access.

I used temporary code like this near line 250
```
    t0 = datetime.utcnow()
    run = self.runs.find_and_modify(**q)
    run_t = datetime.utcnow() - t0
    f = open("/tmp/q.log", "a")
    f.write("new tasks q: "+str(run_t)+"\n")
    f.close()
```
instead of the existing `run = self.runs.find_and_modify(**q)` to log how long the query was taking. It runs quick with or without the index on my db with only a few tests, but on a big db it should help a lot.

Stefano80:
A question: did you try this on a test server? I have to trust you completely, as I dont know anything about MongoDB.

xoto10:
Yes, I have tested it on my home fishtest. I am confident that it will
speed task requests up, but the amount is uncertain - anything from
miniscule to huge is possible. If release isn't until the weekend I will
have done more tests by then as well. I now have a script to create tests,
so I can create lots of 10 game tests and build up the database size a bit.

I have started looking at another index as well, but I don't have any
justification for adding that one yet. More testing tonight/tomorrow might
show if it helps.

If you've seen my comments about timeouts on the other pulls, I now think
that might be the important thing to do, although perhaps the index change
will be first because it's simpler and low impact.

xoto10:
I'm having some trouble with my home fishtest at the moment, so I would hold off for a while.
Obviously if others have tested it and it works ok and you're happy to try it, then go with it - I'm not sure if I'm breaking things in my fishtest or if something else is breaking things but I'm getting various inconsistent results. I am working some things out, but it just takes time ...

xoto10:
The thing is, it doesn't seem to be calling the build_indices() function now for me, so there's no point in adding code in there if it isn't called??
But I think I've had errors out of it before so I was confident that was the place to change, I thought it was being called each time I restart the server. I've been doing most of my index changes directly in the mongo shell or with a small standalone python script to avoid needing to restart.
Confused.

If I create a standalone script to go into the utils directory, like the current scavenge.py, would that be ok? Someone would need to run it manually on the server to update the indexes.

xoto10:
Apologies for my confusion. I can only assume build_indices() is not called from anywhere. Not sure what I was doing a couple of days ago.

I have added a new file index.py in utils to call the build_indices() function. You guys will need to run this manually, the code change in rundb.py will not do anything on its own.
There is also a second new index on the tasks.active field created in build_indices. This is useful for the first query in request_task().

A few false starts here, but I'm comfortable with this now.

Stefano80:
Mmhh... now I am losing confidence. I will wait for the opinion of @ppigazzini .

xoto10:
Up to you guys, of course. My apologies for the confusion, we're all learning here. 
When we restart the server, all the .py files get compiled into .pyc files. In the first day or two when I was getting used to pymongo, I would get syntax errors from changes I had made to the build_indices function, which led me to believe it was being run at server start, but these were just errors as the code was compiled. As I got used to the mongo shell and pymongo, I mostly changed the indexes manually and just left the server running. I had a feeling something wasn't quite right, but mostly I concentrated on checking if the indexes helped or not (including the second one), so timing the request_task() function, not looking closely at build_indices() any more. I was sure (mistake!)  build_indices() was getting run sometimes so I just thought it was some kind of caching issue or something.
In the last day or two I realised there may be an issue, but yesterday when I wanted to be figuring that out, my mongo db got corrupted somehow (possibly due to me inserting tests with a script instead of through the screen) so I spent ages getting that back to a working state instead of figuring out build_indices. I think I'm happy with it now, so I will stop making changes now and let you guys look at it.

ppigazzini:
@Stefano80 I suggest to test this patch on the dev server with a replica of the official db to verify the stability and the speedup
@xoto10 I'm not a python or MongoDB guy, so please suggest as to measure the speedup introduced by your patch

xoto10:
I thought I posted some simple code to output the time taken, but I can't
see it in this thread, perhaps that was somewhere else? (Away from PC at
the moment, so using my phone.)

Most recently, I altered that technique to cover the whole of the
request_task function since there is a query at the start of the function
as well as the one I originally looked at:
- set t0 = datetime.utcnow() at start of function
- at each exit point (2 places, I think); open file /tmp/qt.log for append
('a'), write out a log message including current time - t0, close file

Very low tech, but gives a simple output of time for each call. Of course
there are probably better profiling methods but this worked for me.
The time was only slightly quicker with the indexes on my db with only
10-50 tests in it, but I expect the improvement to be much more obvious on
a large db. (I suspect time taken is proportional to number of tests in db
at the moment, the indexes should make it closer to a constant time.)

Stefano80:
Hi @ppigazzini, yes this sounds like a good way to go.

xoto10:
@Stefano80 @ppigazzini I've pushed a [branch](https://github.com/xoto10/fishtest/tree/index_dbg) to my repo containing some debug mods to rundb.py and a new script to drop / create the indexes which you can use to manually change them if you want (just comment out the drop/create commands you don't want to run each time). The debug in rundb.py writes to /home/fishtest/... so just edit that location if necessary.

For the actual release (at some point, no pressure to do it soon) I suggest you run the utils/index.py created in the pull request. This calls the build_indices() function in rundb.py, then we know that function has accurate definitions of the indexes.

Stefano80:
Hi @xoto10 , I appreciate this contribution, but it seems complex and it adds one additional step to the operations we have to perform. Given that neither @ppigazzini nor myself are mongoDb experts, I feel we should merge only if we can measure a performance improvement.

Can you suggest a way to do it non invasively?

xoto10:
Sure, just use the idx.py script I created on the index_dbg branch I mentioned above. There are 2 drop_index() calls and 2 ensure_index() calls in there. The drop_index() calls drop the new indexes and the ensure_index() calls create the new indexes.

You can just comment out the drop_index() calls and run the script ("python idx.py") to create the indexes, or comment out the ensure_index() calls and run it to drop the indexes. The script lists the indexes on the runs collection after doing its work, so you can comment out the drop_index() _and_ the ensure_index() calls to see the current indexes. (You should see one index on '_id' and one index on 'finished' and 'last_updated' at the moment.)
Note that it could take a few seconds to run, so don't be surprised by that. (Maybe a minute or two at maximum?) Hard to guess, depends on the size of the run collection, strength of hardware, etc.)

I suggest you time the request_task() function to measure the impact, either by using some of the code mods inside request_task() in my index_dbg branch, or by timing the api call from a worker somehow - however you want.

I want to stress that indexing tables is a very basic thing in databases. The queries I looked at run in only 1-2ms on my almost empty database, or about 5ms for the request_task function(), but if there are 10s of 1000s of runs in the database then they could easily take 1000x times that, or 100x at least, basically they will scale linearly. Creating the indexes should speed this up hugely, especially since I used partialFilterExpression to only store the relevant rows in the index, which makes them smaller, and therefore quicker. This is what I do in my day job and huge speedups are commonplace when adding indexes to tables that have none, or least none that can be used. This is definitely something we should fix.

glinscott:
Indexes in the right place are a high-impact, low-risk fix.  If the db manages things properly (and this version of mongo might have a few bugs, but has overall performed quite well), and the index is a good fit, it can be a huge speed up (O(n) to O(1)).

Especially these two seem like perfect fits for an index.  Super safe to try out and see if things get faster.

glinscott:
Also, I should note, once the index is place, the DB will update it for you.  I created the index manually previously, through a temporary script that just called `rundb.build_indices()`

xoto10:
"I created the index manually previously, through a temporary script that just called rundb.build_indices()"

OK, good to hear that. I created utils/index.py in this commit to do the same thing.

Stefano80:
@xoto10, @glinscott : then I say we merge. I am away now for ten days, @glinscott , @ppigazzini feel to merge and restart.

ppigazzini:
@xoto10 @tomtor server updated and restarted (after a DB backup). I ran index.py as suggested.

ppigazzini:
@xoto10 after the update mongod uses 100% CPU (1 of 3 CPU cores). Previously it used max 25%. A restart did not changed the situation.

tomtor:
@ppigazzini It could be building the indices, that can take a while, especially while fishtest is reading/writing concurrently

ppigazzini:
@tomtor reverting the patch not solved the high CPU load from mongod. I restored this patch waiting for @xoto10

tomtor:
@ppigazzini I would expect that stopping fishtest (pserv) would speed up the index build. After the index is build cpu usage should return to normal, unless we hit a bug in this 5 year old mongodb.

Did you try building the index on your test server to see how long it takes there?

ppigazzini:
@tomtor tried in this moment to run index.py on my dev server and got this error, on the official server worked fine. I will be online in 2 hours, I will check the posts on my smartphone while commuting.
```
[fishtest@sfs-0 utils]$ python index.py
  File "index.py", line 16, in <module>
    main()
  File "index.py", line 13, in main
    create_indices()
  File "index.py", line 10, in create_indices
    rundb.build_indices()
  File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 34, in build_indices
    partialFilterExpression = { 'tasks': { '$elemMatch': {'pending': True} } } )
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/collection.py", line 1786, in ensure_index
    self.__create_index(keys, kwargs, session=None)
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/collection.py", line 1655, in __create_index
    session=session)
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/collection.py", line 244, in _command
    retryable_write=retryable_write)
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/pool.py", line 496, in command
    collation=collation)
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/network.py", line 125, in command
    parse_write_concern_error=parse_write_concern_error)
  File "/usr/lib/python2.7/site-packages/pymongo-3.6.0-py2.7-linux-x86_64.egg/pymongo/helpers.py", line 146, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: unsupported expression in partial index: tasks $elemMatch (obj)
    pending $eq true
```

xoto10:
Dev: I said it could be the version, but I can see in the error messages that it is version 3.6, so the partialFilterExpression should be fine. No idea why that error is showing.

Official: As Tomtor says, you might want to stop pserve for a few minutes to see if the db fixes itself / catches up with the workload. Do we know what processes are using all the cpu?

The idx.py script [here](https://github.com/xoto10/fishtest/blob/index_dbg/fishtest/utils/idx.py) might be useful. There are 2 function calls to drop the new indexes, 2 to create the new indexes, and the last line lists what indexes currently exist. I suggest you comment out the drop_index and ensure_index lines and run it to see what indexes it thinks currently exist.

ppigazzini:
@xoto10 My dev server has the latest stable CentOS and MongoDB. When at home I will rebuild the dev server with latest CentOS and a new dev server with latest Ubuntu.@tomtor I don't think that stopping pserve will be very useful, mongod is using only 1 core (at 100%). The others 2 cores are nearly unloaded.

xoto10:
5 year old mongodb? do we know what version?

xoto10:
Stopping pserve would be a useful diagnostic. If the cpu load drops then we know that operations from pserve are causing the load. We can then use idx.py to check the current indexes or drop the new ones or ... whatever ... as appropriate.
If the load stays at 100% with pserve stopped we can suspect the load is from internal mongodb work, and then wait for it to finish / investigate what it is doing.

Another possibility is that the indexes are reducing the I/O so much that the cpu is now the bottleneck. I wouldn't expect this to cause a problem in this case as the cpu work should be minimal, but it's a possibility. (Also disappointing if mongo can only use 1 core). In my day job a hard working cpu is usually the aim (I/O is the slowest part, so the aim is to reduce I/O as much as possible). If this is the case maybe the indexes are counterproductive and we should remove them.

ppigazzini:
@xoto10 I will stop pserve when at home (35 minutes from now)

xoto10:
If you comment out the drop_index and ensure_index calls in idx.py (my debug script mentioned about 5 msgs above this one, not the one in this pull), it will show you the current indexes on the runs collection without changing the indexes. The output should look like this before the indexes are created:
```
fishtest$ python idx_py

Current Indices:
{u'_id_': {u'key': [(u'_id', 1)], u'ns': u'fishtest_new.runs', u'v': 1},
                                          (u'last_updated', -1)],
                                 u'ns': u'fishtest_new.runs',
                                 u'v': 1}}
```
(the two indexes '\_id\_' and 'finished_1_last_updated_-1' were already there) 

and look like this after the 2 new indexes have been created:
```
fishtest$ python idx_py

Current Indices:
{u'_id_': {u'key': [(u'_id', 1)], u'ns': u'fishtest_new.runs', u'v': 1},
                                          (u'last_updated', -1)],
                                 u'ns': u'fishtest_new.runs',
                                 u'v': 1},
 u'tasks.active_1': {u'key': [(u'tasks.active', 1)],
                     u'ns': u'fishtest_new.runs',
                     u'partialFilterExpression': SON([(u'tasks', SON([(u'$elemMatch', SON([(u'active', True)]))]))]),
                     u'v': 1},
 u'tasks.pending_1': {u'key': [(u'tasks.pending', 1)],
                      u'ns': u'fishtest_new.runs',
                      u'partialFilterExpression': SON([(u'tasks', SON([(u'$elemMatch', SON([(u'pending', True)]))]))]),
                      u'v': 1}}
```
(the new indexes 'tasks.active_1' and 'tasks.pending_1' have now been created)

ppigazzini:
Data from the official server:
* mongodb db version v2.4.4
* CPU load from mongod:
  * <  1% with pserve stopped (`pkill -f start_fishtest && pkill -f pserve`)
  *  \> 95% with pserve running

xoto10:
Wow. I don't think those indexes will work then, the server version is too old.

xoto10:
If they are still there somehow, we should drop them. This script should do that:

```
import copy
import os
import pprint
from pymongo import MongoClient, ASCENDING, DESCENDING

from userdb import UserDb
from actiondb import ActionDb
from regressiondb import RegressionDb
from views import parse_tc

import stat_util

db_name='fishtest_new'

# MongoDB server is assumed to be on the same machine, if not user should use
# ssh with port forwarding to access the remote host.
conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
db = conn[db_name]
userdb = UserDb(db)
actiondb = ActionDb(db)
regressiondb = RegressionDb(db)

runs = db['runs']
old_runs = db['old_runs']

chunk_size = 1000


runs.drop_index('tasks.pending_1')
runs.drop_index('tasks.active_1')

```

xoto10:
This script will create simpler indexes allowed in that version of MongoDB:

```
import os
import pprint
from pymongo import MongoClient, ASCENDING, DESCENDING

from userdb import UserDb
from actiondb import ActionDb
from regressiondb import RegressionDb
from views import parse_tc

import stat_util

db_name='fishtest_new'

# MongoDB server is assumed to be on the same machine, if not user should use
# ssh with port forwarding to access the remote host.
conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
db = conn[db_name]
userdb = UserDb(db)
actiondb = ActionDb(db)
regressiondb = RegressionDb(db)

runs = db['runs']
old_runs = db['old_runs']

chunk_size = 1000


# create indexes:
runs.ensure_index([('tasks.pending', ASCENDING)])
runs.ensure_index([('tasks.active', ASCENDING)])


# list indexes
print "\nCurrent Indices:"
pprint.pprint(runs.index_information(), stream=None, indent=1, width=80, depth=None)

```

ppigazzini:
@xoto10 first script applied, the problem seems solved, the CPU load from mongod is normal.

This is the output of idx.py, let me know if it's correct.
```
Current Indices:
{u'_id_': {u'key': [(u'_id', 1)], u'v': 1},
                      u'expireAfterSeconds': 86400,
                      u'v': 1},
                u'expireAfterSeconds': 86400,
                u'key': [(u'deleted', 1)],
                u'v': 1},
                           u'expireAfterSeconds': 86400,
                           u'v': 1},
                                          (u'last_updated', -1)],
                                 u'v': 1},
                                               u'expireAfterSeconds': 86400,
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'v': 1},
                                    u'expireAfterSeconds': 86400,
                                             (u'start_time', -1)],
                                    u'v': 1}}
```

xoto10:
@ppigazzini  [ sorry, didn't see this post for several minutes somehow ]
Interesting. Many more indexes there than I expected - not necessarily a bad thing. Obviously the indexes have been refined over time.

I suggest running the script in my comment directly above yours, the one that mentions "simpler indexes". This will create simple indexes on active and pending. But if you want to stop for now, and continue at some other time, I understand.

xoto10:
I will do some reading on exactly how Mongo handles indexes and review that list above. It looks like there is some redundancy and we may be able to remove one or two of them which would help speed up inserts and updates. But that will be sometime in the future, not today or any time soon ...

ppigazzini:
@xoto10 I restore some recent DB backups on my dev server, all have these the indexes:

```
Current Indices:
{u'_id_': {u'key': [(u'_id', 1)], u'ns': u'fishtest_new.runs', u'v': 2},
                      u'expireAfterSeconds': 86400,
                      u'ns': u'fishtest_new.runs',
                      u'v': 2},
                u'expireAfterSeconds': 86400,
                u'key': [(u'deleted', 1)],
                u'ns': u'fishtest_new.runs',
                u'v': 2},
                           u'expireAfterSeconds': 86400,
                           u'ns': u'fishtest_new.runs',
                           u'v': 2},
                                          (u'last_updated', -1)],
                                 u'ns': u'fishtest_new.runs',
                                 u'v': 2},
                                               u'expireAfterSeconds': 86400,
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'ns': u'fishtest_new.runs',
                                               u'v': 2},
                                    u'expireAfterSeconds': 86400,
                                             (u'start_time', -1)],
                                    u'ns': u'fishtest_new.runs',
                                    u'v': 2}}
```

ppigazzini:
@xoto10 thank you for your help yesterday.
* If I'm not wrong I can revert the commit https://github.com/glinscott/fishtest/pull/167 (in other words, this PR)
* the difference between the outputs of your script idx.py from official server and dev server should be due to the different versions of mongodb
* no problem to apply the script to add the simpler indexes but before provide the script to revert those changes

xoto10:
Yes, you can revert this PR. I will create a new PR with a script to create
2 simple indexes and a script to remove them. That should make the release
process more straightforward.

At a later date I will create some more utility scripts that we can keep in
the /utils/ directory to make this process simpler in the future.

Issue #133, opened 2018-01-25
use https on fishtest


vondele:
as requested by @Stefano80 and discussed here

https://github.com/official-stockfish/Stockfish/issues/1267#issuecomment-360198514

Get rid of the warning (using firefox 57.0.4):
https://support.mozilla.org/en-US/kb/insecure-password-warning-firefox?as=u&utm_source=inproduct
I think using a let's encrypt certificate is the right idea.

d3vv:
It is a hell for admins to upgrade ca-certificates on each server, -1 contributor minimum))

ppigazzini:
From my research Waitress does not natively support SSL, so the solution should be [to setup a proxy server](https://docs.pylonsproject.org/projects/waitress/en/latest/#using-behind-a-reverse-proxy).

tomtor:
@ppigazzini No solution for HTTPS but introducing concurrent request handling by simply running:

`gunicorn --paste production.ini --threads 4`

instead of pserve worked OK for me in small tests.

Just `sudo pip install gunicorn` did the job.

See: https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/deployment/gunicorn.html

Could you try this on your http://94.198.98.239/

Would make adding a proxy server trivial and the single process setup makes efficient caching posible.

ppigazzini:
@tomtor done, I also updated the test DB from official.
As reference:
1. it was also needed `sudo pip install futures`
2. I changed start.sh to continue to use systemd:
```bash
#!/bin/zsh
cd /home/fishtest/fishtest/fishtest
nohup gunicorn --paste production.ini --threads 4 > nohup.out&
```

tomtor:
Except it should be:

`gunicorn --paste production.ini --threads 4 --worker-class gthread`

I corrected my comment (adding --worker-class ghtread) later, so my fault.

Without that clause you get one non threaded worker, according to the docs.

tomtor:
@ppigazzini Do you plan to use this also on the real server in the near future?

ppigazzini:
@tomtor development server updated.
This seems a low risk change: keep in mind that CentOS was never updated (anyway I fixed the yum broken configuration) and I ahave not the credential of the ISP control panel to administer the VM (backup, restart etc.)

The best option in long run is to have a fresh setup either with a brand new VM or with Docker.

mcostalba:
Maybe it could be useful:

[...]

PR #157, opened 2018-03-09
Make rundb thread safe

tomtor:
This is work in progress to make fishtest thread safe

Currently the fishtest server has limited capability to handle requests concurrently.
Some requests, especially those resulting in large amounts of network traffic, will cause the server to block for a while. In addition some views show inconsistent data due to concurrent querying/updating of MongoDb data.

With **gunicorn** fishtest can be started with multiple threads, but master is not thread safe.
The current waitress webserver also has threading capability, but appears to synchronize access to the update APIs.

Note that a multiple process setup will not work either, because we need to share the SPRT/SPSA and task state, so a single process with multiple threads is currently the preferred solution.

See https://github.com/glinscott/fishtest/issues/133 for info about gunicorn.

The code changes in this PR are currently being tested.

Focus is on reliability, not on optimal multi threaded efficiency.

Update: almost 100000 game updates in multiple SPSA runs including the mega SPSA  tune_imbalance3 and running an automated CURL stress query on Overview, no problems detected while running the whole day.

If others want to test/review it, I think it's ready.

### DIFF
--- a/fishtest/fishtest/synchro.py
+++ b/fishtest/fishtest/synchro.py
@@ -1,0 +1,8 @@
+def synchronized(lock):
+    """ Synchronization decorator """
+    def wrap(f):
+        def threadSafeFunction(*args, **kw):
+            with lock:
+                return f(*args, **kw)
+        return threadSafeFunction
+    return wrap
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -3,6 +3,7 @@ import os
 import random
 import math
 import time
+import threading
 from datetime import datetime
 from bson.objectid import ObjectId
 from pymongo import MongoClient, ASCENDING, DESCENDING
@@ -13,6 +14,10 @@ from regressiondb import RegressionDb
 from views import parse_tc
 
 import stat_util
+from synchro import synchronized
+
+# reentrant lock for task management
+lock = threading.RLock()
 
 class RunDb:
   def __init__(self, db_name='fishtest_new'):
@@ -28,9 +33,11 @@ class RunDb:
 
     self.chunk_size = 1000
 
+  @synchronized(lock)
   def build_indices(self):
     self.runs.ensure_index([('finished', ASCENDING), ('last_updated', DESCENDING)])
 
+  @synchronized(lock)
   def generate_tasks(self, num_games):
     tasks = []
     remaining = num_games
@@ -44,6 +51,7 @@ class RunDb:
       remaining -= task_size
     return tasks
 
+  @synchronized(lock)
   def new_run(self, base_tag, new_tag, num_games, tc, book, book_depth, threads, base_options, new_options,
               info='',
               resolved_base='',
@@ -111,6 +119,7 @@ class RunDb:
     }
 
     # Check for an existing approval matching the git commit SHAs
+    @synchronized(lock)
     def get_approval(sha):
       q = { '$or': [{ 'args.resolved_base': sha }, { 'args.resolved_new': sha }], 'approved': True }
       return self.runs.find_one(q)
@@ -123,6 +132,7 @@ class RunDb:
 
     return self.runs.insert(new_run)
 
+  @synchronized(lock)
   def get_machines(self):
     machines = []
     for run in self.runs.find({'tasks': {'$elemMatch': {'active': True}}}):
@@ -150,6 +160,7 @@ class RunDb:
   def get_runs(self):
     return list(self.get_unfinished_runs()) + self.get_finished_runs()[0]
 
+  @synchronized(lock)
   def get_unfinished_runs(self):
     return self.runs.find({'finished': False},
                           sort=[('last_updated', DESCENDING), ('start_time', DESCENDING)])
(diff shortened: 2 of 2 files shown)


fishtest/fishtest/synchro.py
```diff
@@ -0,0 +1,8 @@
+def synchronized(lock):
+    """ Synchronization decorator """
+    def wrap(f):
+        def newFunction(*args, **kw):
```
Stefano80: Could you maybe rename `newFunction` to `threadSafeFunction`?

Stefano80:
Hi @tomtor , I went through the code and it seems safe. I have three small requests:
- a minor one for readability (commented in the code) 
- can you add to the PR a link to the discussion when you setup gunicorn for the dev server for further reference
- in the end could you please smash everything in one single commit?

If @ppigazzini gives its ok, I will merge.

xoto10:
@tomtor Do you know if the framework caches the python modules or something?
I can see it compiles the .py into a .pyc when I restart the fishtest service, but I keep getting the idea it's not running my latest code. Very confusing ...

tomtor:
@Stefano80 Renamed, added gunicorn reference and squashed

tomtor:
@xoto10 Just restarting should be sufficient, I cannot think of a reason that your changes appear to have no effect

xoto10:
@tomtor No worries. The usual thing - I think I understand now I've asked someone else, perhaps writing it down to ask you helped me figure it out.

tomtor:
@Stefano80 I added two info queries to the lock. See commit message. I think that we can continue to use pserve/waitress for minimal impact and apply this PR to production.

I could not detect any substantial performance impact in my test environment. If the impact is larger then expected then it is easy to rollback this change, and add some additional caching before reapplying.

Stefano80:
Hi @tomtor , could you locally verify that with this patch the displayed information improved? You said you wanted to reproduce the garbled output problem locally.

If yes, rebase and squash, I will merge.

tomtor:
I could not reproduce the problem, probably due to insufficient load in my test environment...

So I cannot verify that it fixes things, that will only become clear in production.

tomtor:
@Stefano80 @xoto10 I would like to study https://github.com/Pylons/waitress/tree/master/waitress and run some more performance tests before this is merged. I expect we simply have a resource issue when running expensive request_task and request_spsa APIs, so I would prefer merging https://github.com/glinscott/fishtest/pull/161 first. That should free a considerable amount of CPU/IO resources and leave more for real work.

tomtor:
@Stefano80 See also https://github.com/glinscott/fishtest/pull/163

I think that we need this current PR also for waitress, because waitress DOES execute different API handlers concurrently, default nr of threads is 4.

I am not 100% sure if we must synchronize update_task(), because it is executed so frequently. If we don't synchronize we only risk the garbled last update time. On the other hand, update_task is efficient compared to the other APIs. I think the best approach is to introduce a second lock which is shared between update_task and get_machines.

xoto10:
@tomtor See also my changes to update_task just submitted [here](https://github.com/xoto10/fishtest/commit/6fcf283a4d7990bd93f7b9f5127463fedecfcf76). 
This should stop possible issues where a task can be part saved to the db and then read by another thread.

tomtor:
@xoto10 Looks good, so then it's probably better to remove the lock from update_task completely?

Could you please also check the request_task logic? With a gunicorn tests I noticed that a worker got "No tasks available" while tasks WERE available, probably due to a race condition, which is probably also present in the current waitress setup.

The locks in this PR have a minimal performance impact (I can execute 10000 /test queries in 17s in 10 concurrent CURL sessions), but we should not introduce locks which have no functional/reliability benefits.

xoto10:
I haven't really thought in detail about the locking you've been looking
at, so I'd go with your opinion at the moment. I'm not sure we need lots of
extra locking, but if it doesn't slow things down a lot we could try it.
I'll look at request_task again with your issue in mind ...

xoto10:
@tomtor At first glance I don't see any obvious way that it can miss an available task. The query looks for a task with active == false and updates it to active == true in one atomic operation which sounds good.

We seem to have made progress already, maybe we should just merge the most obvious pulls and then pause for a few days / a week or two to assess how it's performing now, and what issues are still occurring? (Or what new ones we have caused?!) My indexes and your cache and change to the waitress connection limit look like obvious ones to me. Maybe my update_task change as well, I'm fairly confident it's right, but I want to test/think about that a bit more before merging, I've only done fairly simple tests on it so far.

xoto10:
Hmmm, I've just seen [this](https://groups.google.com/d/msg/fishcooking/UPirLLJg8ZM/z6Ew_e3bBQAJ) report from marrco.
I'm not sure if this is something my update_task() changes could fix. At the moment update_task() looks like 2 or 3 separate database updates, each of which might be delayed by locks in the database or scheduling switches. When it is only one database update, there will be fewer opportunities for delay.

In the meantime, perhaps we should increase that timeout as well? (it's still at 5s)

tomtor:
@xoto I believe that is a  result from the 100 connection limit. Update_task has also a retry of 5x in the worker so that is ok.

tomtor:
I am closing this for now...

PR #158, opened 2018-03-10
Increase default SPRT game limit to 1000000

man4:
Currently, SPRT tests have a 128000 game limit, meaning that at most 128
workers can be assigned to a test, unless the limit is manually
increased. With the recent increase of fishtest machines, we are hitting
this limit quite often. The 128 chunks can get assigned quickly when the
test has barely started. As a result, some machines become idle.

This patch allows 1000 chunks to be assigned to a test by default.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -300,7 +300,7 @@ def validate_form(request):
       'drawelo': 240.0,
     }
     # Arbitrary limit on number of games played.  Shouldn't be hit in practice
-    data['num_games'] = 128000
+    data['num_games'] = 1000000
   elif stop_rule == 'spsa':
     data['num_games'] = int(request.POST['num-games'])
     if data['num_games'] <= 0:


Stefano80:
Good idea. Could you please set the number of games to a power of 10. I suggest 1000000

Stefano80:
Hi, last point: squash everything in a single commit, update the description and I will merge.

man4:
Okay, I think I made it one commit now.

xoto10:
The possible downside of this is that workers don't normally get reassigned to other jobs until the end of their 1000 games, and this could be around 400-500 minutes (on 8-cores) if it's an LTC test. (Please check my maths)
I like the idea of using the full power of the server to get a new job through quickly at a quiet time, but someone who comes along just after and then has to wait hours to get a handful of machines for their test might not be so happy. Getting 127 machines to one test is already pretty good.

Stefano80:
You are right but this actually happens only if the framework is empty, so the patch will increase total throughput, which is good.

Stefano80:
Merged and tested. Thank you and congratulations!

Issue #15, opened 2013-05-02
Better error message when password is incorrect


glinscott:

Stefano80:
Just checked, closing.

PR #161, opened 2018-03-11
Add a cache for the Tests view

tomtor:
without any new framework dependencies.

I am a bit in doubt over the cache time, started with 5 seconds, now it's 2 seconds, could even be 1 second

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -8,6 +8,9 @@ import sys
 import json
 import smtplib
 import requests
+import time
+import threading
+
 from email.mime.text import MIMEText
 from collections import defaultdict
 from pyramid.security import remember, forget, authenticated_userid, has_permission
@@ -15,6 +18,20 @@ from pyramid.view import view_config, forbidden_view_config
 from pyramid.httpexceptions import HTTPFound, HTTPBadRequest
 
 import stat_util
+
+# For caching the tests output
+cache_time = 2
+last_tests = None
+last_time = 0
+
+def clear_cache():
+  global last_time
+  last_time = 0
+
+def cached_flash(request, requestString):
+  clear_cache()
+  request.session.flash(requestString)
+  return
 
 @view_config(route_name='home', renderer='mainpage.mak')
 def mainpage(request):
@@ -342,7 +359,7 @@ def tests_run(request):
       run_id = request.rundb.new_run(**data)
 
       request.actiondb.new_run(authenticated_userid(request), request.rundb.get_run(run_id))
-      request.session.flash('Started test run!')
+      cached_flash(request, 'Started test run!')
       return HTTPFound(location=request.route_url('tests'))
     except Exception as e:
       request.session.flash(str(e))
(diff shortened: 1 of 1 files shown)


fishtest/fishtest/views.py
```diff
@@ -347,7 +352,7 @@ def tests_run(request):
       run_id = request.rundb.new_run(**data)
 
       request.actiondb.new_run(authenticated_userid(request), request.rundb.get_run(run_id))
-      last_time = 0
+      clear_cache()
       request.session.flash('Started test run!')
```
Stefano80: Can't you just manually decorate `request.session.flash`?

I mean to create a wrapper function:

```
def cached_flash(request, requestString):
  clear_cache()
  request.session.flash(requestString)
  return 
```
I think it would make the code clearer.

tomtor: Excellent idea!

xoto10:
Removing the time from the title is definitely a good idea :-)
I would have thought 1s was too low, I would start with at least 5 or 10s, but hard to judge. Perhaps start higher and then reduce if people complain, if it's too low how will we know to increase it?

xoto10:
Sure. As you say we're less likely to cause a new problem this way, and we should still gain whenever two or more people hit F5 at roughly the same time. I think it's a win almost whatever time we use, and difficult to tell what the best value is. Let's get it in there and see what happens ...

tomtor:
@xoto10 @Stefano80 Added a semaphore to prevent parallel rebuilds of the cache and disabled caching for pages > 1. I think it is ready.

snicolet:
Slightly off-topic, but to see the progress of the number of games played and LLR, it would be cool too if somewhere in the page there could be a link to the "Elo estimate for SPRT tests" webpage:

http://hardy.uhasselt.be/Toga/live_elo.html?5a186b8e0ebc590ccbb8b1b6

Of course the URL would have to be changed for each test :)

Two ideas for the link:

• either on a new field in the test result page
• or as clickable `DIV` around the grey/green/red block at the top:
````
LLR: 1.39 (-2.94,2.94) [0.00,4.00]
Total: 3197 W: 682 L: 595 D: 1920
````

Note that if we use a clickable `DIV` then we can also add these links on the main "Overview" page, for all SPRT tests.

tomtor:
@snicolet Or integrate it in fishtest instead of just a link (show the meters on the test page).

I often use the Elo estimate page myself, I'll have a look at it once fishtest is a bit more robust with a large number of workers, almost there.

xoto10:
Yes, some kind of link to that page would be good, it's very useful.

Another thing I thought of, how about a "login" link on the left somewhere?
Some operations don't work very smoothly if you're not logged in first. It
would be useful to be able to log in as a separate step before e.g.
stopping/modifying a job.

tomtor:
@xoto The technical waitress error is just a configuration issue. You should be directed to an error page with "Please login".

https://github.com/glinscott/fishtest/issues/15 is probably related. Perhaps some comments should be added there?

Login/Logout buttons would be a logical addition.

Stefano80:
This is not ready, right? Will be back in 10 days, have a nice discussion!

tomtor:
That's a bit unexpected. Did you even read the discussion? That was not related to this patch. It IS ready.

ppigazzini:
@tomtor server updated and restarted.

tomtor:
@ppigazzini Thanks!

PR #162, opened 2018-03-12
Flush output in worker,  also print time taken to fetch version and request a task.

xoto10:
Flush output in worker so that timing of output is more intuitive. Add start message.

Print time taken to fetch version and request so that this can be monitored.

Obviously not a hugely important patch, but I think it makes some useful changes to the worker output. One to include next time we update the worker for some other reason.


Stefano80:
I have to pass on this, will be back in ten days.

vondele:
LGTM

tomtor:
While we are at it, can't you do the same in games.py regarding update_task and request_spsa?

I did similar changes in my local worker.py and raised the timeout to 15s. In times of heavy load I get an response in eg 7 seconds. I would suggest raising the timeout to 15s from 5s in worker.py and games.py, because currently work is completed in the fishtest server and waisted because the worker hangs up before the server can return the result.

If you also bump the version number to 63 here and in fishtest/api.py then it's an easy merge.

xoto10:
Added timing info for update_task().

If many operations are routinely taking 2-3s then it's not surprising that there are occasional timeouts. The timeout increases in this change should help.

Stefano80:
Tom is this ok now? To me it looks good and I would be for merging.

tomtor:
Sure, please merge.

Stefano80:
Hi @xoto10 , sorry, but there seems to be some (trivial) conflicts. Could you double check that this is indeed trivial? I will merge as soon as these are resolved.

xoto10:
Hi @Stefano80 
Yeah, they are trivial. You can see what the idea is, replace each original print command with the new printout() function.

ppigazzini:
@xoto10 games.py was changed by some PRs, so there is a conflicts that must be resolved before the merge of your PR.

I think that should be this commit https://github.com/glinscott/fishtest/commit/e3cb29cb060ecfcbba6a19f83dc72a0e07f46cd6 .

Stefano80:
Also, since we now have a CI solution for the worker: could you add a small unit test for printout?

Stefano80:
I resolved the conflicts manually. Should we merge this and all other worker patches now?

ppigazzini:
@Stefano80 yes

tomtor:
A 15s timeout is OK for me.

Stefano80:
Ok, now I see what happened. I resolved the conflicts manually, and now it passes, but it cannot be rebased because the first commit cannot be applied.

Stefano80:
In fact, it could be squashed, just checked. So in principle, we could apply the 3 worker patches in the pipeline, upgrade the worker version and be done with it. What do you think about this?

ppigazzini:
@Stefano80 To me is OK, please mind that also this PR raises the worker version.

Stefano80:
We are having conflicts here, I will wait for @vondele https://github.com/glinscott/fishtest/pull/258 before I try to solve them.

Stefano80:
I resolved the conflicts, we should test it before merging.

Stefano80:
Locally tests are passed, we should test it on the dev server.

ppigazzini:
@Stefano80 test started w/ dev server and some workers.

Stefano80:
Ok, let me know when can I merge the new worker version on the repository. For now it looks ok, right?

ppigazzini:
@Stefano80 this PR already raises the version number to 63.

Stefano80:
Ah ok. So we can merge this last and close https://github.com/glinscott/fishtest/pull/228 (assumed we are satisfied with it). Right?

Stefano80:
I also updated the worker locally and I am running it vs the official server and it is working as expected.

ppigazzini:
@Stefano80 all test OK, you can merge.

Stefano80:
Squashed and merged.

ppigazzini:
I can updated the production server, OK?

Stefano80:
Yes.

ppigazzini:
Done.

Stefano80:
Something is broken.

Stefano80:
Mmmh... coming back online.

ppigazzini:
Is working ;-)

Stefano80:
That was exciting. We should do a proper post-morten...

Stefano80:
But I do not understand, batches are sill 1000 long.

ppigazzini:
Because those are already started tests.

Stefano80:
Ah ok. I start a new one then.

Stefano80:
But somehow not all workers are coming back.

Stefano80:
@ppigazzini : I have to go now. Let us look tomorrow whether the rest of the workers came back online, otherwise we will have to do something.

PR #163, opened 2018-03-12
Set max waitress connections to max nr of workers

tomtor:
Waitress default is 100.

If we have 200 workers with 8 cores and TC = 10s+0.1s
then this results in about 160 game result updates per second.

If a few expensive requests like the /tests, api/request_run or
api/request_spsa cause a delay of more then 0.625s then requests
will be dropped because the backlog is > 100 requests

Most of these dropped requests could have been processed quickly
and each worker will never have more than one outstanding request.
In addition default worker timeout is 5s.

Raising max connections from 100 to 500 (300 in original proposal, see comments) does better match
current number of workers.

### DIFF
--- a/fishtest/production.ini
+++ b/fishtest/production.ini
@@ -22,6 +22,8 @@ mako.directories = fishtest:templates
 use = egg:waitress#main
 host = 0.0.0.0
 port = 6543
+# Match connection limit with max number of workers
+connection_limit = 500
 
 ###
 # logging configuration


xoto10:
Nice spot! 
Would a higher number be worth considering to get more headroom? (Maybe not very likely to get even more workers, but while we're here ...)

Stefano80:
Yes, I would go to 500 here...

ppigazzini:
@Stefano80 +1 for 500. In case of problems w/ the server will be an one line change.

tomtor:
OK, but we first need https://github.com/glinscott/fishtest/pull/161 in place to have more resources for these game_update and request_* calls. Could be combined with this PR in the same restart.

tomtor:
Changed limit to 500

Stefano80:
Then we wait. Will be back in 10 days, @ppigazzini feel free to merge both as soon as they are done if you feel so.

tomtor:
@ppigazzini You can merge this and #161 
There is no functional dependency.

You could start with #161 (it is ready) and evaluate that first to see if it frees sufficient resources to reduce timeouts.

If you are uncomfortable with this then you could just merge this PR first (or just edit production.ini and restart) and see if that improves the timeout issues.

The index patch of @xoto10 would be a nice next step, but you should discuss that with him first.

ppigazzini:
@tomtor server updated after merging https://github.com/glinscott/fishtest/pull/161
I will wait a couple of hours before updating the server w/ this one.

ppigazzini:
@tomtor server updated w/ this one.

ppigazzini:
@tomtor I'm wondering if connection_limit = 500 is still fine for the fleet...

tomtor:
@ppigazzini I wonder how many connections nginx uses to connect to `pserve`?

Does it map several outside connections to a single pserv connection?

ppigazzini:
@tomtor I enabled on PROD the nginx status monitor, see
https://www.cyberciti.biz/faq/nginx-see-active-connections-connections-per-seconds/

This with 186 active workers:
```
Active connections: 25 
server accepts handled requests
 2340749 2340286 4540308 
Reading: 0 Writing: 2 Waiting: 23 
```

Here some information on pyramid and nginx configuration

- pyramid nginx configuration example
https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/deployment/nginx.html

- nginx parameters
  - http://nginx.org/en/docs/ngx_core_module.html#worker_connections
"Sets the maximum number of simultaneous connections that can be opened by a worker process.
It should be kept in mind that this number includes all connections (e.g. connections with proxied servers, among others), not only connections with clients."

  - http://nginx.org/en/docs/ngx_core_module.html#worker_processes

  - total connections = worker_processes * worker_connections

We are using the default nginx.conf (from top we are using 4 worker processes)
```
user www-data;
worker_processes auto;
pid /run/nginx.pid;
include /etc/nginx/modules-enabled/*.conf;

events {
        worker_connections 768;
        # multi_accept on;
}
```

ppigazzini:
@tomtor If I'm not wrong nginx by default don't limit the connection to the proxied server.
https://www.nginx.com/blog/tuning-nginx
https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns
"limits the maximum number of simultaneous active connections to the proxied server (1.11.5). Default value is zero, meaning there is no limit."

PR #164, opened 2018-03-12
Remove extra calls to run.save() from update_task()

xoto10:
The update_task() function in rundb.py makes several calls to run.save() during [execution], which means that partially updated run data is visible to other processes in the database.

This change removes the intermediate calls and only executes one save() at the end of the function when the whole of the update has finished. To enable this, the stop_run() function is modified so that it can either read and save the run being stopped from/to the database, or update the data in memory if a "run" object is passed in.

Hopefully this will stop some odd effects where inconsistent data is sometimes displayed on fishtest [and also reduce the load on the server].

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -175,7 +175,7 @@ class RunDb:
       
     return result
 
-  def get_results(self, run):
+  def get_results(self, run, save_run = True):
     if not run['results_stale']:
       return run['results']
 
@@ -194,7 +194,8 @@ class RunDb:
 
     run['results_stale'] = False
     run['results'] = results
-    self.runs.save(run)
+    if save_run:
+      self.runs.save(run)
 
     return results
 
@@ -299,12 +300,10 @@ class RunDb:
     if 'spsa' in run['args'] and spsa_games == spsa['num_games']:
       self.update_spsa(run, spsa)
 
-    self.runs.save(run)
-
     # Check if SPRT stopping is enabled
     if 'sprt' in run['args']:
       sprt = run['args']['sprt']
-      sprt_stats = stat_util.SPRT(self.get_results(run),
+      sprt_stats = stat_util.SPRT(self.get_results(run, False),
                                   elo0=sprt['elo0'],
                                   alpha=sprt['alpha'],
                                   elo1=sprt['elo1'],
@@ -312,9 +311,11 @@ class RunDb:
                                   drawelo=sprt['drawelo'])
       if sprt_stats['finished']:
         run['args']['sprt']['state'] = sprt_stats['state']
-        self.runs.save(run)
+        self.stop_run(run_id, run)
 
-        self.stop_run(run_id)
+    if (   not 'spsa' in run['args'] or spsa_games == spsa['num_games']
+        or num_games >= task['num_games'] or len(spsa['w_params']) < 20):
+      self.runs.save(run)
 
     return {'task_alive': task['active']}
 
@@ -333,8 +334,11 @@ class RunDb:
 
     return {}
 
-  def stop_run(self, run_id):
-    run = self.get_run(run_id)
+  def stop_run(self, run_id, run = None):
+    save_it = False
+    if run is None:
+      run = self.get_run(run_id)
+      save_it = True
     prune_idx = len(run['tasks'])
     for idx, task in enumerate(run['tasks']):
       is_active = task['active']
@@ -347,7 +351,8 @@ class RunDb:
     # Truncate the empty tasks
     if prune_idx < len(run['tasks']):
       del run['tasks'][prune_idx:]
-    self.runs.save(run)
+    if save_it:
+      self.runs.save(run)
 
     return {}
 


tomtor:
@xoto10 This is interesting to know if we are hunting for inconsistencies: 

https://blog.meteor.com/mongodb-queries-dont-always-return-all-matching-documents-654b6594a827

Stefano80:
I have to pass here, will be back in 10 days, have a nice discussion!

tomtor:
@xoto10 I think this PR and https://github.com/glinscott/fishtest/pull/154 will really speedup fishtest. 

Especially when stop_run is called from update_task the system currently blocks. An issue related to that is that stop_run can be called concurrently by several concurrent update_task threads (4 in the current setup), which obviously makes no sense. It can and need be stopped only once and stop_run does a LOT of database updates. It is even extremely likely that is is executed concurrently when eg 50 workers are working on the same job.

A low tech approach without introducing a semaphore for each run could be add code at the start of stop_run like e.g.:

```
  stopping = None
  
  def stop_run(self, run_id):
    global stopping
    if run_id == stopping:
      return {}
    stopping = run_id
```
There is obviously still a race condition, but that's no issue, it's just an optimization. What do you think about this?

xoto10:
@tomtor Yes, that could be an issue. Heading to gym now, but will look later. Hmmm, multiple threads updating the same run ... but does that happen? Or do all the (many cores) updates come from one worker?

tomtor:
@xoto10 Yes, it DOES happen. We can have over 200 update_task calls per second from the collective workers, many will be concurrent. If the SPRT stop rule triggers and eg stop_run takes 100ms we will most probably have all 4 threads executing the stop_run code...

xoto10:
Ah, yes! Different workers each on different task, but same test. Hmmm, will think ...
it seems like it must already be interleaving updates on the same run ...

xoto10:
@Tomtor Your thread safe work  might be the right solution here, but just
for update task. Mongo doesn't have built in transactions but we need to
force calls to update task to run sequentially

xoto10:
OK. Going back to your comments about stop_run(). Yes, it will block at stop_run() because of the save() before calling stop_run() and then another save() at the end of stop_run(). This will be different after my changes, but it will still block at the end of update_task() while the update is saved ... and then we get the chance for another thread to read the run data before the update save has finished. :-(

I think we need something to prevent concurrent execution of the whole update_task() function. I was thinking of using a Lock:
```
...
from threading import Lock
...
def __init__  ...
  self.update_lock = Lock()
  ...
def update_task
  self.update_lock.acquire()
  ... <existing code> ...
  self.update_lock.release()
```
but I'm not sure whether we should block or just return if we fail to acquire the lock. Could blocking cause problems if the worker times out at roughly the same time as the update starts getting executed?
Maybe we could sleep for 1s if the acquire fails, retry just once, and then return a fail if the second acquire also fails? (something like this [saves2](https://github.com/xoto10/fishtest/commit/edd3144f6bd14365ff50fbe01e19ebfd2f300801) branch)
What do you think?

tomtor:
@xoto10 This should do the job: https://github.com/glinscott/fishtest/pull/165

xoto10:
Ah yes, nice. We don't want to _always_ block concurrent calls to
update_task, only if they're for the same run.

OK, I'm happy to merge my PR, which should speed things up a bit and remove
some of the odd effects. Then we can merge yours separately, to hopefully
make it really solid.

fishtest/fishtest/rundb.py
```diff
@@ -194,7 +194,6 @@ def get_results(self, run):
 
     run['results_stale'] = False
     run['results'] = results
-    self.runs.save(run)
```
tomtor: This is not related to update_task. 

It stores the result in the database, I think we should keep this line.

xoto10: Hmmm, it's called at line 307 `sprt_stats = stat_util.SPRT(self.get_results(run)` so I can see why I wanted to remove the save, but it's also called a lot from views.py so you're right, the current change is no good.

I could add a boolean 
```
def get_results(self, run, save_run = True):
...
  if save_run:
    self.runs.save(run)
```
 and modify the call in update_task() to pass `False` in. That way we remove the save from update_task() without affecting the other calls?

tomtor: Sounds good

fishtest/fishtest/rundb.py
```diff
@@ -312,9 +309,9 @@ def update_task(self, run_id, task_id, stats, nps, spsa):
                                   drawelo=sprt['drawelo'])
       if sprt_stats['finished']:
         run['args']['sprt']['state'] = sprt_stats['state']
-        self.runs.save(run)
+        self.stop_run(run_id, run)
 
-        self.stop_run(run_id)
+    self.runs.save(run)
```
tomtor: We can have a very important optimization here for expensive SPSA tests:

    if not 'spsa' in run['args'] or spsa_games == spsa['num_games']:
      self.runs.save(run)

because we only need to save the data (and it is a LOT of data for a many parameter SPSA)  at the end of an iteration
or if it is a regular SPRT run!

xoto10: How about:
```
if not 'spsa' in run['args'] or spsa_games >= spsa['num_games']:
  self.runs.save(run)
```
Is >= safer or riskier?

xoto10: Ah, I see. It's the same test as higher up. OK, I go with your version.

xoto10:
Coded, ran a single test, squashed and pushed. Confused since I can't see some comments we made a few minutes ago, does it look ok?
(Will go for a coffee soon, that should wake me up a bit :-)

tomtor:
@xoto10 @ppigazzini Looks GOOD to me. This should give a good performance boost, it will reduce the MongoDb write pressure to perhaps 1/3th and close to 1/14th for big SPSA runs.

tomtor:
@xoto10 Did you also test a short SPSA run?

xoto10:
No, I just did a minimal single test to check it would compile OK. I don't
think I have managed to get an SPSA to work at all on my test server - I
think some other problem came along and distracted me, I haven't tried
recently. I will do one this evening, it should just be a case of copying a
test from live ...

tomtor:
@xoto10 The change higher up is wrong, because we only have the spsa parameters transmitted in the original condition. Now we recompute with non existant parameter data. We can and should just skip the final iteration.

The other change is fine I think. I will also test locally on my server.

tomtor:
@xoto10 And can we run 200 games tests in general? Shouldn't it be a multiple of 1000? Could be the cause of your problem.

xoto10:
@tomtor "So you should leave the original condition higher up in place."
I didn't get this at first, but I have tested and see now. Have reverted to original condition.

"Did the worker continue looping or did it just end after playing those 4 games?"
The worker kept looping because it was being told to stop after 3 games but the 3 games weren't saved, so it was asked to do 4 more games (not sure why 4 when only 3 needed, it was doing batches of 4 ), ... etc

```
if not 'spsa' in run['args'] or spsa_games == spsa['num_games'] or num_games == task['num_games']:
```
Happy to use `==` instead of `>=`, just thought `>=` might be a bit more defensive. I will continue tests as well.

tomtor:
@xoto10 It was asked to do 4 because a complete SPSA mini batch must be an even number (2x nr-of-cores).

It was told to stop after 3 because 197 + 3 == 200. So fishtest ignores these last 3 games because it is not a complete iteration.

Running tests here as well.

xoto10:
I have 2 workers connected to my local fishtest, but one just stopped running the test for some reason. Can't see any errors on server or worker, worker just says:
```
Score of stockfish vs base: 4 - 4 - 12  [0.500] 20
Task exited
```
Test has only run ~1400 out of 2000 games. Any suggestions as to why it might stop? I assume it's nothing to do with our testing, but who knows?? Normally there's an error and trace so you can see what the problem is.

xoto10:
@tomtor pushed to check we're testing the same code

tomtor:
@xoto10 No because I just added:
`or len(spsa['w_params']) < 20 `
with the idea that we do not have to be efficient for small SPSA tests...
Not sure about 20, perhaps 50?

xoto10:
@tomtor You mean on the test deciding whether to save? Something like this?
```
    if (   not 'spsa' in run['args'] or spsa_games == spsa['num_games'] 
        or num_games == task['num_games'] or len(spsa['w_params']) < 20):
      self.runs.save(run)
```
Not sure, my instinct is to avoid the saves wherever possible. But the big tune does seem to be the thing causing trouble, so maybe it is those updates in particular?? Don't know, I say go with 20.

tomtor:
@xoto10 Yes, your code looks fine. Ok, we go with 20, note that I did use your '>=' now

    numgames >= task['num_games']

because that's used earlier in the function.

xoto10:
Looking at the SPSA code and thinking about the large tunes, those updates _are_ the problem. (I'm sure you're convinced of this already!) There must be a huge amount of data in the run documents for the large tunes, with the saved values added every 100 games. Is `spsa['param_history']` the only part of the tuning run data that grows during the test or is there more? We could probably speed up these updates if we could prepopulate this data with some sentinel value when the run is first created, instead of increasing the size of the document every 100 games - this creates a lot more work for the db. (Not saying we do this now, but it's an idea to consider for later.) Not much of an issue if there are only 3 tuning parameters, but it must be a big overhead when there are lots of parameters.

Ah, there's the list of tasks that expands as the test gets more workers over time. I guess this happens to all tests and is 10x less frequent, so the updates for this are less of an issue.

tomtor:
Last push is excellent, we're done here :+1:

xoto10:
Excellent!
Thanks for your help!  :smile:

ppigazzini:
@xoto10 @tomtor server updated after the commit. At the moment there are 2 spsa tasks ([tune_imbalance3](http://tests.stockfishchess.org/tests/view/5a9ccf220ebc590297cb6054) and [tune_pawns](http://tests.stockfishchess.org/tests/view/5aac25380ebc5902997ff07c)) and the workers continue to go off line.

xoto10:
@ppigazzini @tomtor 
Yes, I see that. We do have ~290 workers which is rather a lot, and 2 tuning jobs, so I think it is better than before. But yes, still needs to improve.
See what tomtor says tomorrow, #165 which stops concurrent updates to the same run, and #162 which  increases the timeouts in the worker might be candidates next.
I will work on some more indexing tomorrow; a utils script to show us stats about the database and a couple of index changes.

tomtor:
Hmmm, a bit disappointing. I would have expected better. I suggested to @ppigazzini in #165 to test if further increasing threads helps, but I think we might be really hitting an IO bottleneck here.
The fact that request_version times out hints at all the threads waiting for IO. Perhaps we should just allow one concurrent request_task, to reduce impact of this call until we find a way to speed it up?

Is there some more info about cpu usage of pserve and mongodb? And IO operations of MongoDb?
Currently request task always times out. If we cannot speed that up then we should perhaps limit

tomtor:
@ppigazzini Could you try https://github.com/glinscott/fishtest/pull/171 without merging? So test it by  just saving the current rundb.py and replacing it to see if it improves the load handling?

You could get the new file with

    wget https://raw.githubusercontent.com/tomtor/fishtest/f14da89100e6af5ad435aa6a754f819786490943/fishtest/fishtest/rundb.py

or curl

tomtor:
@xoto10 And something else: Why do we read from MongoDb on each request? We are the only client and will always find what we last wrote? It should be sufficient to read once after booting. What do you think?

xoto10:
@tomtor @ppigazzini
I think the current situation isn't too bad. From experience we know that the updates on large tuning jobs are almost definitely the heavy load, and having merged this one and #165 a large tuning job can now only use all of one thread, which leaves 3 others to do other work. Only allowing 1 update at a time would just reduce our possible I/O throughput I think, making performance worse. I would consider more threads, not less, say 6?

Do we know where the timeouts are now? I would guess only on the large tuning job. Tom, have you got any timings for update_task() on your system, for small/large tests? I think we need to assess the current position as much as we can before rushing into the next work.

Holding the run data in memory? Hmmm. The reads should be cheap, and the DB should be largely in memory anyway, not sure it saves much. It feels like we'd be writing our own DB. I suspect we should be doing update() instead of save(), don't know if that would speed things up at all.

tomtor:
@xoto10 @ppigazzini I was not planning to lower the number of threads, I requested to test with 8. 
Note that #171 does not limit update_task but request_task.
A request_task takes about 2 seconds on a system with low load, if we have a few concurrent then this wil impact all threads and cause them to time out, that's my theory.

xoto10:
@tomtor @ppigazzini
I can see slow responses to the main web page in my log, and a request_task() that timed out from my worker, so it's not just updates to the tuning job that are slow. What does the server load look like, is there any way we can verify if there really are 4 threads working? Do we have just the one RunDB object shared by the whole server? Normally I would expect separate connections to a database, but we have self.conn as part of a single RunDB, are we doing all this on one thread?

tomtor:
@xoto10 Note that threads have been raised to 8, I requested a test with 16.

tomtor:
@xoto10 We have a single rundb, but that's by design. We want to share state. Note that MongoDb serialiazes writes by itself for the same document, so more connections do not improve the situation.

xoto10:
@tomtor @ppigazzini  Ah, I misread. 
2s!!! If that's right then that is a big deal. I will work on the indexing then, we should be able to fix that ...
I think the timeout for empty workers failing request_task() should be higher, otherwise we get the problem of many empty workers causing a load on the system, perhaps I should change that in #162.
I counted the workers in the list just now, it said 216 workers, but I counted only ~150 that had updated "seconds ago", not sure if that means anything.

tomtor:
@xoto10 The slow request_task is really a bottleneck when a run has ended and eg 100 workers request a new task within 30 seconds. I really think that #171 limits the impact on other runs

xoto10:
@tomtor @ppigazzini "Note that MongoDb serializes writes by itself for the same document" 
Of course. 
I misread your #171 as being on update_task() which I was thinking of as the expensive call. This is why your #165 is good because it stops us trying to do parallel updates which can't be executed concurrently anyway. I didn't see that (my view of) #171 gained anything. My misreading there.

"so more connections do not improve the situation."
If we only have one connection then MongoDB might be serialising all threaded calls through that one connection. The database is the device for sharing state, is there any evidence that we have 8 active threads? Did moving from 4 to 8 change anything? I suspect we only have one. (Happy to  proved wrong.)

tomtor:
@xoto10 We really have 8 threads. I see that when debugging eg the tests/ View. The point is that when API calls take long then that will consume these threads also for simple API tasks. So we must either speed up these calls (update_task often takes just 25 ms) or prevent expensive API calls from consuming all threads. This work in these expensive API calls is waisted, because the worker gets a timeout before the expensive result is delivered. #171 deals with that. We count raise the Semaphore count to 2 if we see that the system behaves better.

tomtor:
@ppigazzini Thanks for the changes! A recent request_task from my worker took 4900ms, so this API is REALLY slow. Other APIs run fine (update_task around 30ms but sometimes a lot more). I think its good that we now run only one request_task in parallel. Let's keep the system running for a while to learn more.

ppigazzini:
@tomtor @xoto10 Yesterday I reinstalled my [dev server](http://94.198.98.239/tests) from zero (CentOS 7.x, w/ a copy of the official fishtest DB), I ran some tests of https://github.com/glinscott/fishtest/pull/164# using a spsa test w/ very low time control, and a couple of workers (31 threads linux and 7 threads WSL). I didn't know as measuring the difference from the two settings (w/o abd w/ the PR), so I tested only to be sure to not have big problems.
Here are the updated scripts to install a fishtest server, I'm using CentOS to have a setting similar to the official server (feel free to comment out the diff patch used to create the two users):
* [CentOS](https://gist.github.com/ppigazzini/8d3608910616eba855f649dd57dbe0f9)
* [Ubuntu](https://gist.github.com/ppigazzini/0c7968edd80929a8d4ad94345177e460)

xoto10:
@tomtor @ppigazzini My suspicion is that some of the slow requests are only slow because they are delayed by others. Experience suggests the system runs ok without the large tuning job, so doesn't that suggest the large update_task() calls are the root of the problem?
Also you mentioned I/O limits earlier, I suspect our habit of using save() on the whole document instead of update() on only the fields being updated is creating unnecessary I/O.

I have created #174 to create an index that may speed up request_task(). I suggest we try this and see if it helps. There is possibly #175 as well, I separated them into 2 scripts so that we can tell if either of them causes a problem.

We should plan an upgrade to MongoDB, the current version is old and we might a performance boost from the upgrade. Once on a newer version we will have more sophisticated indexes available.

tomtor:
@tomtor Hi, on my modified worker request_task just required 7000ms. So on a not modified worker it would have gotten a timeout. Note that without the big SPSA task a request_task still takes 2000ms...

tomtor:
@xoto10 Yes, updating the whole document is probably killing. Just updating modified fields sounds good.

tomtor:
@ppigazzini @xoto10 This should reduce the huge SPSA history: https://github.com/glinscott/fishtest/pull/176

PR #165, opened 2018-03-14
Make update_task synchronous for each active run

tomtor:
This prevents:

- A slow down due to parallel execution of the stop run and SPSA logic
  which overwrites the computed results of other threads
- Concurrent overwriting saves to the MongoDb
- Stopped runs which display an LLR of <= 2.94 instead of >= 2.95
- Display artifacts in Last Updated time

but still allows concurrent updates by tasks of different runs.

This is important because update_task is the most executed (> 99%) API call.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -3,6 +3,8 @@ import os
 import random
 import math
 import time
+import threading
+
 from datetime import datetime
 from bson.objectid import ObjectId
 from pymongo import MongoClient, ASCENDING, DESCENDING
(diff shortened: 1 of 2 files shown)


tomtor:
@ppigazzini Hi, I prefer that you first test https://github.com/glinscott/fishtest/pull/164 because it gives a bigger performance improvement than this PR!
This PR could be next.

ppigazzini:
@tomtor ok, I will check https://github.com/glinscott/fishtest/pull/164 firts. Thank you

tomtor:
@ppigazzini could you please try threads = 16 in production.ini to see if that improves the timeouts?

ppigazzini:
@tomtor sure, done https://github.com/glinscott/fishtest/pull/173
Server updated and restarted.

PR #166, opened 2018-03-14
Revert "add index on tasks.pending to speed up query of new tasks"

ppigazzini:
Reverts glinscott/fishtest#154


ppigazzini:
https://github.com/glinscott/fishtest/pull/154 causes a MongoDB high CPU load.

PR #167, opened 2018-03-14
Revert "Revert "add index on tasks.pending to speed up query of new tasks""

ppigazzini:
Reverts glinscott/fishtest#166


ppigazzini:
https://github.com/glinscott/fishtest/pull/167 was recommitted to debug the MongoDB high CPU load problem.

PR #170, opened 2018-03-16
Improve tests cache implementation

tomtor:
and add some comments.

This gives a minor improvement under heavy load, because it prevents parallel cache builds when we don't have an old cache yet, e.g. after a server restart.


tomtor:
@ppigazzini Hi, do you have a text suggestion for the comment?

ppigazzini:
I think should be corrected "build" in "built", e.g. : "Another thread has built" or "Another thread built"

tomtor:
Ah, you are right! Will fix it, thanks.

PR #171, opened 2018-03-17
Limit concurrent expensive request_task API

tomtor:
Limit impact of request_task API on throughput


ppigazzini:
@tomtor according your request I'm testing by hand this PR. Let me know if I must commit it.

PR #174, opened 2018-03-17
create index on run.tasks.pending

xoto10:
This index aims to speed up the request_task() function with less risk than the index on tasks.active. (I suspect the index on tasks.active may cause a decrease in performance on some queries.)

There are comments in the script indicating how to drop the index if necessary.

PR #175, opened 2018-03-17
create index on tasks.active

xoto10:
This aims to speed up the request_task() function and get_machines() function, but has a risk of slowing down another query in request_task().
#174 should be tried before this one and the impact assessed for a while before considering this one.

Edit: no need to restart the server when changing indexes, the index create might take a short time, but once the script finishes and the new index is listed it is active. Obviously if it does cause a problem follow the instructions in the script to drop the index.

### DIFF
--- a/fishtest/utils/index_active.py
+++ b/fishtest/utils/index_active.py
@@ -1,0 +1,43 @@
+
+# index_active.py - create new index on runs.tasks.active
+#
+# Run this script manually to create the index, it could take a few seconds/minutes
+# to run.
+#
+# If the index needs to be removed for any reason, comment out the create_index()
+# and uncomment the drop_index() and rerun.
+
+
+import os
+import sys
+import pprint
+from pymongo import MongoClient, ASCENDING, DESCENDING
+
+
+db_name='fishtest_new'
+
+# MongoDB server is assumed to be on the same machine, if not user should use
+# ssh with port forwarding to access the remote host.
+conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
+db = conn[db_name]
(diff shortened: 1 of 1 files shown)


xoto10:
@ppigazzini @tomtor 
It looks like this one has caused a problem. I am getting continual exceptions in update_task but it was OK earlier. I suggest you drop this index, there are instructions in the script, you need to comment out the create_index() call and uncomment the drop_index() call and then run it again.

ppigazzini:
@xoto10 done

```
Dropping index ...

Current Indexes:
{u'_id_': {u'key': [(u'_id', 1)], u'v': 1},
                      u'expireAfterSeconds': 86400,
                      u'v': 1},
                u'expireAfterSeconds': 86400,
                u'key': [(u'deleted', 1)],
                u'v': 1},
                           u'expireAfterSeconds': 86400,
                           u'v': 1},
                                          (u'last_updated', -1)],
                                 u'v': 1},
                                               u'expireAfterSeconds': 86400,
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'v': 1},
                                    u'expireAfterSeconds': 86400,
                                             (u'start_time', -1)],
                                    u'v': 1},
 u'tasks.pending_1': {u'key': [(u'tasks.pending', 1)], u'v': 1}}
```

xoto10:
OK, thanks.

ppigazzini:
@xoto10 let me know if I must:
* retest this index after the https://github.com/glinscott/fishtest/commit/0a14bc413341289e8b8c7665eb956fa7d6f53bde
* revert this commit

PR #176, opened 2018-03-17
Optimize the stored SPSA parameters

tomtor:
Record less for > 20 parameters runs and limit max size

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -481,8 +481,18 @@ class RunDb:
         'c': c,
       })
 
-    # Every 100 iterations, record the parameters
+    # Every 100/1000 iterations, record the parameters
+    # and stop at 500000 or 200000
     if 'param_history' not in spsa:
       spsa['param_history'] = []
-    if len(spsa['param_history']) < spsa['iter'] / 100:
-      spsa['param_history'].append(summary)
+    if len(spsa['params']) < 20:
+      freq = 100
+      maxlen = 5001
+    else:
+      freq = 1000
+      maxlen = 201
+    if len(spsa['param_history']) < maxlen:
+      if len(spsa['param_history']) < spsa['iter'] / freq:
+        spsa['param_history'].append(summary)
+    elif len(spsa['param_history']) > maxlen:
+        spsa['param_history'] = spsa['param_history'][-maxlen:]


tomtor:
Note that I am still tuning this...

ppigazzini:
@tomtor ok, let me know when it's ready to be commited.

tomtor:
This is ready now, but I think you should also commit now https://github.com/glinscott/fishtest/pull/171
because that last one tested OK and request_task is still a bit too slow?
@xoto10 Do you agree with keeping/merging #171

xoto10:
To be honest, I'm not convinced request_task() is slow. My worker is getting some calls returned in 0.1s : 
```
$ grep "Task requested" work.log
Task requested in 3.157374s
Task requested in 0.118925s
Task requested in 0.095153s
Task requested in 2.409761s
Task requested in 1.948533s
Task requested in 2.118051s
Task requested in 0.157814s
Task requested in 0.092477s
Task requested in 0.103471s
Task requested in 2.339293s
```
If some calls are being returned in 0.1s that makes me think that the slow ones are being caused by some other thing, not this function.

But I don't see #171 doing much harm other than maybe causing a few timeouts on request_task(), so go for it if you like - if it does help that will tell us something.

I think reducing the data volume in the spsa writes (and/or frequency of the writes) will help - I suspect that as the root cause of our problems.

tomtor:
@xoto10 Note #171 is currently in production, it is only not committed. Since it is there almost all machines have 'last updated' set to seconds, instead of many old zombie workers. 
I agree that the real root cause is in unneeded large writes, but we have not fixed that yet.
If it is fixed and request_spsa is always fast (like update_task < 50ms) then we can revert #171

ppigazzini:
@tomtor @xoto10 I will commit https://github.com/glinscott/fishtest/pull/171, if and when proved not useful anymore, I will be happy to revert it.

tomtor:
@locutus2 This has an impact on your current test. The SPSA history will no longer grow, but SPSA parameters itself will still be tuned (see code). If you start a new tuning session it will record SPSA every 1000 iterations, so it is perhaps best to start a new one, which will have less impact on the framework and a proper history.

ppigazzini:
@tomtor server updated

tomtor:
@xoto10 
Task requested in 0.118925s
Task requested in 0.095153s
Task requested in 2.409761s
Task requested in 1.948533s
Task requested in 2.118051s
Task requested in 0.157814s
Task requested in 0.092477s
Task requested in 0.103471s

Pretty sure those fast responses are "no task". Never seen a response < 1s the last 4 days with a task....

PR #178, opened 2018-03-18
Fix missing runs from Overview page

tomtor:
This is caused by a MongoDb index combined with a search
which skips some entries while being updated, see:

https://blog.meteor.com/mongodb-queries-dont-always-return-all-matching-documents-654b6594a827

Performance impact should be small because this is only called from the cached Overview page.

Edit: Note that there is still a smaller race condition, because this prevents acquiring new update locks while the overview is building, but updates might be already in flight. So it will reduce the chance but not completely reduce it to 0. Acquiring all run locks looks like overkill.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -153,7 +153,8 @@ class RunDb:
     return list(self.get_unfinished_runs()) + self.get_finished_runs()[0]
 
   def get_unfinished_runs(self):
-    return self.runs.find({'finished': False},
+    with self.run_lock:
+      return self.runs.find({'finished': False},
                           sort=[('last_updated', DESCENDING), ('start_time', DESCENDING)])
 
   def get_finished_runs(self, skip=0, limit=0, username='', success_only=False):


tomtor:
@vondele @xoto10 Thanks for the thumbs up, but see the edit. It does not fix the issue completely, let's see if this patch is sufficient

xoto10:
I thought there was an option on find() to make it read consistent, which would be the good way to do it, but a quick look just now suggests not. Working with a Mongo DB we need to do more of this than I'm used to with a relational db. If this helps, than that's good, if it's not perfect, well, it's only the display that's incorrect, and a refresh solves that. If it was an error storing data, I would be more interested in getting it more reliable.

tomtor:
@ppigazzini How often and when is scavenge.py script run? This is also modifying the database.

ppigazzini:
@tomtor 
```
~$ crontab -l
# Backup mongodb database every hour and upload to s3
2 * * * * ~/fishtest/fishtest/utils/backup.sh

# Remove clients that are not updating any more
*/5 * * * * ~/fishtest/fishtest/utils/scavenge.py

# Update the users table
*/29 * * * * ~/fishtest/fishtest/utils/update_users.py

# Restart server every six hours
20 */6 * * * ./start.sh
```

ppigazzini:
@tomtor I'm waiting for your OK (after the info about the scavenge.py) to merge this PR. Let me know

tomtor:
@ppigazzini I think you can merge, the scavenge.py only runs once per hour so that will not be the reason for the missing runs. Let's see if this PR reduces this problem or makes it disappear completely.

ppigazzini:
@tomtor  scavenge.py runs at every 5th minute, so 12 times per hour. Let me know.

tomtor:
@ppigazzini I could just produce the error at 1651. It took about 20 attempts so please merge.

ppigazzini:
@tomtor OK, server updated and restarted.

tomtor:
@ppigazzini We are down due to my patch?

ppigazzini:
@tomtor It seems so.

tomtor:
@ppigazzini Best to checkout the previous version. Perhaps a dead lock I missed?

ppigazzini:
@tomtor I reverted locally two commits to return up, so the candidate is this https://github.com/glinscott/fishtest/commit/a56517123249f736dc862f2d1dfdc3e7a09ae7ef

```
~/fishtest$ git log --oneline
6911fb9 Fix race condition between user action and cache
0530a9f Optimize the stored SPSA parameters
f855f7d Limit concurrent expensive request_task API
e9c2b5e Improve tests cache implementation
8be09e3 create index on run.tasks.pending
```

ppigazzini:
@tomtor I'm testing your patches on my [dev server](http://94.198.98.239/tests) and they seems fine. Do Have you suggestion to replicate the problem? More pressure from the workers? More running time?

```
[fishtest@dfts-0 fishtest]$ git log --oneline
45dc32b Fix missing runs from Overview page
a565171 Fix a race condition on periodic lock cleanup
6911fb9 Fix race condition between user action and cache
0530a9f Optimize the stored SPSA parameters
f855f7d Limit concurrent expensive request_task API
e9c2b5e Improve tests cache implementation
```

ppigazzini:
@noobpwnftw could add some of your workers to my [dev server](http://94.198.98.239/tests) to tests the latest two commits?

tomtor:
@ppigazzini Only more pressure from workers... I will study it tomorrow.

noobpwnftw:
How many workers do you need @ppigazzini ?

ppigazzini:
@noobpwnftw start with 10 workers. This is a copy of the official DB, simply use your usr & pwd.

tomtor:
@ppigazzini Did you run a565171 a while in production or did you only try this and 45dc32b together, after which the boot failed? Especially a565171 is only a local change, hard to imagine a locking error on that one.

ppigazzini:
@tomtor I update the server w/ a script in order to have logs. From sync_repo.log.20180318165402 I merged both the commits, rebooted the VM and checked for a minute the web page.

```
diff --git a/fishtest/fishtest/rundb.py b/fishtest/fishtest/rundb.py
index 9b6a78c..ce6eaa6 100644
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -153,8 +153,7 @@ class RunDb:
     return list(self.get_unfinished_runs()) + self.get_finished_runs()[0]

   def get_unfinished_runs(self):
-    with self.run_lock:
-      return self.runs.find({'finished': False},
+    return self.runs.find({'finished': False},
                           sort=[('last_updated', DESCENDING), ('start_time', DESCENDING)])

   def get_finished_runs(self, skip=0, limit=0, username='', success_only=False):
@@ -289,15 +288,13 @@ class RunDb:
     with self.run_lock:
       self.purge_count = self.purge_count + 1
       if self.purge_count > 100000:
-        old = time.time() - 10000
-        self.active_runs = {k: v for k, v in self.active_runs.iteritems() if v['time'] >= old}
+        self.active_runs.clear()
         self.purge_count = 0
       if id in self.active_runs:
-        active_lock = self.active_runs[id]['lock']
-        self.active_runs[id]['time'] = time.time()
+        active_lock = self.active_runs[id]
       else:
         active_lock = threading.Lock()
-        self.active_runs[id] = { 'time': time.time(), 'lock': active_lock }
+        self.active_runs[id] = active_lock
       return active_lock

   def update_task(self, run_id, task_id, stats, nps, spsa):
Updating 6911fb9..45dc32b
Fast-forward
 fishtest/fishtest/rundb.py | 11 +++++++----
 1 file changed, 7 insertions(+), 4 deletions(-)
repo sync finished, reboot the server with:
sudo reboot
```

I reverted https://github.com/glinscott/fishtest/commit/45dc32bacf569d2c32193cc54e429427549c04e0, rebooted the VM, mongod and python was running in top, but the webpage was still down.
So I reverted also https://github.com/glinscott/fishtest/commit/a56517123249f736dc862f2d1dfdc3e7a09ae7ef, rebooted the VM, and the webpage was finally up.

Let me know.

tomtor:
So it looks as if a565171 is faulty. It is cosmetic, but strange it causes a hang. It becomes active after 100000 updates, and 10000 seconds. You could change that 100000 to 10 to see what happens. You need less testing load that way. You may also change the 10000 seconds to 10 seconds. That is how I tested locally.

I cannot imagine that it is related to an old python version on the production server.

ppigazzini:
```
Python 2.6.6 (r266:84292, Jul 23 2015, 15:22:56)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux2
```

I will change the dev server according your suggestion.

ppigazzini:
@tomtor the [dev server](http://94.198.98.239/tests) is working fine w/ your two patches also w/ 10 updates and 10 seconds. There are ~200 cores w/ 2 SPSA tests @ very short tc.

A fresh install of the official server will help to develop and to test properly...

tomtor:
I found this, perhaps a difference between 2.6 and 2.7

    # Python 2.7 and 3.x
    mydict = { k:v for k,v in mydict.items() if k!=val }
    # before Python 2.7
    mydict = dict((k,v) for k,v in 
    mydict.iteritems() if k!=val)

tomtor:
Perhaps best to revert just a565171 until we have a newer server and retry 45dc32b alone? Sorry for the trouble...

ppigazzini:
Quick and dirty test

Official server:
```
Python 2.6.6 (r266:84292, Jul 23 2015, 15:22:56)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux2
>>> mydict = { k:v for k,v in mydict.items() if k!=val }
  File "<stdin>", line 1
    mydict = { k:v for k,v in mydict.items() if k!=val }
                     ^
SyntaxError: invalid syntax
```

Dev server:
```
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> mydict = { k:v for k,v in mydict.items() if k!=val }
NameError: name 'mydict' is not defined
```

tomtor:
Apparently line 311 should have been written (for Python 2.6):

    self.active_runs = dict((k,v) for k, v in self.active_runs.iteritems() if v['time'] >= old)

Do we want to test this or just postpone this?

If we want to test:

https://github.com/glinscott/fishtest/pull/181

ppigazzini:
Reverted https://github.com/glinscott/fishtest/commit/a56517123249f736dc862f2d1dfdc3e7a09ae7ef and retried https://github.com/glinscott/fishtest/commit/45dc32bacf569d2c32193cc54e429427549c04e0

Server rebooted

```
:~/fishtest$ git log --oneline
2f47264 Revert "Fix a race condition on periodic lock cleanup"
45dc32b Fix missing runs from Overview page
a565171 Fix a race condition on periodic lock cleanup
6911fb9 Fix race condition between user action and cache
0530a9f Optimize the stored SPSA parameters
```

ppigazzini:
@noobpwnftw you can stop your workers from the dev server, thank you for your help

PR #180, opened 2018-03-18
Use separate connections instead of one shared connection

xoto10:
This is a discussion PR, rather than one to implement immediately, although the change here may be useful on its own.
I believe we are not using MongoDBs connection pool ability correctly since self.runs is defined in the single RunDB object. This will mean that a lot of database work will be serialised when it could execute more concurrently. This change makes a start at correcting this by using a local runs object in functions in the rundb.py source file.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -20,18 +20,24 @@ class RunDb:
   def __init__(self, db_name='fishtest_new'):
     # MongoDB server is assumed to be on the same machine, if not user should use
     # ssh with port forwarding to access the remote host.
-    self.conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
-    self.db = self.conn[db_name]
+    self.client = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
+    self.db = self.client[db_name]
+
     self.userdb = UserDb(self.db)
     self.actiondb = ActionDb(self.db)
     self.regressiondb = RegressionDb(self.db)
+
+    # self.runs and self.old_runs should be phased out and local definitions used in 
+    # the functions where they are needed
+    # Note: these are currently still used outside this source file
     self.runs = self.db['runs']
     self.old_runs = self.db['old_runs']
 
     self.chunk_size = 1000
 
   def build_indices(self):
-    self.runs.ensure_index([('finished', ASCENDING), ('last_updated', DESCENDING)])
+    runs = self.db['runs']
+    runs.ensure_index([('finished', ASCENDING), ('last_updated', DESCENDING)])
 
   def generate_tasks(self, num_games):
     tasks = []
@@ -66,6 +72,7 @@ class RunDb:
     if start_time == None:
       start_time = datetime.utcnow()
 
+    runs = self.db['runs']
     run_args = {
       'base_tag': base_tag,
       'new_tag': new_tag,
@@ -115,7 +122,7 @@ class RunDb:
     # Check for an existing approval matching the git commit SHAs
     def get_approval(sha):
       q = { '$or': [{ 'args.resolved_base': sha }, { 'args.resolved_new': sha }], 'approved': True }
-      return self.runs.find_one(q)
+      return runs.find_one(q)
     base_approval = get_approval(resolved_base)
     new_approval = get_approval(resolved_new)
     allow_auto = username in ['mcostalba', 'jkiiski', 'glinscott', 'lbraesch'] 
@@ -123,11 +130,12 @@ class RunDb:
       new_run['approved'] = True
       new_run['approver'] = new_approval['approver']
 
-    return self.runs.insert(new_run)
+    return runs.insert(new_run)
 
   def get_machines(self):
+    runs = self.db['runs']
     machines = []
-    for run in self.runs.find({'tasks': {'$elemMatch': {'active': True}}}):
+    for run in runs.find({'tasks': {'$elemMatch': {'active': True}}}):
       for task in run['tasks']:
         if task['active']:
           machine = copy.copy(task['worker_info'])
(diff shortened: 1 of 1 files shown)


tomtor:
But see https://github.com/glinscott/fishtest/pull/178

If it was completely serialised we would not have that issue?

xoto10:
Maybe. Or maybe individual reads / updates block and interleave as they wait for I/O?
I notice we don't seem to get much above 100% cpu, shouldn't that be 200%, 300% if we have an overloaded threaded system?

I gather you have a decent load test ability, how much cpu use do you see used under load? Just try this change and see? Anecdotally, I stopped the server while a test was running to restart with updated code, and the queued up updates seemed to go in very quickly, but it wasn't much work and I didn't check that again more formally. Perhaps that is one way to check, I will try that ...

Reading docs, I found something saying the result of the MongoClient() call should be global, which implied to me that the rest should be more local, but I couldn't find a good code example.

tomtor:
I think we have low cpu in pserv because we are mostly waiting for MongoDb. MongoDb itself gets many update requests for the same run object and will serialize these and wait for IO. We should reduce the update pressure, I do not think database connection (pooling) is the problem.

xoto10:
But if it serialises updates on the one busy run object that should leave
spare threads to service the reads, e.g. the main page?

tomtor:
@ppigazzini What does "iostat 5" show ?

tomtor:
@xoto10 Yes, that's what is happening. Many threads are active, that's why we can do many concurrent update_tasks each taking < 50ms while we wait 2000ms for a request_task.

But note that in the end most threads are either waiting on a MongoDb operation directly or on a thread (with lock:) which is waiting on a MongoDb operation.

ppigazzini:
@tomtor 
```
~$ iostat 5
Linux 2.6.32-042stab127.2 (tests.stockfishchess.org)    03/18/18        _x86_64_        (32 CPU)

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           2.48   41.82    1.86    0.01    0.00   53.82

Device:            tps   Blk_read/s   Blk_wrtn/s   Blk_read   Blk_wrtn

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           0.00   35.08    1.48    0.00    0.00   63.45

Device:            tps   Blk_read/s   Blk_wrtn/s   Blk_read   Blk_wrtn

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           0.00   33.76    1.28    0.00    0.00   64.97

Device:            tps   Blk_read/s   Blk_wrtn/s   Blk_read   Blk_wrtn

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           0.00   41.81    1.95    0.00    0.00   56.24

Device:            tps   Blk_read/s   Blk_wrtn/s   Blk_read   Blk_wrtn

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
           0.00   31.28    1.14    0.00    0.00   67.58

Device:            tps   Blk_read/s   Blk_wrtn/s   Blk_read   Blk_wrtn


```

tomtor:
@ppigazzini I would expect to see disk devices like:

avg-cpu:  %user   %nice %system %iowait  %steal   %idle
          44,05    0,00    0,20    0,00    0,00   55,75

Device:            tps    kB_read/s    kB_wrtn/s    kB_read    kB_wrtn
nvme0n1           0,00         0,00         0,00          0          0
sda               0,00         0,00         0,00          0          0
sdb               0,00         0,00         0,00          0          0
sdc               0,00         0,00         0,00          0          0

So you should add something like /dev/sda1 or the device mongodb stores its data on

noobpwnftw:
The thing with having a front end proxy like nginx before pserve is still worthwhile even if pserve is multi threaded, for example in a busy job queue of mine, I ran apache to host scripting and yet a local nginx front end would give some performance improvement, the way I see it by doing so is the backend(apache, pserve) can be free of congestion from connection handling(use HTTP 1.1 on nginx upstream to get persistent connections).

From my previous local test I created some real load with my 100 workers on 1 instance of pserve, it used 90% CPU usage in one thread(maybe more after that worker = 500 adjustment), then I switch to nginx + 4 pserve they use 60% CPU each, so 240% in total vs 90%. This is something you can try after DB concurrency issues are solved. And each pserve process would have its own DB connections.

tomtor:
@noobpwnftw Note that multiple pserv processes no longer share the SPRT and SPSA state and will overwrite each others computation in MongoDb without a major rewrite of the current software. Also the handing out of tasks to workers need to be synchronized. This is easy in a single multithreaded pserv, but not in a multi pserv setup with MongoDb. 

Using nginx in itself is a good idea for https and some caching, but using it as a frontend for multiple pserv instances has implications.

noobpwnftw:
Then one pserve instance + nginx it is.
However, I can probably triple the framework throughput from what it has now and I'm willing to do so, but it would almost certainly cause major issues in many way. So we should try to tackle those stuff...

xoto10:
@tomtor Thanks for stress-version.sh 
I want to work the db a bit, so I modified it to get the /tests page, at least that doesn't error and does return the correct result. (Hmmm, I didn't check that it worked ok ... well it makes mongodb show on `top` anyway so I assume it works.) I'm trying running a few copies of Stefan's big tune, but it spends a lot of time compiling so the updates are less frequent than I expected. I hacked the code to do the update_task request 10 times for each update, but it's not enough - I'll try using another machine tomorrow ...

tomtor:
@xoto10 Don't forget that /tests (The complete overview) is cached for 2 seconds. So it will not fetch data from MongoDb, just do the template expansion. Fetching a specific test page is not cached.

tomtor:
@xoto10 I think that a good test approach would be to skip the whole web layer while testing. Why don't we just access rundb.py directly from a test python script (like the create_index) and do mongodb find_run and save_run directly to create a test load and time things?

First step would be to create a few big runs so we have some run IDs to use in the test script.

Next step would be to add threads and test the connection to see if that is a bottleneck.

xoto10:
@tomtor Yes, that would be a good way to test the database throughput.

PR #186, opened 2018-03-23
Speed up request_task logic

tomtor:
by replacing the get_machines() MongoDb query with some Python code.

This is not critical but will reduce the request_task API time from about 700ms to probably < 200ms.

PR #187, opened 2018-03-23
Retire worker reconnect logic

tomtor:
This is open for discussion. If anyone is a fan of the current code I'd like
to hear why it is useful.

When a worker disconnects from fishtest because of errors
and reconnects without being restarted (it gets a new ID
on restart) it is handed it's old job, unless it was
already assigned to another worker.

This situation is rare and the logic requires scanning all
tasks handed out.

https://github.com/glinscott/fishtest/pull/186 increased
the speed by remembering active task for each worker,
but this information is lost on a fishtest reboot.

Benefit from the current code is minimal so retire it.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -269,7 +269,6 @@ class RunDb:
 
   task_time = 0
   task_runs = None
-  worker_task = {}
 
   def request_task(self, worker_info):
     if self.task_semaphore.acquire(False):
@@ -291,22 +290,6 @@ class RunDb:
 
     max_threads = int(worker_info['concurrency'])
     exclusion_list = []
-
-    # Does this worker have a task already?  If so, just hand that back
-    for runt in self.task_runs:
-      key = worker_info['unique_key']
-      if key in self.worker_task and self.worker_task[key] == runt['_id']:
-        run = self.get_run(runt['_id'])
-        task_id = -1
-        for task in run['tasks']:
-          task_id = task_id + 1
-          if task['active'] and task['worker_info'] == worker_info:
-            if task['pending']:
-              return {'run': run, 'task_id': task_id}
-            else:
-              # Don't hand back tasks that have been marked as no longer pending
-              task['active'] = False
-              self.buffer(run, True)
 
     # We need to allocate a new task, but first check we don't have the same
     # machine already running because multiple connections are not allowed.
@@ -353,8 +336,6 @@ class RunDb:
         runt['args']['internal_priority'] = run['args']['internal_priority']
         self.task_runs.sort(key=lambda r: (-r['args']['priority'], -r['args']['internal_priority'], r['_id']))
         break
-
-    self.worker_task[worker_info['unique_key']] = run['_id']
 
     return {'run': run, 'task_id': task_id}
 


Stefano80:
I am not sure of why it is that way. Actually, I also do not understand why batches are assigned to other workers and would retire this as well.

tomtor:
@Stefano80 Partial competed task slots disturb the number of games and priority calculations. So that's the reason batches are reassigned to other workers. For some reason this does not work for spsa like http://tests.stockfishchess.org/tests/view/5ab621870ebc590295d884ef

So we have to manually increase the number of to be played games in that situation. By the way, it would be nice if I got moderator rights on fishtest so I would be able to adjust things if needed and approve tests

ppigazzini:
@tomtor I think that the decision about the approver's rights is on @snicolet, the maintainer of Stockfish.
I will be very glad to update your profile, of course :)

snicolet:
@ppigazzini 

It is of course OK, in my opinion, if all of the fishtest maintenance team members, and more generally all the people who have improved the fishtest server significantly in the last few weeks, get approver's rights for tests in fishtest. 

So you can add at least these ones:
@tomtor 
@Stefano80 
@ppigazzini 
@xoto10 
@ianfab

And while we are at it, I would like to propose to...
@Rocky640
@syzygy1 
@noobpwnftw
@jerrydonaldwatson
... if they would accept to become approvers on fishtest too.

The task is mainly to verify that the proposed tests do not have a security hole for the workers machine, and politely ask the authors to adjust the workload of the server in the 1% of the cases which are not handled automatically.

ppigazzini:
@snicolet OK, I started adding the fishtest approver rights to @tomtor
For other people I wait their acceptance: anyway I need to know their fishtest username.

jerrydonaldwatson:
@snicolet @ppigazzini 
I would be honoured to be an approver. My fishtest username is: jdonald

ppigazzini:
@snicolet added fishtest approver rights to @jerrydonaldwatson

jerrydonaldwatson:
Thank you!

syzygy1:
@snicolet 
I accept to become an approver.

ppigazzini:
@snicolet added fishtest approver rights to @syzygy1 (fishtest username: syzygy)

xoto10:
No thanks, guys, although it's nice of you to offer. (Apologies for being
quiet recently, I have had a high temperature and headaches for the last 6
days. And I still have it.)

It was good to help get fishtest working quicker, although it did start to
feel a bit like work at times.

My current plans are to keep working on stockfish now and then, but this
may change later in the year/or sometime, so I'd rather not add extra roles.

Cheers!

ppigazzini:
@tomtor @Stefano80 is this ready to be merged?

tomtor:
@ppigazzini  Yes, for me it is ready...

ppigazzini:
Official server updated.
Thank you @tomtor

Vizvezdenec:
I think that fishtest needs more approvers from USA/Canada since usually there is a trouble with no patches approval at the time of like 5 am CEST. (like it is now, for example https://i.imgur.com/KahEb9z.png )
Maybe add @AndyGrant ?

PR #188, opened 2018-03-24
Fix periodic flush

tomtor:
Improve long interval between sprt overview updates

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -155,7 +155,7 @@ class RunDb:
       run = self.runs.find_one({'_id': ObjectId(id)})
       if not run:
         run = self.old_runs.find_one({'_id': ObjectId(id)})
-      self.run_cache[id] = { 'rtime': time.time(), 'wtime': time.time(), 'run': run, 'dirty': False }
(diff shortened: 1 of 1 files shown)


tomtor:
@ppigazzini This is ready for a small test. I tested already locally.

ppigazzini:
@tomtor test started on the [dev server](http://94.198.98.239/tests). Use the default user/pwd user01/user01 user00/user00 to insert and approve test runs.
I have updated the [wiki fishtest server setup](https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup) with the latest version of the CentOS/Ubuntu setup scripts.

snicolet:
I see a strange bug on the main server at the moment:

(A) when I load the main page http://tests.stockfishchess.org/tests , my test "iid_tweak2" always show the same number of games (1028)

(B) but when I click on the test to see the detailed statistics, it shows 10768 games : http://tests.stockfishchess.org/tests/view/5ab643cb0ebc590295d88548

(C) I have tried reloading the main page every two seconds during two minutes, but it doesn't show any progress compared to (A)

ppigazzini:
@snicolet I confirm the problem. I updated by hand the official server with this PR and the problem seems solved.
@tomtor let me know when you are done with the test on the dev server: i will commit the PR and updated the official server.

ppigazzini:
official server updated and rebooted.

PR #184, opened 2018-03-19
Create utils/current.py to list current slow operations

xoto10:
@tomtor @ppigazzini This script attempts to list the current operations that are slow (seconds_running > 0.3) and with a known query. Perhaps running this occasionally will give us an idea which operations are slow.
It could probably be enhanced to do `uptime` to show the load averages and maybe other diagnostics?

PR #190, opened 2018-03-24
Update adapters.py

noobpwnftw:
1. The client side HTTP connection pool creates unnecessary load on server because when connection pool is not fully occupied, previously created connections are not reused.
2. We are reporting to server single-threaded, meaning that we'd never use more than one connection ever, pooling more than one of them makes no sense.
3. Make use of internal retry mechanism instead of throwing exceptions right away to reduces chances of getting stuck at 0.00.

### DIFF
--- a/worker/requests/adapters.py
+++ b/worker/requests/adapters.py
@@ -26,7 +26,7 @@ from .cookies import extract_cookies_to_jar
 from .exceptions import ConnectionError, Timeout, SSLError
 
 DEFAULT_POOLSIZE = 1
-DEFAULT_RETRIES = 3
+DEFAULT_RETRIES = 0
 
 
 class BaseAdapter(object):
--- a/worker/requests/adapters.py
+++ b/worker/requests/adapters.py
@@ -25,8 +25,8 @@ from .packages.urllib3.exceptions import HTTPError as _HTTPError
 from .cookies import extract_cookies_to_jar
 from .exceptions import ConnectionError, Timeout, SSLError
 
-DEFAULT_POOLSIZE = 10
-DEFAULT_RETRIES = 0
+DEFAULT_POOLSIZE = 1
+DEFAULT_RETRIES = 3
 
 
 class BaseAdapter(object):


Stefano80:
Can'te we then remove the exeptions throwing logics and save some LOCs?

noobpwnftw:
Some other parts of worker code rely on the exceptions to handle HTTP errors, and we don't want to make it retry forever.

So basically we can allow a few retries internally in the requests and are not exposed to the worker code, then we could save some job re-initialization in case of short network anomalies. And when it really fails after retrying the current behavior is not changed.

tomtor:
Eliminating the pool is good.

The retry is dangerous however because in the case of a timeout fishtest often DID process the request but could not send the reply in time. You can see that because some workers with timeouts still had their game status updated. Workers at 0.00 indeed never had their request processed. 

Retry of the same request will increase the fishtest load for those requests which got serviced but timed out.

noobpwnftw:
In cases when server actually processed the request but didn't reply in time(timeout increased to 10 secs in another PR), then we might need to fix the server to reduce such latency.

It would reduce the chances that the worker resetting to waiting state and issue another resource heavy request_task & building process when recoverable by a retry. In cases that it is not recoverable, for example, a retry would get a different run_id, it is functionally equivalent to having 3 attempts of a 10 second delay loop for a previously exception case which then waits 10~60 seconds and do the same.

It would not create backwards submission problems because it is a synchronous call and single-threaded.

Note that this change is in requests covers more than just fishtest API, including Github downloading, etc.

tomtor:
@xoto10 It would be nice if https://github.com/glinscott/fishtest/pull/184 got adapted with the strategy in the previous comment.

Stefano80:
Could we agree that we first get rid of the pool and postpone using internal retry mechanisms until we have agreed? 

The first one seems to be just for free, the second requires carefully switching on of the community brain.

Stefano80:
By the way, reading [this](https://urllib3.readthedocs.io/en/1.3/managers.html) it becomes difficult for me to understand why we use a PoolManager at all, since we all connect serially first to github, then to fishtest and so on.

Maybe there is a fundamentally easier way to solve this.

noobpwnftw:
Updated PR, only reducing pool size to 1.

Stefano80:
Thanks @noobpwnftw , last small change: please squash all in a single commit, push everything, then I will merge (it is better for reverting and keeping granular history).

PR #191, opened 2018-03-24
Increase timeout on HTTP requests

noobpwnftw:
This should give more time to the server when it is under heavy load.


tomtor:
this overlaps with https://github.com/glinscott/fishtest/pull/162

noobpwnftw:
Integrated elsewhere, closing.

PR #192, opened 2018-03-24
Adjust spsa history limits

tomtor:
Still troubles with big spsa, limit history more

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -571,9 +571,12 @@ class RunDb:
     if len(spsa['params']) < 20:
       freq = 100
       maxlen = 5001
-    else:
+    elif len(spsa['params']) < 50:
       freq = 1000
       maxlen = 201
+    else:
+      freq = 10000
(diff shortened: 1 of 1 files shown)


ppigazzini:
server updated and rebooted

tomtor:
@ppigazzini It's a pity that this will not really reduce the current active spsa run,  allthough it will prevent it from growing much larger. New runs however will be limited. A bit surprised about the difference compared to the test server, which really could handle the load.

ppigazzini:
@tomtor @Stefano80 the difference could be or on the guest side (up to date CentOS and packages) or on the host side (virtuozzo HW/SW still to 2013).
In the first case the server could be rebuilt in 30 minutes (10' to rebuild the VM, 10' to run the setup script, 10' minutes to restore the DB and setup some configs by hand).
In the second case we should think to switch the official fishtest server on a newer infrastructure (e.g. on my ISP), but in this case we need to talk w/ Marco, Joona, Gary (e.g. I don't know who control the DNS).

PR #193, opened 2018-03-24
Clear hist

tomtor:
Sorry, fior this new PR but I think that this is slightly better to deal with the running big test.

Probably a new MongoDb would allow us to relax these restrictions in the future.


tomtor:
Lets first see how the previous turns out...

PR #194, opened 2018-03-25
Retire script active_index.py

ppigazzini:
We got some exceptions in update_task function after
adding the MongoDB index on the official fishtest server.
Index alredy dropped on the official server, now retire the script.
https://github.com/glinscott/fishtest/pull/175


ppigazzini:
@Stefano80 simply merge and restart to clean up the "utils" folder

PR #195, opened 2018-03-25
Drop useless instruction in start.sh

ppigazzini:
There is not a start_fishtest process to be killed


Stefano80:
Looks good to me. If you tested it on the dev server, you can just merge from my side.

ppigazzini:
@Stefano80 server updated and rebooted

PR #197, opened 2018-03-25
Update README.md information

ppigazzini:
Add information about the fishtest workflow,
delete outdated setup instruction,
link the Wiki to get information and setup instruction.


ppigazzini:
@Stefano80 yes, if there are not suggestions.

PR #198, opened 2018-03-26
Retire engine build server code

ppigazzini:
The workers are configured to build themselves the engine binaries.


ppigazzini:
@Stefano80 @tomtor I tested this PR on the dev server and it seems OK, but I'm not so proficient with python, so this needs a double check. Thank you.

tomtor:
@ppigazzini My Python experience is also limited but it looks good to me. If the dev server boots then I am pretty sure everything is OK.

Stefano80:
Hi @ppigazzini, thanks for this awesom reduction of code. It looks good to me, and I think this does not depend on python.

I can remember that a very long time ago, windows workers could not build themselves. This is not anymore the case, right? Just to be safe, did you have some windows workers running for the dev server?

ppigazzini:
@Stefano80 the Windows workers are able to build the engines, there are:
- [Wiki instructions](https://github.com/glinscott/fishtest/wiki/Running-the-worker-on-Windows) to install MinGW-w64 by msys2 (written and maintained by me)
- the [portable worker](https://github.com/glinscott/fishtest/wiki/A-portable-fishtest-worker-for-Windows-64-bit)

I tested the PR on the [dev server](http://94.198.98.239/tests) for some hours with a Linux worker and a Windows worker. 3 test runs: SPSA, SPRT, Fixed, all with very short TC 0.2+0.02

Stefano80:
Merged. Server restarted. Monitored for some minutes until windows machines come back online, seems ok.

PR #199, opened 2018-03-26
Rewrite history logic

tomtor:
Store the history every 'freq' iterations.
More tuned parameters result in a lower update frequency,
so that the required storage (performance) remains constant.

Limit history to 250000 iterations (500000 games).

History is written e.g. every 100 iterations (<= 4 parameters),
every 250 iterations (10 parameters) and every 2500 iterations (100
parameters).

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -564,21 +564,15 @@ class RunDb:
         'c': c,
       })
 
-    # Every 100/1000 iterations, record the parameters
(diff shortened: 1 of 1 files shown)


tomtor:
@Stefano80 Updated to more universal and compacter code.

Stefano80:
Did you test it locally or on dev server? I will merge as soon as I get a yes from you...

tomtor:
@Stefano80 Yes, tested locally.

PR #201, opened 2018-03-27
Resort internal priority after each request_task

tomtor:
This functionality was lost after the MongoDb speedup modifications
which did only recompute the priority once every minute.


tomtor:
@Stefano80 If this_ is the defacto (fishtest/Python) standard for an iterator then that could be better, but then we should change it in other places too in a separate renaming style commit.

Stefano80:
Ok, let us do it later. Merged, restarted (with both of this and .https://github.com/glinscott/fishtest/pull/199)

PR #202, opened 2018-03-27
Update .gitignore

ppigazzini:
Delete the build folder, useless after the retirement of the engine build server.
Add the log of the pserve process.


ppigazzini:
@Stefano80 take a look, should be OK

Stefano80:
Yes is ok

PR #203, opened 2018-03-27
Reduce HTTP connection pool size

noobpwnftw:
The client side HTTP connection pool creates unnecessary load on server because when connection pool is not fully occupied, previously created connections are not reused.
We are reporting to server single-threaded, meaning that we'd never use more than one connection ever, pooling more than one of them makes no sense.


noobpwnftw:
A remake of https://github.com/glinscott/fishtest/pull/190

PR #210, opened 2018-03-29
Fix double listing of recently finished run

tomtor:
Needs testing on the Dev server

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -827,7 +827,8 @@ def tests(request):
         request.rundb.buffer(run, True)
         post_result(run)
 
-    runs[state].append(run)
+    else:
+      runs[state].append(run)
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor dev server ready with the PR code, some workers and a couple of runs (SPRT and Fixed).
user01 to add runs, user00 to approve runs.

ppigazzini:
@tomtor the test on the dev server ran w/o problems, if you think that this PR is ready I will merge.

tomtor:
@ppigazzini Just saw the double listing again so my fix apparantly does not work. I will have a look at the code again.

tomtor:
Ok, I saw the double listing while #211 was in test. So #210 is most likely ok. It is cosmetic so committing is low risk and ok for me.

ppigazzini:
@tomtor official server updated and restarted

PR #211, opened 2018-03-29
Use cache lock

tomtor:
This is a folluwup to

https://github.com/glinscott/fishtest/pull/178

Now that we have a buffer cache we should use that lock
instead of the update_task lock. This should handle
the periodic flush thread and give a small performance boost.

Also add a simple Travis-CI test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 python:
-  - "2.6"
+  - "2.7"
+services:
+  - mongodb
+before_script:
+  - sleep 5
 # command to run tests
 script: 
-  - cd worker && python setup.py test
+  - (cd worker && python setup.py test) && python setup.py test
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -143,6 +143,7 @@ class RunDb:
   # Cache runs
   run_cache = {}
   run_cache_lock = threading.Lock()
+  run_cache_write_lock = threading.Lock()
 
   timer = None
 
@@ -166,7 +167,8 @@ class RunDb:
       id = str(run['_id'])
       if flush:
         self.run_cache[id] = { 'dirty': False, 'rtime': time.time(), 'ftime': time.time(), 'run': run }
-        self.runs.save(run)
+        with self.run_cache_write_lock:
+          self.runs.save(run)
       else:
         if id in self.run_cache:
           ftime = self.run_cache[id]['ftime']
@@ -175,6 +177,8 @@ class RunDb:
         self.run_cache[id] = { 'dirty': True, 'rtime': time.time(), 'ftime': ftime, 'run': run }
 
   def flush_buffers(self):
+    if self.timer is None:
+      return
     with self.run_cache_lock:
       now = time.time()
       old = now + 1
@@ -189,7 +193,8 @@ class RunDb:
       if not oldest is None:
         if int(now) % 60 == 0:
           self.scavenge(self.run_cache[oldest]['run'])
-        self.runs.save(self.run_cache[oldest]['run'])
+        with self.run_cache_write_lock:
+          self.runs.save(self.run_cache[oldest]['run'])
         self.run_cache[oldest]['dirty'] = False
         self.run_cache[oldest]['ftime'] = time.time()
       self.timer = threading.Timer(1.0, self.flush_buffers)
@@ -205,7 +210,7 @@ class RunDb:
     return list(self.get_unfinished_runs()) + self.get_finished_runs()[0]
 
   def get_unfinished_runs(self):
-    with self.run_lock:
+    with self.run_cache_write_lock:
       return self.runs.find({'finished': False},
                           sort=[('last_updated', DESCENDING), ('start_time', DESCENDING)])
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+pyramid
+bson
+pymongo
+scipy
+requests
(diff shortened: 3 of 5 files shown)


ppigazzini:
@tomtor test started on the dev server

ppigazzini:
@tomtor no, this the script used to be sure to make a clean install
```
#!/bin/bash
# to update a fishtest server on CentOS, simply run:
# sudo -H bash update-fishtest.sh | tee update-fishtest.sh.log
#
# to use fishtest open a browser and connect to:
# http://<ip-address>

STAMP=$(date +%Y%m%d%H%M%S --utc -d '1 hour')

systemctl stop fishtest

# download and prepare fishtest
sudo -i -u fishtest << EOF
tar -czvf fishtest.$STAMP.tgz fishtest
rm -rf fishtest
git clone https://github.com/glinscott/fishtest.git

cd fishtest
git config --global user.email "you@example.com"
git config --global user.name "Your Name"

# add here the upstream branch or the PR to be tested
git pull --no-edit origin pull/211/head

# pull some PRs mandatory for an up to date installation
# comment out if a PR will be committed upstream
# https://github.com/glinscott/fishtest/pull/43
# https://github.com/glinscott/fishtest/pull/106
git pull --no-edit origin pull/43/head
git pull --no-edit origin pull/106/head
EOF

# setup fishtest
cd /home/fishtest/fishtest/fishtest
python setup.py develop

echo "repo sync finished, start the server with:"
echo "sudo systemctl start fishtest"
```

tomtor:
@Stefano80 Sure, it will be nice to have some automated tests. I will write some code which exercises some of the APIs. Might take a while before I have something sufficient, but this patch is not urgent.

tomtor:
@Stefano80 @ppigazzini On second thought, how do we deal with the required MongoDb interaction? Is there some standard MongoDb stub setup?  Not sure how to proceed to get a minimal useful test.

Stefano80:
This is a good question, in Python 3 we could use the mock framework, here we have to think about it.

tomtor:
@Stefano80 There are some samples, I have just added the MongoDb service.

- We should add some users
- Create a test run with user A
- Approve with user B
- Execute request_task
- Simulate a worker calling update_task until "task no longer needed"
- Check the main page for the end result

ppigazzini:
@tomtor @Stefano80 I'm an absolute beginner but what about using [docker w/ Travis](https://docs.travis-ci.com/user/docker/) to setup a fishtest server and a fishtest worker ?

Stefano80:
Hi, I am also a beginner here. Whatever solution we choose, it would be cool if we would still be able to test locally using `python setup.py test` because this really speed up working.

Stefano80:
Ah, I just recognized that the test in the root directory is not actually executed. If I try to execute I again have the problem that it does not find the requests module (no matter what I do), as can be seen here https://travis-ci.org/Stefano80/fishtest/builds

Stefano80:
I fixed it, see https://github.com/glinscott/fishtest/pull/213

Stefano80:
Hi @tomtor , if you want you test to run, you have to specify it in the setup too.

tomtor:
@Stefano80 The travis test now runs OK. I did merge an old version of your #213. 

I added a test iterator, I think we can merge this one and add your later changes to #213 in a separate merge?

Stefano80:
Did you try to start the worker locally? I had problems with adding the
requests package and had to separate the worker has an own application

Il sab 31 mar 2018, 17:53 Tom Vijlbrief <notifications@github.com> ha
scritto:

tomtor:
@Stefano80 I have only ran your old version of this morning, that ran OK. See also the travis report from this PR. Note that I need Python 2.7 to run these tests, Python 2.6 fails.

Stefano80:
I had the effect that the tests would run but locally would fail, and could only reproduce it by adding a test running the worker has a script using subprocess

Stefano80:
I will give it a go later today.

Stefano80:
I mean running the worker locally, not the tests.

Stefano80:
So Tom, I tried out. If I just checkout your patch and try to run the worker locally, I get the following error
```
  File "worker.py", line 15, in <module>
    from games import run_games
  File "/home/stefano/Documents/stefano/programming/chess/fishtest_my/worker/games.py", line 8, in <module>
    from . import requests
ValueError: Attempted relative import in non-package
```
This is exactly the problem what plagued me in the last two days. The problem is that, seen from root, currently fishtest is configured both a module, a package and an application which can be run. The only solution I found was to consider the worker as an application, which means move `setup.py` in the worker directory, remove `__init__.py` from that directory and say .travis to use `cd worker && python setup.py test` as the test script.

In the end, it also makes sense, since fishtest really consists of two components: the worker and the server. These are not just two modules but two different standalone applications.

Hope this helps you.

I think this discussion is also of interest for @ppigazzini .

tomtor:
@ppigazzini @Stefano80 Ok, so we can simply use:

    (cd worker && python setup.py test) && python setup.py test

to run both?

Stefano80:
Probably, I don't know how idiomatic this would be.

tomtor:
@ppigazzini @Stefano80 Ok, I merged #214 into this PR and invoke both setup.py scripts now. Looks good to me, so I think this is ready too. Note that I use Python 2.7 and not 2.6, but we will want to move to 2.7 soon, so that should be OK for the Travis tests.

ppigazzini:
@tomtor @Stefano80 tested successfully w/ dev-server and windows, linux, WSL workers.
Official server updated, fishtest service restarted.

PR #213, opened 2018-03-30
Specify test suite and fix some imports for that

Stefano80:
Now I was able to fix the fact the only test was not actually running.

Since I did some change in the production code (the syntax of two imports is slightly changed), it would be cool if someone could quickly test on dev server, alternatively I'll do it tomorrow.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,8 @@
 from setuptools import setup, find_packages
 
 setup(
-    name="fishtest",
-    version="0.1",
-    packages=find_packages()
+    name = "fishtest",
+    version = "0.1",
+    packages = find_packages(),
+    test_suite = "test_worker"
 )
--- a/test_worker.py
+++ b/test_worker.py
@@ -3,6 +3,7 @@ import worker
 import worker.worker
 import os
 import os.path
+import subprocess
 
 class workerTest(unittest.TestCase):
 
@@ -10,7 +11,6 @@ class workerTest(unittest.TestCase):
     if os.path.exists('foo.txt'):
       os.remove('foo.txt')
    
-    
   def test_config_setup(self):      
     config = worker.worker.setup_config_file('foo.txt')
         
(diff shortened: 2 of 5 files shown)


ppigazzini:
@Stefano80 dev server updated, linux & windows workers updated

* worker: linux & windows

errore message:
```
  File "fishtest-master/worker/worker.py", line 15, in <module>
    from games import run_games
  File "/home/fishtest/fishtest-master/worker/games.py", line 8, in <module>
    from . import requests
ValueError: Attempted relative import in non-package
```

* server
The dev server works correctly with the workers form master, the `setup.py` change should be fine.

Stefano80:
Ok, I can reproduce the error locally. I am trying to understand what the real problem is here.

Stefano80:
So @ppigazzini @tomtor , I am doing some slow progress (and learn a lot in the process).

Now the worker it is turned in a true package, such that at least it works locally and the tests are correctly performed.

What I still have to do is in the test_worker to catch the fact that I must have [username] [password] in the standard output. 

Any easy to implement idea would be appreciated, and if @ppigazzini could check whether this would now work on a fresh install I would be immensely grateful.

Stefano80:
Ok, this is done from my side. I now also check that the worker can be correctly started and asks for username and password.

My only last doubt is whether it works with a fresh install and whether we have to update the worker version.

ppigazzini:
@Stefano80 fresh install of the dev server with `update-fishtest.sh`, fresh install of the worker with
```
#!/bin/bash
cd fishtest
rm -rf worker
git fetch upstream
git pull  --no-edit upstream pull/213/head
cd worker
python worker.py --host 94.198.98.239 --concurrency 3 user01 user01
```

Error message on Windows worker and WSL worker

```
Worker version 62 connecting to http://94.198.98.239:80
Will fetch task soon...

Exception running games:
  File "/root/_git/fishtest/worker/games.py", line 375, in run_games
    setup_engine(new_engine, worker_dir, run['args']['resolved_new'], repo_url, worker_info['concurrency'])
  File "/root/_git/fishtest/worker/games.py", line 115, in setup_engine
    f.write(requests.get(github_api(repo_url) + '/zipball/' + sha, timeout=HTTP_TIMEOUT).content)
NameError: global name 'requests' is not defined
Task exited
```

Stefano80:
Ah, interesting thank you. This means I am making progresses but I am not 100% there. Thank you, by the way, for this kind of worker tests the dev server is not needed, it is possible to do it locally, I would have just needed to wait some more time.

Stefano80:
New PR is a better version of that.

PR #215, opened 2018-04-02
Add hyperlink to hardy.uhasselt.be/Toga/live_elo

tomtor:
and a proper stop() to rundb.

Also some minor tweaks to the Travis test.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -176,10 +176,15 @@ class RunDb:
           ftime = time.time()
         self.run_cache[id] = { 'dirty': True, 'rtime': time.time(), 'ftime': ftime, 'run': run }
 
-  def flush_buffers(self):
-    if self.timer is None:
-      return
+  def stop(self):
     with self.run_cache_lock:
+      self.timer = None
+    time.sleep(1.1)
+
+  def flush_buffers(self):
+    with self.run_cache_lock:
+      if self.timer is None:
+        return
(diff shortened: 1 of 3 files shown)


ppigazzini:
@tomtor updated dev-server and workers, test started.

tomtor:
@ppigazzini Thanks!

tomtor:
@Stefano80 It deletes (x) the run. The point is that it should be the last test run. test_delete_run was run as second test. It is executed in alphabetic order. An alternative is prepending 00, 01, 02 to the test name?

tomtor:
@Stefano80 Added 10, 20, 30 as prefix, as is done in other scripts where order is important...

Stefano80:
Ah ok. It is specified in unit test that the order is important? I would actually solve this by writing helper methods and call them in the desired order in the test methods.

You can also make use of the setup methods in the class.

tomtor:
@Stefano80 Not sure how to rewrite it, please post the setup class code you prefer

Stefano80:
For me it's ok like it is now, if in python unit test it is specified that the order is alphabetical. In the next days I will see whether I can improve on this.

Stefano80:
Commited, restarted service. Thank you @tomtor , this is a simple yet effective improvement for UX.

tomtor:
It was a feature request by @snicolet a while ago

snicolet:
Supercool :-)

PR #219, opened 2018-04-05
Limit request_task concurrency

tomtor:
A request_task call takes about 150ms (7 per second).

The worker randomizes calls in a 1 to 30s interval.

With 200 workers or more the request queue might consume
all threads, so limit concurrency to 4 requests.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -266,14 +266,20 @@ class RunDb:
 
   # Limit concurrent request_task
   task_lock = threading.Lock()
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor the dev-server and the workers are running fine since 4 hours. I'm waiting the @Stefano80 review before merging this PR.

tomtor:
@Stefano80 renamed

Stefano80:
Thanks. For me it's ok.

Stefano80:
Restarted.

PR #220, opened 2018-04-06
Clean up and add new links to menu and more task info

tomtor:
Remove some old code with hardwired usernames and old
"Regression link" and add

- Login link
- Slow (LTC) query
- System info to tasks in test view


Stefano80:
@tomtor I am not acquainted with this part of the server code, so I have a dumb question: is there also code for the regression page that can be removed, beside the link?

tomtor:
@Stefano80 @snicolet @ppigazzini There is a lot of Regression code in views.py but that is not related to this specific link. Perhaps we can remove that code too, not sure about it.

tomtor:
@snicolet LTC tests are matched with regexp [4-9][0-9], that's OK with you? VLTC (eg 300) will not be matched. I'm open for other regexp suggestions.

ppigazzini:
@tomtor @Stefano80 @snicolet [this](http://35.161.250.236:6543/regression) is how Fabian is reporting the regression tests for Multi Variant SF (there are also the data for standard chess). Pragmatically: very big error bars, but the gain trend is showed correctly.

tomtor:
@Stefano80 Deleted old regression code

Stefano80:
Hi,

the page from Fabian is well-done, as usual, thx to @tomtor to dig into the code. I see two options now:
1) we leave the obsolete page as it is now and we get it live again. I would keep only the regression stuff and leave all other information away.
2) we remove and we build it up again from scratch. We should collect the statistics from the database, in some automatic way.

Any thoughts?

vondele:
I wonder if we should also remove the 'chat' link, it certainly doesn't seem to be valuable in the current state.

ppigazzini:
@Stefano80 there is a 'stats' group, perhaps is used to control the user rights for the regression data.

tomtor:
@Stefano80 I already removed the old code (https://github.com/glinscott/fishtest/pull/220/commits/c89abb7c0a25e3fd9e35ed11512d1f884a2c46e0) and suggest simply adding a link to Fabians page? A fancier approach should be in a separate PR

Stefano80:
I think this is ok now. Let me give it a look in more detail, and then we can merge.

tomtor:
@Stefano80 Oops, missed this one https://github.com/glinscott/fishtest/pull/220/commits/38f63b74958fc87510b9c7bcab320c48a15993ac

ppigazzini:
@tomtor @Stefano80 If we want to link an external page then I suggest to use [nextchessmove](https://nextchessmove.com/dev-builds)

I introduced the Fabian's regression page only to ask if makes sense to compute internally some regression data with relaxed error bars to save some framework CPU.

snicolet:
I suggest to link to a wiki page, where we can add link much more easily than via a pull request for the server.

Apart from, that, thanks for working this, everybody, it will improve the user interface and give developers more introspection tools for dat-mining the past patches!

Stefano80:
I think @snicolet is right, we should add a wiki page on Stockfish repository with resources like ccrl, nextchessmove etc so we don't have to make a PR every time a new page pops up. I propose to do in a separate PR.

Proposal: Tom, could you remove again the link to Fabian page and squash everything if you're done? Then we can look at it on the dev server and then merge.

tomtor:
@Stefano80 Squashed and add link to Wiki https://github.com/glinscott/fishtest/wiki/Regression-Tests

ppigazzini:
@tomtor I just merged https://github.com/glinscott/fishtest/pull/221 and this has created a conflict with this PR. Sorry, next time I will try to follow the presentation order.

tomtor:
@ppigazzini No problem, rebased....

Stefano80:
Hi @tomtor , thank you for the wiki page, I added links to SPCC and the CCRL pages for Stockfish.

ppigazzini:
@tomtor @Stefano80 test started on the dev server.
I updated the wiki [setup scripts](https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup) to install the python application as user fishtest (this solve also the ownership of the egg folder):
```
python setup.py develop --user
```

DragonMist:
Just a note, I've contacted Chad of NextChessMove, and he will add my graph data on SF progression tests to his graph, will be nice to see both at the same time and compare them.

ppigazzini:
@tomtor @Stefano80 official server updated and fishtest service restarted.

tomtor:
I see that the task summary lists a bit more workers then are really active. Will have a look at that.

fishtest/fishtest/rundb.py
```diff
@@ -118,8 +119,7 @@ def get_approval(sha):
       return self.runs.find_one(q)
     base_approval = get_approval(resolved_base)
     new_approval = get_approval(resolved_new)
-    allow_auto = username in ['mcostalba', 'jkiiski', 'glinscott', 'lbraesch']
```
tomtor: @Stefano80 Old code for auto approvers, regarding #229

fishtest/fishtest/views.py
```diff
@@ -378,6 +331,7 @@ def tests_run(request):
   if 'base-branch' in request.POST:
     try:
       data = validate_form(request)
+      data['auto_approve'] = has_permission('approve_run', request.context, request)
```
tomtor: @Stefano80 New code for auto approval (#229 )

ppigazzini:
@tomtor @Stefano80 @snicolet IMO there should not be a self-approve right.

tomtor:
Well, it is not self approve.  The code was already approved by someone earlier and it is resubmitted, eg for LTC. Self approving has always been possible by creating a second account which you use to enter the test...

tomtor:
If we really value security we should create new users blocked and let them be unblocked by approvers. The eMail adres can currently also be left blank....

tomtor:
@ppigazzini I initially considered removing the existing code completely, but wanted some feedback first. 

Apparently most approvers did not notice the new situation or are happy about it.

In my opinion we could auto approve for **everyone**, in the early days of fishtest it was that way. It makes no sense to require reapproval for eg an LTC run (how is that different from changing eg the priority?), but perhaps I am missing something. On the other hand, LTC resubmittals are rare, so the gain of auto approval is minimal.

ppigazzini:
In this [fishcooking thread](https://groups.google.com/forum/#!topic/fishcooking/TZDYHEW61Pw) was advanced a concern about the security of the LTC auto approve feature for passed STC tests.

tomtor:
https://github.com/glinscott/fishtest/commit/07af327e68bb74eb8996644471c818463b09efaf 
was the procedure when I registered....

Stefano80:
Hi @tomtor : I do not have anything against auto-reapproval of LTC, but I noticed that my STC was autoapproved. I do not have a good insight in the code, but I somehow suspect that whether it is a resubmission or not is only checked based on the branch name. This is bad, since many people (for example I) make different attempts with the same branch.

Of course fishtest is not very secure, but this is a really bad hole. I am of course not criticizing you for anything, I actually very happy that you exposed that before something bad happened!

Personally: I would completely remove auto-approval.

vondele:
I also think there is no need for auto-approval, there is now a rather active community of approvers.

tomtor:
@Stefano80 According to the code it searches for the sha, but anyway, I have no problem with removing this old code. I will write a PR.

PR #221, opened 2018-04-06
Refactoring of Travis script

Stefano80:
This is a refactoring of the setups, folder structure and travis script to enhance modularity of the worker/server construct.

If we merge this, we have two separate applications, the worker and the server, which are tested separately in a common environment. Eventually, we could even think of separating the two applications completely.


tomtor:
@Stefano80 Looks good to me

ppigazzini:
@Stefano80 I'm syncing the DB on the dev-server, then I will start a fast test for this PR (just to be safe) and more longer for the other waiting PRs

ppigazzini:
official server updated and fishtest service restarted

ppigazzini:
@Stefano80 @tomtor I need support from you python experience.

After merging this PR on the dev server that there are now two egg.info folders:
* `/home/fishtest/fishtest/fishtest/fishtest.egg-info` cloned by git, not created by setup. If I delete this folder the fishtest server does not start anymore, no hints from the logs
* `/home/fishtest/fishtest/fishtest/fishtest_server.egg-info` created after running setup.py with this PR. The folder and the folder content have root as owner (I updated my scripts to change the ownership)

I think that the old folder should be deleted but there are some reference elsewhere that should be updated.

tomtor:
@ppigazzini The issue is in production.ini:

Change at the top to:
    [app:main]
    use = egg:fishtest_server

tomtor:
@ppigazzini  @Stefano80 Yes, if you do not run the setup.py on the official server then renaming fishtest.egg-info should probably work AND be needed. We should however remove the old fishtest.egg-info from the git repository, because it will no longer be used in a normal install where setup.py is run.

tomtor:
@Stefano80 @ppigazzini And we can also remove requirements.txt from the root, because that is included in the new setup.py. So should I adapt #220 with these changes?

test_server.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+cd fishtest
+python setup.py test
+cd ..
```
tomtor: The
    cd ..
at the end of the scripts can be removed

Stefano80:
Hi, Tom, Pasquale, thank you for the fast troubleshooting.

Tom if you could include this in the PR it would be great

ppigazzini:
@tomtor I make the changes by hand on the production server and all is running fine.
To me is OK to introduce these changes in https://github.com/glinscott/fishtest/pull/220

test_server.sh
ppigazzini: @tomtor I agree

tomtor:
@ppigazzini ok, .gitignore updated

Issue #218, opened 2018-04-04
Worker stuck in build loop when someone deletes the Github fork after tasks assigned


noobpwnftw:
It would then not cleaning up the tmp folder and will eventually fill up the disk.

A try... finally... block should be added to clean up intermediate files upon failure.

Stefano80:
Has being worked on in https://github.com/glinscott/fishtest/pull/222

PR #222, opened 2018-04-06
Add exception handling of download & build process

noobpwnftw:
In case of an exception occurred during the download & build process, add a catch block so that temp directory cleanup is not skipped.

This also gives an opportunity to retry a few times instead of going through the expensive request_task process in case of a recoverable failure.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -110,31 +110,35 @@ def setup_engine(destination, worker_dir, sha, repo_url, concurrency):
   if os.path.exists(destination): os.remove(destination)
   """Download and build sources in a temporary directory then move exe to destination"""
   tmp_dir = tempfile.mkdtemp()
-  os.chdir(tmp_dir)
+  
+  try:
+    os.chdir(tmp_dir)
+    with open('sf.gz', 'wb+') as f:
+      f.write(requests.get(github_api(repo_url) + '/zipball/' + sha, timeout=HTTP_TIMEOUT).content)
+    zip_file = ZipFile('sf.gz')
+    zip_file.extractall()
+    zip_file.close()
 
-  with open('sf.gz', 'wb+') as f:
-    f.write(requests.get(github_api(repo_url) + '/zipball/' + sha, timeout=HTTP_TIMEOUT).content)
-  zip_file = ZipFile('sf.gz')
-  zip_file.extractall()
-  zip_file.close()
+    for name in zip_file.namelist():
+      if name.endswith('/src/'):
+        src_dir = name
+    os.chdir(src_dir)
 
-  for name in zip_file.namelist():
-    if name.endswith('/src/'):
-      src_dir = name
-  os.chdir(src_dir)
+    custom_make = os.path.join(worker_dir, 'custom_make.txt')
+    if os.path.exists(custom_make):
+      with open(custom_make, 'r') as m:
+        make_cmd = m.read().strip()
+      subprocess.check_call(make_cmd, shell=True)
+    else:
+      subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency), shell=True)
 
-  custom_make = os.path.join(worker_dir, 'custom_make.txt')
-  if os.path.exists(custom_make):
-    with open(custom_make, 'r') as m:
-      make_cmd = m.read().strip()
-    subprocess.check_call(make_cmd, shell=True)
-  else:
-    subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency), shell=True)
-
-  shutil.move('stockfish'+ EXE_SUFFIX, destination)
-  os.chdir(worker_dir)
-  shutil.rmtree(tmp_dir)
-
+    shutil.move('stockfish'+ EXE_SUFFIX, destination)
+  except:
+    raise Exception('Failed to setup engine for %s' % (sha))
+  finally:
+    os.chdir(worker_dir)
+    shutil.rmtree(tmp_dir)
+    
 def kill_process(p):
   if IS_WINDOWS:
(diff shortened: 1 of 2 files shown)


noobpwnftw:
Should fix #218

Stefano80:
Ok, I have to take some time on this. Thank you!

Stefano80:
Hi @noobpwnftw , I assume that this is a fix for https://github.com/glinscott/fishtest/issues/218, right?

If I understand it correctly, you were seeing problems in `setup_engine` on your machines, due to the fact that somebody deleted the fork. Is this what happened?

For the rest, it looks ok, I have one question: why do you want to retry 3 times? Should not be enough to just do not throw any exceptions?

noobpwnftw:
@Stefano80 It is there because it is overall cheaper to retry here, if one of the two downloads failed from network issues.

Stefano80:
@ppigazzini : so it looks it does not work yet?
@noobpwnftw : I see, can we first make the try / except construct change alone? Otherwise it is difficult to judge whether we are progressing or not.

noobpwnftw:
@ppigazzini Maybe delete the entire forked repo? I don't know how to get there but I got full lot of tmpXXXX in the /tmp folder and those disappeared from active worker list.

@Stefano80 done.

ppigazzini:
@noobpwnftw @Stefano80 
* when deleting the whole forked repo I got this error message
```
Worker version 62 connecting to http://94.198.98.239:80
Will fetch task soon...
Downloading 2moves_v1.pgn ...
Downloading cutechess-cli-linux-64bit.zip ...

Exception running games:
  File "/root/_git/fishtest/worker/games.py", line 406, in run_games
    base_nps = verify_signature(new_engine, run['args']['new_signature'], remote, result, games_concurrency * threads)
  File "/root/_git/fishtest/worker/games.py", line 65, in verify_signature
    busy_process = subprocess.Popen([engine], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
Task exited
``` 

* when deleting the branch the workers are able to download the source code: GitHub has the option to restore a deleted branch, so I suppose that the branch reference is always valid. To be sure I pushed a branch changing a piece value in order to have different base_signature & test_signature.

ATM I can only test the code for regression, open to suggestions

noobpwnftw:
@ppigazzini Prior to this patch the worker would left in the /tmp folder an empty directory or a 0 byte sf.zip file. In my case it ate disk(or filesystem) capacity leading to problems.

After patch it should not leave anything in that folder upon failure.

ppigazzini:
@noobpwnftw I'm simply reporting that I'm not able to test some functions of the patch. I'm not so proficient in python to approve the code, so I left this task to @Stefano80

worker/games.py
```diff
@@ -110,28 +110,32 @@ def setup_engine(destination, worker_dir, sha, repo_url, concurrency):
...
+    if os.path.exists(custom_make):
+      with open(custom_make, 'r') as m:
+        make_cmd = m.read().strip()
+      subprocess.check_call(make_cmd, shell=True)
+    else:
+      subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency), shell=True)
+
+    shutil.move('stockfish'+ EXE_SUFFIX, destination)
+  except:
+    pass
```
Stefano80: Here, if you just pass, later it will fail with some strange exceptions in `run_games` or `verify_signature`. Better, it would be to use the same construct as in `verify_signature` and say the API to stop the run, I past the code from `verify_signature` as an example 
```
    if int(bench_sig) != int(signature):
      message = 'Wrong bench in %s Expected: %s Got: %s' % (os.path.basename(engine), signature, bench_sig)
      payload['message'] = message
      requests.post(remote + '/api/stop_run', data=json.dumps(payload), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT)
      raise Exception(message)
```

noobpwnftw:
Did 1 & 3.
I have no clue how to write 2.

Stefano80:
Hi @noobpwnftw , now I have gone through your patch and did two things

1) some proposal for minor changes in `games.py`
2) write a small test

You can find everything in the commit https://github.com/Stefano80/fishtest/commit/5d048775b04e473689d9d0ee2e427f9b5f8b7806 

Could you give it a look and include the changes or build up on them?

noobpwnftw:
I've learned some Python now: your code raised an exception in the except block, which I was trying to avoid under the impression that doing so would have the finally block skipped.
I think your patch is fine and please just do PR on that instead of mine.

Stefano80:
Hi @noobpwnftw , I would prefer you change this one if it is not too much work:
1) it gives you proper credit
2) it keeps track of the discussion
So, if you have time, merge my proposal, squash everything and I will merge.

Stefano80:
Thx! Could you please squash all commits into 1? This can be done by git rebase master -i and choosing squash for all commits apart of the first one. Yes, and then you have to force the push.

@ppigazzini, this is dive

noobpwnftw:
One moment, I try to will figure this out.

ppigazzini:
@Stefano80 test started

Stefano80:
Looks like is running fine, I will merge later today.

ppigazzini:
@Stefano80 fine also to me. This worker patch should be merged with others worker PRs already tested and pushed to all active workers changing the version number.

A conflict is blocking the PR https://github.com/glinscott/fishtest/pull/162: that can be solved only by @xoto10 ?

PR #224, opened 2018-04-08
Add a few pyramid tests and some minor fixes

tomtor:
- Add test iterator for server tests
- test /signup and /login
- Fix Tasks worker summary spelling: with 1 coreS
- prevent creating a user with an existing name
- update development.ini

### DIFF
--- a/fishtest/tests/test_rundb.py
+++ b/fishtest/tests/test_rundb.py
@@ -3,17 +3,19 @@ import datetime
 
 from fishtest.rundb import RunDb
 
+import util
+
 rundb = None
 run_id_stc = None
 run_id = None
   
-class CreateServerTest(unittest.TestCase):
+class CreateRunDBTest(unittest.TestCase):
 
   def tearDown(self):
     # Shutdown flush thread:
     rundb.stop()
 
-  def test_10_create_run(self):
+  def test_10_create_run(self): 
     global rundb, run_id, run_id_stc
     rundb= RunDb()
     # STC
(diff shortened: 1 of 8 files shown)


ppigazzini:
The [dev-server](http://94.198.98.239/tests) and the workers are running fine since 3 hours.
I tested the registration of:
* an already registered user: OK (not permitted)
* a new user: OK (permitted)

@Stefano80 to me the merge is OK

fishtest/tests/test_rundb.py
```diff
@@ -49,16 +53,15 @@ def test_40_list_LTC(self):
...
-      w['pending'] = False
-    rundb.buffer(run, True)
+    for run in rundb.get_runs():
+      if run['args']['username'] == 'travis' and not 'deleted' in run:
+        print('del ')
+        run['deleted'] = True
+        run['finished'] = True
+        for w in run['tasks']:
+          w['pending'] = False
+        rundb.buffer(run, True)
```
Stefano80: Should'nt we assert that we have no runs running anymore?

fishtest/tests/util.py
```diff
@@ -0,0 +1,7 @@
+from fishtest.rundb import RunDb
+
+def find_run():
```
Stefano80: Maybe we could rename this `find_travis_run` or even have a function `find_user_run(userName)` which is called with `travis` as an argument.

fishtest/tests/test_users.py
```diff
@@ -0,0 +1,57 @@
+import unittest
```
Stefano80: Thanks for testing, this is the spirit! Awesome!

fishtest/tests/util.py
tomtor: Good suggestion, will change that

fishtest/tests/test_rundb.py
tomtor: update_task will check this, so that should never be a problem

ppigazzini:
Started the test w/ the dev-server

tomtor:
@ppigazzini Thanks, but the only change was in the Travis tests, there should be no change in the actual running code since your test of yesterday. I will squash the commits.

ppigazzini:
server updated, fishtest service restarted.

PR #225, opened 2018-04-09
SLightly wider column

vondele:
on the main page, colored boxes containing LLR results can overflow if the number of tests exceeds 100000, causing the numbers to be split on two lines.

This patch (unfortunately untested), widens the column by a bit to give space for 2 more digits.

Would need somebody to have a look at the result on the dev server if the result is fine.

### DIFF
--- a/fishtest/fishtest/templates/base.mak
+++ b/fishtest/fishtest/templates/base.mak
@@ -42,8 +42,8 @@
           <li><a href="https://github.com/glinscott/fishtest" target="_blank">Github</a></li>
           <li><a href="https://groups.google.com/forum/?fromgroups=#!forum/fishcooking" target="_blank">Forum</a></li>
           <li><a href="https://groups.google.com/forum/?fromgroups=#!forum/fishcooking_results" target="_blank">History</a></li>
-          <li><a href="https://hxim.github.io/Stockfish-Evaluation-Guide/" target="_blank">EvalGuide</a></li>
-	  <li><a href="https://github.com/glinscott/fishtest/wiki/Regression-Tests" target="_blank">Regression</a></li>
+          <li><a href="https://hxim.github.io/Stockfish-Evaluation-Guide/" target="_blank">Eval&shy;Guide</a></li>
+	  <li><a href="https://github.com/glinscott/fishtest/wiki/Regression-Tests" target="_blank">Regres&shy;sion</a></li>
           <li><a href="http://abrok.eu/stockfish/" target="_blank">Compiles</a></li>
           <li><a href="https://github.com/official-stockfish/Stockfish" target="_blank">SF-github</a></li>
           <li class="nav-header">Admin</li>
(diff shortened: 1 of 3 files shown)


tomtor:
Regarding cosmetics, the Regression link text (in base.mak) is too wide on my mobile phone.

Could you change it to Regres&shy;sion so that it can be tested together with your change?

vondele:
done 2x.

ppigazzini:
@vondele [dev-server](http://94.198.98.239/tests) updated. I started a run fixed 10000000 games @ ridiculous TC of 0.05+0.005. You can view anyway all the historic runs of fishtest.

tomtor:
The soft hyphens look/work great on my phone (thanks).

I find the colored boxes a bit wide now but I guess i will get used to that.

tomtor:
http://94.198.98.239/tests/view/5ac883a9e44b06447afa3d3b is still broken

vondele:
yes, it is a bit too wide. I don't know how to get the proper estimate. Currently on page 18, there is the test with 219664 games http://94.198.98.239/tests/view/5ab6abea0ebc5902932cbcfa it suggests I could remove 5 more pixels.

@tomtor, broken in which sense ? Looks normal to me.

ppigazzini:
server updated

tomtor:
The box on top of the test page on my phone:

    Total: 35480 W: 12109 L: 12035 D: 113
    36

vondele:
I see, the updated version looks OK on firefox, but not on chrome. I guess the same issue might be playing on phone vs. desktop.. I guess it is just not OK to specify in pixels as the needed width will be non-standard. I've pushed another version with 41 characters... maybe this works ? Sorry for the many versions, I'm not too experienced with html.

tomtor:
Line 17 at tests_view.mak should be changed, eg to:

    <div class="span5">

or add a similar min-width?

vondele:
@ppigazzini ~~good design is always a matter of taste~~ ugh, no, I'm afraid I have to pass.

tomtor:
@vondele See my editted comment 3 up. Note that you could also save the HTML page and edit/view it on your computer locally or use some advanced stuff (Shift-CTRL-I in eg chrome).

vondele:
Thanks for the help @tomtor ... I give it a last try. After that it seems it is best done by somebody who knows...

ppigazzini:
server updated

vondele:
OK, I've reverted back to 305px, which improves on the current version (285px), seems to avoid wrapping on the desktop, but doesn't fix things on all devices.

tomtor:
@vondele could you or @ppigazzini please also change line 17 of tests_view.mak from span4 to span6

I think then this PR is fine.

vondele:
@tomtor done.. @ppigazzini it would be good to have a final view of the result on the dev server, just to be sure.

ppigazzini:
server updated

Stefano80:
On my phone the coloured box is still one digit short.

tomtor:
Looks good to me for up to 200000 games (the 10000000 @ 0.05+0.005 run is still split). Wider fixed width is not very elegant so I propose to merge this PR. I have an alternative variable width for a separate PR.

Stefano80:
Ok.

ppigazzini:
official server updated and fishtest service restarted

PR #226, opened 2018-04-09
Various changes to games result box

tomtor:
Followup to https://github.com/glinscott/fishtest/pull/225

- Make the box variable size and as small as possible without overflow
- Add some more info to Pending runs (SPRT bounds)
- Link to Live ELO page for SPRT runs

### DIFF
--- a/fishtest/fishtest/templates/base.mak
+++ b/fishtest/fishtest/templates/base.mak
@@ -3,6 +3,7 @@
 <head>
   <title>Stockfish Testing Framework</title>
   <link href="/css/bootstrap.min.css" rel="stylesheet">
+  <%block name="meta"/>
 
   <script src="http://code.jquery.com/jquery-1.8.3.js"></script>
   <script src="/js/bootstrap.min.js"></script>
--- a/fishtest/fishtest/templates/elo_results.mak
+++ b/fishtest/fishtest/templates/elo_results.mak
@@ -1,4 +1,4 @@
-<%page args="run"/>
+<%page args="run, show_gauge=False"/>
 
 <%
   def get_run_style(run):
@@ -7,4 +7,24 @@
     return ''
 %>
 
-<pre style="${get_run_style(run)};" class="elo-results">${'\n'.join(run['results_info']['info'])}</pre>
+%if 'sprt' in run['args'] and not 'Pending' in run['results_info']['info'][0]:
+<a href="${'/html/live_elo.html?' + str(run['_id'])}" style="text-decoration:none">
+%endif
+%if show_gauge:
+<div id="chart_div_${str(run['_id'])}" style="width:90px;float:left;"></div>
+%if 'sprt' in run['args'] and not 'Pending' in run['results_info']['info'][0]:
+<div style="margin-left:90px;padding: 30px 0;">
+%else:
+<div style="margin-left:90px;">
+%endif
+%endif
+<pre style="${get_run_style(run)};white-space:nowrap;" class="elo-results">
+${run['results_info']['info'][0]}<br/>
+${run['results_info']['info'][1] if len(run['results_info']['info']) > 1 else ''}
+</pre>
+%if show_gauge:
+</div>
+%endif
+%if 'sprt' in run['args'] and not 'Pending' in run['results_info']['info'][0]:
+</a>
+%endif
--- a/fishtest/fishtest/templates/run_table.mak
+++ b/fishtest/fishtest/templates/run_table.mak
@@ -1,13 +1,45 @@
-<%page args="runs, show_delete=False"/>
+<%page args="runs, show_delete=False, active=False"/>
 
 <%namespace name="base" file="base.mak"/>
+
+%if active:
+<script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
+<script type="text/javascript">
+      google.charts.load('current', {'packages':['gauge']});
+      google.charts.setOnLoadCallback(drawCharts);
+
+      function drawCharts() {
+
+        var a= -2.94, b= 2.94;
+        var options = {
+          redFrom: a, redTo: 0,
+          greenFrom: 0, greenTo: b,
+          min: a, max: b,
+          minorTicks: 5
+        };
+        
+        %for run in runs:
+          %if 'sprt' in run['args']:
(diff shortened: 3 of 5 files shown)


ppigazzini:
@tomtor @Stefano80 @vondele
[dev-server](http://94.198.98.239/tests)
[official](http://tests.stockfishchess.org/tests)

ppigazzini:
On the page of the individual test the box length seems to take always half page.

tomtor:
@vondele Regarding the box on individual pages, we have a few options:
- Original the colored box was span4 (halve of the detail section below it which is span8)
- I made it span6 so that it is bigger (current version on dev server)
- We can make it span8, so it is as wide as the detail section
- We can drop the span so it will extend over the whole page width (span11, the menu bar on the left is span1), so it is also above the "Actions" header

I personally prefer span8 or complete width

tomtor:
image

tomtor:
image

This is a bit too much, so span8 (previous image) looks best?

tomtor:
@ppigazzini Yes, that's correct, it's currently always halve the page. The last option is that I make it as small as possible...

vondele:
my taste would be span8 (i.e. as the details section) or as small as possible. But I don't have too strong preferences.

tomtor:
Ok, I am going for the as small as possible option for consistency with the main view. Also made some more room for the test comment section, by making TC slightly smaller and minimizing space for the approve/delete buttons. @ppigazzini Perhaps good to have this latest squashed version on the dev server for a final evaluation?

ppigazzini:
@tomtor @Stefano80 @vondele : dev server updated. To me seems OK.
[dev-server](http://94.198.98.239/tests)
[official](http://tests.stockfishchess.org/tests)

tomtor:
@vdbergh Hi Michel, I inspected your nice live_elo.js script which has this regexp:

        var sprt=/>LLR: [0-9\-\.]* \([0-9\-\.,]*\) \[[0-9\-\.,]*\]\s*Total: [0-9]* W: ([0-9]*) L: ([0-9]*) D: ([0-9]*)<\/pre>/g;

which will no longer match after this PR because `\s*` will not match the `<br/>` and some extra whitespace has been added.

        var sprt=/>\s*LLR: [0-9\-\.]* \([0-9\-\.,]*\) \[[0-9\-\.,]*\].*Total: [0-9]* W: ([0-9]*) L: ([0-9]*) D: ([0-9]*)\s*<\/pre>/g;

should match both old and new HTML. Could you adapt your script?

Related to this, are you interested in making your live_elo script an integral part of fishtest (by a PR) or would you object if I did that by copying your code? That would allow showing the info as part of an individual test page or showing a compact version on the main page for running tests.

tomtor:
@vdbergh I made a branch with a copy, hope you don't mind: 

https://github.com/tomtor/fishtest/tree/LiveELO

Stefano80:
That would be a great enhancement!

snicolet:
Two small ideas for improvement:

• the yellow box itself could be clickable, linking to the LiveElo page. I think technically it would be sufficient to add a `<div></div>` around the yellow box, and a link around that : `<a href="blah"> <div></div> </a>`  is valid HTML5. And if we have that, we could reuse it everywhere, including from the overview page.

• For "Pending.." patches, it seems to me that the displayed delay (1 hour....) is not very useful info, maybe we could replace it with the SPRT bounds of the pending patch?

Stefano80:
I agree with the second suggestion. For the first one we have to see how be it looks like.

vdbergh:
@tomtor No I certainly don't mind! I have always realized that screen scraping is very fragile. However at the time I wrote live_elo.html there was no other option since fishtest development had stalled.

I have no time to maintain this now, so please make any changes you want!

tomtor:
@vdbergh Thanks! One more question: sprt_elo.js is asm.js code generated by emscripten from a C source. Do you have link to this source?

vdbergh:
This is the source code for the C++ version.

http://hardy.uhasselt.be/Toga/live_elo.tar.gz

If you want to run it on the server then it might also consider using the python version

http://hardy.uhasselt.be/Toga/sprt_elo.tar.gz

Stefano80:
Is this done?

tomtor:
Yes, it is done, but it is not evaluated on the dev server yet.

ppigazzini:
Started the test on the dev server.

tomtor:
@ppigazzini Strange, it looks different from my test server (missing gauge and LLR bounds from pending), but auto refresh and box size are OK:

screenshot from 2018-04-13 11-12-47

ppigazzini:
@tomtor dev-server updated. I prefer only compact text information, but this is just me.

tomtor:
@ppigazzini Yes, that's why it is important to get feedback first. How do you feel about the auto refresh, perhaps both should be options for the user to switch on/off?

tomtor:
@ppigazzini In addition the box is too high when no gauge is shown. @vondele @Stefano80 What do you think about auto-refresh and gauges?

ppigazzini:
@tomtor I am not a typical fishtest user, so my preference should not be kept in consideration on this topic.

tomtor:
The auto refresh can be disturbing when the layout changes and you want to click somewhere. The best option would be to refresh only the active game info boxes by using a run_info API which we also need for the other big live ELO display.  I will work some more on this of there is interested in a live update in the first place on this overview...

vondele:
@tomtor , the gauge on the main page is not really my taste (also the llr is already available, so if a gauge, I'd go for elo), will be confusing if we have 10 active tests. However, integrating the 'live_elo' page in the page of the individual test (as opposed to an extra page) would be something neat. On that page, autorefresh would be nice as well. Just my 2 cents of course.

tomtor:
I removed the auto refresh (don't like the current restless implementation) and fixed the layout for active runs without gauge.
Regarding the gauges, I like that I can see the LLR state of various tests at a glance but without automatic updates that is a bit less usefull.

tomtor:
Ok, I squashed and disabled the gauge. For me it is ok to merge, perhaps I will reenable it in the future with dynamic updates and eg showing (also) ELO.

ppigazzini:
Tested working on the dev server. Official server updated.
Thank you @tomtor

PR #228, opened 2018-04-11
New worker version

Stefano80:
Update worker version to 63.

To be merged after all worker PR have been merged, before restarting.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -113,7 +113,7 @@ def request_version(request):
   token = authenticate(request)
   if 'error' in token: return json.dumps(token)
 
-  return json.dumps({'version': 62})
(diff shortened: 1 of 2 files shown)


Stefano80:
If I am correct, merging this after all other worker PR have been merged will solve all problems, right?

ppigazzini:
@Stefano80 I agree that this is the correct workflow to update the worker.

Stefano80:
Before merging, we should:
- test on the dev server after merging locally this one
- then merge also xoto10 patch and test again on the dev server
- then merge this one on the repository, retest on the dev server
- go to production

ppigazzini:
@snicolet the update process seems a little fragile, e.g. on windows w/ msys2 never worked (the worker continue to cycle w/ an error).

PR #230, opened 2018-04-12
Remove no longer used regression database code

tomtor:
I somehow missed these files in #220


ppigazzini:
Test started on the dev-server

ppigazzini:
Official server updated, fishtest service restarted.
Thank you @tomtor

PR #231, opened 2018-04-12
Improve message

vondele:
minor nit.. make clear the job is submitted to the queue, will get active (start) later.


ppigazzini:
Official server updated and fishtest service restarted.
Thank you Joost.

PR #232, opened 2018-04-13
Retire auto approval

tomtor:
See discussion at #220


ppigazzini:
Tested working on the dev-server, official server updated.
Thank you @tomtor

PR #233, opened 2018-04-14
Fix showing LLR bounds for pending runs

tomtor:
when no runs are active

This is a minor fix


ppigazzini:
@tomtor @Stefano80 @snicolet [dev-server](http://94.198.98.239/tests) updated. Let me know if this is corrett.

tomtor:
http://94.198.98.239/tests does now show [0,5] below Pending... when no active workers so it is OK for me.

ppigazzini:
official server updated.

Stefano80:
Hi, I am just back at home. I noticed that the display of the remaining time is there for the production server and removed on the development server. Is this thought to be so.

PR #234, opened 2018-04-14
Implement more user management

tomtor:
- Allow users to change their password and email
- More checking on signup form
- New users start blocked
- Add registration time to user data
- List new pending/blocked users for approvers

In addition add a check to update_task so that workers
can only update tasks of the same user.


tomtor:
@ppigazzini I see I have to adapt the unit tests.

Regarding logout, should that be different to entering a another name (of a non-approver) at /login? 

I agree that a logout action is best practice. I will see if the framework has a real logout which forgets the user.

ppigazzini:
@tomtor yes, when I need to force a logout (as during the user management tests) I simply login w/ another user. On the dev server there are all the fishtest users and the two new users:
* user00 (pwd: user00) as approver
* user01 (pwd: user01) as normal user

tomtor:
@ppigazzini I added a real /logout which forgets the current user. Profile after /logout requires a new login now... Thanks for this suggestion

ppigazzini:
@tomtor mmm, under Pending there are the new users and the banned users. An approver could unblock a banned user thinking that was a new registration.

tomtor:
@ppigazzini That's true, that's one of the reasons I added a registration date. 

I expect that real banned users are rare and notorious and recognized by approvers. Also blocked users are expected to be unblocked after a while. I could add a real pending flag but not sure that is worth the trouble. Would we need both a real blocked and a real new user list?

Perhaps it's better to have a real drop user button for users which we do not want to unblock soon?

vondele:
do I understand correctly you want new users to be blocked initially ? I don't think this is such a good idea. It is not very welcoming to new users and as such might increase the barrier to starting contribution. Attracting new devs and contributors is very important, and being able to 'start immediately' as one signs up can help.

I agree it is important that approvers check a bit more carefully patches of new devs. I think this happens already now. To make that more prominent, what one could do is to display on the approval page some info that is already on http://tests.stockfishchess.org/users namely the 'tests submitted'.

just my 2c.

tomtor:
@vondele @snicolet @Stefano80  There is no site in the world that allows new users without at least eMail verification and/or gotchas and that is not without good reasons. All those sites also really love to grow their user population. I will not list all risks of the current relaxed user creation. This PR is still very relaxed. No email verification, no gotchas, just an approver who activates the new user.

We approve every test, why not approve at least once in a lifetime a new user?

This and the lack of https really need to be fixed.

tomtor:
The main point is that blocking a malicious cpu contributor makes no sense when you can generate 1000 active accounts each second. I had to mail Gary for an account and received a strong password and username a few years ago. 

This PR is much more user friendly. You should really play with it on the dev server, it is still user friendly.
The references to the Forum could be removed now that pending users are directly visible for approvers.

tomtor:
@vondele Yes, a new user/test delay from a specific IP could be added in a later PR, but protecting against a real attack is hard and I think not needed for a low profile site as fishtest.

I will remove the references to the forum. I added these before the pending menu. They are not needed anymore. Thanks for your feedback.

tomtor:
@ppigazzini @vondele I removed the Forum references. I would like to add other improvements/suggestions in later PRs after we gained more experience with the changes in this PR.

tomtor:
@ppigazzini Could we add this to fishtest on the current server: https://pypi.python.org/pypi/ratelimit/1.1.0

ppigazzini:
@tomtor ok, suggestion ignored. I have updated the dev server to your last commit.

ppigazzini:
@tomtor ok, dev server updated.

tomtor:
@ppigazzini Fixed travis tests, I am done here.

ppigazzini:
official server updated.

@tomtor thank you for this improvement.

vondele:
@tomtor from the first few days of having the 'pending' users on fishtest, it seems like there are multiple registrations per day, mostly of mostly random (russian) email addresses. I wonder if you can see on the server the number of people that have registered in the past (it most certainly is much larger than what http://tests.stockfishchess.org/users shows).

Should we have at least a captcha ? (There are chess captchas, but seemingly they are a bit controversial).

tomtor:
@vondele Just activated and mailed a .ru user myself... 

Only @ppigazzini can access the user database. The /users only shows users that have either created a test or contributed 1 hour with their worker.
Perhaps we should add to the pending form all users which are not listed in /user with a Delete button?
Adding a captcha is also interesting/usefull @ppigazzini What's your opinion?

tomtor:
@ppigazzini This would list those Idle users under pending: 
https://github.com/tomtor/fishtest/tree/CleanUsers

ppigazzini:
@tomtor merged your branch, now the dev-server shows the users

vondele:
@ppigazzini yes, that list looks like I was afraid it would look like ....

tomtor:
If you repull you have a drop button (one click for each user :D  )

vondele:
Somehow there should be only human action from our side for those users that tried to connect a machine to fishtest (i.e. run the worker), or submit a test, otherwise is doesn't scale (don't feel like clicking >3000 drop buttons :-))

ppigazzini:
dev-server updated
@vondele I envy you, I haven't the approver rights on the official server :(

tomtor:
Hmm also a user with an empty name...

I could add a drop all button but what is the fun in that.

PR #235, opened 2018-04-15
Add active_runs and get_run APIs

tomtor:
These JSON APIs make it possible to write new (HTML) apps
without the need to parse the HTML output of fishtest.

In addition rewrite the Live ELO page to use these APIs with a lot of support from Michel Van den Bergh.

### DIFF
--- a/fishtest/fishtest/__init__.py
+++ b/fishtest/fishtest/__init__.py
@@ -63,6 +63,8 @@ def main(global_config, **settings):
   config.add_route('api_stop_run', '/api/stop_run')
   config.add_route('api_request_version', '/api/request_version')
   config.add_route('api_request_spsa', '/api/request_spsa')
+  config.add_route('api_active_runs', '/api/active_runs')
+  config.add_route('api_get_run', '/api/get_run/{id}')
 
   config.scan()
   return config.make_wsgi_app()
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -33,6 +33,30 @@ def get_username(request):
 def authenticate(request):
   return request.userdb.authenticate(get_username(request), request.json_body['password'])
 
+def strip_run(run):
+  if 'tasks' in run:
+    del run['tasks']
+  if 'bad_tasks' in run:
+    del run['bad_tasks']
+  if 'spsa' in run['args'] and 'param_history' in run['args']['spsa']:
+    del run['args']['spsa']['param_history']
+  run['_id'] = str(run['_id'])
+  run['start_time'] = str(run['start_time'])
+  run['last_updated'] = str(run['last_updated'])
+  return run
+
+@view_config(route_name='api_active_runs', renderer='string')
+def active_runs(request):
+  l = {}
+  for run in request.rundb.get_unfinished_runs():
+    l[run['_id']] = strip_run(run)
+  return json.dumps(l)
+
+@view_config(route_name='api_get_run', renderer='string')
+def get_run(request):
+  run = request.rundb.get_run(request.matchdict['id'])
+  return json.dumps(strip_run(run.copy()))
+
 @view_config(route_name='api_request_task', renderer='string')
 def request_task(request):
   token = authenticate(request)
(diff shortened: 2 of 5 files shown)


vdbergh:
> Note that a future PR will use these APIs to remove the HTML-parsing from the Live ELO page.

You can (should!) also remove the tabs and the input box. I provided those since it was convenient for copy pasting the URL of a test page.  Now that you can invoke the script directly via the sprt link I suspect these features are no longer used and they greatly clutter the javascript code.

tomtor:
@vdbergh Yes, I will do that. @ppigazzini Also found an issue with the get_run API so this is not ready to merge/test yet.

tomtor:
@ppigazzini I have rewritten the Live ELO page, it's ready for testing...

ppigazzini:
[dev-server](http://94.198.98.239/tests) updated.

tomtor:
Note that one may have to clear the browser cache to reload the changed JS and HTML pages 
(Shift reload does the trick on Chrome)

ppigazzini:
@tomtor I played w/ the dev server and it seems OK, but I don't know very well the changes done and what test thoroughly. Let me know if in your tests the PR is OK.

vdbergh:
Wow! Much cleaner now!

Two small comments. 

(1) I found out recently that the three src lines at the bottom of live_elo.html are superfluous.  Nothing of that is used.

(2) In live_elo.js, it seems to me that the large ItemEnum can also go now. Those were error codes for the parsing. They should now be unused.

vdbergh:
Small bug:
```                                          
LLR | -2.96 [-2.94,2.94] (undefined)
```
(the "undefined").

tomtor:
@vdbergh Thanks, just spotted the `(undefined)` myself. Should be fixed now, togeter with your other suggestions.

Another potential issue is that the sprt link is available on pending runs (it is disabled on the Results box). When you click it you get something that looks OK to me, we used to get the Message "Still pending". What;s your opinion on that?

ppigazzini:
dev-server updated.

tomtor:
@ppigazzini Sorry for the trouble, had to apply another fix...

vdbergh:
I do not really have a strong opinion. I assume that now the page will automatically start showing data once the test starts, which is nice. 

However the infinities don't look so nice in the number of games. Zeroes would be better. Note that one would have actually expected NaN. The fact that infinity is shown is because of a hack. Internally sprt_elo.js regularizes 0 values for W,D,L to 0.001. The total number of games is rounded back to an integer value at the end, but I forgot to do that for the individual W,D,L values (this is a very borderline case!).

So to show zeroes j.W,j.D,j.L should be rounded and one should divide by something like (j.games+0.001).

ppigazzini:
dev-server updated ;)

tomtor:
@ppigazzini Thanks! The travis test is failing sometimes due to download errors of python libraries by Travis. I will squash tomorrow, that will force a rebuild and probably fix that...
@vdbergh I added your suggestion, should show 0.0% now (Fractions are nice). It tested and it also starts when it is no longer pending, so it should be OK now.

ppigazzini:
@tomtor OK, I forced a couple of rebuilds and at last the Travis check passed.

Stefano80:
Wow, @tomtor , this is awesome! Thank you so much for this.

vdbergh:
I have one more small remark. I am putting it here for reference in case it turns out to be a problem.

My version of the live_elo page stopped updating once the test was accepted or rejected.

The idea was that people might open many live_elo pages in different tabs and then forget about them. This would uselessly stress the server. 

The relevant line was 
```
                    if(items.state=="-"){
                        follow_live.timer_once=setTimeout(follow_live,20000,true);
                    }
```
So the rescheduling of `folllow_live` is guarded by a check on the state. I assume in the current version this would become
```
                    if(!items.args.sprt.state){
                        follow_live.timer_once=setTimeout(follow_live,20000,true);
                    }
```

tomtor:
@vdbergh Ah, i wondered if we would need such a feature. So it was there and I stripped it out. It is now in the final version. Thanks for all the support. 
@ppigazzini Could you reload the test server for a final test please?

ppigazzini:
@tomtor dev-server updated

tomtor:
@ppigazzini I checked that a completed sprt is not reloading so for me it is OK to merge.

ppigazzini:
Official server updated.
Thank you @tomtor

PR #241, opened 2018-04-18
Make API stop_run synchronous

tomtor:
stop_run is called when a wrong run is detected by workers.

Prevent that with many workers this call is executed in multiple threads
simultaneously.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -126,11 +126,12 @@ def stop_run(request):
   if not user or user['cpu_hours'] < 1000:
     return ''
 
-  run = request.rundb.get_run(request.json_body['run_id'])
-  run['stop_reason'] = request.json_body.get('message', 'No reason!')
-  request.actiondb.stop_run(username, run)
+  with request.rundb.active_run_lock(str(request.json_body['run_id'])):
+    run = request.rundb.get_run(request.json_body['run_id'])
+    run['stop_reason'] = request.json_body.get('message', 'API request')
+    request.actiondb.stop_run(username, run)
 
-  result = request.rundb.stop_run(request.json_body['run_id'])
+    result = request.rundb.stop_run(request.json_body['run_id'])
   return json.dumps(result)
 
 @view_config(route_name='api_request_version', renderer='string')
--- a/fishtest/tests/test_users.py
+++ b/fishtest/tests/test_users.py
@@ -1,4 +1,5 @@
 import unittest
+import datetime
 
 from pyramid import testing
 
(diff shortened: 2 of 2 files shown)


tomtor:
@ppigazzini @vondele @Stefano80 This is a bit of tedious work to test with real workers and real failing bench numbers so I wrote a complete unit test to test the single line I added. Also an interesting exercise in testing the API with authentication.

I am pretty sure it should work as expected, but feel free to create a real failing test on the dev server.

fishtest/fishtest/api.py
```diff
@@ -126,11 +126,12 @@ def stop_run(request):
   if not user or user['cpu_hours'] < 1000:
     return ''
 
-  run = request.rundb.get_run(request.json_body['run_id'])
-  run['stop_reason'] = request.json_body.get('message', 'No reason!')
-  request.actiondb.stop_run(username, run)
+  with request.rundb.active_run_lock(str(request.json_body['run_id'])):
+    run = request.rundb.get_run(request.json_body['run_id'])
+    run['stop_reason'] = request.json_body.get('message', 'No reason!')
```
Stefano80: I would replace `'No reason!'` by `'api request'` or similar.

fishtest/tests/test_users.py
```diff
@@ -61,6 +67,40 @@ def test_logins(self):
...
+    # Shutdown flush thread:
+    global rundb
+    rundb.stop()
+    testing.tearDown()
+
+  def test_stop_run(self):
+    with testing.testConfig() as config:
+      config.add_route('api_stop_run', '/api/stop_run')
+      self.assertEqual(stop_run(self.request), '{}')
+      self.assertEqual(self.request.rundb.get_run(self.request.json_body['run_id'])['stop_reason'], 'travis')
```
Stefano80: This is really awesome! Cool that you found out about this `testing` module!

Stefano80:
Hi @tomtor, this is very nice the way you tested, thank you. 

By the way, to be idiomatic, the PR should be called "Make API stop_run synchronous **again**"

tomtor:
@Stefano80 

> By the way, to be idiomatic, the PR should be called "Make API stop_run synchronous again"

What do you mean with again? Fishtest was always multithreaded, but with a lot of race-conditions. This is one of the last to be removed.

vondele:
Well, I'd guess this is a *HUGE* patch. It will make fishtest great again! 

;-)

fishtest/fishtest/api.py
tomtor: Only the first stop_run is a bit expensive because all empty tasks are removed, the other calls are quite cheap when they are executed non parallel. Ignoring/skipping later stop_runs with additional code adds complexity with little gain. Note that  a stopped_run may also be rescheduled later...

tomtor:
@Stefano80 Changed 'No reason!' and squashed again.

ppigazzini:
official server updated.
thank you @tomtor

PR #243, opened 2018-04-20
Add captcha for signup form

tomtor:
See: https://www.google.com/recaptcha/admin

Store the secret key in

    ~/fishtest.captcha.secret

Store the public key in

    templates/signup.mak

in the line near the end:

    <div class="g-recaptcha" data-sitekey="PUBLIC-KEY-HERE"></div>

screenshot_20180421-085541


ppigazzini:
official server updated
great work @tomtor :-)

Issue #237, opened 2018-04-16
keep a history of actions


vondele:
currently the page
http://tests.stockfishchess.org/actions
only shows the last 100 actions (<24h). 
Sometimes it would be useful to have more history along the lines of
http://tests.stockfishchess.org/actions?page=N
would be a nice feature if easy to do.

vondele:
fixed with #244. Thanks.

PR #244, opened 2018-04-21
Add search to actions

tomtor:
This is an alternative for https://github.com/glinscott/fishtest/issues/237

Allow searching on action type and/or username.

Also clean up some named parameter white space issues.

### DIFF
--- a/fishtest/fishtest/actiondb.py
+++ b/fishtest/fishtest/actiondb.py
@@ -7,8 +7,13 @@ class ActionDb:
     self.db = db
     self.actions = self.db['actions']
 
-  def get_actions(self, max_num):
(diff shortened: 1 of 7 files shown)


ppigazzini:
dev server updated: the "Actions" page is a little slow.

vondele:
The interface is very nice and useful. It is indeed rather slow.

ppigazzini:
dev server updated setting actions limit to 100.

vondele:
LGTM

tomtor:
@ppigazzini Adding a mongodb index to action['username'] (should/could be done outside this PR) should speed user searches up (but those are rare), so I think this is OK with the old 100 limit.

ppigazzini:
official server updated
@tomtor thank you! If a search should be slow we can try with the mongodb index.

PR #245, opened 2018-04-22
Fix error on modify/approve test when not logged in

tomtor:
Fix error on approve/modify test when not logged in

After login return to the main page.

Without this fix:

    Internal Server Error

    The server encountered an unexpected internal server error

    (generated by waitress)

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -424,8 +424,7 @@ def tests_modify(request):
     request.actiondb.modify_run(authenticated_userid(request), before, run)
 
     cached_flash(request, 'Run successfully modified!')
-    return HTTPFound(location=request.route_url('tests'))
-  return {}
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor please suggest the steps to replicate the problem on the dev server (I tried unsuccessfully some combination)

tomtor:
@ppigazzini On the unpatched fishtest:
- logout
- try modify an existing test

vondele:
actually a similar error happens when trying to approve a test when not logged in. Maybe this fix also fixes that, otherwise would be great if you could look into that as well.

ppigazzini:
@tomtor yes, I stopped always at step 3.
Sorry for the late answer but I wrote a little patch to unblock the default users on the dev server and update the install scripts :-)

official server updated.

PR #246, opened 2018-04-23
Fix scavenge.py

tomtor:
See #242

### DIFF
--- a/fishtest/utils/scavenge.py
+++ b/fishtest/utils/scavenge.py
@@ -28,7 +28,8 @@ def get_idle_users(days):
          or u['registration_time'] < datetime.utcnow() - timedelta(days=days):
         idle[u['username']] = u
   for u in rundb.userdb.user_cache.find():
-    del idle[u['username']]
+    if u['username'] in idle:
+      del idle[u['username']]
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor official server updated, but I got this error:

```
 python fishtest/fishtest/utils/scavenge.py

  File "fishtest/fishtest/utils/scavenge.py", line 47, in <module>
    main()
  File "fishtest/fishtest/utils/scavenge.py", line 44, in main
    scavenge_users(scavenge=True)
  File "fishtest/fishtest/utils/scavenge.py", line 40, in scavenge_users
    rundb.userdb.users.delete_one({'username': u['username']})
  File "/usr/lib64/python2.6/site-packages/pymongo/collection.py", line 1422, in __call__
    self.__name.split(".")[-1])
TypeError: 'Collection' object is not callable. If you meant to call the 'delete_one' method on a 'Collection' object it is failing because no such method exists.
```

On the dev-server it works fine.

tomtor:
@ppigazzini Thanks for the upgrade efforts. I cannot think of another reason why it should fail on the production server.

Issue #247, opened 2018-04-23
Wrong message


Rocky640:
Until recently, if I submit a new test and the branch had not been pushed to the remote, 
I would get  a message similar to  "Unable to find branch" 

Now, the message only says "commit"
![image](https://user-images.githubusercontent.com/8949638/39142543-000a64bc-46f9-11e8-850d-b7bf60809623.png)

tomtor:
I get "Cannot find branch in developer repository" in #248 but this issue probably IS present in the current master. Thanks for reporting.

PR #248, opened 2018-04-23
Enter commit info into test info field when left blank

tomtor:
In addition jump to the created test page
instead of the home page after creating a test.

Should also fix https://github.com/glinscott/fishtest/issues/247

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -273,9 +273,12 @@ def users_monthly(request):
 def get_master_bench():
   bs = re.compile('(^|\s)[Bb]ench[ :]+([0-9]{7})', re.MULTILINE)
   for c in requests.get('https://api.github.com/repos/official-stockfish/Stockfish/commits').json():
+    if not 'commit' in c:
+      return None
     m = bs.search(c['commit']['message'])
     if m:
       return m.group(2)
+  return None
 
 def get_sha(branch, repo_url):
   """Resolves the git branch to sha commit"""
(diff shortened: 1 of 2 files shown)


tomtor:
@vondele I could not find an efficient way to find the most recent branch with the github API so no prefill of the 'test branch' field. Your other suggestion (jump to the created test page) was easy to add.

ppigazzini:
dev-server updated

vondele:
excellent! works as advertised.

vondele:
#247 is also fixed on the dev-server

ppigazzini:
official server updated.
@tomtor thank you.

fishtest/fishtest/views.py
```diff
@@ -324,18 +327,24 @@ def validate_form(request):
...
-      data['new_signature']= m.group(2)
+    if not 'commit' in c:
+      raise Exception('Cannot find branch in developer repository')
+    if len(data['new_signature']) == 0:
+      bs = re.compile('(^|\s)[Bb]ench[ :]+([0-9]{7})', re.MULTILINE)
+      m = bs.search(c['commit']['message'])
+      if m:
+        data['new_signature']= m.group(2)
+    if len(data['info']) == 0:
+        data['info'] = ('STC: ' if re.match('^[012][0-9][^0-9].*', data['tc']) else 'LTC: ') \
```
snicolet: The 'STC: ' strings are a little bit hurting the readability of the test comments, while bringing no information (less than 1 bit of information on average :-))

Maybe this would be sufficient:

````
data['info'] = ('' if re.match('^[012][0-9][^0-9].*', data['tc']) else 'LTC: ') \
````

tomtor: I agree...

tomtor:
@vondele suggested this feature and the texts so I suggest that @snicolet or @vondele create a PR.

vondele:
yes, will take care.

PR #250, opened 2018-04-24
Use old mongodb method to drop users

tomtor:
This fixes scavenge.py for the current MongoDb software in production.

Followup for #246

### DIFF
--- a/fishtest/utils/scavenge.py
+++ b/fishtest/utils/scavenge.py
@@ -37,7 +37,7 @@ def scavenge_users(scavenge=True, days=28):
     for u in get_idle_users(days):
       print(u['username'])
       if scavenge:
-        rundb.userdb.users.delete_one({'username': u['username']})
+        rundb.userdb.users.remove({'_id': u['_id']})
 
 def main():
   scavenge_tasks(scavenge=True)


tomtor:
@ppigazzini I tested this, it still works on the newer MongoDb software, so you can test it on the dev server as well.

ppigazzini:
It works fine on the dev-server, but it fails on the official server...

```
OscarAmono
edwardpeng91
juergen
bingodoc
  File "fishtest/fishtest/utils/scavenge.py", line 47, in <module>
    main()
  File "fishtest/fishtest/utils/scavenge.py", line 44, in main
    scavenge_users(scavenge=True)
  File "fishtest/fishtest/utils/scavenge.py", line 38, in scavenge_users
    print(u['username'])
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 8: ordinal not in range(128)

```

tomtor:
@ppigazzini Some are already dropped, you can just comment (or remove) the print:

    #print(u['username'])

ppigazzini:
@tomtor no error, please check if the users are dropped

tomtor:
Yes, they are gone :grinning:

ppigazzini:
@tomtor please open a new PR dropping the print

Issue #253, opened 2018-04-25
Large tuning still cripples the server


noobpwnftw:
It is however strange that I do not see any actual network failures on my proxy between workers to the server, so it is just the server isn't giving responses in time.

tomtor:
And #254

tomtor:
@ppigazzini I have fishtest running with pypy:
- download and extract https://bitbucket.org/pypy/pypy/downloads/pypy2-v5.10.0-linux64.tar.bz2 in $HOME
- install virtualenv (distribution version or with pip)
- virtualenv -p ~/pypy2-v5.10.0-linux64/bin/pypy pypy
- source pypy/bin/activate 
- pip install numpy
- pypy setup.py install
- pypy /usr/local/bin/pserve production.ini

tomtor:
@ppigazzini An additional advantage of pypy installation is that you also get new versions of all used Python packages in a separate environment.

noobpwnftw:
I see, is it then possible to limit the number of concurrent tuning tasks? Like we could have something similar to the throughput setting so that no more than X tasks should run. It is also useful to other tasks, like we don't want to have every worker assigned to one single long-running test.

tomtor:
@noobpwnftw 

> I see, is it then possible to limit the number of concurrent tuning tasks? Like we could have something similar to the throughput setting so that no more than X tasks should run. It is also useful to other tasks, like we don't want to have every worker assigned to one single long-running test.

Sure, but not just as fix to the current issue which should/could be solved by a different Python stack.

The issue with limiting the number of workers for each run is that the only situation in which a run gets all the workers is simply when there are no other runnable runs.  So what should we do in that situation? Leave 80% of the workers idle? Once workers are allocated to a run they should finish their allocated task, because the credits and priority calculations are currently based on workers completing their 1000 game tasks. Fixing those calculations would allow stopping workers without these drawbacks. Lowering the batch size to eg 500 is also an option as has been proposed on the forum.

ppigazzini:
@tomtor thank you for the suggestion. I will have some spare time during the week end to test this (and the mongo db upgrade) on a new dev server installed with CentOS 6.8.

ppigazzini:
@tomtor anyway, what is the advantage of using pypy? Also with python 2.7 from the software collections should be possible to have a clean fishtest install with updated packages.

tomtor:
[...]

PR #254, opened 2018-04-25
Cleanup SPSA calculation

tomtor:
by not computing the summary when it is not stored.

Also remove some old optimization code which does no
longer make sense after the introduction of the run cache.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -402,9 +402,7 @@ class RunDb:
         self.stop_run(run_id, run)
         flush = True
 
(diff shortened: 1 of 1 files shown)


Stefano80:
Hi @tomtor , it looks good to me. 

Just for my curiosity: how much speed-up does it give for the update_spsa function? Is the bottleneck really the computing of the summary? And is `update_spsa` a bottleneck in the SPSA sessions?

tomtor:
No, the speedup is minimal. It is more of a cleanup (the removal of the condition in update-task) and not computing the summary when we do not use it. See #253 for a discussion. This code is not the bottle neck for spsa.

ppigazzini:
official server updated

PR #255, opened 2018-04-25
Add numpy to setup.py

tomtor:
Missing requirement


Stefano80:
Merged, this is necessary and nothing bad can happen.

PR #256, opened 2018-04-25
Fix race condition in get_pending()

tomtor:
Found while stress testing with pypy

### DIFF
--- a/fishtest/fishtest/userdb.py
+++ b/fishtest/fishtest/userdb.py
@@ -1,5 +1,7 @@
 import sys
 import time
+import threading
+
 from datetime import datetime
 from pymongo import ASCENDING, DESCENDING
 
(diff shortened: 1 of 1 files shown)


Stefano80:
Hi @tomtor , thank you for this, I will merge as soon as @ppigazzini gives its ok. Could you maybe post or commit the code you used for stress testing? By the way, would it make sense to add it Travis?

tomtor:
@Stefano80 that's really simple, utils/stress-main.sh:

    for i in 1 2 3 4 5 6 7 8 9 10; do curl http://localhost:6543/tests?[1-1000] > /dev/null & done
    wait

Not sure if that should be added to the repo.
I could perhaps add something to travis later, those travis scripts do need some love.

Stefano80:
Ok, then let me merge, at latest tomorrow (unless @ppigazzini precedes me).

ppigazzini:
official server updated

Issue #257, opened 2018-04-25
cache builds on the worker


vondele:
I have been thinking about the issue of having smaller batches run on fishtest (maybe as small as 100). AFAIK, the main obstacle is the overhead of building the stockfish binaries. However, a solution that might be easy to implement would be to keep a worker-local cache of binaries (probably in the form of stockfish.SHA files in the worker directory), which a worker would use as possible. This would remove the building overhead largely (a worker would at most need to build all running patches, but that's only a handful). In fact, it could even speedup with the current batch size, as e.g. master is often the same for sequences of patches, tuning builds things twice, and a fast worker might do multiple batches already now. By keeping only the 25 latest binaries, the size of the cache would be 10Mb or so.

ppigazzini:
@vondele w/ msys2 on windows the binary is 3455 KB, so caching 25 runs requires 25*2*3.5 MB = 175 MB (if the base can be shared the disk space is halved).

vondele:
Really... ? Linux: 344K

Is that equivalent to a static build (Linux static is 3.2Mb) ? If so is that really needed when building a local version ? I see that the makefile adds -static for mingw.

Even 200Mb is actually still on the semi-reasonable size of a cache, but is not ideal anymore.

vondele:
I tried on linux to cross compile to mingw (https://github.com/glinscott/fishtest/wiki/Building-stockfish-on-Windows#cross-compilation-from-ubuntu) and the non-static stripped binary is 384K, so reasonable. It looks like the static build includes libwinpthread-1.dll and libstdc++-6.dll, which would explain the size.

ppigazzini:
`libgcc_s_seh-1.dll  libstdc++-6.dll  libwinpthread-1.dll` to be correct, view the [wiki](https://github.com/glinscott/fishtest/wiki/Building-stockfish-on-Windows#building-stockfish-with-link-time-optimization-lto)

vondele:
In case we can't get rid of -static for local builds, we can use ```strip --strip-all``` which reduces the mingw builds to 1.2M which would reduce a 25 element cache to ~30Mb, which I think is more or less OK.

ppigazzini:
`strip` and `strip --strip-all` make the same binary w/ my msys2

msys2 binary
- 3456 KB static w/o strip
- 1138 KB static w/ strip
- 911 KB dynamic w/o strip
- 281 KB dynamic w/ strip

vondele:
[...]

PR #258, opened 2018-04-26
Keep a cache of compiled SF binaries.

vondele:
As discussed in issue #257

keep the 25 most recent binaries on the worker, and reuse instead of recompile.

untested, for discussion.


Stefano80:
Ok, tomorrow in the evening I will have time and try to go through it.

ppigazzini:
1. dev-server setup with chunk_size=10

https://github.com/glinscott/fishtest/blob/f2308e6097a8d4212d44a34b3242434965230259/fishtest/fishtest/rundb.py#L28

2. workers started with:
```
#!/bin/bash
cd fishtest
rm -rf worker
git fetch upstream
git pull  --no-edit upstream pull/258/head
cd worker
python worker.py --host 94.198.98.239 --concurrency 3 user01 user01
```

3. error message on WSL
```

Exception running games:
  File "/root/_git/fishtest/worker/games.py", line 447, in run_games
    ['-each', 'proto=uci', 'tc=%s' % (scaled_tc)] + nodestime_cmd + threads_cmd + book_cmd
TypeError: 'str' object is not callable
```
4. error message on Windows msys2
```
C:\msys64\mingw64\bin\strip.exe: 'stockfish': No such file
make: *** [Makefile:427: strip] Error 1

Exception running games:
  File "D:\__test_worker\ori\worker01\\worker.py", line 91, in worker
  File "D:\__test_worker\ori\worker01\games.py", line 377, in run_games
    setup_engine(new_engine, worker_dir, sha_new, repo_url, worker_info['concurrency'])
  File "D:\__test_worker\ori\worker01\games.py", line 129, in setup_engine
    subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency) + ' strip', shell=True)
  File "C:\Python27\lib\subprocess.py", line 186, in check_call
    raise CalledProcessError(retcode, cmd)
CalledProcessError: Command 'make COMP=mingw ARCH=x86-64-modern -j 7 strip' returned non-zero exit status 2
Task exited
```

vondele:
@ppigazzini Thanks! 
- Number 3. should be fixed by the 1d4a616 commit.
- Number 4. Windows msys2 .... seems like a stockfish makefile issue. Can you do in the stockfish tree
```
make COMP=mingw ARCH=x86-64-modern -j 7 profile-build
make COMP=mingw ARCH=x86-64-modern -j 7 strip
```
The error message seems to suggest that strip doesn't find the binary.

tomtor:
@ppigazzini if you lower the chunksize you should also lower the max number of games default : eg to 
1000 * chunksize 
Otherwise the list gets very big.

ppigazzini:
@tomtor thank you, i viewed only now your message. I set the chunksize=100 because the server was unresponsive.
@vondele IMO number 4 is missing the extension .exe

vondele:
@ppigazzini I think so as well, but that ought to be done in the Makefile. The strip target is thus not working ? Any chance you could test the following change to the SF Makefile (both profile-build and strip):
```diff
diff --git a/src/Makefile b/src/Makefile
index 25436e25d..34264e78e 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -22,7 +22,11 @@
 ### ==========================================================================
 
 ### Executable name
+ifeq ($(COMP),mingw)
+EXE = stockfish.exe
+else
 EXE = stockfish
+endif
 
 ### Installation dir definitions
 PREFIX = /usr/local
@@ -437,7 +441,7 @@ clean: objclean profileclean
 
 # clean binaries and objects
 objclean:
-       @rm -f $(EXE) $(EXE).exe *.o ./syzygy/*.o
+       @rm -f $(EXE) *.o ./syzygy/*.o
 
 # clean auxiliary profiling files
 profileclean:
```

ppigazzini:
@vondele could you open a PR on Stockfish to correct the Makefile?

vondele:
sure, the makefile issue belongs there. Just wondering if it actually fixes the issue. Cross-compiling & stripping from linux seems to work this way.

ppigazzini:
@vondele windows worker works fine with the commit https://github.com/glinscott/fishtest/pull/258/commits/7dc785d19fa40b0e76aa0609e871aaa4f79fa258

vondele:
I think it is working correctly indeed. Obviously we still want to strip under mingw eventually, but this try/pass would still be needed if we want to test old version of SF (regression tests).

I wonder if this change will also protect somewhat against people running two workers in the same directory and producing wrong benches. Now, the binaries all have unique names, and the right binary will be picked anyway.

tomtor:
https://github.com/glinscott/fishtest/blob/f2308e6097a8d4212d44a34b3242434965230259/fishtest/fishtest/views.py#L403

Should be

    data['num_games'] = 1000 * request.rundb.chunk_size

In addition because request_task is an expensive server operation (with a random delay at the worker of 1-30s) I suggest a chunk size of minimal 250.

vondele:
250 @tomtor might indeed make a lot of sense (also in light of the random delay). 

However, I think it would be good to make the change to the worker and the eventual reduction of batch size independent. 

Since I don't know the server, if we change the chunk_size, does it need special care, e.g. to have the attribution of played games, etc. correct ?

tomtor:
Regarding chunk size, this also effects SPSA runs. A worker will play mini batches of 2 * Nr-Processors games. So a 32 core worker will play 64 games in an SPSA batch. The last games (chunk_size % (2*Nr-Processors)) from a chunk_size task are currently waisted. They are played but do not contribute to an SPSA iteration. This is not a major issue but we should look at it in a separate PR.

tomtor:
@vondele See https://github.com/glinscott/fishtest/pull/261/commits/781fcc9afef3f08b133a6a50909735bececb49ea for a fix of the travis test

vondele:
@tomtor thanks! Integrated in the patch.

vondele:
Done. Note that this is now a change to both worker and server. The worker might thus need a version bump.

ppigazzini:
@vondele please consider to write this PR also for the [Multi Variant Fishtest](https://github.com/ianfab/fishtest) run by @ianfab :-)

noobpwnftw:
Please let me know: do I need a particular version of worker to test things?

ppigazzini:
I run this script
```
#!/bin/bash
# run once then comment out
git clone https://github.com/glinscott/fishtest.git
cd fishtest
git config --global user.email "you@example.com"
git config --global user.name "Your Name"
cd ..

# run always
cd fishtest
rm -rf worker
git fetch upstream
git pull  --no-edit upstream pull/258/head
cd worker
python worker.py --host 94.198.98.239 --concurrency 8 user01 user01
```

Stefano80:
So, now I have some conflicts. @vondele : could you solve them? One is somewhat clear, the other one is impacting the logic, so I don't know what to do.

Stefano80:
Sorry for closing.

Stefano80:
Somehow I understood that this is ready for merging, now I realise that it probably is not. No harm done, as I could not merge and server has not been restarted.

vondele:
@Stefano80 it was ready for merging, but it looks like it needs a rebase... let me have a look.

noobpwnftw:
If I did things correctly, my workers are running this branch. I have manually fixed one conflict in games.py

worker/games.py
```diff
@@ -149,7 +145,11 @@ def setup_engine(destination, worker_dir, sha, repo_url, concurrency):
         make_cmd = m.read().strip()
       subprocess.check_call(make_cmd, shell=True)
     else:
-      subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency), shell=True)
+      subprocess.check_call(MAKE_CMD + ' -j %s' % (concurrency) + ' profile-build', shell=True)
```
Stefano80: Ok for now. But could you open a issue in order to fix this? I see two possible fixes:
- check that we get correct fail and then pass or
- only avoid stripping after checking which kind of makefile we got

worker/games.py
```diff
@@ -380,25 +380,28 @@ def parse_options(s):
   if not os.path.exists(testing_dir):
     os.makedirs(testing_dir)
 
-  new_engine = os.path.join(testing_dir, 'stockfish' + EXE_SUFFIX)
-  base_engine = os.path.join(testing_dir, 'base' + EXE_SUFFIX)
+  # clean up old engines (keeping the 25 most recent)
+  engines = glob.glob(os.path.join(testing_dir, 'stockfish_*' + EXE_SUFFIX))
+  engines.sort(key=os.path.getmtime)
+  if len(engines) > 25:
```
Stefano80: Without checking, I think that 
```
for old_engine in engines[26:]:
   os.remove(old_engine)
```
could work too.

worker/games.py
```diff
@@ -380,25 +380,28 @@ def parse_options(s):
...
+    for old_engine in engines[:len(engines)-25]:
+      os.remove(old_engine)
+
+  # create new one
```
Stefano80: > create new ones

should be the right comment...

Stefano80:
The CI build is failing. Does it pass locally? I retrigger it here online.

worker/games.py
vondele: I'm not sure what to do here. Do you want to catch specific exceptions (calledProcessError) only?

Stefano80: Yes, I would only catch specific exceptions and rethrow all others. By the way, what happens if you `strip` a linux build?

Stefano80:
Mmmh... it keeps failing.

vondele:
@Stefano80 I can only guess why CI might fail, don't yet know how to test locally. It looks like that line is related to a previous patch. Maybe we have exceeded the number of github api calls / hour ?

worker/games.py
noobpwnftw: then it will lose debug info, which I believe it is even beneficial since we are caching them.

vondele: I'm a python newbie actually.... wouldn't mind if you did the catching & rethrowing...

Stripping linux binaries works fine, yields some space savings.

noobpwnftw:
It says test_item_download function failed, it is in the worker code that expecting certain behavior.

vondele:
```
  tree = requests.get(github_api(FISHCOOKING_URL) + '/git/trees/setup', timeout=HTTP_TIMEOUT).json()
  for blob in tree['tree']:
```
it seems like 'tree' is not available in tree['tree'], which could happen if the github_api call was refusing to return what is needed.

Stefano80:
I just checked the master locally, it passes (by the way the correct command is `test_worker.sh`). It sounds more like too much requests. Let me merge your patch locally and check whether it works for me.

Stefano80:
Ok, locally your patched passes, so it is probably due to too many github requests. If you squash I will merge and restart the builds in some time.

worker/games.py
Stefano80: Ok, let us open an issue with this `try/catch` stuff. I don't know whether I can do it on my linux machine, but it could be that somebody picks it up.

Stefano80:
I am now going offline for some time, coming back in 30 min and wait for @ppigazzini . If you dont have the time to squash, I will do, we can work on giving you authorship later.

vondele:
should be squashed now...

Stefano80:
Ah, you catched me in the last time. Mergin.

PR #261, opened 2018-04-27
Do not POST SPSA parameters from worker

tomtor:
but store them in the fishtest server to reduce network traffic
and significant unmarshalling time for SPSA runs with many parameters.

This reduces #253 significantly and is a followup to #155 

Only drawback is an increase in memory usage of the server, about 50 bytes per SPSA parameter per worker. So 200 workers working on a 500 parameter run adds 5 megabyte to the worker memory usage.

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -96,6 +96,7 @@ def update_task(request):
   if 'error' in token: return json.dumps(token)
 
   result = request.rundb.update_task(
+    worker=request.json_body['unique_key'],
     run_id=request.json_body['run_id'],
     task_id=int(request.json_body['task_id']),
     stats=request.json_body['stats'],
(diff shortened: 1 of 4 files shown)


ppigazzini:
@tomtor dev server updated, started a SPSA run w/ some workers.

ppigazzini:
updated

tomtor:
@ppigazzini Sorry, I made some more changes, could you please update again?

ppigazzini:
updated ;-)

tomtor:
@ppigazzini Thanks, Note that you must also update the worker you use for testing, I forgot that just myself ;-)

Stefano80:
Hi,
I would like to merge this and the other worker versions today. 

@ppigazzini if you are online, could you drop a line? I think for worker versions it is better to troubleshoot live, so I will leave it to you to restart. I will join the party and report any issues from my local worker.

ppigazzini:
@Stefano80 I'm commuting now, I will be on-line at 22.00 CEST

Stefano80:
Ok, then I will make sure to be online then.

tomtor:
It is a bit tricky code but it's good to merge now with the other workers changes.

PR #262, opened 2018-04-27
Remove default STC from fill, as suggested in #248

vondele:
see subject.


ppigazzini:
production server updated, thank you @vondele @snicolet

PR #264, opened 2018-04-28
Cache MongoDb user lookups

tomtor:
because threads are waiting for lookups > 50% of the time

See http://vmprof.com/#/48225b9c-d12c-4d14-96ce-bed6442eeaea?id=1,0,0,1,0,0,0,0,0,0,0,0,0,0,0,2&view=flames (before this PR and #261 update_task takes almost 30% of the time)

and

http://vmprof.com/#/0c662771-0c65-4f67-bcc4-8483b546f2ff?id=1,0,0,1,0,0,0,0,0,0,0,0,0,0,0,3&view=flames (now 10% of the time)

### DIFF
--- a/fishtest/fishtest/userdb.py
+++ b/fishtest/fishtest/userdb.py
@@ -17,8 +17,28 @@ class UserDb:
   def init_collection(self):
     self.users.create_index('username', unique=True)
 
+  # Cache user lookups for 60s
+  user_lock = threading.Lock()
+  cache = {}
+
+  def find(self, name):
+    with self.user_lock:
+      if name in self.cache:
+        u = self.cache[name]
+        if u['time'] > time.time() - 60:
+          return u['user']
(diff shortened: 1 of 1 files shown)


ppigazzini:
dev-server and workers updated, restarted the failed Travis check by hand.

tomtor:
@ppigazzini Thanks! Those travis build are quite unreliable in this context. This build failed in a way I had not seen before.

tomtor:
@Stefano80 @ppigazzini @vondele I think it would be a good idea with the recent speedups and 250 chunk size to change the worker start delay from 1-30s to 1-10s

vondele:
@tomtor I agree with the reduction in delay.... it is something I wanted to propose as well. I think with all the improvements in the infrastructure, this has become unproblematic.

ppigazzini:
Production server updated.
Thank you @tomtor

PR #265, opened 2018-04-28
Adjust worker timings

tomtor:
With recent changes to smaller chunk sizes and speedups reduce the startup time
to 1-10s from 1-30s for increased efficiency with fast workers.

Also remove the progressive quadratic delay after failed update_task
calls. These made no real sense, because cutechess-cli keeps on playing
while we wait and will be kiled after about 3*TC seconds.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -118,7 +118,7 @@ def setup(item, testing_dir):
         blob_json = blob_request.json()
         if blob_request.status_code == requests.codes.ok:
           break
-        sleep(5)
+        sleep(HTTP_TIMEOUT)
       with open(os.path.join(testing_dir, item), 'wb+') as f:
         f.write(b64decode(blob_json['content']))
       break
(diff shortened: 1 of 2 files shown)


tomtor:
Note that we actually have a two stage delay, if we get no_task because more then 4 workers are queued during the first 1-10s window, we enter a second delay:

https://github.com/glinscott/fishtest/blob/8a2da7b2abb691a67e2418502d9996312a20641f/worker/worker.py#L94

This should be quite robust AND efficient.

noobpwnftw:
I usually lower some of those delays and it never caused any problem.
In case of trouble, I have a proxy that throttles communication from my workers to the server.

tomtor:
@noobpwnftw Yes #219 makes it almost impossible to overload the server with request_task...

Stefano80:
Ok, but could we overhaul the worker update before rolling this out? Yesterday we had some exciting moments with @ppigazzini ....

noobpwnftw:
I probably know why my updates failed.
My workers are running CentOS7, and they need a script mentioned in the Wiki to have devtoolset-6 enabled, that might has affected the CWD path that caused self restart cannot locate itself.

tomtor:
I propose #267 

@ppigazzini I have nothing against a git solution, so please feel free to suggest an alternative

Stefano80:
Feel free to merge, I documented my approval. I just wanted to remember us that we have to fix something.

Stefano80:
I will be online in the evening.

Issue #125, opened 2017-12-30
fishtest can't be run on linux with /tmp noexec mount option


d3vv:
$ mount|grep "/tmp"
shm on /tmp type tmpfs (rw,nosuid,nodev,noexec,relatime)

/tmp/tmpGOdCbf/locutus2-Stockfish-c5bcbd4/src $ ./stockfish
-bash: ./stockfish: Permission denied

d3vv:
UPD. If /tmp is out of space or mounted with noexec option you could to use:

export TMPDIR=</path/to/my/temp/dir>

Stefano80:
Is this a fishtest or a stockfish problem?

d3vv:
Fishtest if fishtest-program try to compile sources and execute pgo into /tmp by default

vondele:
The fishtest worker (games.py) uses ```tmp_dir = tempfile.mkdtemp()```

The python docs state that
```
 tempfile.mkdtemp([suffix=''[, prefix='tmp'[, dir=None]]])

    Creates a temporary directory in the most secure manner possible. There are no race conditions in the directory’s creation. The directory is readable, writable, and searchable only by the creating user ID.
```
There indeed no guarantee we can execute stuff there. We could do the build in 'fishtest/worker/testing' which we anyway need to be executable.

tomtor:
This can be closed now?

ppigazzini:
Yes

PR #266, opened 2018-04-28
Compile in worker subdir.

vondele:
fixes the cornercase described in issue #125 for those systems where /tmp/ is noexec. Since we anyway need the worker dir to be writable, and exec, we can as well compile there.


ppigazzini:
added at the running test w/ dev-server and workers

PR #267, opened 2018-04-28
Update updater

tomtor:
add `import requests` to updater.py

**NOTE: This should be done by HAND for current installations or update will fail :/
It was lost in the last update...**

and "rm -r requests/" 5 year old directory.

Add requests dependency to setup.py

Re-add a fresh requests library copy to the repository as a temporary band aid in anticipation of a new updater/installer which installs libraries/dependencies in a standard way.


Stefano80:
This would be great. Let us check it on the development server and then merge readily.

ppigazzini:
@tomtor I'm testing w/ dev-server and these workers updated w/ git:
- ubuntu: OK
- windows msys2 and windows WSL
```
 delete mode 100644 worker/requests/status_codes.py
 delete mode 100644 worker/requests/structures.py
 delete mode 100644 worker/requests/utils.py
  File "worker.py", line 9, in <module>
    import requests
ImportError: No module named requests
```
Copying a `requests` folder solved the problem w/ both the windows workers

tomtor:
Note the edited BOLD text in the PR description. We should tell fishtest users in advance on the forum that they have to edit their updater.py file, before we can roll out a new worker version.

tomtor:
@ppigazzini I edited the wiki, you need to install `python-requests`

ppigazzini:
@tomtor I needed a refresh of the page to view the **BOLD** test

tomtor:
@ppigazzini @Stefano80 At least we cannot bump the version until users repaired updater.py

We can merge without bumping the version, that would allow user to fetch the new version from github.com

tomtor:
@ppigazzini `sudo apt install python-setuptools` first

noobpwnftw:
I'm actually against the removal of included requests package:
1. You may get unknown behavior if the system package is problematic.
2. We had some customized parameters there like timeouts, connection pools, etc that would then need to be specified at many places in our code to override the default behavior.

tomtor:
@noobpwnftw We could update the version used, but it is bad style (and a potential security risk because we do not get updates) to include such a large popular Python library for internet communication. We do not do that for other dependencies like gcc.

noobpwnftw:
Then we will need to configure https://github.com/glinscott/fishtest/commit/793ad5d5cfe4758fe9b6fa6b8e1045354dd30007 outside the requests package, probably the retry too.

ppigazzini:
@tomtor @noobpwnftw how to solve the problem for the windows workers?

tomtor:
@noobpwnftw According to http://urllib3.readthedocs.io/en/1.2.1/managers.html connections to the same host will use the same connection?

ppigazzini:
@tomtor update:
* Ubuntu: `sudo apt install python-setuptools` solved the problem for `python setup.py install --user`
* WSL: setptools was alredy installed and `python setup.py install --user` fails with the same error.

noobpwnftw:
@tomtor It apparently wasn't the case when that change was made, probably because of Github APIs may be from different hosts?

tomtor:
@ppigazzini Please note that running `setup.py` is not needed, nor documented in the Wiki for normal worker users. They only need to install `python-requests` with the Linux/WSL/msys2 package manager...

tomtor:
@noobpwnftw We don't care about accessing Github APIs by the worker, we just should trust that an up-to-date requests library as used by thousands of Python programs does the right thing. If the requests library does strange things when accessing the fishtest server (it does not for as far as we currently now) then we have a fishtest server programming issue...

ppigazzini:
@tomtor deleting the `requests` folder breaks an already installed Windows worker and block a new installation on Windows (just tested with a clean VM).
The new installation starts working after copying an old `requests` folder into the `worker` folder.

Stefano80:
Would it be possible to add some type of workaround in the `main()` which is only used once for getting rid of the requests package and having a clean install and which is cleaned-up with a subsequent PR in some weeks?

ppigazzini:
@Stefano80 if I'm not wrong in this moment the function to update the worker should be crippled. IMO is more simple to correct the bug and ask to the users to make a fresh install or to make a simple correction by hand.

tomtor:
I am all in for providing the end user easy access to the request library, but having a 5 year or 5 day old copy in the fishtest git repository is simply not the way forward. In addition we have the situation that updater.py is broken, so we have to offer a new easy way to install updated fishtest workers for windows users anyway...

Probably an updated Windows one-click-installer and a Linux/Mac/WSL install oneliner script, something like:

    curl github.com/fishtest/... | sh

ppigazzini:
@tomtor to solve the emergency we could:
* have a PR withe the working updater.py for the new installation (but with the requests library)
* ask the already active users to copy the updater.py file or to make a fresh install

With the emergency solved we could decide the best way to install the worker and to push the updates.

To be clear: I would vote to merge this PR is it worked also for the Windows users.

Stefano80:
Can we give the active users a script or something at least partially automated to comply with our request? The very minimum would a step-by-step guide.

tomtor:
We could point users to (after merging):

https://raw.githubusercontent.com/glinscott/fishtest/master/worker/updater.py

and tell them to save the file from their browser-menu in the worker directory and confirm the "file already exists" warning.

I would prefer that we point Window users instead to a new version of https://github.com/glinscott/fishtest/wiki/A-portable-fishtest-worker-for-Windows-64-bit which also includes/installs the standard requests library so that users have a better environment to start with.
At a later time we could then drop requests/ from the fishtest tree and the standard requests library would take its place.

Stefano80:
Would it be possible to add the link in the error message the best people are getting?

tomtor:
We can only fix:

    f.write(requests.get(WORKER_URL).content)
    NameError: global name 'requests' is not defined

Not an error like:

    SSLError: [Errno 1] _ssl.c:504: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert 
     protocol version

which is caused by old OS versions

Stefano80:
Ok, I understand, thank you for the analysis. 

If I am not mistaken, in principle we could catch those two errors in `main` and rethrow them adding the hint to a fresh install. Is that possible?

ppigazzini:
dev-server and worker updated: all workers are running fine.

tomtor:
> If I am not mistaken, in principle we could catch those two errors in main and rethrow them adding the hint to a fresh install. Is that possible?

Regarding the second error, that could add some value for the end-user, but the only report has been from @snicolet.

I don't see any reason to provide code for the first error, because how do we add that code to main() without a working updater?

If a user has a working updater/or fresh install then the first error will not occur anyway...

Stefano80:
Hi @tomtor , thank you for the answer. I see that my reasoning has some weak points. 

I try to elaborate: f I understand correctly, the users see an error when `update` is run, called by the worker `main`. Thus, if we [here](https://github.com/glinscott/fishtest/blob/17210c6d7ce1670f4a6f070c8950fbdabd7d2feb/worker/worker.py#L75) add a meaninfgul `try/catch` block, we could hint the user. I am thinking on something on the lines
```
      printout('Updating worker version to %d' % (req['version']))
      try:
          update()
          printout("Worker version checked successfully in "+str((datetime.utcnow()-t0).total_seconds())+"s")
     catch:
          printout("Failure in the worker update. If problem persists a fresh install following LINK could help. If the problem still persists, please open an issue on LINK")
```
And then we can either rethrow the error or just carry on with the old worker version. Any thoughts on that?

tomtor:
@Stefano80 I see, yes that is possible ofcourse...

Stefano80:
Hi, ok for me, I will be available from 20:00 CEST, but if you are monitoring it maybe makes sense to start earlier.

tomtor:
Allthough we have currently a non functioning updater, all worker changes are backward compatible (no non-backward compatible API changes), so there is no real need to bump the VERSION.

If we DO bump the version we would have to inform users that they have to patch by end.
If we DO NOT bump the version, many workers will not pick up the fixes.

Perhaps we could do it in two stages. Merge now (advanced users will fetch from github) and announce on the Forum the update procedure by hand. Bump the version in 7 days.

Stefano80:
This sounds sensible.

tomtor:
@Stefano80 @ppigazzini  It just occured to me (perhaps this was obvious to you) thaf we should raise the version number in the worker now and after 7 days in the server...

So we can see who updated by hand already.

tomtor:
This is also interesting (for later): https://stackoverflow.com/questions/12332975/installing-python-module-within-code?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

ppigazzini:
@tomtor I don't know the code server side, but I just tested that a worker is able to works in fishtest with a version number higher respect to the server one.
So you are proposing:
* today we will raise the worker version, so we can view from the web page the users who update the worker w/ git. 
* next days we will try to push the other users to update the worker by hand (forum)
* next week we will raise the server version: the workers with a lower version number will stop working until an updated by hand

To me this is fine.

ppigazzini:
@tomtor regarding your link, I don't know if pip is available on our windows setup (python by python.org or by msys2)

tomtor:
@ppigazzini https://github.com/msys2/msys2/wiki/Packages lists:
- mingw/mingw-w64-python2-requests
- mingw/mingw-w64-python2-pip

and pip is standard Python2: https://docs.python.org/2/installing/index.html

I expect that `python -m ensurepip` will install pip on a Python2 Windows installation, but probably it is already available and you can just use `python -m pip install requests`

ppigazzini:
@tomtor 

* msys2 python2:
```
$ python2
Python 2.7.13 (default, Feb 14 2017, 14:46:01)
[GCC 6.3.0] on msys
>>> python -m ensurepip
  File "<stdin>", line 1
    python -m ensurepip
                      ^
SyntaxError: invalid syntax
>>> python -m pip install requests
  File "<stdin>", line 1
    python -m pip install requests
                ^
SyntaxError: invalid syntax
>>>
```

* python.org
```
Python 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)] on win32
>>> python -m ensurepip
  File "<stdin>", line 1
    python -m ensurepip
                      ^
SyntaxError: invalid syntax
>>> python -m pip install requests
  File "<stdin>", line 1
    python -m pip install requests
                ^
SyntaxError: invalid syntax
>>>
```

From the [Zen of Python](https://www.python.org/dev/peps/pep-0020/):
* There should be one-- and preferably only one --obvious way to do it.
* Although that way may not be obvious at first unless you're Dutch.

If I'm not wrong you are Dutch, so please explain. :)

tomtor:
@ppigazzini You are entering your commands in the Python interpreter, you should enter them on the commandline/shell

ppigazzini:
@tomtor 
msys2
```
vmuser@WIN-7BP4TF83HU9 MINGW64 ~
$ python2 -m pip install requests
/usr/bin/python2: No module named pip

vmuser@WIN-7BP4TF83HU9 MINGW64 ~
$ python2 -m ensurepip
Collecting setuptools
Collecting pip
Installing collected packages: setuptools, pip
Successfully installed pip-9.0.1 setuptools-28.8.0

vmuser@WIN-7BP4TF83HU9 MINGW64 ~
$ python2 -m pip install requests
Collecting requests
  Downloading https://files.pythonhosted.org/packages/49/df/50aa1999ab9bde74656c2919d9c0c085fd2b3775fd3eca826012bef76d8c/requests-2.18.4-py2.py3-none-any.whl (88kB)
...
```

tomtor:
@ppigazzini That looks better :-)
Could you test the one-click-installer?

ppigazzini:
@tomtor one-click-installer already downloaded. It seems that I haven't the permission to upload the one-click-installer on s3, though.

tomtor:
@ppigazzini I think the current one-click-installer image is already OK and has pip installed....

ppigazzini:
@tomtor Yes, python from the one-click-installer is OK:
```
C:\Fishtest-Worker-Portable\Python27>python -m ensurepip
Ignoring indexes: https://pypi.python.org/simple
Requirement already satisfied (use --upgrade to upgrade): setuptools in c:\fisht
est-worker-portable\python27\lib\site-packages
Requirement already satisfied (use --upgrade to upgrade): pip in c:\fishtest-wor
ker-portable\python27\lib\site-packages

C:\Fishtest-Worker-Portable\Python27>python -m pip install requests
Collecting requests
  Downloading https://files.pythonhosted.org/packages/49/df/50aa1999ab9bde74656c
2919d9c0c085fd2b3775fd3eca826012bef76d8c/requests-2.18.4-py2.py3-none-any.whl (8
8kB)
...
```

tomtor:
@ppigazzini Thanks for testing! We could use this in a future version of updater.py

ppigazzini:
@tomtor OK, I'm merging the already approved worker PRs, then I will update only the worker version.

ppigazzini:
@tomtor I prepared a new Fishtest-Worker-Portable archive, with new packages (python3) and a revised installer logic:
 - download fishtest from github (and delete "requests" folder)
 - install an updated "requests" packages

I'm waiting the s3 access key to replace the old files.

PR #269, opened 2018-04-29
Unique_key from task + 2 fixes

tomtor:
A small optimization for #261 

In addition add a lock to get_worker_key() because it has a race condition.

Fix print issue introduced in #162

8a2da7b

PR #270, opened 2018-04-30
Fix SPSA games end time calculation

tomtor:
This caused premature worker exits


ppigazzini:
dev-server and workers updated with #267 #269 #270, should be fine.

tomtor:
@ppigazzini Made a change to fix, please reload, sorry...

ppigazzini:
Done, no problem. Travis restarted twice by hand.

PR #272, opened 2018-04-30
[WIP] Catch kill process error

ppigazzini:
Print an error message and pass the execution if the kill of the process fails.
The kill of the process is used as safety measure against a hanging process.
At the moment all the running processes seems to run and close fine,
so we are always trying to kill an already closed process.
With a WSL worker this is a problem during SPSA test,
because after the exception the execution is stopped for some seconds.


ppigazzini:
This is for discuss the problem.

From my first local tests:
* linux worker: `p.kill()` seems to handle nicely the case of killing an already closed process, no system error raised at all
* windows worker + msys2: the 'try: except: pass' seems not able to catch the system error, but fortunately there is not a time out of the execution during the SPSA test
* WSL worker: the 'try: except: pass' works nicely, catching the error and passing the execution

ATM the WSL worker runs the same code of the Linux worker.

tomtor:
Do we have any idea why this:

  File "/home/travis/build/glinscott/fishtest/worker/test_worker.py", line 47, in test_item_download
    games.setup('polyglot.ini', '.')
  File "/home/travis/build/glinscott/fishtest/worker/games.py", line 112, in setup
    for blob in tree['tree']:
KeyError: 'tree'

fails 9 of 10 times in Travis?

vondele:
My guess this is the git api not returning what we expect ? Presumably some error message after having used it too often ? Maybe we should catch the error, and print the content of 'tree'.

tomtor:
It's probably just the rate limiting, but for some reason the server Travis tests API calls are a bit more relaxed then this one.

ppigazzini:
@tomtor can you review this PR? I'm not Dutch ;-)

tomtor:
@ppigazzini I'll try ;-)
Note that @vondele also speaks Dutch (but is from Belgium) so maybe he can understand Guido van Rossum's way of thinking too ;-)

vondele:
@tomtor right... but I'm not very pythonic (yet?). 

Just a minor nit from my side... I'd just change the error message to a note (since this is not an error people should worry about, and it is recurring on fishcooking):
printout('Note: ' + str(e) + ' killing the process pid: ' + str(p.pid) + ', possibly already terminated')

PR #275, opened 2018-05-01
Make the worker Python3 compatible

tomtor:
because Python2 will be EOL 1 jan 2020

Also make the updater.py runnable stand alone: "python updater.py"

To my surprise very few changes to the code were needed.

Tested with python3 AND python2:
- Downloading testing/ contents
- building
- running runs
- updating...

And fix the Travis worker_test for when we hit the github.com API limit
and add an `updater.py` Travis test.

Show Python version number to server.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from __future__ import print_function
 
 import datetime
 import json
@@ -25,7 +26,7 @@ try:
 except ImportError:
   from queue import Queue, Empty  # python 3.x
 
-# Global beacuse is shared across threads
+# Global because is shared across threads
 old_stats = {'wins':0, 'losses':0, 'draws':0, 'crashes':0, 'time_losses':0}
 
 IS_WINDOWS = 'windows' in platform.system().lower()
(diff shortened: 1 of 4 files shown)


tomtor:
Merging this makes sense because users can no longer use a wrong Python for running the worker.

ppigazzini:
@tomtor where applicable is better to use `printout()`, correct?

tomtor:
@ppigazzini Using printout() was probably needed on Windows to prevent mixing output from cutechess and the Python program. The print statements that were not changed to printout ()  by @xoto10 are executed when cutechess is not running? Not sure if changing those is good style/needed...

ppigazzini:
The message of @xoto10 in #162 was "replace each original print command with the new printout() function." . I'm asking simply because I would like to know when using printout() (and why) and when using print().

xoto10:
Perhaps I've missed the first part of this thread. My aim with the
printout() function was to call flush() after each print so that progress
messages were not delayed by buffering. The simplest way was to create a
new function which included the flush() and then change every print "..."
into printout(...). I coded and tested this in Ubuntu.
Did I leave some prints unchanged? I don't remember doing that
deliberately. It's possible I missed them by accident, although searching
for print is easy so I don't know how I managed that.

tomtor:
@xoto10 I see. Normally on Linux terminal IO is line buffered, so I thought that it was needed for Windows. I doubt if it makes a difference on any regular OS.

tomtor:
@ppigazzini This is reliable for Python2 but I discovered that with Python3 the subprocess communication is unreliable...

tomtor:
@ppigazzini If you have some spare time (absolutely no need to hurry), could you test this on Windows? Linux works OK (Python2 and Python3) in my local testing setup...

ppigazzini:
@tomtor Test started w/ the dev-server (master) and:
* your branch [Worker3](https://github.com/tomtor/fishtest/archive/Worker3.zip)
* windows worker: python 3.6.5 by python.org and mingw_w64 by msys2
* WSL worker (ubuntu 16.04 based): python 3.5.2

ATM the workers are running fine.

tomtor:
@ppigazzini Yes, we are killing sometimes twice, because it hard to determine the exact state and if we miss a kill we have a looping python interpreter. 

Note that a standard way to test for a process existence is to see if you can kill it/send it a 0 signal, so it is quite clean.

I silenced the warning in the last commit...

ppigazzini:
@tomtor on windows worker the message "ERROR: The process "7912" not found." is not silenced. On WSL worker I restored the output to inspect the double killing.

tomtor:
@ppigazzini I removed a call to kill() after normal exit and restored the warning...

ppigazzini:
@tomtor if this is ready please squash the commits.

Issue #277, opened 2018-05-02
Main web page is stuck


Stefano80:
As described here, at the moment is not possible to update the main page, although everything else is fine

https://groups.google.com/forum/m/?fromgroups#!topic/fishcooking/jf63d0tysFw

Stefano80:
Interestingly, this is stuck in the same place from all devices from which I tried to access it. It froze when the cluster test was at

LLR: 1.86 (-2.94,2.94) [-3.00,1.00]
Total: 139951 W: 28293 L: 28452 D: 83206 

The test has passed already 

http://tests.stockfishchess.org/tests/view/5ae79a150ebc5926dba90de5

Stefano80:
The user pages are ok, and it is possible to look and approve tests by going through the users.

Stefano80:
It also has some random unresponsiveness when going to specific pages.

Stefano80:
Restarted the server, it looks like it helped, I let it open for a post-mortem.

noobpwnftw:
https://github.com/glinscott/fishtest/blob/1f5fb49db60d0cca7fc0139bee5551fbea5ee66e/fishtest/fishtest/views.py#L1015
Where is the acquire() for this release()?

tomtor:
@noobpwnftw line 878

https://github.com/glinscott/fishtest/blob/1f5fb49db60d0cca7fc0139bee5551fbea5ee66e/fishtest/fishtest/views.py#L878

noobpwnftw:
It happens again at this moment.

ppigazzini:
production server rebooted

tomtor:
I see, it happens when a new batch of workers is allocated. This happens much faster now because we don't have to fetch the user from MongoDb. I expect something goes wrong with the MongoDb driver.

tomtor:
You could comment out this line and reboot:
Line 199: rundb.py

    def get_unfinished_runs(self):
      #with self.run_cache_write_lock:
      return self.runs.find({'finished': False},

tomtor:
@Stefano80 @ppigazzini  See previous comment

ppigazzini:
@tomtor Done.

tomtor:
@ppigazzini Thanks!
It happened at 02-05-18 14:44:55 when SC approved http://tests.stockfishchess.org/tests/view/5ae9ce0e0ebc590b101c0e8f
[...]

PR #278, opened 2018-05-03
Catch exception when building Overview

tomtor:
and release the semaphore

Should reduce the impact of transient errors, but does not fix the root cause.

See #277 

Note that diff looks impressive, but "git diff -w" shows that only a "try/except" was added...


ppigazzini:
production server updated, finger crossed.

tomtor:
@ppigazzini Thanks, I am very interested if an (mongodb) exception shows up in the pserv log

PR #279, opened 2018-05-04
Fix race condition on worker_info

tomtor:
Do not make active until worker_info set.

Relates to #277


ppigazzini:
Production server updated.
Thank you @tomtor

PR #280, opened 2018-05-04
Reduce mongodb write pressure

tomtor:
and cache non dirty runs for 60s (was 10s)

See  #277


ppigazzini:
dev-server and workers updated for a fast test with #279 and #280

ppigazzini:
Production server updated.
Thank you @tomtor

tomtor:
@ppigazzini Thanks! Let's monitor the log for one or two days and if all looks quiet (no exceptions) then the cron reboot can be removed.

PR #281, opened 2018-05-04
Flush pserve buffered output

ppigazzini:
pserve log is always empty.
Flush the pserve buffered output according
this [stackoverflow post](https://stackoverflow.com/questions/25674613/python-nohup-out-dont-show-print-statement)

While here use /bin/sh for better generality.


ppigazzini:
I'm already testing this on the production server, hand made change.
I wait anyway review and suggestions.

PR #284, opened 2018-05-06
Fix error for stop/delete when not logged in

tomtor:
Follow up to #245


vondele:
somewhat related, if I press approve when not logged in, I'm redirected to login, but after logging in I'm back to the main screen, without the test being approved.

tomtor:
@vondele Yes, I know that is hard(er) to fix...

ppigazzini:
dev-server and workers updated.

tomtor:
The bug behaviour is the same as in #245. I checked on the dev server, it is OK to merge.

ppigazzini:
Production server updated.

PR #285, opened 2018-05-07
Drop /bin/zsh and use /bin/sh

ppigazzini:
Use /bin/sh in scripts to simplify the server requirements.

Add the line buffering for the nohup output in the commented out instruction.

### DIFF
--- a/fishtest/start_dev.sh
+++ b/fishtest/start_dev.sh
@@ -1,4 +1,4 @@
-#!/bin/zsh
+#!/bin/sh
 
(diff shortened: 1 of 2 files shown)


ppigazzini:
@Stefano80 Already tested on the dev-server and on the production server (hand made changes).

ppigazzini:
@Stefano80 On the production server I changed the user default shell to /bin/bash, added the needed variables in .bashrc, copied the .zsh_history content in .bash_history. On CentOS /bin/sh is a symbolic link to /bin/bash.

Issue #287, opened 2018-05-08
[Improvement] Add link to 'my tests' page


mrobaczyk:
Currently i have to find myself in 'Users' and click on number of my tests. There is link 'Profile' but it leads to password change...
Cheers

ppigazzini:
#288 merged and production server updated.
Thank you @tomtor

PR #288, opened 2018-05-08
Add link to user tests on Profile page

tomtor:
See #287


ppigazzini:
Production server updated.

Mmmm, Travis keeps failing with:
```
File "/home/travis/build/glinscott/fishtest/worker/games.py", line 112, in setup
    for blob in tree['tree']:
KeyError: 'tree'
```

Issue #293, opened 2018-05-14
highlight your machines and show limit info when logging on fishtest


candirufish:
It is quite a hassle if you run many machines on fishtest and searching yours in the list and seeing if it is running ok. I suggest highlight all of them when user is logged and put them on top of the list together with info "total machines running", "total CPUs" etc...

Another thing is if you're a newbie, you might not know there is a limit imposed on the number of machines you can put on fishtest and might think it is some error or fishtest is broken and might thwart that contributor from putting more cpus. I guess displaying this limit info when logging in is better with a button messaging admin to raise it.

tomtor:
We show machine-limit on the user profile page for approvers. I agree that it makes sense to show that info for all users.

The overview is shared/cached between all users, but a sort button is possible.

PR #295, opened 2018-05-14
Add sorting to table headers

tomtor:
Other part of #293


tomtor:
Does not work on mobile. Will have a look at that...

tomtor:
It DID work on mobile after all, but I changed it so sorting works on any table on each page by clicking the table headers.

ppigazzini:
@tomtor let me know if this is ready to be merged.

tomtor:
@ppigazzini Ready

ppigazzini:
Production server updated.
@tomtor Thank you!

PR #298, opened 2018-05-17
Fix worker empty book download

tomtor:
See #296

PR #294, opened 2018-05-14
Show more info on user profile page

tomtor:
Implements part of #293

PR #282, opened 2018-05-05
Implement PGN upload/download and database clone

tomtor:
See #140 

PGN-files are stored gzip compressed in a new MongoDb collection ('pgns') with as only additional attribute the creating run_id-task_id.

Currently only a minimal API to retrieve the PGNs and a link in the task list.

About 200 kbyte compressed storage for a 250 game task (compression to 25% original size).

Also add a simple clone script which uses the REST APIs to make a copy of the run and PGN data in a fishtest database.

PR #299, opened 2018-05-18
Fix LTC tag for generated test info

### LINKED ISSUE #297: Unexpected behaviour in automatic submit message
I just noticed that the prefix "LTC:" is always used for non STC, even in the case in which the time control is shorter, see http://tests.stockfishchess.org/tests/view/5afdd81d0ebc591fdf408f2d

tomtor:
Fix #297

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -349,7 +349,7 @@ def validate_form(request):
       if m:
         data['new_signature']= m.group(2)
     if len(data['info']) == 0:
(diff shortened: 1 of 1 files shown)


ppigazzini:
dev-server and workers updated, this PRs under test: #282 #294 #295 #298 #299

tomtor:
Thanks, I tested the new regexp, so this PR should be ready to merge too...

ppigazzini:
Production server updated.
@tomtor Thank you!

PR #305, opened 2018-05-25
Force rebuild of main page after change

tomtor:
In some rare cases an approved test was still shown as unapproved to the approver


ppigazzini:
Production server updated.
Thanks you @tomtor

Issue #304, opened 2018-05-24
username 2 letters


mrobaczyk:
Just a quick question: Is there any reason we are displaying only 2 letters of an username on overview page?

Cheers 
Mike

tomtor:
To safe display space. Those two letters where fine when there were only a few workers and testers. I agree it is a bit restricting now and  some contributors share the same two letter prefix.

mrobaczyk:
Yes, some testers have the same prefix, so maybe it's time to reorganize the space a little bit and expand username column :) 
Mike

tomtor:
Expand it to 4 or even 8 characters? I suggest 4 and a mouse-over which shows full name. I will write a PR for this in the near future unless someone has a better proposal.

tomtor:
I tried 4 but found that visually not pleasing, so I changed it to 3. Combined with the mouse-over that should be sufficient.

mrobaczyk:
Don't we have space for entire user name? They usually are not so long :) This would be the easiest solution :)

ppigazzini:
@mrobaczyk the PR #306 shows the complete username with a mouse-over popup

PR #306, opened 2018-05-25
Usernames from 2 to 3 characters

tomtor:
and add username mouse-over popup

See #304

### DIFF
--- a/fishtest/fishtest/templates/run_table.mak
+++ b/fishtest/fishtest/templates/run_table.mak
@@ -65,7 +65,7 @@
     </td>
     %endif
     <td style="width:6%">${run['start_time'].strftime("%y-%m-%d")}</td>
-    <td style="width:2%"><a href="/tests/user/${run['args'].get('username','')}">${run['args'].get('username','')[:2]}</td>
+    <td style="width:2%"><a href="/tests/user/${run['args'].get('username','')}" title="${run['args'].get('username','')}">${run['args'].get('username','')[:3]}</td>
     <td style="width:16%"><a href="/tests/view/${run['_id']}">${run['args']['new_tag'][:23]}</a></td>
     <td style="width:2%">${base.diff_url(run)}</td>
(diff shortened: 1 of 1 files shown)


ppigazzini:
dev-server and workers updated w/ PRs #305 and #306

tomtor:
@ppigazzini Stil just 2 characters shown, so the PR is not active?

ppigazzini:
@tomtor here the `fishtest/fishtest/templates/run_table.mak` on the dev-server

```
    <td style="width:6%">${run['start_time'].strftime("%y-%m-%d")}</td>
    <td style="width:2%"><a href="/tests/user/${run['args'].get('username','')}" title="${run['args'].get('username','')}">${run['args'].get('username','')[:3]}</td>
    <td style="width:16%"><a href="/tests/view/${run['_id']}">${run['args']['new_tag'][:23]}</a></td>
 ```

**Comment by tomtor** [comment]
@ppigazzini Yes, it's OK now...

**Comment by ppigazzini** [comment]
@tomtor tried with a reboot of the VM. Have you any idea because this can have worked?

**Comment by tomtor** [comment]
@ppigazzini Running "setup.py install" and restarting the Python process should be sufficient. You did the first step, so for some reason your restart procedure does not kill the running Python. I suggest just checking the PID of Python before and after the restart attempt to see if it changed. If not just kill it with "kill PID". The same happened to the test of #302 and probably #300

**Comment by ppigazzini** [comment]
@tomtor I don't understand why the fishtest stopping was commented out in the script, it makes absolutely no sense ...
The problem should be solved now.

```
#!/bin/bash
# to update a fishtest server simply run:
# sudo -H bash update-fishtest.sh 2>&1 | tee update-fishtest.sh.log
#
# to use fishtest open a browser and connect to:
# http://<ip-address>

STAMP=$(date +%Y%m%d%H%M%S --utc -d '1 hour')

#systemctl stop fishtest
```

ppigazzini:
Production server updated.
Thank you @tomtor

PR #310, opened 2018-05-26
Stricter check on username special characters

### LINKED ISSUE #309: Error on pending user
http://tests.stockfishchess.org/pending gives currently an error when selecting the single pending user.

@ppigazzini what shows the server log?

tomtor:
Fix #309


ppigazzini:
@tomtor dev-server updated, added your DNS entry to recaptcha.
The PR seems OK to me, I wait your OK to merge it. 

Tested that scavenge.py is able to delete the bad usernames.
As note I set a password with a trailing space character, I was able to authorize the user and to login with that password.

PR #311, opened 2018-05-28
Add worker options

tomtor:
for min_threads, max_memory and max_concurrency

PR #312, opened 2018-05-30
Pass on failing to remove an old binary.

vondele:
As discussed here:

https://groups.google.com/d/msg/fishcooking/fDE27JcbG1g/BKKMZS5tAQAJ

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -383,7 +383,11 @@ def run_games(worker_info, password, remote, run, task_id):
   if len(engines) > 25:
     engines.sort(key=os.path.getmtime)
     for old_engine in engines[:len(engines)-25]:
-      os.remove(old_engine)
+      try:
+         os.remove(old_engine)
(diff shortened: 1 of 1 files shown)


ppigazzini:
dev-server and workers updated for a fast test (I will lower by hand on a worker the limit of 25 engines).

ppigazzini:
Production server updated with PRs #312 and #311 .
The latest PR also raised the worker version number to 67.
Thank you @vondele

vondele:
@ppigazzini out of curiosity ... did you do something to trigger the failure to remove a file ? I still have no idea why that would be the case .

ppigazzini:
@vondele I simply started the binary in a console to trigger the failure.

vondele:
OK... yes, that would be a reason, and it is strange that users are seeing this, if they are not running two workers in the same dir. Anyway, this works around.

PR #315, opened 2018-06-05
Keep the zipped MongoDB dump

ppigazzini:
The size of the DB is increasead, so keep locally a zipped copy of
the MongoDB dump in order to have more free HD space for the server maintenance.
This speed up also the copy of the DB on the development server.

### DIFF
--- a/fishtest/utils/backup.sh
+++ b/fishtest/utils/backup.sh
@@ -1,13 +1,17 @@
 #!/bin/sh
 
+# Load the AWS variables when scheduled with cron
+source ~/.bashrc
+
 cd ~/backup
-~/mongodb/bin/mongodump
-rm -f dump.tar.gz
-tar cvzf dump.tar.gz dump
+~/mongodb/bin/mongodump && \
+rm -f dump.tar.gz && \
+tar -czvf dump.tar.gz dump && \
+rm -rf dump
 
 DAY=$(date +%Y%m%d --utc -d '1 hour')
 mkdir -p archive/$DAY
 mv dump.tar.gz archive/$DAY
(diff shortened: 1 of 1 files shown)


ppigazzini:
Tested working on the production server (change by hand).

tomtor:
@ppigazzini So S3 still has a dump for each day?

ppigazzini:
@tomtor During some tests `s3put` stopped working with file bigger than 5 GB. With #316 the dump.tar.gz is 3.7 GB, so we are able to upload to S3 each hour (but we keep only one dump/day). Unfortunately `fetch_file` stops during the download of big file from S3, I don't know the working limit.

ppigazzini:
@tomtor I found the bug (introduced by me...) that stopped the upload to S3 from the 08.05.2018.
I already changed the file by hand on the production server and I will keep this PR open to collect suggestions.
There is still the problems of:
* upload : 5 GB size limit with `s3put`
* download: unknown size limit with `fetch_file`

tomtor:
@ppigazzini I suggest testing recent software: eg https://github.com/boto/boto3

ppigazzini:
@tomtor the update of boto from 2.9.5 to 2.48.0 not solved the limit problems.
I will test boto3 on the dev server.

tomtor:
See https://www.quora.com/How-will-you-upload-a-file-greater-than-100-megabytes-in-Amazon-S3

ppigazzini:
@tomtor thank you.
I installed a [dev-server on CentOS 6.9](http://94.198.98.16:6543/tests) with the same packages of the production server.
As first step I want to test the mongoDB upgrade.

PR #317, opened 2018-06-05
Break long words in table info

tomtor:
Fix layout on narrow devices


ppigazzini:
dev-server updated.

tomtor:
@ppigazzini Ok to merge. You should look in pending tests to the right most info column.  When there is a long link in the text (a reference to a commit) then not all text is shown (and no scrollbar is added) on a narrow display without this patch...

tomtor:
@ppigazzini See fixStatsOverflow in production.

ppigazzini:
Production server updated.
Thank you @tomtor

PR #322, opened 2018-07-09
Limit num_games to max 1000000

tomtor:
to prevent server overload

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -425,6 +425,10 @@ def validate_form(request):
     if data['num_games'] <= 0:
       raise Exception('Number of games must be >= 0')
 
+  max_games = 4000 * request.rundb.chunk_size
+  if data['num_games'] > max_games:
+    raise Exception('Number of games must be <= ' + str(max_games))
+
   data['threads'] = int(request.POST['threads'])
   data['priority'] = int(request.POST['priority'])
   data['throughput'] = int(request.POST['throughput'])
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -487,6 +487,11 @@ def tests_modify(request):
       request.session.flash('Unable to modify number of games in a fixed game test!')
       return HTTPFound(location=request.route_url('tests'))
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
On the dev server the control is working for "Fixed games tests", but I was able to schedule a 5.000.000 games SPSA.
Is it working as intended?

tomtor:
@ppigazzini Yes, you're right. I pushed a new version which should work for all cases.

Thanks for testing.

ppigazzini:
@tomtor now it works also for SPSA. I wait a couple of hours before merging.

tomtor:
@ppigazzini We still can MODIFY the number over 1000000. Not sure if we should prevent that as well. I think the most important is preventing new tests, but I could add an additional check for modify...

ppigazzini:
@tomtor I lack the SF testing experience to state if the default limit of 1e6 games is fine and if/when should be necessary to modify that limit. @Stefano80 @snicolet are more indicated than me to make an educated review.

ppigazzini:
@tomtor dev-server updated with "Prevent modify num_games > 4000 chunks"

tomtor:
@ppigazzini I modified a test to 1250000 before the last commit was in place. If you can modify a new test > 1000000 then something is wrong...

tomtor:
@ppigazzini 

> IMO is more clean to have only one value for all the test types.

I fully agree, but I cannot reproduce setting SPRT to 1,250,000. A new test has 250,000 as default for me.

Stefano80:
Hi, I think we never had a SPRT over 1e6, but some on the range of 500k. With increasing strength this could become a problem.

I also think that 1e6 spsa is more than enough. I would go for 1e6 limit, and, if this is somehow possible, prevent modification.

I mean, we could ever prevent modification of number of games altogether and be done with it. I cannot imagine a case in which this could really help.

ppigazzini:
@tomtor correct, 250,000 is the SPRT default. The 1,250,000 should be a byproduct of my multitasking... Sorry for the noise.

ppigazzini:
Production server updated.
Thank you @tomtor

PR #325, opened 2018-07-15
Delete the pgns collection from the backup

ppigazzini:
The pgns collection has a big size.
This can trigger:
* the bandwidth limit of the production server
* the size limit of the boto package used to upload to s3


ppigazzini:
Tested working fine in the last month.
Production server updated.

PR #341, opened 2018-12-18
Update default hash size for STC

xoto10:
This makes the default hash size 8 for STC.
I have not tested this as my local fishtest is not working currently but the change is minimal and hopefully easy enough for others to check.
Let me know if I need to do anything else.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -17,7 +17,7 @@
   <div class="control-group">
     <label class="control-label">Test options:</label>
     <div class="controls">
-    <input name="new-options" value="${args.get('new_options', 'Hash=4')}">
+    <input name="new-options" value="${args.get('new_options', 'Hash=8')}">
     </div>
   </div>
   <div class="control-group">
@@ -35,7 +35,7 @@
   <div class="control-group">
     <label class="control-label">Base options:</label>
     <div class="controls">
-    <input name="base-options" value="${args.get('base_options', 'Hash=4')}">
+    <input name="base-options" value="${args.get('base_options', 'Hash=8')}">
     </div>
   </div>
   <div class="control-group">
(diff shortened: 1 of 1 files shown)


ppigazzini:
@mcostalba @snicolet @tomtor @vondele this patch was tested successfully on the dev server with some workers and is ready to be merged. Let me know if is fine to change the hash default.

mcostalba:
@ppigazzini I don't know what is the rationale7advantages behind this change, maybe someone cold explain me.

Chess13234:
@mcostalba Less hash pressure, more ttHits, adds 4elo.
It allow for more fair tests, closer to real world conditions, where no one uses 4mb except for hypershort time controls like game in 1 second.

mcostalba:
What was the reason we moved to current limits and why those reasons are no more valid now?

I am not against, but I'd like people to understand that what we have now is the result of a past analysis. Maybe is no more valid now, but we need to know it to make conscious choices.

Chess13234:
@mcostalba They were raised to current limits from 2/32 mb
 https://github.com/glinscott/fishtest/commit/5a3353e5b71e3a1343994258e2ea0f19f8e9027f

CoffeeOne:
@mcostalba 
The current setting was a result of test series done by user lucasart.
Goal was too find the lowest possible TT size, which does not lose Elo.
So the size was reduced and it was tested always with [-4, 0]. It ended up with the TT size of 4MB for STC and 64MB for LTC. 4 MB is no more OK for STC, it's too low with current stockfish. I made a test with TT size of 8MB against 4MB [0, 4] which passed, so the opposite test that was made when it was introduced.

mcostalba:
Ok thanks guys for the explanations. I am fine with this merging. Thanks

ppigazzini:
Patch merged and production server updated. Thank you for the good work guys.

PR #342, opened 2018-12-23
Change SPRT parameters for standard test to [0.5, 4.5] + [0,3.5]

mcostalba:
Following the discussion here https://github.com/official-stockfish/Stockfish/issues/1859, we aim at

- Improve precision, i.e. discard zero or negative patches and accept positive ones with higher margin

- Manage resource consumption.

After various simulations, here are results for different sets of SPRT parameters:

| Limits | [0,5] + [0,5]  | [0.5, 4.5] + [0,3.5] | [1,4] + [0,3]  | [0.5, 4] + [0, 3] | [0.5, 4] + [0, 3.5] | [0.4, 4.4] + [0, 3.2] |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 0 ELO pass prob | 0.0025 | 0.00123 | 0.00036 | 0.0011 | 0.0011 | 0.0014 |
| 1 ELO pass prob | 0.0744 | 0.1002 | 0.0805 | 0.1451 | 0.1183 | 0.1276 |
| +1 ELO rejectance ratio | 0.8201 | 0.7547 | 0.7625 | 0.6718 | 0.7049 | 0.7137 |
| +1.5 ELO rejectance ratio | 0.6310 | 0.5160 | 0.4865 | 0.3968 | 0.4249 | 0.4679 |
| +2 ELO rejectance ratio | 0.3936| 0.2709 | 0.2026 | 0.1700 | 0.1819 | 0.2374 |
| total ELO gain ratio | 1.0 | 1.3192 | 1.2373 | 1.7596 | 1.5664 | 1.5523 |
| -0 ELO acceptance ratio | 2.5 e-04 | 9.1e-05 | 2.2e-0.5 | 1.0 e-04 |  7.77 e-05 | 1.0 e-04 |
| Avg. STC cost | 18431 | 24456 | 34574 | 30936 | 30936 | 25110 |
| Avg. STC + LTC Cost | 27931 | 38039 | 45286 | 50604 | 46093 | 42990 |

From the table above, there is consensus that [0.5, 4.5] + [0, 3.5] pair is the best balance between improved sensitivity and resource consumption.

This patch makes the above parameters the default.

A big thank to all the people involved, particularly to @vondele and @vdbergh for developing and tuning the simulation tool, @Alayan-stk-2, @Chess13234, @xoto10, @Vizvezdenec and @NKONSTANTAKIS for the great and high quality discussion.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -64,7 +64,8 @@
     <label class="control-label">SPRT bounds:</label>
     <div class="controls">
       <select name="bounds">
-        <option value="standard">Standard [0,5]</option>
+        <option value="standard">Standard STC [0.5, 4.5]</option>
+        <option value="standard">Standard LTC [0, 3.5]</option>
         <option value="tweak">Parameter Tweak [0,4]</option>
         <option value="regression">Non-regression [-3,1]</option>
         <option value="simplification">Simplification [-3,1]</option>
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -64,8 +64,8 @@
     <label class="control-label">SPRT bounds:</label>
     <div class="controls">
       <select name="bounds">
-        <option value="standard STC">Standard STC [0.5, 4.5]</option>
-        <option value="standard LTC">Standard LTC [0, 3.5]</option>
+        <option value="standard STC">Standard STC [0.5,4.5]</option>
+        <option value="standard LTC">Standard LTC [0,3.5]</option>
         <option value="tweak">Parameter Tweak [0,4]</option>
         <option value="regression">Non-regression [-3,1]</option>
         <option value="simplification">Simplification [-3,1]</option>
@@ -205,7 +205,8 @@ Cowardice,150,0,200,10,0.0020"""})['raw_params']}</textarea>
 $(function() {
   var update_bounds = function() {
     var bounds = $('select[name=bounds]').val();
-    if (bounds == 'standard') { $('input[name=sprt_elo0]').val('0'); $('input[name=sprt_elo1]').val('5'); }
+    if (bounds == 'standard STC') { $('input[name=sprt_elo0]').val('0.5'); $('input[name=sprt_elo1]').val('4.5'); }
+    if (bounds == 'standard LTC') { $('input[name=sprt_elo0]').val('0'); $('input[name=sprt_elo1]').val('3.5'); }
     if (bounds == 'tweak') { $('input[name=sprt_elo0]').val('0'); $('input[name=sprt_elo1]').val('4'); }
     if (bounds == 'regression') { $('input[name=sprt_elo0]').val('-3'); $('input[name=sprt_elo1]').val('1'); }
     if (bounds == 'simplification') { $('input[name=sprt_elo0]').val('-3'); $('input[name=sprt_elo1]').val('1'); }
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -64,8 +64,8 @@
     <label class="control-label">SPRT bounds:</label>
     <div class="controls">
       <select name="bounds">
-        <option value="standard">Standard STC [0.5, 4.5]</option>
-        <option value="standard">Standard LTC [0, 3.5]</option>
+        <option value="standard STC">Standard STC [0.5, 4.5]</option>
+        <option value="standard LTC">Standard LTC [0, 3.5]</option>
         <option value="tweak">Parameter Tweak [0,4]</option>
         <option value="regression">Non-regression [-3,1]</option>
         <option value="simplification">Simplification [-3,1]</option>


Alayan-stk-2:
For the record : the best bounds are a factor of what testers actually submit to fishtest, of how close we are to resource limits, of the actual elo distribution of submitted patches, of how independent the patches are from each other...

The simulation tool is great, but work by @vondele and @vdbergh shows that we still have limitations in the modelling.

And of course, we don't know how tester behavior will adapt to changes.

Therefore, it is still a likely possibility that we could get even better testing parameters (the table itself could be used to argue this point with the [0.4, 4.4] + [0, 3.2] bounds example), but given the uncertainty, the smaller change that is [0.5, 4.5] + [0, 3.5] is safer when it comes to resource usage while already giving a good benefit.

In a few months, we will be able to collect new data (did we run in a resources bottleneck or not ? how much elo did it helped to gain ? how did testers adapt their behavior ? has patch elo distribution changed meaningfully (gainer are expected to be slightly rarer with time as SF gets better) ?) and will have much more solid data to test new bounds through an updated simulation tool and see if yet another change would be beneficial or if sticking to [0.5, 4.5] + [0, 3.5] is the best choice.

NKONSTANTAKIS:
I think I prefer the [0.5, 4] + [0, 3.5], higher quality standard at a higher cost but atm framework is often empty. I guess that the changes that we made are a good first step to test the waters and evaluate.

Now we need to choose new tuning bounds, appropriate with the characteristics of parameter changes.
**The main element of those is imo their risk-free nature and constantly changing optimae.**
So we do not worry at all for loading our back with obfuscation code and any concerns about statistical error should be minimal, as values will get plenty future opportunities to retune.
**A proper strategy imo would be to 1. prioritize saving resources and 2. lower the elo gaining demands**

ianfab:
@mcostalba Since the values of the bounds are not parsed from the option's html text, the corresponding javascript code also needs to be adjusted for the change to take effect, see https://github.com/glinscott/fishtest/blob/3623e83387cb6ff107d044c6d74b6f3e6879a4e4/fishtest/fishtest/templates/tests_run.mak#L206-L208

mcostalba:
@ianfab Thanks. Updated.

NKONSTANTAKIS:
@vondele I am talking about the need for new parameter tweak bounds. Its awkward to have [0.4] for tweaks and [0,3.5] for code. Imo new tweak bounds should be made easier to pass and also lower the average game usage in order to facilitate frequent adjustment.

FauziAkram:
completely agree with @NKONSTANTAKIS 
Tuning bounds should be easier to pass than code patches.

mcostalba:
I have submitted the first test with new parameters (for now setting manually):

http://tests.stockfishchess.org/tests/view/5c20a0af0ebc5902ba12cd1d

vdbergh:
I wonder if it would be possible to automatically flag LTC tests which have not passed STC? This would make it much easier to do statistical analysis afterwards (now manual inspection of descriptions is necessary, which is obviously not pleasant and also error prone).

MJZ1977:
Thanks to all contributors. I completely agree with the changes. We will see a big increase in ELO gain curve of SF I hope !

SFfan876:
[0, 3.5] for LTC very stupid. It will lead to the fact that in the future SF will be worse to scale on LTC. Bounds for LTC should be the same as for STC or even higher

MJZ1977:
I have a question to SPRT experts : often, we want just to make a test to have the trend and we hope it will not go to 200k games ... What SPRT parameters can we use to stop the test quickly ? [1..3] for example ?
I think that a good idea to save ressources is to have such a test "Fast STC" for these cases. It will not change anything in the validation procedure that will follow the normal STC and LTC steps

Alayan-stk-2:
You can stop your own tests manually.

MJZ1977:
I know that, but it is better to let fishtest do it alone : 

- to have green / yellow / red colors
- to compare the tests with the same procedure
- because we are not always following the test results behind the screen :-)

vondele:
@MJZ1977 the wider the bounds, the more quickly sprt tests terminate, of course if the bounds are far away from the actually elo value, things terminate quickly as well (e.g. [5, 10] will be quick for most patches). 

However, to get a 'quick trend' typically few games give no valid information, you can as well do a dice roll. 40k games have a +-2 Elo error, typically, what we're looking for are differences of the order of 1 Elo. If you have a single parameter to explore, best is to submit a couple of standard STC tests with various values of the param, and pick what passes... If you have multiple parameters to explore, it is very difficult to do it by hand.

MJZ1977:
40k games are fine for me. I am worried about 100k+ games.
Can I make tests with [-1..6] bounds for my own camparisons ?

vondele:
I would rather not.. will result in a mess if everybody starts doing that. I would do the normal bounds, and if you care, set the maximum number of games to e.g. 40k.

MJZ1977:
I tried one time to change the maximum number of games, but it didn't work. Perhaps a bug when test is already started.

vondele:
possible, you might not be able to reduce. It does work if you set if from the beginning (AFAICT).

ppigazzini:
@mcostalba patch tested successfully on the dev server with some workers. Let me know if it's OK to merge it.

mcostalba:
@ppigazzini Of course! Thanks a lot and Good 2019 :-)

ppigazzini:
@mcostalba merged, official server updated. I wish you (and the other guys out there improving SF) a great 2019 :-)

PR #355, opened 2019-01-19
Use the pentanomial frequencies for fixed length tests.

vdbergh:
Use the pentanomial frequencies for fixed length tests.
    
The pentanomial error bars are a bit smaller for the type of tests that are run on fishtest.
    
To correctly appreciate this one should use the observation that the amount of effort needed to reach certain error bars is _inversely quadratic_ in the size of the error bars.

### DIFF
--- a/fishtest/fishtest/stat_util.py
+++ b/fishtest/fishtest/stat_util.py
@@ -1,4 +1,4 @@
-import math
+import math,copy
 
 def erf(x):
   #Python 2.7 defines math.erf(), but we need to cater for older versions.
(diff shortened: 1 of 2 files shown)


vdbergh:
Thanks! From the 1+0.01 test you scheduled one can see that the saving in resources by using the pentanomial model is about 5% (2.5% in error bars) at this ultra fast TC. For STC it is about 10% and for LTC it is about 15%,

I am currently working on SPRT. The fishtest implementation is actually rather trivial but it is more work to correctly validate the results (a.o. not to compare apples and oranges). To do this I wrote a simulator (for empirical verification) as well as a corresponding version of sprta.py (for theoretical verification). The empirical and theoretical results match very well.

PR #357, opened 2019-01-23
Change default throughput to 10000

xoto10:
[This](https://groups.google.com/d/msg/fishcooking/BxWmrJ9yK64/ltnoKx0rFQAJ) discussion on fishcooking  suggested changing the default throughput for tests.

Possible values of 5000 or 10000 were mentioned. This patch currently changes the default to 10000, but if people want a different value, it could obviously be changed before being merged.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -170,7 +170,7 @@ Cowardice,150,0,200,10,0.0020"""})['raw_params']}</textarea>
   <div class="control-group">
     <label class="control-label">Throughput:</label>
     <div class="controls">
-      <input name="throughput" value="${args.get('throughput', 1000)}">
+      <input name="throughput" value="${args.get('throughput', 10000)}">
     </div>
   </div>
   <div class="control-group">


snicolet:
@ppigazzini
 I am fine with the change

ppigazzini:
@xoto10 to test your PR yesterday I set other runs to priority=0 and throughput=10000, then I created this new [run](http://dfts-0.pigazzini.it/tests/view/5c490c0bded3222c3d217d3b). Unfortunately is still in pending state.
I created a new SPRT run, please, have a look.

To submit/approve runs on the dev-server use these accounts (username/password):
user00/user00 - approver
user01/user01 - user

xoto10:
From a quick look, the priority of this job is lower than the other ones so
it will be last to run. This is mostly because the other jobs were created
earlier and they get priority. You can see there is a roughly 43000 gap to
the next job which was created 12 hours earlier - until workers are
allocated the priority is set from the time since the epoch. As workers are
allocated, the priority of other jobs should get lower until this job
eventually starts.
With only a few workers on dev, I think the new throughput value will seem
too high, and will generally make jobs run in start time order. On live
fishtest this is what we want to do - shift priority towards longer running
jobs.

ppigazzini:
Official server updated. Thank you @xoto10

Vizvezdenec:
tbh currently we have the opposite of problem we had before this test.
Previously test would never finish and now tests are never starting - it's pretty easy for test to drag 0 workers for like 10 hours because 2-3 tests eat all workers instead of it. It's pretty frustrating to test this way - if you want to quickly see if idea is not a complete garbage you still need to wait for 5+ hours before you get even a single core into doing it. Maybe 10k is too much?

xoto10:
Yes, I was thinking this might be the case yesterday too, but I wondered if
it was an unusual result because of the tests running at the time (a few
big LTC tests running at the same time and only 900-1000 cores instead of
the ~1200 we have had at times).

It looked to me that we had capacity to run up to 3 LTCs and a few other
tests, but at this level new tests struggled to start until existing tests
finished.

I suggest we keep an eye on it for a day or two, but if necessary a change
to a smaller value would be easy to do. If we change again, what are
candidate values, maybe 6000? (divides down easily, presumably boosts
number of tests that can run at once by 2/3?)

PR #358, opened 2019-01-23
Sprt

vdbergh:
This is a PR on top of PR #355.
```
Make SPRT(...) in stat_util.py return the pentanomial LLR if possible.
    
We also implement the usual correction for discrete time.
See David Siegmund, Sequential Analysis, Tests and Confidende
Intervals, p50.
```
The math behind this PR has been verified by simulation. See 

for an extensive discussion.

Of course I have also tested it on my local copy of fishtest.

### DIFF
--- a/fishtest/fishtest/stat_util.py
+++ b/fishtest/fishtest/stat_util.py
@@ -1,3 +1,4 @@
+from __future__ import division
 import math,copy
 
 def erf(x):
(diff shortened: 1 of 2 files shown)


ppigazzini:
@vdbergh dev server updated with all the pentanomial PRs, added some Fixed and SPRT tests, started some workers.

vdbergh:
Thx! 

On my local copy of fishtest there seems to be a test not showing pentanomial frequencies despite the fact it should have them. Unfortunately I cannot reproduce it. I think the server may have generated a spurious empty task. Is there a good way to generate debug prints from fishtest? Or else is there a way to query the database directly?

ppigazzini:
@vdbergh log the mongo output with:
```
mongo | tee mongo-`date +%Y%m%d%H%M%S`.log
use fishtest_new
db.runs.find( ).pretty()
exit
```

vdbergh:
@ppigazzini  Wow thanks! That looks great.

vdbergh:
Mystery (probably) solved. I must have accidentally ran a task with a worker from master since the db says version 67:2 for one task, and 68:2 for the other tasks. I guess I am just very tired now.

vdbergh:
BTW In case you are wondering why the SPRT(0,20) test already stopped at LLR 2.91. This is the discrete time correction kicking in. This will not be visible with the types of bounds normally used on fishtest since it will be absorbed by the rounding of the LLR to 2 decimal digits.

PR #359, opened 2019-01-24
Flush the subprocess stdin buffer

ppigazzini:
Required by python 3.x
Bug reported by Guy Vreuls

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -63,6 +63,7 @@ def verify_signature(engine, signature, remote, payload, concurrency):
     busy_process = subprocess.Popen([engine], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
     busy_process.stdin.write(enc('setoption name Threads value %d\n' % (concurrency-1)))
     busy_process.stdin.write(enc('go infinite\n'))
+    busy_process.stdin.flush()
 
   try:
     bench_sig = ''


gvreuls:
Here is a link to the FishCooking discussion: [https://groups.google.com/forum/#!topic/fishcooking/0usGsUbOyZg](https://groups.google.com/forum/#!topic/fishcooking/0usGsUbOyZg)

(It wasn't about staking a claim but about the context :) )

ppigazzini:
I followed this [stackoveflow thread](https://stackoverflow.com/questions/27571305/subprocess-stdin-buffer-not-flushing-on-newline-with-bufsize-1) to fix the bug.
An alternative fix could be to call popen with unbuffered write/read avoiding the flush():
```
busy_process = subprocess.Popen([engine], stdin=subprocess.PIPE, stdout=subprocess.PIPE, bufsize=0)
```

From the [popen documentation](https://docs.python.org/3.7/library/subprocess.html) it seems than different versions of python have different defaults for bufsize, for python 2.x the default should be bufsize=1 (i.e. buffered read/write ...)

```
bufsize will be supplied as the corresponding argument to the open() function when creating the stdin/stdout/stderr pipe file objects:

0 means unbuffered (read and write are one system call and can return short)
1 means line buffered (only usable if universal_newlines=True i.e., in a text mode)
any other positive value means use a buffer of approximately that size
negative bufsize (the default) means the system default of io.DEFAULT_BUFFER_SIZE will be used.
Changed in version 3.3.1: bufsize now defaults to -1 to enable buffering by default to match the behavior that most code expects. In versions prior to Python 3.2.4 and 3.3.1 it incorrectly defaulted to 0 which was unbuffered and allowed short reads. This was unintentional and did not match the behavior of Python 2 as most code expected
```

PR #361, opened 2019-01-26
Live elo

vdbergh:
This a PR on top of PR #358.

The live_elo page now gets its elo estimation from a new api_call "get_elo", which is almost the same as "get_run", except that the result contains an extra dictionary with elo information.

I have now put all the stats files in a separate directory "stats" to keep things tidy.

The next and last step should be to use the pentanomial frequencies in the live_elo page.


ppigazzini:
@vdbergh dev-server updated w/ all pentanomial PRs

vdbergh:
Thks!!

Issue #362, opened 2019-01-27
Optimal throughput value


Vizvezdenec:
Recently throughput was changed from 1000 to 10000 by default and it produced the opposite from previous problem - before we had a lot of tests not finishing because they don't get any workers and now we have a lot of tests that don't start because they don't get any workers because all workers are used by 2-3 tests that have like 50-100k games already but don't converge within this number of games.
So even after approval tests can get 0(!) workers for 2-3-4 hours. I think it's a serious issue because in most of the cases you need like 5-10k games to see if idea is a complete garbage or not but now you can wait like 5+ hours for your test to get this number of games even if fishtest is not on a heavy load.
Maybe 10000 default throughput is too high?

vondele:
I do agree that 10000 instead of 1000 is a big change. 

One further reason for the behavior you see is manual changes in priority or throughput. If one changes e.g. priority from -1 to 0, all the 'waiting time' at prio -1 is taken into account at prio 0, and such a test will attract all workers. Similar for changes in throughput.

vdbergh:
Long ago I proposed a credit system. Running tests would cost credit, but successful patches would earn credit (elo gainers would gain more than simplifications).  Contributing games would also earn credit. The costs would depend on the available resources (e.g. tests that are run while the framework is empty would be free).

This would a kind of regulated economy. One could also think of a true supply and demand system where people bid credits for available resources but that would be too adversarial I think.

Vizvezdenec:
> 
> 
> I do agree that 10000 instead of 1000 is a big change.
> 
> One further reason for the behavior you see is manual changes in priority or throughput. If one changes e.g. priority from -1 to 0, all the 'waiting time' at prio -1 is taken into account at prio 0, and such a test will attract all workers. Similar for changes in throughput.

Usually this changes are made to manually throw "least useful" patches into pile of idling patches. Basically yesterday I did some manual fishtest hacking to get idling for 2-4 hours some workers.

19-01-27 01:55:36 | Vizvezdenec | soft_mc | modify: priority changed from 1 to 0
19-01-27 01:45:53 | Vizvezdenec | soft_mc | modify: priority changed from 0 to 1
19-01-27 01:43:10 | Vizvezdenec | restrictedTw1 | modify: priority changed from 1 to 0
19-01-27 01:35:17 | Vizvezdenec | restrictedTw1 | modify: priority changed from 0 to 1
[...]

PR #364, opened 2019-01-30
Reduce default throughput from 10000 to 3000

xoto10:
See also the discussion #362 

The recent change to the default throughput seems to have been overdone, and older jobs are now preventing new ones from starting for extended periods.
This change reduces the default throughput so that is higher than the original 1000, but lower than the current 10000.


ppigazzini:
@xoto10 production server updated.

PR #367, opened 2019-02-02
Raise minimum worker version to 68

ppigazzini:


ppigazzini:
@noobpwnftw I raised the worker version to 68 to fix some worker bugs. It seems that your workers need a restart.

Issue #368, opened 2019-02-02
Updater code broken


ppigazzini:
Forcing a worker updated requires a worker restart by hand.

I'm collecting here the debug data to write a patch.
- TypeError: 'bool' object is not callable

Worker output:
```
~$ python fishtest-master/worker/worker.py
Worker starting ...

Memory: 34025091072
Worker version 67 connecting to http://tests.stockfishchess.org:80
Fetch task...
Updating worker version to 68
Exception accessing host:
Traceback (most recent call last):
  File "fishtest-master/worker/worker.py", line 102, in worker
    update()
  File "/home/fishtest/fishtest-master/worker/updater.py", line 45, in update
    restart(worker_dir)
TypeError: 'bool' object is not callable
```
The error is due to using `restart` as name for both a function and a variable.

https://github.com/glinscott/fishtest/blob/13b3dc34acae0b0b229e51fd1c60725298bf2617/worker/updater.py#L12-L45

- [Errno 2] No such file or directory

Worker output:
```
~$ python fishtest-master/worker/worker.py
Worker starting ...

Memory: 34025091072
Worker version 64 connecting to http://tests.stockfishchess.org:80
Fetch task...
Updating worker version to 68
['/usr/bin/python', 'fishtest-master/worker/worker.py']
/usr/bin/python: can't open file 'fishtest-master/worker/worker.py': [Errno 2] No such file or directory
```

The updater code restarts worker.py using the worker.py folder as working folder, but the working folder (where is written the configuration file `fishtest.cfg`) could be different.

We should use the true working folder in this way:
https://stackoverflow.com/questions/5137497/find-current-directory-and-files-directory

PR #369, opened 2019-02-03
fix restart name conflict

tomtor:
Should fix https://github.com/glinscott/fishtest/issues/368

### DIFF
--- a/worker/test_worker.py
+++ b/worker/test_worker.py
@@ -13,7 +13,7 @@ class workerTest(unittest.TestCase):
       os.remove('foo.txt')
     if os.path.exists('polyglot.ini'):
       os.remove('polyglot.ini')
-   
+
   def test_item_download(self):
     try:
       games.setup('polyglot.ini', '.')
@@ -21,9 +21,9 @@ class workerTest(unittest.TestCase):
     except KeyError:
       pass
 
-  def test_config_setup(self):      
+  def test_config_setup(self):
     config = worker.setup_config_file('foo.txt')
-        
+
     self.assertTrue(config.has_section('login'))
     self.assertTrue(config.has_section('parameters'))
     self.assertTrue(config.has_option('login', 'username'))
@@ -31,7 +31,7 @@ class workerTest(unittest.TestCase):
     self.assertTrue(config.has_option('parameters', 'host'))
     self.assertTrue(config.has_option('parameters', 'port'))
     self.assertTrue(config.has_option('parameters', 'concurrency'))
-        
+
   def test_worker_script(self):
     p = subprocess.Popen(["python" , "worker.py"], stderr = subprocess.PIPE)
     result = p.stderr.readline()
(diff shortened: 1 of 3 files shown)


tomtor:
@vdbergh I fixed the travis build by forcing scipy 1.1.0 (until the fixed 1.2.1 https://github.com/scipy/scipy/issues/9606) but I am not familiar with scipy. Do you think using 1.1.0 is safe?

tomtor:
Hi @ppigazzini I am not sure I understand the second problem.

You think that there is a restart problem when the user did start:

    python worker/worker.py

instead of:

    cd worker; python worker.py

ppigazzini:
@tomtor yes, that is the problem. We cannot enforce as a user launch the worker. What about forcing an os.chdir() in the worker.py folder at the start of worker.py?

PS: I will be at home in the afternoon.

tomtor:
> @tomtor yes, that is the problem. We cannot enforce as a user launch the worker. What about forcing an os.chdir() in the worker.py folder at the start of worker.py?
> 

Forcing a different working directory would prevent using different configuration files. I think just restoring the original working directory is the best approach for the cleanest `do_restart()`

I did this in the last commit.

tomtor:
@ppigazzini It should be OK now. I think the last Travis failure is a transient error...

tomtor:
@ppigazzini It's not ready yet, I'll have a look at it later...

tomtor:
@ppigazzini It should be ready for testing now. 

I tested it myself by forcing an update by changing line 100 in `worker.py` from:

to

    if True:

and starting:

         python3 worker/worker.py

It did do a succesfull update, i.e. restore my working copy to master.

ppigazzini:
@tomtor the production server (CentOS 6) and the dev-server (CentOS 7) are using numpy and scipy from distutils installed with yum.
On the dev/sever I successfully uninstalled numpy and scipy with yum and installed the latest versions with pip, but I'm not confident to try this on the fly on the production server because I have very little time in these days to debug a server problem (in the past some package updates with pip broke the production server). 

Please open a new PR for scipy and numpy and keep this PR only for the updater problem.

tomtor:
@ppigazzini Sure, I understand.

I removed the scipy commit from this PR.

There is no need for a new PR because in time scipy 1.2.1 will fix this Travis build issue.

ppigazzini:
@tomtor the update of running workers (Linux, Windows, Windows Subsystem for Linux) went fine on the dev server. I will test other combinations before merging this PR. Great work Tom.

ppigazzini:
@tomtor all my tests went fine, ready to be merged.
@noobpwnftw tomorrow at 02:15 PM CET I would like to force a worker update. Please let me know if there is a problem with the scheduling.

noobpwnftw:
@ppigazzini It's fine with me.

tomtor:
@ppigazzini thanks for your test work and diagnosis of the working directory issue!

Note that the workers would need a last manual restart. After that newer updates should properly restart. Not sure that is clear for everybody. Will the update of tomorrow carry the new @vdbergh pentamonial worker updates?

noobpwnftw:
Just a quick question:
Is it true that the workers will *not* automatically update and restart? Or there is a chance.

If the former, can you bump the version number when ready so I can manually update them earlier?

ppigazzini:
@noobpwnftw no problem, I can raise the worker version tonight at 11 pm CET.
@tomtor I wrote a post in fishcooking group to make clear that will be required a restart by hand.
I think that the pentanomial is still a WIP, the @vdbergh 's dev laptop should return in action tomorrow after a serious injury :)

noobpwnftw:
@ppigazzini My workers have updated with a clean install to latest master.

PR #370, opened 2019-02-04
Raise worker version to 69 (only worker side)

ppigazzini:


ppigazzini:
@noobpwnftw worker version raised to 69.

PR #372, opened 2019-02-07
Pentanomial elo

vdbergh:
This is a PR on top of PR #361. We now calculate the elo estimates for SPRT tests from the pentanomial frequencies. The estimates have been verified by simulation using code that is here

This PR requires quite some changes under the hood (restricted to the stats stuff!) but from the user's point of view the change is almost invisible (the elo estimates will be slightly different from those calculated from the trinomial frequencies).

Despite the fact I said this would be the last PR, I will yet make another PR with a template generated page which contains all the raw stats so that it is easier to audit the LLR and elo estimates as provided by the "results" widget and the live_elo page, and also to compare them with the old (obsolete) trinomial quantities.


ppigazzini:
@vdbergh dev-server updated and started a couple of tests.

ppigazzini:
@vdbergh dev-server updated yesterday night for test your latest commit https://github.com/glinscott/fishtest/pull/372/commits/9a83c3ad0b89834452a205ffac40c58c6f62e65f

vdbergh:
Thks!

PR #276, opened 2018-05-01
Make fishtest server python3 compatible

tomtor:
Also very few changes required.

Tested on Python2 AND Python3

PR #373, opened 2019-02-11
Pentanomial stats

vdbergh:
Pentanomial statistics!
    
We put the stats files in a separate directory.
We use our own brentq. In this way the stats files are self contained.
    
Update get_elo() in stat_util.py. The new get_elo() will now accept
both trinomial and pentanomial frequencies. Use the new get_elo() for
fixed length tests in views.py.
    
Make SPRT(...) in stat_util.py return the pentanomial LLR if possible.
    
 New get_elo api call.
    
 Render live_elo.html with the data from the api call.
 and eliminate the emscripten code.
    
 Pentanomial elo estimation.
    
 New template generated page /tests/stats/id.
 A link added in /test/view/id.
    
 Show the difference between the trinomial and pentanomial variance.
 This difference provides information about the bias in the
 opening book (measured by RMS) through the "accounting identity"
    
  sigma^2(trinomial)=sigma^2(pentanomial)+RMS_bias^2.
    
 Confidence intervals for various statistical quantities.


ppigazzini:
@vdbergh dev-server updated.

ppigazzini:
@vdbergh dev-server updated.

ppigazzini:
@vdbergh the two pages "Live Elo" and "Raw Statistic"perhaps could be merged to have all the info in only one place.

vdbergh:
@ppigazzini In my mind the two pages serve quite different purposes. Live_elo.html is for the user (so it only reports interesting things (e.g. hiding the needed conversions bayes elo<->logistic elo) and tries to look nice). /tests/stats/id is for auditing and so it reports a lot of stuff which would be completely uninteresting and even ununderstandable for most people and makes no attempts at cosmetics. Moreover as I wrote, I would like to add some explanations to it, again something that most people will not care about.

I considered hiding /tests/stats/id (i.e. making it only accessible by using the explicit URL) but this is a little 
too inconvenient.

One idea could be to have the page in the format of a jupyter notebook so that the data and the math could be mixed. However I have no experience with jupyter notebooks and I do not know if it feasible.

ppigazzini:
@vdbergh Ok, I understand the hype page :)

Some notes:
1. the Elo Live is missing the pentanomial result
2. pentanomial result is shown [0-2], the trinomial is shown [1; 0; 0.5]. For the hype the penta should be written [2-0] and for consistency the trinomial should be written [1; 0.5; 0]. (BTW I prefer [0-2])
3. a little inconsistency in Raw Stastitc page: SPRT test in queue with 0 games the page shows the trinomial statistic but don't show the pentanomial statistic.

vdbergh:
> @vdbergh Ok, I understand the hype page :)
> 
> Some notes:
> 
> 1. the Elo Live is missing the pentanomial result

@ppigazzini I do not feel strongly about this and wouldn't mind including them. The reason I did not include them was (ok, besides lazyness) that I thought they did not give a lot of information. The trinomial probabilities have some information that people care about (the draw ratio, and whether or not there are "more wins than losses"). Interpreting the pentanomial probabilities is not so easy. I did include the pentanomial frequencies in the elo result widget as a flag that they were present and being used, not so much for their actual values.

> 2. pentanomial result is shown [0-2], the trinomial is shown [1; 0; 0.5]. For the hype the penta should be written [2-0] and for consistency the trinomial should be written [1; 0.5; 0]. (BTW I prefer [0-2])

This is true I assume. If we include the pentanomial frequencies in the live_elo page (as I said above, I am neutral) then both trinomial and pentanomial frequencies should probably be written in a consistent way. But I am worried about writing them as [2-0]. The only efficient way to enter the pentanomial frequencies into a tool is by copy pasting, and for that to work well they should always be written in the same way.

> 3. a little inconsistency in Raw Stastitc page: SPRT test in queue with 0 games the page shows the trinomial statistic but don't show the pentanomial statistic.

Yes I agree. This has something to do with initialization. I need to think about it.

ppigazzini:
@vdbergh GitHub is able to [process the .ipynb files](https://help.github.com/articles/working-with-jupyter-notebook-files-on-github/), [here](https://github.com/ibm-et/jupyter-samples) some notebook examples on GitHub to view the output.

ppigazzini:
@vdbergh I thought that after some testing the trinomial should be totally replaced by the pentanomial. but perhaps I'm wrong and the two type must coexist (at least temporary).
Anyway I vote for the natural [0-2] and [0-1] presentation.

vdbergh:
@ppigazzini In fact the trinomial quantities still contain important information. For one thing the difference between the trinomial and pentanomial variance is approximately given by the square of the (RMS)bias of the book (this assumes additivity of elo, which, no doubt, is at most only approximately true).

I am still working on this (I am writing a text explaining the various statistical quantities, and at the same time I also plan to add more information to the raw stats page - some more confidence intervals, and also sensitivity). However I made a preliminary commit which also gives the difference between the two variances (rather than just the ratio) and converts it to elo.

ppigazzini:
@vdbergh dev server updated. Thank you for your explanation.

vdbergh:
@ppigazzini  Would it be possible to update the dev server with the latest commit? I now also calculate sensitivity (with confidence interval). As an experiment I would like to test the new proposed 2 moves book.

The last thing I need to do is to include some sort of confidence interval for the bias estimate for the book and to write documentation for the formulas. I am working on this but the computation for the confidence is a bit involved and I have been quite busy both in my private and professional life.

ppigazzini:
@vdbergh dev server updated.

ppigazzini:
@vdbergh you suggested to use scipy/numpy for chi2 computation, so I'm wondering why you are not using scipy functions, e.g. 
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brentq.html
https://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.norm.html

vdbergh:
> @vdbergh you suggested to use scipy/numpy for chi2 computation, so I'm wondering why you are not using scipy functions, e.g.
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brentq.html
> https://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.norm.html

I don't feel strongly about this, but the code I already had did not depend on scipy/numpy, which I thought was rather nice since AFAIK scipy/numpy is hard to get working on Windows. The implementation of the normal distribution in stat_util.py is due to Lucas. 

I am currently rebasing the PRs. I would prefer not to make any changes for now, unless I discover real bugs.

ppigazzini:
@vdbergh I agree to merge the well tested code, I was asking only for some well known incompatibility/problem in scipy. We could decide in the future if simplify the codebase using the functions from scipy.

vdbergh:
I tried to install a fresh copy of fishtest (I would like to test on localhost for now). My laptop has Ubuntu 16.04. I tried to run the setup script setup-fishtest.sh (which says it is for 18.04). It more or less seems to have worked except that I get an error  

``` Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 16, in <module>
    from views import format_results
  File "/home/fishtest/fishtest/fishtest/fishtest/views.py", line 39, in <module>
    with open(os.path.join('fishtest', 'static', 'favicon.ico'), 'r') as f:
IOError: [Errno 2] No such file or directory: 'fishtest/static/favicon.ico'
```
I do have favicon.ico...

ppigazzini:
@vdbergh I always viewed that error running a python script. This was solved running the script from `${HOME}/fishtest/fishtest/`
.I will test the latest setup script on Ubuntu 18.04

vdbergh:
Is there a short cut where you simply run "git clone" and then start the server with some command? That would be much easier for me.

tomtor:
@vdbergh After creating a virtual-env (optional) I cd to the fishtest directory (with `development.ini` file) and run:

`python setup.py install`

and start fishtest with:

`pserve development.ini`

tomtor:
@vdbergh The only additional thing from the setup script which is needed is creating the test users:

```
import os, sys
sys.path.append(os.path.expanduser('~/fishtest/fishtest/fishtest'))
from rundb import RunDb
rdb = RunDb()
rdb.userdb.create_user('user00', 'user00', 'user00@user00.user00')
rdb.userdb.add_user_group('user00', 'group:approvers')
user = rdb.userdb.get_user('user00')
user['blocked'] = False
user['machine_limit'] = 100
rdb.userdb.users.save(user)
rdb.userdb.create_user('user01', 'user01','user01@user01.user01')
user = rdb.userdb.get_user('user01')
user['blocked'] = False
user['machine_limit'] = 100
rdb.userdb.users.save(user)
```

vdbergh:
This now gives me a similar error
```Traceback (most recent call last):
  File "create_users.py", line 3, in <module>
    from rundb import RunDb
  File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 16, in <module>
    from views import format_results
  File "/home/fishtest/fishtest/fishtest/fishtest/views.py", line 20, in <module>
    import fishtest.stats.stat_util
  File "/home/fishtest/fishtest/fishtest/fishtest/__init__.py", line 8, in <module>
    from rundb import RunDb
  File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 16, in <module>
    from views import format_results
  File "/home/fishtest/fishtest/fishtest/fishtest/views.py", line 39, in <module>
    with open(os.path.join('fishtest', 'static', 'favicon.ico'), 'r') as f:
IOError: [Errno 2] No such file or directory: 'fishtest/static/favicon.ico'```

The "import fishtest.stats.stat_util" seems suspicious (I moved all stats files to a new directory).

I will investigate.

**Comment by vdbergh** [comment]
Ok I have rebased all PRs. There are now 4 remaining PRs. 351, 352, 354 and 373. Apparently 474 is also necessary to resolve some path related issue. I did a quick check on my local copy of fishtest that everything seems to be behaving as it should.

The best thing I think is now to have the PRs run on the dev server for a bit so that I can double check with my scripts that the various statistical quantities (listed at /tests/stats/<id>)
are correctly computed. I can connect two 6 core workers.

Note that it is very important that all workers that connect are updated. If in a run one old worker connects then the pentanomial statistics will be disabled for the whole run.

**Comment by ppigazzini** [comment]
@vdbergh https://github.com/glinscott/fishtest/pull/474 already merged.

DEV updated with these PRs, already started 5 workers with https://github.com/glinscott/fishtest/pull/351
```
# pentanomial
git pull --no-edit origin pull/351/head
git pull --no-edit origin pull/352/head
git pull --no-edit origin pull/354/head
git pull --no-edit origin pull/373/head
```

**Comment by ppigazzini** [comment]
@vdbergh this the new [DEV address](https://dfts-2.pigazzini.it)

**Comment by ppigazzini** [comment]
@vdbergh the trinomial/pentanomial number of games are different. I reload a fresh copy of the DB to not mix data from previous test.
```
LLR: 0.53 (-2.94,2.94) [-1.50,4.50]
Total: 1058 W: 432 L: 403 D: 223
Ptnml(0-2): 75, 83, 193, 94, 84
```

vdbergh:
Yes this is normal. The pentanomial data can only be updated after a game pair has finished. But the games finish asynchronously in case there are multiple threads (this is usually the case). So the pentanomial data will usually lag behind a little. If a task is interrupted then it is even possible that a few games are lost in this way.

vdbergh:
A more basic comment is that the pentanomial statistics count game pairs. So the total should be multiplied by two.

ppigazzini:
At the moment there is a difference of 1000 games with 4000 total games. I'm using 5 workers at 3 cores, so I'm expecting a max difference of 15 games.

ppigazzini:
Ah, ok, now is clear.

vdbergh:
I did some tests on the dev server and some sanity checks on the results, and it seems everything is as it should be...

ppigazzini:
@vdbergh I start testing on DEV the other PRs to be merged with the pentanomial ones, your workers should be able to run.

ppigazzini:
PROD server updated, great work, thank you @vdbergh !

tomtor:
Yes, impressive job @vdbergh 

Note that I made some changes to the Python3 PR to handle the new stats/ directory: https://github.com/glinscott/fishtest/pull/276/commits/7cbeb7781828b42640cc9d05b10ad1190756fa5b

I tested this by visiting the live_elo page until all runtime errors were resolved.

MJZ1977:
This is a very nice job !

I have just a minor comment : the order seems opposite to W/L/D order. It should be more intuitive to invert the number so if ELO is positive, the first numbers are bigger than last numbers !

snicolet:
I have a small question about the pentanomial0-2) displayed on the main page.

The following were snapshots taken during the test http://tests.stockfishchess.org/tests/view/5e1703cc0ac9162cdd5f3a24

LLR: -0.04 (-2.94,2.94) {0.00,2.00} 
Total: 328 W: 45 L: 45 D: 238 
Ptnml(0-2): 4, 25, 88, 30, 1

LLR: 0.48 (-2.94,2.94) {0.00,2.00} 
Total: 719 W: 112 L: 90 D: 517 
Ptnml(0-2): 9, 50, 198, 76, 9
As I understand it, in the Ptnml(0-2) row we have we have 5 numbers which are numbers of pairs LL, LD, {DD or WL}, WD, WW.

My question is: why do the numbers in the Ptnml line do not match the numbers in the Total line? :-)

ppigazzini:
@snicolet because are pairs and because pentanomial values are incremented only when a pair is completed (2 games).
So sum(Ptnml)*2 <= Total

View https://github.com/glinscott/fishtest/pull/373#issuecomment-566683092 :)

PR #388, opened 2019-09-05
Add Github API authentication

### LINKED ISSUE #385: With 4000+ cores on fishtest, "Base signature" is set to "None" (because GitHub API limit)
When there are this much cores on fishtest I often (and not only I) get this error:
"unable to find branch" and this is master branch which autosets bench to "None" (can't be fixed manually with inputting correct one https://i.imgur.com/U1rkPah.png
This usually autofixes after some time but it's pretty annoying.

tomtor:
Fetch ID/SECRET from server environment

Should fix  #385 

See https://developer.github.com/v3/#rate-limiting


tomtor:
No longer needed, we use .netrc for API rate limit basic authentication

PR #394, opened 2019-09-15
Setup Pyramid to work with nginx

ppigazzini:
Setup Pyramid according to [The Pyramid Community Cookbook](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/deployment/nginx.html).

### DIFF
--- a/fishtest/production.ini
+++ b/fishtest/production.ini
@@ -20,8 +20,14 @@ mako.directories = fishtest:templates
 
 [server:main]
 use = egg:waitress#main
-host = 0.0.0.0
-port = 6543
+host = 127.0.0.1
+port = %(http_port)s
+
+trusted_proxy = 127.0.0.1
+trusted_proxy_count = 1
+trusted_proxy_headers = x-forwarded-for x-forwarded-host x-forwarded-proto x-forwarded-port
+clear_untrusted_proxy_headers = yes
+
 # Match connection limit with max number of workers
 connection_limit = 500
 threads = 16


ppigazzini:
@tomtor I initialized a new Pyramid project w/ [CookieCutters](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/cookiecutters.html) to check the configuration files and production.ini has [`listen`](https://docs.pylonsproject.org/projects/waitress/en/stable/arguments.html) syntax on all addresses vs `host/name` syntax on 127.0.0.1
Both syntax work fine on the dev server.

```
[server:main]
use = egg:waitress#main
listen = *:%(http_port)s
```
[Here an old issue](https://github.com/Pylons/pyramid/issues/489) about 0.0.0.0 vs. 127.0.0.1

Any suggestion for the final setup?

tomtor:
> EDIT_000 : the production server doesn't work w/ `host = 127.0.0.1`

@ppigazzini What do you mean with "production server"? I would expect that if you use nginx (like the test server) then 127.0.0.1 works fine. Without nginx (like the current production server) you really need 0.0.0.0 to listen on the public ethernet iterface.

ppigazzini:
@tomtor I use the term "production server" for the VPS running http://tests.stockfishchess.org
I need to study a bit to understand your answer...

I will be only in a couple of hours, have a nice afternoon :)

ppigazzini:
@tomtor I thought that the iptable instructions redirected port 80 to localhost:6543 , but perhaps I'm wrong (no iptable knowledge, only google), or the iptable configuration used now is different.

https://github.com/glinscott/fishtest/blob/5a94e7434d1f1d8c5037d84c431f6531bd273e93/SETUP#L41-L45

tomtor:
@ppigazzini I am not an iptables expert but redirecting to localhost is described here:

https://superuser.com/questions/661772/iptables-redirect-to-localhost

Note that if you do that then there is no reason to not use 0.0.0.0 as pserve listen address.

Specifying 127.0.0.1 is considered more secure as pserve listen address because you can not connect from outside the machine. The iptable configuration makes this again possible so there is no security gained.

ppigazzini:
@tomtor I opened this PR because I prefer to have references to this repo in the server setup script.

And to have the confirm that this cannot be merged on the production server (w/o nginx) to simplify the setup script. These break all:
```
trusted_proxy = 127.0.0.1
trusted_proxy_count = 1
trusted_proxy_headers = x-forwarded-for x-forwarded-host x-forwarded-proto x-forwarded-port
clear_untrusted_proxy_headers = yes
```

tomtor:
@ppigazzini Just a note that /tests triggers the auto_purge feature and DOES write the mongodb when new tasks are generated.

https://github.com/glinscott/fishtest/blob/40ad53f259f0d608dcb1315a29f85e8fb7237180/fishtest/fishtest/views.py#L950

So ideally this code is moved to update_task in rundb.py, but until that is done, new tasks will be delayed by 60s (cache timeout for request_task and run buffer), which should be ok.

vondele:
I would be in favour of removing auto_purge, especially since we can do it by hand, and it should rarely be needed. I also agree it really is hiding some info if it is done automatically. What could be useful as a replacement is some small indication in the test output that purging would change the result (e.g. add a * after the LLR/Elo score or similar).

snicolet:
I am fine with the removal: auto_purge was historically nice at the beginning of fishtest when the specs of the testing machines varied wildly, which is no longer the case.

Issue #395, opened 2019-09-22
Fishtest in an infinite loop


snicolet:
The Fishtest site is not working at the moment.

From what I can see or infer from the state just before the pages stop responding,
this may be due to the test "KingAndPawnEvaluator" not compiling properly on some
workers. 

This happens sometimes, tests which cannot compile suck the workers but never
release them. These tests show that they have many active workers but no games.

The problem today is that it seems impossible to stop this test...

ppigazzini:
@snicolet the Fishtest site seems working fine in this moment, no action by me on the vps.

tomtor:
I assume it needs a similar change as #390.

ppigazzini:
@snicolet fixed, thank you for the alert.

snicolet:
Thanks to both!

PR #396, opened 2019-09-22
Immediately stop runs

tomtor:
instead of relying on some auto_purge actions

See #395


ppigazzini:
@tomtor dev server updated but no time to make tests now. I will be online in a couple of hours.

ppigazzini:
@tomtor evil runs signed failed at first wrong bench. Let me know if you want some other tests or I can merge on the fly.

tomtor:
This should be low risk so merging is ok for me.

ppigazzini:
Prod server updated, thank you @tomtor

PR #397, opened 2019-09-23
Fix games per hour statistic for grand total

tomtor:
@ppigazzini This one sneaked through. The games/hour grand total should be cleared but it was not, so in the users report the games/hour continuously grows. The first incremental run should fix the grand total, no need to run a full scan.


ppigazzini:
Prod server updated, script already ran by hand.

tomtor:
@ppigazzini That was quick! Thanks!

PR #393, opened 2019-09-11
Write a more efficient script to update the user statistics

tomtor:
See #391 

`update_users.py` reads the whole MongoDb database to build user statistics from scratch.

This PR creates a new script `delta_update_users.py` which incrementally updates the statistics and only reads runs from the last 31 days.

Edit: Note that the first run will be as slow as `update_users.py`

PR #398, opened 2019-09-23
Optimize purge_pgn.py

tomtor:
Skip runs older than 30 days

See also https://github.com/glinscott/fishtest/pull/393#issuecomment-534154127


ppigazzini:
Thank you @tomtor.
Test by hand on the prod server
```
$ time python purge_pgn.py
PGN runs/tasks saved:      0/      0
PGN runs/tasks purged:     0/      0

real    0m0.461s
user    0m0.322s
sys     0m0.113s
```

PR #399, opened 2019-09-26
Do not skip deleted runs in statistics

tomtor:
because the users DID work on them and it
disturbs delta update calculation

@ppigazzini Can you run a full scan before testing or installing on production?

### DIFF
--- a/fishtest/utils/delta_update_users.py
+++ b/fishtest/utils/delta_update_users.py
@@ -14,8 +14,6 @@ skip = False
 
 def process_run(run, info, deltas=None):
   global skip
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor prod server updated. In this case should I start a `delta_update_users.py scan_all` ?

tomtor:
Yes please do!

Thanks!

Op do 26 sep. 2019 11:41 schreef ppigazzini <notifications@github.com>:

ppigazzini:
@tomtor done, lower users values. Next week I will check the users values against the `scan_all` users values.

PR #406, opened 2019-09-28
Update bounds when clicking STC/LTC buttons

### LINKED ISSUE #405: LTC button
Maybe i'm the only one who gets caught with this but...

When we reschedule a STC test, and press LTC, time and hash are changed correctly,
but sprt stays the same which is unexpected behaviour.

if sprt was 0.5...4.5 (STC) it should change to 0....3.5 (LTC)

tomtor:
Should fix #405

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -206,7 +206,7 @@ $(function() {
   var update_bounds = function() {
     var bounds = $('select[name=bounds]').val();
     if (bounds == 'standard STC') { $('input[name=sprt_elo0]').val('0.5'); $('input[name=sprt_elo1]').val('4.5'); }
-    if (bounds == 'standard LTC') { $('input[name=sprt_elo0]').val('0'); $('input[name=sprt_elo1]').val('3.5'); }
+    if (bounds == 'standard LTC') { $('input[name=sprt_elo0]').val('0.0'); $('input[name=sprt_elo1]').val('3.5'); }
     if (bounds == 'tweak') { $('input[name=sprt_elo0]').val('0'); $('input[name=sprt_elo1]').val('4'); }
     if (bounds == 'regression') { $('input[name=sprt_elo0]').val('-3'); $('input[name=sprt_elo1]').val('1'); }
     if (bounds == 'simplification') { $('input[name=sprt_elo0]').val('-3'); $('input[name=sprt_elo1]').val('1'); }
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor I don't view difference after applying this PR to the [dev server](https://dfts-0.pigazzini.it)

By the way I think that @Rocky640 is asking to auto set the "SPRT bounds" after pushing the "Long (LTC)" time control button.

tomtor:
> By the way I think that @Rocky640 is asking to auto set the "SPRT bounds" after pushing the "Long (LTC)" time control button.

I do both. If you had STC bounds and press LTC you get LTC bounds. If you have LTC bounds and press STC you get STC-bounds. I wonder why it is not working for you. Browser cache refresh?

ppigazzini:
@tomtor after clicking "Reschedule" you get "Custom bounds"

tomtor:
> after clicking "Reschedule" you get "Custom bounds"

Ok, then nothing changes. Only when you have selected STC and press LTC bounds change.

tomtor:
@ppigazzini please test again

ppigazzini:
@tomtor it works when rescheduling a STC test.
If I'm not wrong the code is only about the STC case (in other words: it doesn't work when rescheduling a LTC test)

tomtor:
> If I'm not wrong the code is only about the STC case (in other words: it doesn't work when rescheduling a LTC test)

Now I am confused. If you reschedule a test that was already LTC then all bounds are already correct?

ppigazzini:
@tomtor LTC reschedule -> new test form w/ "custom bounds" "0.0 - 3.5".
Pushing "short STC", "long LTC" buttons, time and hash change, but SPRT bounds are always fixed to "custom bounds" "0.0 - 3.5"

tomtor:
@ppigazzini I got it this time (I hope :D ), thanks for the feedback!

ppigazzini:
@tomtor LTC doesn't work, check the dev server.

> Sounds like something that could happen to me ;-)

It happened, you are Tom before learning to code :D

tomtor:
@ppigazzini It works for me, after rescheduling eg https://dfts-0.pigazzini.it/tests/run?id=5d8938130ebc59600ae5649f I can toggle with the STC/LTC buttons?!

ppigazzini:
@tomtor rescheduling a STC and pushing "long LTC" change the bounds.
Rescheduling a LTC and pushing "short STC" doesn't change the bounds.

https://dfts-0.pigazzini.it/tests/view/5d8fb8b0612e02ce742bf20b

tomtor:
@ppigazzini Ah, it was '0' vs '0.0' when rescheduling an existing LTC test....

Please test again, fingers crossed.

ppigazzini:
@tomtor it works! :)
Can I merge?

tomtor:
> Can I merge?

Yes, please do! Thanks again for your thorough testing!

Perhaps I should convert the fields to floats, but it is OK for me now.

ppigazzini:
Done, thanks @tomtor :)

PR #409, opened 2019-10-05
Small cleanups

tomtor:
remove test for obsolete worker version

remove redundant line for tests_stop()

leading 0's for Played (better sorting)

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -160,7 +160,7 @@ Gaussian Kernel Smoother&nbsp;&nbsp;<div class="btn-group"><button id="btn_smoot
   <%
     stats = task.get('stats', {})
     if 'stats' in task:
-      total = stats['wins'] + stats['losses'] + stats['draws']
+      total = str(stats['wins'] + stats['losses'] + stats['draws']).zfill(3)
     else:
       continue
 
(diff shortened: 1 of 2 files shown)


ppigazzini:
@tomtor dev server updated.
https://github.com/glinscott/fishtest/pull/409/commits/80870b72b046681b1e11da0f78ac7c519b9e70dd verified, I don't know how to check the other two changes.

tomtor:
@ppigazzini not related to this patch but I see that a worker is not  stopped immediately?

tomtor:
@ppigazzini I assume the test server has the nginx multi process setup with separate process for /tests? This means that seeing the stop in the main view is a bit delayed (about 1 or 2 seconds) after pressing STOP. So you must reload /tests, that is the disadvantage of the multiple processes.

I pushed a change so that this effect should be reduced, can you please reload the server?

tomtor:
> sentenced to 50 PRs!

@ppigazzini I think I am already close to that or did you mean NEW PRs? ;-)

ppigazzini:
@tomtor server updated
ps: NEW PRs upon MY requirements. First one: "Stop button" floating position > 50 pixel from the mouse if the user is Dutch.

ppigazzini:
@tomtor dev server setup with the latest version of the [wiki script](https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup)

tomtor:
@ppigazzini Ok, I have tested and stopping works fine now.

The other change was for showing the download link and that is working too. So merging is OK for me.

ppigazzini:
Prod server updated.

tomtor:
Thanks!

PR #411, opened 2019-10-07
Show number of cores for active tests

tomtor:
So one does not have to visit the individual test page


tomtor:
@ppigazzini I improved the layout "cores:" always on a new line, that should look better...

Can you please restart the test server?

ppigazzini:
@tomtor done, probable bug.

tomtor:
Hmm, yes certainly. I pushed again....

ppigazzini:
@tomtor server updated

Issue #412, opened 2019-10-07
Gracefully decommissioning workers


noobpwnftw:
I wonder what is the best way to decommission a large number of workers, say after they finish their current task and not to request the next one.
How can I reduce disruption of running tests when I need to turn off some workers without either aborting their tasks or waiting for fishtest to recycle their task slots?

tomtor:
Would require eg a file flag (simple and portable) or implementing an Unix SIGHUP handler for the worker.

I will put it on my to-do list, but for now I suggest not to worry about it.

PR #413, opened 2019-10-08
exit on existance of signal file

tomtor:
See #412 

@noobpwnftw I have not tested this, but it could be as simple as this?

Is this sufficient and could you test it?

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -25,6 +25,7 @@ from optparse import OptionParser
 from games import run_games
 from updater import update
 from datetime import datetime
+from os import path
 
 WORKER_VERSION = 69
 ALIVE = True
@@ -165,12 +166,13 @@ def worker(worker_info, password, remote):
   return success
 
 def main():
-  print("Worker started in " + os.getcwd() + " ...\n")
+  worker_dir = path.dirname(path.realpath(__file__))
+  print("Worker started in " + worker_dir + " ...\n")
 
   signal.signal(signal.SIGINT, on_sigint)
   signal.signal(signal.SIGTERM, on_sigint)
 
-  config_file = 'fishtest.cfg'
+  config_file = path.join(worker_dir, 'fishtest.cfg')
   config = setup_config_file(config_file)
   parser = OptionParser()
   parser.add_option('-n', '--host', dest='host', default=config.get('parameters', 'host'))
@@ -232,6 +234,8 @@ def main():
   success = True
   global ALIVE
   while ALIVE:
+    if path.isfile(path.join(worker_dir, 'fish.exit')):
+      break
     if not success:
       time.sleep(HTTP_TIMEOUT)
     success = worker(worker_info, args[1], remote)


tomtor:
@noobpwnftw Ok, the worker will cleanly exit after finishing the current active batch of 250.

ppigazzini:
@tomtor cross platform temp folder
https://stackoverflow.com/questions/847850/cross-platform-way-of-getting-temp-directory-in-python

worker/worker.py
```diff
@@ -232,6 +233,8 @@ def main():
   success = True
   global ALIVE
   while ALIVE:
+    if path.isfile("/tmp/fish.exit"):
```
ppigazzini:
```
import tempfile

    if path.isfile(path.join(tempfile.gettempdir(), 'fish.exit')):
```

noobpwnftw: or maybe in testing folder instead? like custom_make.txt.

ppigazzini: @noobpwnftw I approve your proposition. A user running hundreds of workers is able to code a script to write/delete the file. On unix this should work
```bash
touch dir{00..99}/worker/fish.exit
rm dir{00..99}/worker/fish.exit
```

ppigazzini: @tomtor file in the worker dir, this should work:
```
   if path.isfile(path.join(os.path.dirname(os.path.realpath(__file__)), 'fish.exit')):
```

tomtor: I pushed your code suggestion, thanks

ppigazzini: @tomtor I edited the suggestion to set `worker_dir` at the script start because after the first 250 game batch `os.path.dirname(os.path.realpath(__file__)=/<some_path>/worker/testing`

ppigazzini:
@tomtor sorry, wrong copy and paste on the fly from my script...

worker/worker.py
```diff
@@ -166,6 +167,7 @@ def worker(worker_info, password, remote):
 
 def main():
   print("Worker started in " + os.getcwd() + " ...\n")
+  worker_dir = path.join(path.dirname(path.realpath(__file__)))
```
tomtor: @ppigazzini Is worker_dir different from os.getcwd() in the previous line?

worker/worker.py
```diff
@@ -232,6 +234,9 @@ def main():
   success = True
   global ALIVE
   while ALIVE:
+    print(path.join(worker_dir, 'fish.exit'))
```
tomtor: Should this print be removed or is it an aid for stopping the worker (so user knows where to place 'fish.exit' ?)

worker/worker.py
ppigazzini: @tomtor I copied that code from games.py

worker/worker.py
ppigazzini: @tomtor I wrote that print for debug

worker/worker.py
ppigazzini: @tomtor difference between `os.getcwd()` and `path.dirname(path.realpath(__file__))`

https://stackoverflow.com/questions/5137497/find-current-directory-and-files-directory

worker/worker.py
```diff
@@ -234,7 +234,6 @@ def main():
   success = True
   global ALIVE
   while ALIVE:
-    print(path.join(worker_dir, 'fish.exit'))
     if path.isfile(worker_dir, 'fish.exit'):
```
ppigazzini: if path.isfile(path.join(worker_dir, 'fish.exit')):

ppigazzini: @tomtor I beg your pardon for wasting your time. Wrong copy&past, editing w/o , comments too short. I promise to review the code only if having a proper time and not during a multitasking frenzy...

ppigazzini:
@tomtor  notes after some readings and experimenting.

1. `python ./fishtest/worker/worker.py`:
 - `os.getcwd()=/home/user`, i.e. is the path where you start the command
 - `path.dirname(path.realpath(__file__))=/home/user/fishtest/worker`,  i.e. is the path of the started script

Now `fishtest.cfg` is written in `os.getcwd()`, IMO this should be next to `custom_make.txt`, `fish.exit`

2. `path.dirname(path.realpath(__file__))` changes under Linux Ubuntu 18.04 and WSL with the `os.chdir()` used in `games.py` (it doesn't change under Windows), so we need to set a worker_dir variable at the start of `worker.py`.

I propose this snippet of code to fix also the path of `fishtest.cfg` (tested working under Linux, WSL, Windows, but please double check because is a copy&paste):

```python
def main():
  worker_dir=path.dirname(path.realpath(__file__))
  print("Worker started in " + worker_dir + " ...\n")
  
  signal.signal(signal.SIGINT, on_sigint)
  signal.signal(signal.SIGTERM, on_sigint)

  config_file = path.join(worker_dir, 'fishtest.cfg')

...

  success = True
  global ALIVE
  while ALIVE:
    if path.isfile(path.join(worker_dir, 'fish.exit')):
      break
    if not success:
      time.sleep(HTTP_TIMEOUT)
    success = worker(worker_info, args[1], remote)

if __name__ == '__main__':
  main()
```

tomtor:
@ppigazzini I looked at the code and it looks OK, but I did not actually test it, because you are much better at that (testing in multiple environments).

One concern is that changing the location of 'fishtest.cfg' from CWD to directory of 'worker.py' might surprise some users. I start fishtest just with 'python3 worker.py' so for me there would not be a difference.

ppigazzini:
@tomtor commuting, give me one hour. A worker update requires always a forum post.

tomtor:
@ppigazzini I do not think there is a hurry for this PR because @noobpwnftw is probably the only user which will use this new (minor) feature.

PR #416, opened 2019-10-17
Handle test stop, restart in statistics update

tomtor:
When a test is stopped and restarted (by a modify) this
can cause many runs (30 days period) to be counted twice.
This happens only when the restarted test was the only new
finished test at the last execution of this script.
When the test is restarted before the script is executed
then there is also no error condition.

This can be fixed by storing also all older run-ids,
with a small performance impact (bigger MongoDb table).

Note that tasks in the restarted test are counted twice
when the error condition happens. This should be very rare.

@ppigazzini This has low priority. The performance impact should be minimal. Note that linenumbers have changed so you should adapt your upgrade procedure patch.

### DIFF
--- a/fishtest/utils/delta_update_users.py
+++ b/fishtest/utils/delta_update_users.py
@@ -122,6 +122,7 @@ def update_users():
     current += step_size
 
   if new_deltas:
+    new_deltas.update(deltas)
     rundb.deltas.remove()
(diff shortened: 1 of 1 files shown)


ppigazzini:
Prod server updated, just started `delta_update_users.py full`
Thank you @tomtor

PR #417, opened 2019-10-19
Change STC bounds to [-1..5]

snicolet:
This pull request proposes to change the STC bounds to [-1..5]

At the date of october 19th, the pourcentage for speculative LTC of the last 100 LTC Elo-gaining tests was 70%, from all active Stockfish developers.

This gives some feedback after a few months of intensive use with the changes introduced in glinscott/fishtest#342, and shows that the new STC bounds were too strict for our community since even the most motivated members bypass them.

Related discussions:
glinscott/fishtest#342
https://github.com/official-stockfish/Stockfish/issues/2283
https://github.com/official-stockfish/Stockfish/pull/1804#issuecomment-445429885
https://github.com/official-stockfish/Stockfish/issues/1859

PR #418, opened 2019-10-19
Remove yellow for STC tests

snicolet:
The yellow test results at STC are often abused as a pretext to make speculative LTC.

If the STC bounds change of pull request https://github.com/glinscott/fishtest/pull/417 is merged, then STC tests will get a green passed status more easily, and we can restrict the yellow color to LTC test results.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -672,8 +672,10 @@ def format_results(run_results, run):
 
   result['info'].append('Total: %d W: %d L: %d D: %d' % (sum(WLD), WLD[0], WLD[1], WLD[2]))
 
+  tc = run['args']['tc']
+
   if state == 'rejected':
-    if WLD[0] > WLD[1]:
+    if WLD[0] > WLD[1] and tc == '60+0.05':
       result['style'] = 'yellow'
     else:
       result['style'] = '#FF6A6A'


snicolet:
NB: I don't have a local fishtest server to test this change, please review carefuly for syntax errors, etc.

Alayan-stk-2:
I don't think there is a point in changing the color coding.

If approvers are asked to enforce a "no spec LTC for yellows" rule, certainly they would be able to.

With the wider bounds, looking at yellows for ideas that might be worth exploring further is not as useful because the elo estimates get quite unreliable, but it's still nicer than a stream of reds.

The argument used for spec. LTC is not so much about the color-coding than about the elo estimate that is over +1 for a bunch of those tests that end up running as spec. LTC. Calling the STC a long yellow is just a shorthand for this idea.

NKONSTANTAKIS:
From extensive usage of spec LTC to zero usage is a bit extreme. How about limiting it to just 1 at a time and only if no other tests are running? A lot of elo came from spec LTC (albeit at a high price), and its something handy to keep the network busy at idle hours. With the new bounds and rules I suspect it will be idle often.

tomtor:
@snicolet I suggest to test for LTC with:

`import re`
`...`
`re.match('^([6-9][0-9])|([1-9][0-9][0-9])', run['args']['tc'])`

I am also in favor of maintaining the current STC yellow because it hints at a patch with potential, not for speculative LTC, but  for trying some variant with another STC run.

xoto10:
I also would like to keep the yellow as it highlights tests that might have potential.

If the change is made, using parse_tc() might be neater, see e.g. https://github.com/glinscott/fishtest/blob/3316c0d1e56e64e0191a4a28daba1792a91b02d2/fishtest/fishtest/views.py#L980

PR #419, opened 2019-10-19
Use blue instead of green for passed simplifications

snicolet:
Use blue instead of green for passed simplifications

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -678,7 +678,10 @@ def format_results(run_results, run):
     else:
       result['style'] = '#FF6A6A'
   elif state == 'accepted':
-    result['style'] = '#44EB44'
+    if (float(sprt['elo0']) + float(sprt['elo1'])) < 0.0:
+      result['style'] = '#66CCFF'
+    else:
+      result['style'] = '#44EB44'
   return result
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
@snicolet [dev-server](https://dfts-0.pigazzini.it) updated. I think that we should also change:
- the color for "tests" pages
- the menu voice from "Greens" to "Passed" or "Accepted"

@tomtor suggestions?

ianfab:
Due to a hack in the displaying of the list of succesful tests using the color, this change also needs to be reflected there: https://github.com/glinscott/fishtest/blob/a227e3953bb98f4e26d2f4eb1e59fd888d6428d3/fishtest/fishtest/rundb.py#L225-227

tomtor:
Its obvious that the fishtest colouring needs some refactoring love if more colours are introduced.
If we agree on the blue colour then it's easy to change the PR in an agile way so that:

> * the color for "tests" pages
I propose that only NEW tests get the blue color

> * the menu voice from "Greens" to "Passed" or "Accepted"
Just let this unchanged, so Greens returns only the greens (and not the blues :-)

ppigazzini:
Prod server updated, thank you @snicolet

snicolet:
thanks :-)

PR #432, opened 2019-11-19
Raise worker version to 70 (only worker side)

ppigazzini:


ppigazzini:
@noobpwnftw server and worker now support also https protocol.
Please test the updated worker also using python3.

noobpwnftw:
@ppigazzini workers updated to version 70 and python3, not using https due to another layer of reverse proxy at my side.

ppigazzini:
@noobpwnftw OK, I'm keeping http enabled in nginx.
The worker will never drop http because it's a nightmare to use https/self-signed-certificate/python on a developer workstation.

Issue #408, opened 2019-10-04
VPS upgrade to Ubuntu 18.04


ppigazzini:
## People
- [x] @glinscott - VPS build
- [x] @ppigazzini - other tasks

### Date
16.11.2019
### Start time pre-upgrade task
10:30 UTC

### Start time upgrade task
11:00 UTC

### pre-upgrade
- [x] write setup_fishtest.sh
  - [x] with passwords and keys
  - [x] disable the fishtest start
  - [x] comment out all the crontab jobs
- [x] write the mongodb restore script
```bash
#!/bin/bash
# mongodb_restore.sh

bkp_dir=${HOME}/backup/restore
[ ! -d ${bkp_dir} ] && mkdir -p ${bkp_dir}
mv dump.tar.gz ${bkp_dir}
cd ${bkp_dir}
mongo fishtest         --eval 'db.dropDatabase()'
mongo fishtest_new     --eval 'db.dropDatabase()'
mongo fishtest_testing --eval 'db.dropDatabase()'

tar -xzvf dump.tar.gz && rm -f dump/*/*.json && mongorestore --drop
```
- [x]  write delete_old_users.sh
```bash
#!/bin/bash
# delete_old_users.sh
# remove old pending users (remove lines 154 to 158)

cd ~/fishtest/fishtest/utils
cp delta_update_users.py delete_old_users.py
vi delete_old_users.py << EOF
:154,158d
:154
2X
:155
.
:wq
EOF
$VENV/bin/python delete_old_users.py
```
- [x] utility scripts backup
- [x] make a gmail account for the Let's Encrypt certificate
- [x] forum post to alert for:
 - the downtime
 - the switch to https
- [x] comment out crontab jobs
- [x] stop fishtest
```bash
#!/bin/sh
cd ${HOME}/fishtest/fishtest
cp nohup.out ${HOME}/logs/nohup.out.$(date +%Y%m%d%H%M%S --utc)
pkill -f pserve
```
- [x] logs backup
- [x] mongodb backup on s3
```
bash ~/fishtest/fishtest/utils/backup.sh
```
- [x] mongodb dump download

### upgrade
- [x] new VPS Ubuntu 18.04 LTS
 - ssh access as or "root" or a new user in sudo group:
```bash
#!/bin/bash
user_name='vpsuser'
user_pwd='<vpsuser_password>'
useradd -m -s /bin/bash ${user_name}
echo ${user_name}:${user_pwd} | chpasswd
[...]

PR #433, opened 2019-11-19
Speed up main page by sorting in Python

tomtor:
and not with MongoDb which reads the whole database.

@ppigazzini Please test on dev server at a convenient time.

See also https://github.com/glinscott/fishtest/issues/408#issuecomment-555445376

Someone knows what happened to Travis builds?

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -226,8 +226,9 @@ class RunDb:
       # not currently is the color!
       q['results_info.style'] = '#44EB44'
 
-    c = self.runs.find(q, skip=skip, limit=limit, sort=[('last_updated', DESCENDING)])
+    c = self.runs.find(q, skip=skip, limit=limit, sort=[('$natural', DESCENDING)])
     result = [list(c), c.count()]
+    result[0].sort(key=lambda r: (r['last_updated']), reverse=True)
 
     if limit != 0 and len(result[0]) != limit:
       c = self.old_runs.find(q, skip=max(0, skip-c.count()), limit=limit-len(result[0]))


ppigazzini:
@tomtor dev-server updated w/ the PR.
Are welcome suggestions about how to benchmark this PR and the https://github.com/glinscott/fishtest/issues/408#issuecomment-555445376

ppigazzini:
@tomtor no idea about travis

tomtor:
@ppigazzini I do not see a significant speedup (I wonder why), so I guess moving all but the most recent 1000 tests to `old_runs` is the only option? Do you agree or do you find the main page delay acceptable? One issue now is that approved/created tests take a long time to appear.

Should I write something like https://stackoverflow.com/questions/27039083/mongodb-move-documents-from-one-collection-to-another-collection/41483981#41483981

tomtor:
https://travis-ci.org/glinscott/fishtest/pull_requests shows the up to date Travis build status  including this PR but https://github.com/glinscott/fishtest/pulls does not.

@vondele Do you have a suggestion?

vondele:
@tomtor this seems similar to
https://github.com/official-stockfish/Stockfish/issues/2314#issuecomment-533812660

ppigazzini:
I have access to https://travis-ci.org/glinscott/fishtest/settings but not to https://github.com/glinscott/fishtest/settings/

I wrote a mail to @glinscott

tomtor:
I am closing this because of https://github.com/glinscott/fishtest/pull/434

glinscott:
From what I see, travis should still be active:

image

It looks like the previous integration we used is now deprecated though.

PR #436, opened 2019-11-22
Python script to time some db queries

xoto10:
This script creates an easy way to see how quickly some of the key database queries are taking. This may help diagnose problems at times when fishtest gets slow.

More queries can be added to make the script a more thorough.

### DIFF
--- a/fishtest/utils/test_queries.py
+++ b/fishtest/utils/test_queries.py
@@ -1,0 +1,59 @@
+
+# test_queries.py - run some sample queries to check db speed
+#
+
+from __future__ import print_function
+
+import os
+import sys
+import pprint
+import time
+from pymongo import MongoClient, ASCENDING, DESCENDING
+
+sys.path.append(os.path.expanduser('~/fishtest/fishtest'))
+from fishtest.rundb import RunDb
+
+
+db_name='fishtest_new'
+rundb = RunDb()
+
+# MongoDB server is assumed to be on the same machine, if not user should use
+# ssh with port forwarding to access the remote host.
+conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
+db = conn[db_name]
+runs = db['runs']
+pgns = db['pgns']
+
+
+
+def printout(s):
+  print(s)
+  sys.stdout.flush()
+
+
+
+printout("\nFetching unfinished runs ...")
+start = time.time()
+unfinished_runs = rundb.get_unfinished_runs()
+end = time.time()
+
+printout(str(end-start) + "s\nFetching machines ...")
+start = time.time()
+machines = rundb.get_machines()
+end = time.time()
+
+printout(str(end-start) + "s\nFetching finished runs ...")
+start = time.time()
+finished, num_finished = rundb.get_finished_runs(skip=0, limit=50, username='',
+                                                 success_only=False, ltc_only=False)
+end = time.time()
+
+printout(str(end-start) + "s\nRequesting pgn ...")
+if (len(finished) == 0):
+    finished.append({'_id':'abc'})
+start = time.time()
+pgn = rundb.get_pgn(str(finished[0]['_id']) + ".pgn")
+end = time.time()
+
+printout(str(end-start) + "s\n")
+


ppigazzini:
@xoto10 I'm writing with my phone on the train.
The latest script version used yesterday hanged, If I'm not wrong there was an error about thread. I will test this PR when at home.

Please add compatibility to python 3 and keep the script not executable:

```python
#!/usr/bin/env python

from __future__ import print_function

print(s)   # python 3
```

tomtor:
@xoto10 I am worried about the request_task test call because it injects nonsense data in the (production) mongodb database: illegal ip and non-existing user.

xoto10:
> @xoto10 I am worried about the request_task test call because it injects nonsense data in the (production) mongodb database: illegal ip and non-existing user.

Ah yes, this is a good point, I will think about that one some more. At the time I recognised that the server would allocate a task to a fictitious worker, but I didn't think that was a big deal. But I agree, I am not comfortable with writing that info into the db (and therefore it could be seen when viewing the details of a test).

xoto10:
@tomtor @ppigazzini 
I removed the request_task call. It looks like the main db access it does is get_unfinished_runs() which is already in the script so request_task() isn't needed.

ppigazzini:
@xoto10 running on the dev-server I got this error:
```
$ ${VENV}/bin/python test_queries.py

Fetching unfinished runs ...
0.000307083129883s
Fetching machines ...
0.0265979766846s
Fetching finished runs ...
3.35118889809s
Requesting pgn ...
  File "test_queries.py", line 55, in <module>
    pgn = rundb.get_pgn(finished[0]['_id'])
  File "/home/usr00/fishtest/fishtest/fishtest/rundb.py", line 130, in get_pgn
    id = id.split('.')[0] # strip .pgn
AttributeError: 'ObjectId' object has no attribute 'split'
```

fishtest/utils/test_queries.py
```diff
@@ -0,0 +1,59 @@
...
+start = time.time()
+finished, num_finished = rundb.get_finished_runs(skip=0, limit=50, username='',
+                                                 success_only=False, ltc_only=False)
+end = time.time()
+
+printout(str(end-start) + "s\nRequesting pgn ...")
+if (len(finished) == 0):
+    finished.append({'_id':'abc'})
+start = time.time()
+pgn = rundb.get_pgn(finished[0]['_id'])
```
tomtor: @xoto10 Try `rundb.get_pgn(str(finished[0]['_id']) + ".pgn")`

ppigazzini: @xoto10 it works with the @tomtor suggestion:
```
$ $VENV/bin/python test_queries.py 
Fetching unfinished runs ...
0.000203847885132s
Fetching machines ...
0.0169911384583s
Fetching finished runs ...
3.15122318268s
Requesting pgn ...
0.000858068466187s
```

tomtor:
Regarding indices, I have the impression that some are unneeded:

                u'key': [(u'deleted', 1)],
                u'ns': u'fishtest_new.runs',
                u'v': 2},
                           u'ns': u'fishtest_new.runs',
                           u'v': 2},
                                          (u'last_updated', -1)],
                                 u'ns': u'fishtest_new.runs',
                                 u'v': 2},
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'ns': u'fishtest_new.runs',
                                               u'v': 2},
                                             (u'start_time', -1)],
                                    u'ns': u'fishtest_new.runs',
                                    u'v': 2},

I expect that we only need 2 from the 5 above:

                           u'ns': u'fishtest_new.runs',
                           u'v': 2},
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'ns': u'fishtest_new.runs',
                                               u'v': 2},

because the multi-attribute indices will also work when querying a subset, although perhaps slightly slower. Less indices will result in faster updates and less space in the cache, so in the end it could be more efficient.

tomtor:
Index stats:

    db.runs.aggregate( [ { $indexStats: { } } ] )              

shows 3 unused indices (the first is used 1 time, the others 0 times):       
                                                                                
    { "name" : "deleted_1", "key" : { "deleted" : 1 }, "host" : "tests.stockfishchess.org:27017", "accesses" : { "ops" : NumberLong(1), "since" : ISODate("2019-11-24T09:59:34.119Z") } } 
    { "name" : "last_updated_-1_start_time_-1", "key" : { "last_updated" : -1, "start_time" : -1 }, "host" : "tests.stockfishchess.org:27017", "accesses" : { "ops" : NumberLong(0), "since" : ISODate("2019-11-24T09:59:34.119Z") } }
    { "name" : "tasks.pending_1", "key" : { "tasks.pending" : 1 }, "host" : "tests.stockfishchess.org:27017", "accesses" : { "ops" : NumberLong(0), "since" : ISODate("2019-11-24T09:59:34.119Z") } }

xoto10:
> Regarding indices, I have the impression that some are unneeded:

Yes, I always supected this, but I didn't have any way of knowing for sure.

> because the multi-attribute indices will also work when querying a subset, although perhaps slightly slower. Less indices will result in faster updates and less space in the cache, so in the end it could be more efficient.

My experience with other dbs is that an index can only be used if the 1st, 1st and second, ... fields are known (compare with looking up a word in a dictionary, you have to know the 1st letter, or 1st 2 letters, etc). Once you you reach a field that is not known or is being matched with like 'X%' then later fields in the index cannot be used.
But also note, a full scan of an index can be useful if the index only includes a few rows, e.g. active : True that we mentioned earlier.

If glinscott can view load averages over time, then perhaps a simple way is to just drop a candidate index, wait for 24h (if required) and then view the load average graph. If no obvious problem, then assume it's ok to remove that index.

As a rule-of-thumb, usually the multi-field indexes are overkill, the main gain comes from indexing one field, or at most 2 fields.

PR #437, opened 2019-11-22
Create index on runs.tasks.active

xoto10:
This patch simplifies the query for active tasks in rundb.py and creates an index on runs.tasks.active.
The script index_pending.py is renamed to index_change.py since it is not only used for the index on pending but as a generic script to alter any index.
The speedup can be measured with #436


xoto10:
I can remove the file rename if that makes things easier.

xoto10:
Also, I imagine the main gain comes from the index, so that could be tried quickly on its own, if you prefer not to update rundb.py just now.

fishtest/fishtest/rundb.py
```diff
@@ -116,7 +116,7 @@ def new_run(self, base_tag, new_tag, num_games, tc, book, book_depth, threads, b
 
   def get_machines(self):
     machines = []
-    for run in self.runs.find({'tasks': {'$elemMatch': {'active': True}}}):
+    for run in self.runs.find({'tasks.active': True}):
```
tomtor: If changed here we should change all `$elemMatch` for consistency.

I prefer to leave it unchanged

tomtor:
Adding/testing the index is a good idea

fishtest/fishtest/rundb.py
xoto10: My concern is that the elemmatch is a more complex operation and may be slower than the simpler test. But I don't know how it's implemented, it's possible the query optimiser sees through the un-necessary complexity and implements both operations the same way under the hood.
I will remove the rundb.py change for now while we investigate ...

fishtest/utils/index_change.py
```diff
@@ -0,0 +1,55 @@
...
+# ssh with port forwarding to access the remote host.
+conn = MongoClient(os.getenv('FISHTEST_HOST') or 'localhost')
+db = conn[db_name]
+runs = db['runs']
+pgns = db['pgns']
+
+
+
+def printout(s):
+  print s
```
tomtor: Please use python3 `print(...)` and the corresponding import for python2.6+ compatibility

tomtor:
@xoto10 An issue with this index is that it will be huge, because of the many tasks.

Speedup by changing:

https://github.com/glinscott/fishtest/blob/6a4fb798b66a668a467b8cd2d0c476c14ebb4c45/fishtest/fishtest/rundb.py#L119

and adding `{'finished': False}` condition could be tried first

xoto10:
I'm not too worried about it being large, but if it's not needed then
better not to create it. Using finished : false seems like a good idea, I
agree we should try that.

ppigazzini:
@xoto10 @tomtor first test result, I redo the test with a longer loop.

* master (after #441):
```
> db.runs.totalIndexSize()
6139904

$ time bash curl_loop.sh urls_list.txt
real    2m22.558s
user    0m7.531s
sys     0m13.859s
```

* pr:
```
> db.runs.totalIndexSize()
6459392

$ time bash curl_loop.sh urls_list.txt
real    2m5.426s
user    0m7.313s
sys     0m14.422s
```

script output:
```
$ ${VENV}/bin/python index_change.py
Creating tasks.active index ...

Current indexes on runs:
{u'_id_': {u'key': [(u'_id', 1)], u'ns': u'fishtest_new.runs', u'v': 2},
                      u'ns': u'fishtest_new.runs',
                      u'v': 2},
                u'key': [(u'deleted', 1)],
                u'ns': u'fishtest_new.runs',
                u'v': 2},
                           u'ns': u'fishtest_new.runs',
                           u'v': 2},
                                          (u'last_updated', -1)],
                                 u'ns': u'fishtest_new.runs',
                                 u'v': 2},
                                               u'key': [(u'finished', 1),
                                                        (u'last_updated',
                                                         -1),
                                                        (u'start_time', -1)],
                                               u'ns': u'fishtest_new.runs',
                                               u'v': 2},
                                             (u'start_time', -1)],
                                    u'ns': u'fishtest_new.runs',
                                    u'v': 2},
 u'tasks.active_1': {u'key': [(u'tasks.active', 1)],
                     u'ns': u'fishtest_new.runs',
                     u'v': 2},
 u'tasks.pending_1': {u'key': [(u'tasks.pending', 1)],
                      u'ns': u'fishtest_new.runs',
                      u'v': 2}}
```

ppigazzini:
@xoto10 the conflict disable the merge for this PR. I'm able to make tests on the dev-server, though.

ppigazzini:
@xoto10 @tomtor my script is not able to detect a speedup. Anyway the index size seems to be low.

* master (after #441):
```
$ time bash curl_loop.sh urls_list.txt
real    11m53.043s
user    0m34.719s
sys     1m10.719s

> db.runs.stats()
(...)
        "nindexes" : 8,
        "indexBuilds" : [ ],
        "totalIndexSize" : 6131712,
        "indexSizes" : {
                "_id_" : 1339392,
                "finished_1_last_updated_-1" : 741376,
                "finished_1_last_updated_-1_start_time_-1" : 1359872,
                "last_updated_-1_start_time_-1" : 1359872,
                "deleted_1_finished_1" : 327680,
                "deleted_1" : 335872,
                "args.username_1" : 344064,
                "tasks.pending_1" : 323584
        },
        "scaleFactor" : 1,
        "ok" : 1
}
```

* pr:
```
$ time bash curl_loop.sh urls_list.txt
real    11m18.867s
user    0m35.344s
sys     1m12.031s

> db.runs.stats()
(...)
        "nindexes" : 9,
        "indexBuilds" : [ ],
        "totalIndexSize" : 6434816,
        "indexSizes" : {
                "_id_" : 1339392,
                "finished_1_last_updated_-1" : 741376,
                "finished_1_last_updated_-1_start_time_-1" : 1359872,
                "last_updated_-1_start_time_-1" : 1359872,
                "deleted_1_finished_1" : 327680,
                "deleted_1" : 335872,
                "args.username_1" : 344064,
                "tasks.pending_1" : 323584,
                "tasks.active_1" : 303104
        },
        "scaleFactor" : 1,
        "ok" : 1
}
```

ppigazzini:
@xoto10 anyway this PR is a speedup wrt the master **before** #441 (only 10 loops to keep the test short)

* master ( **before** #441):
```
$ time bash curl_loop.sh urls_list.txt
real    3m54.867s
user    0m7.125s
sys     0m13.625s
```

* pr:
```
$ time bash curl_loop.sh urls_list.txt
real    2m6.206s
user    0m6.719s
sys     0m13.906s
```

xoto10:
@ppigazzini @tomtor 
> first test result, I redo the test with a longer loop.
> 
>     * master (after #441):
> 
> 
> ```
> > db.runs.totalIndexSize()
> 6139904
> 
> $ time bash curl_loop.sh urls_list.txt
> real    2m22.558s
> user    0m7.531s
> ```
> 
>     * pr:
> 
> 
> ```
> > db.runs.totalIndexSize()
> 6459392
> 
> $ time bash curl_loop.sh urls_list.txt
> real    2m5.426s
> user    0m7.313s

I was hoping for a much more obvious change, usually a correct index change will give more than 10x improvement (often >100x). I am not sure if mongodb is caching the first result and then returning that instantly in later queries, this is possible if it is inspecting a timestamp and it knows no updates have been done since the original query.
Anyway, I will create an alternative PR with @tomtor 's suggestion regarding `active : False` , that seems preferable to this one.

ppigazzini:
@xoto10 
* master before #441 
```
$ ${VENV}/bin/python test_queries.py

Fetching unfinished runs ...
0.000458002090454s
Fetching machines ...
11.9291341305s
Fetching finished runs ...
4.52183890343s
Requesting pgn ...
0.000951766967773s
```

* master before #441 + #437 
```
$ ${VENV}/bin/python test_queries.py

Fetching unfinished runs ...
0.000166893005371s
Fetching machines ...
0.00169205665588s
Fetching finished runs ...
3.16469192505s
Requesting pgn ...
0.000974893569946s
```

* master after #441 
```
$ ${VENV}/bin/python test_queries.py

Fetching unfinished runs ...
0.00016188621521s
Fetching machines ...
0.0117638111115s
Fetching finished runs ...
3.29121494293s
Requesting pgn ...
0.00100016593933s
```

* master after #441 + #437
```
$ ${VENV}/bin/python test_queries.py

Fetching unfinished runs ...
0.000179052352905s
Fetching machines ...
0.00248193740845s
Fetching finished runs ...
3.17063689232s
Requesting pgn ...
0.000839948654175s
```

ppigazzini:
@xoto10 please mind that I wrote that my script was not able to measure speedup:
```
# curl_loop.sh
#!/bin/bash
if [[ ${#} -ne 1 ]]; then
  echo usage: bash ${0} page_list.txt
  exit 1
fi
url_base="https://dfts-2.pigazzini.it"
tot=10
for ((i=${tot}; i>0; --i)); do
  while read -r url_page; do
    curl -L ${url_base}${url_page} &> /dev/null &
  done < ${1}
  echo ${i}
  wait
done
``` 

```

/
/tests
/login
/signup
/users/monthly
/users
/tests?ltc_only=1
/tests?ltc_only=10
/tests?ltc_only=100
/tests?success_only=1
/tests?success_only=10
/tests?success_only=100
/tests?page=1
/tests?page=10
/tests?page=100
/tests?page=1000
/tests/view/5d6ed7890ebc595bbb6ed5e6
/tests/view/5dd5dbebf55d99269e1883a1
/tests/view/5dd54282f531e81cf278eae0
/tests/view/5dd5c0540c135b64c8c4d560
/html/live_elo.html?5d6ed7890ebc595bbb6ed5e6
/html/live_elo.html?5dd5dbebf55d99269e1883a1
/html/live_elo.html?5dd54282f531e81cf278eae0
/html/live_elo.html?5dd5c0540c135b64c8c4d560
/api/pgn/5dd5c0350c135b64c8c4d55e-52.pgn
/api/pgn/5dd5c0350c135b64c8c4d55e-62.pgn
/tests/user/snicolet
/actions
/actions?action=&user=snicolet
/actions?action=delete_run&user=snicolet
```

ppigazzini:
@xoto10 @tomtor here the test on the prod-server (I dropped the index after the test)

* master :
```
${VENV}/bin/python fishtest/fishtest/utils/test_queries.py

Fetching unfinished runs ...
0.0002601146698s
Fetching machines ...
0.105093955994s
Fetching finished runs ...
8.98656988144s
Requesting pgn ...
0.00114798545837s
```

* master + #437 :
```
$ ${VENV}/bin/python fishtest/fishtest/utils/test_queries.py

Fetching unfinished runs ...
0.000392913818359s
Fetching machines ...
0.155565023422s
Fetching finished runs ...
3.96374607086s
Requesting pgn ...
0.000836849212646s
```

tomtor:
> Fetching finished runs ...
> 3.96374607086s

We should focus on this last slow call now.

The query is:
     q = {'finished': True, 'deleted': {'$exists': False}}  
     c = self.runs.find(q, skip=skip, limit=limit, sort=[('last_updated', DESCENDING)])

So we need an index on finished/deleted/last_updated and we do NOT have that one...

tomtor:
Added:

    db['runs'].ensure_index([('finished', ASCENDING), ('deleted', DESCENDING), ('last_updated', DESCENDING)])  

on prod but no speedup?

xoto10:
> > Fetching finished runs ...
> > 3.96374607086s
> 
> We should focus on this last slow call now.

Agreed. 

> The query is:
> q = {'finished': True, 'deleted': {'$exists': False}}
> c = self.runs.find(q, skip=skip, limit=limit, sort=[('last_updated', DESCENDING)])
> 
> So we need an index on finished/deleted/last_updated and we do NOT have that one...

Ah. In the db I work with in my day job, queries with ` and f != v ` (roughly similar to this) can't (/don't generally) use an index on f.

So I think we want an index on finished (asc/desc, doesn't matter) and last_updated (desc may be better, I don't know). The condition on `deleted` should just be filtered out as the rows are returned. And I think we already have that one?
I will do some reading ...

xoto10:
Perhaps this is relevant?

[mongodb.com](https://docs.mongodb.com/manual/tutorial/ensure-indexes-fit-ram/)

> For the fastest processing, ensure that your indexes fit entirely in RAM so that the system can avoid reading the index from disk.
> 
> To check the size of your indexes, use the db.collection.totalIndexSize() helper, which returns data in bytes:


Another idea: Could we do the `'deleted': {'$exists': False}` test in python, so the query is the simpler, ` finished : True, sort by ... ` ? Perhaps this query would use the index to fetch and sort the results?

tomtor:
> So I think we want an index on finished (asc/desc, doesn't matter) and last_updated (desc may be better, I don't know). The condition on `deleted` should just be filtered out as the rows are returned. And I think we already have that one?

@xoto10 Yes:

    (u'last_updated', -1)],                                                                                                                    u'ns': 
    u'fishtest_new.runs',                                                                                                                        u'v': 
    2},

xoto10:
@tomtor 

https://docs.mongodb.com/manual/reference/operator/meta/explain/index.html

$explain is probably the best way to find out what is really happening

xoto10:
Ah, I've just realised the obvious thing. If the query optimisation is quite good in this version, it will notice that ` finished : True ` is almost always true and it will choose not to use the index because it thinks the index is useless.

We may be able to use a hint to persuade it:  [mongodb hint](https://docs.mongodb.com/manual/reference/method/cursor.hint/)

tomtor:
I found the cause for the slow finished query, will create a PR.

PR #439, opened 2019-11-23
sort old runs for viewing

tomtor:
Currently the last page shows some tests of 100 days old.

And at the break: https://tests.stockfishchess.org/tests?page=134


tomtor:
@ppigazzini Merged this one and restarted pserve

ppigazzini:
@tomtor 👍

PR #444, opened 2019-11-26
Enhance test_queries.py to include explains and to test hints

xoto10:
Enhance test_queries.py to include explains and to test hints.

### DIFF
--- a/fishtest/utils/test_queries.py
+++ b/fishtest/utils/test_queries.py
@@ -30,30 +30,102 @@ def printout(s):
   print(s)
   sys.stdout.flush()
 
+def qlen(c):
+  if (c): return len(list(c))
+  else  : return 0
 
 
+# Tests: Hint some queries
+
+printout("\nHinted queries:")
+printout("\nFetching unfinished runs old hint ...")
+start = time.time()
+c = runs.find({'finished': False}, sort=[('last_updated', DESCENDING), ('start_time', DESCENDING)]
+             ).hint('finished_1_last_updated_-1')
+end = time.time()
+
+printout("{} rows {:1.4f}".format(qlen(c), end-start) + "s\nFetching unfinished runs new hint ...")
+start = time.time()
+c = runs.find({'finished': False}, sort=[('last_updated', DESCENDING)]).hint('finished_1_last_updated_-1')
+end = time.time()
+
+printout("{} rows {:1.4f}".format(qlen(c), end-start) + "s\nFetching machines hint ...")
+start = time.time()
+c = runs.find({'finished': False, 'tasks': {'$elemMatch': {'active': True}}}).hint('finished_1_last_updated_-1')
+end = time.time()
+
+printout("{} rows {:1.4f}".format(qlen(c), end-start) + "s\nFetching finished runs hint ...")
+start = time.time()
+q = {'finished': True}
+c = runs.find(q, limit=50, sort=[('last_updated', DESCENDING)]).hint('finished_1_last_updated_-1')
+end = time.time()
+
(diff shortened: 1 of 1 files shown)


xoto10:
The explains show ` u'stage' : u'COLLSCAN' ` in my local fishtest since it has no data, but on prod they should show which index is being used.

ppigazzini:
@xoto10 here the log from the dev-server:
[test_queries.zip](https://github.com/glinscott/fishtest/files/3893187/test_queries.zip)

xoto10:
@ppigazzini @tomtor 
Interesting!

Looking at the explain plans, (see winningPlan for each one and above that executionTimeMillis) it looks like @tomtor 's change makes the finished query use the finished_1_last_updated_-1 index with a runtime of 2ms instead of the finished_1_last_updated_-1_start_time_-1 index which had a runtime of 13ms. (runtimes maybe not believable but the index change looks definite)
  
Looking at the hinted queries at the end, either there's a bug or they ran a lot faster than the originals at the top of the log file. :-)

These suggest to me that would we could consider one of:
- dropping the finished_1_last_updated_-1_start_time_-1 index since perhaps it isn't useful
- adding `.hint('finished_1_last_updated_-1')` to these queries in the code

There is a possibility that these results are influenced by caching, so you could try cutting and pasting the `# Tests: Hint some queries` section from the end to before the `# Time some important queries` so that the hinted versions have the disadvantage of going first, and run the test again.

xoto10:
@ppigazzini 
Those runtimes for the hinted queries are buggy, I will update the PR ...

xoto10:
@ppigazzini @tomtor 
No, perhaps the hints really are working! :-)
I've moved the hinted queries to the start of the script and added the #rows returned so we can check the results match between the original and hinted versions.

xoto10:
Updated this to use the current definitions of the queries.

tomtor:
Merged, we should write a rebuild indices script which deletes unneeded and creates needed indices in the future.

xoto10:
> a rebuild indices script

Agreed. A standard technique is to code it to drop all the existing indexes (either with an exhaustive list or even better by querying the db to get the list) and then recreate the ones that are wanted. Ideally this makes it re-runnable in almost any circumstance and the end state is always the same.

xoto10:
@ppigazzini @tomtor 
Can we run this latest test_queries.py to see how the queries are being executed now and to get some benchmark times now the system is running fairly smoothly?

ppigazzini:
@xoto10 [test_queries.zip](https://github.com/glinscott/fishtest/files/3912755/test_queries.zip)

xoto10:
@ppigazzini @tomtor 
Can we try dropping the index on finished, last_updated, start_time ?
I suspect the other index on finished and last_updated is as good / maybe better (it should be smaller), and there's no need to have both. 

If am wrong, it is possible the opposite could work, dropping the index on finished, last_updated, but I prefer to try dropping the larger index first.

We can run test_queries.py, and look at the load averages to see if the change helps or makes things worse.

tomtor:
> Can we try dropping the index on finished, last_updated, start_time

This is the sort order, so I expect that dropping the other index would be better, but timing both options on dev is a good idea

tomtor:
@ppigazzini @xoto10 I dropped by hand on prod:

    db['runs'].drop_index('finished_1_last_updated_-1') 

without bad effects... 

@xoto10 `test_queries.py` will now fail:

    pymongo.errors.OperationFailure: error processing query: ns=fishtest_new.runsTree: finished $eq false                                                                                        Sort: { last_updated: -1, start_time: -1 }                                                                                                                                                   
    Proj: {}                                                                                                                                                                                      
    planner returned error :: caused by :: hint provided does not correspond to an existing index

Can you file a PR to fix that?

PR #447, opened 2019-11-27
Fix delete run

tomtor:


tomtor:
@ppigazzini Merged, restarted and they are gone...

tomtor:
@ppigazzini Have to fix old_runs too....

ppigazzini:
@tomtor I'm commuting now, can you revert?

tomtor:
@ppigazzini Sure, I will combine it with this PR

ppigazzini:
@tomtor in effect also the DEV crashes, but unfortunately I thought than that code was triggered by the workers and I didn't know how to test it. My bad.

PR #449, opened 2019-11-28
Fix "divide by zero" warnings improving chi2 computation for unusual cases

ppigazzini:
Extend to matrix operations the original idea by @xoto10

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -753,20 +753,38 @@ def get_chi2(tasks, bad_users):
 
   observed = numpy.array(users.values())
   rows,columns = observed.shape
-  df = (rows - 1) * (columns - 1)
-  column_sums = numpy.sum(observed, axis=0)
-  row_sums = numpy.sum(observed, axis=1)
-  grand_total = numpy.sum(column_sums)
-  if grand_total == 0:
+  # Results only from one worker: skip the test for workers homogeneity
+  if rows == 1:
+    results = {'chi2': float('nan'), 'dof': 0, 'p': float('nan'), 'residual': {}}
     return results
 
+  column_sums = numpy.sum(observed, axis=0)
+  columns_not_zero = sum(i > 0 for i in column_sums)
+  df = (rows - 1) * (columns - 1)
+
+  if columns_not_zero == 0:
+    return results
+  # Results only of one type: workers are identical wrt the test
+  elif columns_not_zero == 1:
+    results = {'chi2': 0.0, 'dof': df, 'p': 1.0, 'residual': {}}
+    return results
+  # Results only of two types: workers are identical wrt the missing result type
+  # Change the data shape to avoid divide by zero
+  elif columns_not_zero == 2:
+    idx = numpy.argwhere(numpy.all(observed[..., :] == 0, axis=0))
+    observed = numpy.delete(observed, idx, axis=1)
+    column_sums = numpy.sum(observed, axis=0)
+
+  row_sums = numpy.sum(observed, axis=1)
+  grand_total = numpy.sum(column_sums)
+
   expected = numpy.outer(row_sums, column_sums) / grand_total
-  diff = observed - expected
-  adj = numpy.outer((1 - row_sums / grand_total), (1 - column_sums / grand_total))
-  residual = diff / numpy.sqrt(expected * adj)
+  raw_residual = observed - expected
+  std_error = numpy.sqrt(expected * numpy.outer((1 - row_sums / grand_total), (1 - column_sums / grand_total)))
+  adj_residual = raw_residual / std_error
   for idx in range(len(users)):
-    users[users.keys()[idx]] = numpy.max(numpy.abs(residual[idx]))
-  chi2 = numpy.sum(diff * diff / expected)
+    users[users.keys()[idx]] = numpy.max(numpy.abs(adj_residual[idx]))
+  chi2 = numpy.sum(raw_residual * raw_residual / expected)
   return {
     'chi2': chi2,
     'dof': df,


ppigazzini:
@tomtor works fine on [DEV](https://dfts-2.pigazzini.it), I just started learning numpy.

tomtor:
@ppigazzini Looks good to me!

vondele:
```
  if numpy.min(expected * adj) == 0:
    return results
``` 
might have to be '<=0' ?

ppigazzini:
@tomtor if I'm not wrong the estimated standard error matrix (the denominator with the square root) can have some 0 values when: 
 * `column_sums`: total wld, has 0/0s (in this case `expected` has a column of 0s for each 0 in `column_sums` vector)
 * `adj` has 0/0s. This is possible when:
    * there is only 1 username/worker in `row_sums`
    * there is only 1 result type in `column_sums`  (e.g. only victories)

In any case negative values should be impossible.

xoto10:
@ppigazzini @tomtor @vondele 
I would just calculate the divisor `div = numpy.sqrt(expected * adj)` first and then test it for zero, but I'm sure there are many valid choices.

ppigazzini:
@xoto10 I'm searching for the correct name of the divisor: textbooks/papers use different names.

xoto10:
@ppigazzini
I understand. I don't know this matrix maths stuff so I can't help. I wouldn't worry too much about the name of a local variable that's only used for a couple of lines.

ppigazzini:
@xoto10 correct naming is helpful to understand the calculus.
At example in this script "residual" should be "standardized_residual" (according to A. Agresti) or "adjusted_standardized_residual" according to other authors and "diff" should be "residual".

ppigazzini:
@tomtor I viewed only now that there are 2 potential division by zero:
```
  residual = diff / numpy.sqrt(expected * adj)

  chi2 = numpy.sum(diff * diff / expected)
```

By the way before committing this PR we should check the effect of the division by zero and the effect of this PR.

ppigazzini:
@tomtor
https://dfts-2.pigazzini.it/tests/view/5de00f96ad7a0cbdc76e3cab
( log in PROD and comment out the code)
* master shows:
```
Stats
chi^2	nan or 0.0.0 (total wld has 0s or results only from one worker)
dof	0
p-value	nan%
```
and the worker has `Residual nan` in red

* pr shows this:
```
chi^2 | 0.00
-- | --
dof | 0
p-value | 0.00%
```
and the worker has `Residual 0.000` in green.

I'm wondering if this PR can change the tests acceptance/rejection whit results only from one worker.

tomtor:
> I'm wondering if this PR can change the tests acceptance/rejection logic whit results only from one worker

This code is used for purging, not for LLR calculations, so changing it will have no effect unless purging behaviour changes significantly. Purging is currently rare.

ppigazzini:
@tomtor the old PR version returned `p=0` so autopurge was always enabled. IMO this was wrong.

These are the "divide by zero" cases:
* only one worker (): residual, chi2 and p-value make no sense, but the test result (IMO) has meaning
```python
  results = {'chi2': float('nan'), 'dof': 0, 'p': float('nan'), 'residual': {}}
```
* wld has only 1 type of result: workers have same behavior: residual, chi2 and p-value make no sense, but the test result has meaning
```python
  results = {'chi2': 0.0, 'dof': 0, 'p': float('nan'), 'residual': {}}
```
* wld has only 2 types of result: workers could have different behavior: residual, chi2 and p-value should be computed discarding the zero column.

I'm starting testing this new version on DEV.

IMO is a valid alternative to keep a simpler code and the Warnings.

ppigazzini:
@tomtor here some tests on DEV, `nan` values are written by the PR to avoid wrong auto-purge, no "divide by zero" is involved.

* test with two workers and only 2 wld types (here the PR deletes the column of zeros)
https://dfts-2.pigazzini.it/tests/view/5de152068ba733521756a6ca

* test finished with only a worker
https://dfts-2.pigazzini.it/tests/view/5de0dba36fbc771bd1c76fad

* test with only 1 wld type
https://dfts-2.pigazzini.it/tests/view/5de153ee8ba733521756a6ce

ppigazzini:
@tomtor check the random test on DEV and PROD (has also #451): on PROD the worker names have links, on DEV no.

tomtor:
> @tomtor check the random test on DEV and PROD (has also #451): on PROD the worker names have links, on DEV no.

Only when logged in as approver links are shown (this is by design).

I see links on DEV.

ppigazzini:
@tomtor I need some rest :)

tomtor:
> IMO is a valid alternative to keep a simpler code and the Warnings.

Hmm, I prefer no runtime errors/warnings over adding some code.

Statistics is not my thing, perhaps @vondele or @snicolet can check this PR.

tomtor:
If we remove the hacky auto-purge can this code then be made more simple or do we still need it for manual purge?

ppigazzini:
@tomtor the code should be called for both manual and auto purge.

I added a WIP in the PR title, I want to search in the old forum/github posts the reason for some choices (variables naming, complementary p-value etc.). Before merging I would like to rename some variables to have a more readable code.

Hoping for@vondele @snicolet feedbacks :)

vondele:
I had a look, and I think the PR does what you want to do, but concerning the proper stats @vdbergh might be the best person to judge?

ppigazzini:
@vondele the history of the chi2 computation is basically all contained in this forum thread:
https://groups.google.com/forum/#!msg/fishcooking/91-01oWX3Vo/A4z5jPhN2E4J

There is also this thread with some info by Lyudmil Antonov
https://groups.google.com/forum/#!msg/fishcooking/jpzc9c7KlzM/Ucph9g_eCAAJ

The variable names were not changed after switching from raw residuals to adjusted residuals.

ppigazzini:
@vondele IMO it's wrong to lower the df when missing a result type, tonight I will push a new commit to correct my mistake.

ppigazzini:
@vondele @tomtor `observed` is always a random variable, column(s) of zeros is(are) due to the small sample size wrt the ratio P(result type higher probability)/P(result type higher probability) with P() probability. So it's always `df = (rows - 1) * (columns - 1)` w/ `columns = 3`.

In this script I verified the equivalence of chi2, p-value, adjusted residuals, computed in these two ways:
* adding at the column(s) of zeros a random variable w/ uniform distribution and very low mean value to solve the small sample size and avoid the divide by zero problem
* dropping the column of zeros w/o changing the df

```python
from __future__ import print_function
import numpy
import scipy
import scipy.stats

def get_chi2(observed):
  print("observed:\n%s" % observed)
  rows, columns = observed.shape
  column_sums = numpy.sum(observed, axis=0)
  columns_not_zero = sum(i > 0 for i in column_sums)
  print("columns not zero %s" % columns_not_zero)
  df = (rows - 1) * (columns - 1)

  if columns_not_zero == 1:
    residual = numpy.zeros((rows, columns))
    print("chi2 = %s df = %s p-value = %s"% (0.0,df,1.0))
    print("residual:\n%s" % residual)
    return residual
  elif columns_not_zero == 2:
    idx = numpy.argwhere(numpy.all(observed[..., :] == 0, axis=0))
    observed = numpy.delete(observed, idx, axis=1)
    column_sums = numpy.sum(observed, axis=0)

  row_sums = numpy.sum(observed, axis=1)
  grand_total = numpy.sum(column_sums)

  expected = numpy.outer(row_sums, column_sums) / grand_total
  diff = observed - expected
  adj = numpy.outer((1 - row_sums / grand_total), (1 - column_sums / grand_total))
  residual = diff / numpy.sqrt(expected * adj)
  chi2 = numpy.sum(diff * diff / expected)
  p = 1 - scipy.stats.chi2.cdf(chi2, df)
  print("residual:\n%s" % residual)
  print("chi2 = %s df = %s p-value = %s"% (chi2,df,p))

  return residual

rows, columns = 10, 3

print("\ntwo type results")
observed = numpy.random.randint(130,200,(rows,columns))
observed = observed.astype(float)

observed[:,0] = numpy.random.rand(rows)/1.0e6
residual3 = get_chi2(observed)

observed[:,0] = numpy.zeros(rows)
residual2 = get_chi2(observed)

print("residual3-residual2:\n%s" % (residual3[:,1:]-residual2))

print("\none type result")
observed = numpy.random.randint(130,200,(rows,columns))
observed = observed.astype(float)

observed[:,0] = numpy.random.rand(rows)/1.0e6
observed[:,1] = numpy.random.rand(rows)/1.0e6
residual3 = get_chi2(observed)

observed[:,0] = numpy.zeros(rows)
observed[:,1] = numpy.zeros(rows)
residual2 = get_chi2(observed)

print("residual3-residual2:\n%s" % (residual3-residual2))
```

ppigazzini:
@tomtor @vondele merged and PROD updated. I updated the list of useful links in previous comment  https://github.com/glinscott/fishtest/pull/449#issuecomment-559960140

vdbergh:
FWIW I think the changes are ok. It seems to me that in a contingency table the chi^2 test, including the calculation of residuals, will lead to division by zero if and only if ([there is a zero row] or [there is a zero column] or [there is only one row] or [there is only one column]). The PR seems to deal with all possibilities except the first one, but I guess this can simply not happen here.

ppigazzini:
@vdbergh thank you for your time.
The first case [there is a zero row] was already dealt by https://github.com/glinscott/fishtest/blob/16a291aaeae50aa0b5e583c004230d1e08ff7f63/fishtest/fishtest/views.py#L751-L752 

In fishtest these 4 unusual cases happen usually at the start of the test, or when the test is stopped after few games.

PR #455, opened 2019-12-01
Add numba for SPSA inner loop calculation

tomtor:
The SPSA calculation is the most CPU intensive part in fishtest.

Rewrite inner loop for Numba JIT compilation.

Note that this PR requires Python 3 Numba version, so it will be merged at a later date.

EDIT: default clipping is `old`, so gain is somewhat limited.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -7,6 +7,9 @@ import threading
 import zlib
 import re
 
+import numba
+from numba import jit
+
 from datetime import datetime, timedelta
 from bson.objectid import ObjectId
 from bson.binary import Binary
@@ -637,7 +640,10 @@ class RunDb:
       R = w_params[idx]['R']
       c = w_params[idx]['c']
       flip = w_params[idx]['flip']
-      param['theta'] = self.spsa_param_clip_round(param, R * c * result * flip, spsa['clipping'], 'deterministic')
+      #param['theta'] = self.spsa_param_clip_round(param, R * c * result * flip, spsa['clipping'], 'deterministic')
+      param['theta'] = jit_spsa_param_clip_round(
+          float(param['theta']), float(param['min']), float(param['max']),
+          R * c * result * flip, spsa['clipping'])
       if grow_summary:
         summary.append({
           'theta': param['theta'],
@@ -647,3 +653,20 @@ class RunDb:
 
     if grow_summary:
       spsa['param_history'].append(summary)
+
+@jit(nopython=True)
+def jit_spsa_param_clip_round(theta, pmin, pmax, increment, clipping):
+  value = 0.0
+
+  if clipping == 'old':
+    value = min(max(theta + increment, pmin), pmax)
+  else: #clipping == 'careful':
+    inc = min(abs(increment), abs(theta - pmin) / 2, abs(theta - pmax) / 2)
+    if inc > 0:
+        inc_sgn = 0 if increment == 0 else increment / abs(increment)
+        value = theta + inc_sgn * inc
+    else: #revert to old behavior to bounce off boundary
+        value = min(max(theta + increment, pmin), pmax)
+
+  return value
+
--- a/fishtest/setup.py
+++ b/fishtest/setup.py
@@ -14,7 +14,8 @@ requires = [
     'numpy<1.16',
     'scipy<1.2',
     'requests',
-    'boto'
+    'boto',
+    'numba',
     ]
 
 setup(name='fishtest-server',


tomtor:
I have not been able to get a speedup with a 40 parameter tune so I am closing this for now

Time per SPSA computation (complete outer loop including logging is 0.0001s, JIT compiled is 4 times slower for some reason

fishtest/fishtest/rundb.py
```diff
@@ -647,3 +653,20 @@ def update_spsa(self, worker, run, spsa_results):
...
+@jit(nopython=True)
+def jit_spsa_param_clip_round(theta, pmin, pmax, increment, clipping):
+  value = 0.0
+
+  if clipping == 'old':
+    value = min(max(theta + increment, pmin), pmax)
+  else: #clipping == 'careful':
+    inc = min(abs(increment), abs(theta - pmin) / 2, abs(theta - pmax) / 2)
+    if inc > 0:
+        inc_sgn = 0 if increment == 0 else increment / abs(increment)
```
ppigazzini: if `increment == 0` then `inc == 0` so the ternary operator can be dropped => `value = theta + inc * increment / abs(increment)`

fishtest/fishtest/rundb.py
```diff
@@ -647,3 +653,20 @@ def update_spsa(self, worker, run, spsa_results):
 
     if grow_summary:
       spsa['param_history'].append(summary)
+
+@jit(nopython=True)
+def jit_spsa_param_clip_round(theta, pmin, pmax, increment, clipping):
+  value = 0.0
```
ppigazzini: @tomtor IMO the initialization can be dropped

ppigazzini:
@tomtor I have a speedup using `sorted` to clip in a range
```
python3 test_range_clip.py
sorted = 0.24389982223510742
minmax = 0.3744676113128662
```

```python
import time
import random

a=[[random.random() for i in range(3)] for j in range(1000000)]
# force x[0]=pmin < x[2]=pmax for minmax
for x in a:
  x[2]+=x[0]

t=time.time()
for x in a:
  sorted(x)[1]
print("sorted = %s" % (time.time()-t))

t=time.time()
for x in a:
  min(max(x[0], x[1]), x[2])
print("minmax = %s" % (time.time()-t))
```

fishtest/fishtest/rundb.py
tomtor: Yes, you are correct. Changes should be applied to original `spsa_param_clip_round` around line 520 because for some reason the JIT compiler didn't speed-up.

fishtest/fishtest/rundb.py
tomtor: Yes, you are correct! But most (all?) tests are run with `clipping == 'old'` so no speedup in general case.

Will you create a new PR for `spsa_param_clip_round`?
Note that the rounding code can also be removed (line 538) from that function

tomtor:
> I have a speedup using `sorted` to clip in a range

I rewrote with adding the results to a variable but it is indeed slightly faster to sort...

fishtest/fishtest/rundb.py
ppigazzini: @tomtor for the general case we can try replacing `min(max())` w/ `sorted` and bench for a speedup.
I'm not the best person to submit a new PR:
- I don't know the whole SPSA code and how to bench the speedup for that code
- I don't know how to submit a new SPSA test (always reschedule a recent one)

tomtor: OK, but I will not introduce the `sorted` because the speed difference is small and creating the list takes time.

ppigazzini:
@tomtor best performance using `if`, tested also in the worst case of values already ordered.

```
$ python3 test_range_clip.py
sorted:
0.2563669681549072
minmax:
0.3586390018463135
if:
0.17473316192626953
```

```python
import time
import random

a=[[random.random() for i in range(3)] for j in range(1000000)]
a=[i if i[0]<=i[2] else i[::-1] for i in a]
#a=[list(range(3)) for j in range(1000000)]

t=time.time()
for x in a:
  sorted(x)[1]
print("sorted:\n%s" % (time.time()-t))

t=time.time()
for x in a:
  min(max(x[0], x[1]), x[2])
print("minmax:\n%s" % (time.time()-t))

t=time.time()
for x in a:
  if x[1]<x[0]:
    x[1]=x[0]
  elif x[1]>x[2]:
    x[1]=x[2]
print("if:\n%s" % (time.time()-t))
```

ppigazzini:
@tomtor I'm using this PR to learn some inside of python, e.g. in place `.sort()` is faster than built-in `sorted()`, but the C style `if` is always the fastest. I rewrote the test script mimic the real code to account for the overhead, but I don't know if the speedup will translate to the whole SPSA algorithm, though.

* input random values
```
sort:
0.3505716323852539
minmax:
0.4158294200897217
if:
0.24533867835998535
```
* input ordered values (should be the higher frequency case)
```
sort:
0.37754106521606445
minmax:
0.46552610397338867
if:
0.22386813163757324
```

``` python
import time
import random

a=[[random.random() for i in range(3)] for j in range(1000000)]
a=[i if i[0]<=i[2] else i[::-1] for i in a]
#a=[[1.0, 2.0, 3.0] for j in range(1000000)]

a=[{'min': i[0], 'theta': i[1], 'max': i[2]} for i in a]
inc = 0.001

t=time.time()
for x in a:
  x=[x['min'], x['theta']+inc, x['max']]
  x.sort()
print("sort:\n%s" % (time.time()-t))

t=time.time()
for x in a:
  min(max(x['min'], x['theta']+inc), x['max'])
print("minmax:\n%s" % (time.time()-t))

t=time.time()
for x in a:
  y=x['theta']+inc
  if y<x['min']:
    y=x['min']
  elif y>x['max']:
    y=x['max']
print("if:\n%s" % (time.time()-t))
```

tomtor:
> but I don't know if the speedup will translate to the whole SPSA algorithm

Yes, it will. It is in the inner loop so it will be a speedup, but it is unclear how much without measuring it. I expect a very modest total speedup because there is a lot of additional code.

I instrumented by hand the begin/end of the calling function and the inner function to test the Numba speedup. It is still unclear to me why the Numba change did not work. Pypy is also still interesting to experiment with.

Note that the current performance is actually more than sufficient for the current tuning runs...
If it really was a bottleneck I would rewrite the SPSA APi in Rust (I LOVE rust 👍 https://www.rust-lang.org/ )

ppigazzini:
@tomtor "a lot of additional code" in pythonic sense I suppose, because it's a simple `if elif
` compound statement that avoid two function calls. By the way this is the C code style that Numba should be able to speedup ;)
 
https://github.com/glinscott/fishtest/blob/685a7188a81b8f1c051bc8fed6f8f80afa7b26b7/fishtest/fishtest/rundb.py#L519-L527

```python
  def spsa_param_clip_round(self, param, increment, clipping, rounding):
    if clipping == 'old':
      #value = min(max(param['theta'] + increment, param['min']), param['max'])
      value = param['theta'] + increment
      if value < param['min']:
        value = param['min']
      elif value > param['max']:
        value = param['max']
    else: #clipping == 'careful':
      inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
      if inc > 0:
          value = param['theta'] + inc * increment / abs(increment)
      else: #revert to old behavior to bounce off boundary
          #value = min(max(param['theta'] + increment, param['min']), param['max'])
          value = param['theta'] + increment
          if value < param['min']:
            value = param['min']
          elif value > param['max']:
            value = param['max']
```

tomtor:
I meant "additional code" in the calling loop. But indeed the if/else is optimal for numba

ppigazzini:
Reading now that "additional code" while commuting, IMO was wrote without thinking about performance.

ppigazzini:
For each parameter we repeat over and over the same conditional jumps about clipping and rounding. Should be better to move out the logic from the loop and:
* retire `spsa_param_clip_round()` and insert `spsa_param_clip_old()`, `spsa_param_clip_careful()`, `spsa_param_clip_old_round()`, `spsa_param_clip_careful_round()`. A 
* or rewrite `spsa_param_clip_round()` as closure
https://late.am/post/2015/05/07/optimize-python-with-closures.html

ppigazzini:
@tomtor I wrote matlab code, so I dislikes `if` `for` `while`.

https://github.com/glinscott/fishtest/blob/685a7188a81b8f1c051bc8fed6f8f80afa7b26b7/fishtest/fishtest/rundb.py#L532-L537
```
$ python3 test_rand.py
master:
0.030910253524780273
pr fast:
0.020459413528442383
pr fastest:
0.01808762550354004
```

```python
import time
import random
import math

a=[[20 * (random.random() - 0.5), random.uniform(0,1)] for i in range(100000)]

t=time.time()
for value, rand_u in a:
  fl = math.floor(value) #greatest integer <= value, thus works for negative.
  if rand_u < value - fl:
    value = fl + 1
  else:
    value = fl
print("master:\n%s" % (time.time()-t))

t=time.time()
for value, rand_u in a:
  value = math.floor(value + 1 - rand_u) # same results like master, a bit slower
print("pr fast:\n%s" % (time.time()-t))

t=time.time()
for value, rand_u in a:
  value = math.floor(value + rand_u) # same probabilistic effect but different results wrt master, fastest
print("pr fastest:\n%s" % (time.time()-t))
```

ppigazzini:
@tomtor learning some python. I was wrong, with closure there is not speedup
```
$ python3 test_param.py

clip: old  round: deterministic
0.36438441276550293 - master
0.27333641052246094 - closure

clip: old  round: randomized
0.8396246433258057 - master
0.6558735370635986 - closure

clip: careful  round: deterministic
0.9201669692993164 - master
0.8676719665527344 - closure

clip: careful  round: randomized
1.418668270111084 - master
1.1879796981811523 - closure
```


```python
import time
import random
import math

def spsa_param_clip_round_maker(clipping, rounding):

  def f0(param, increment):
    value = param['theta'] + increment
    if value < param['min']:
      value = param['min']
    elif value > param['max']:
      value = param['max']
    value = math.floor(value + random.uniform(0,1))
    return value

  def f1(param, increment):
    value = param['theta'] + increment
    if value < param['min']:
      value = param['min']
    elif value > param['max']:
      value = param['max']
    return value

  def f2(param, increment):
    inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
    if inc > 0:
        value = param['theta'] + inc * increment / abs(increment)
    else: #revert to old behavior to bounce off boundary
        value = param['theta'] + increment
        if value < param['min']:
          value = param['min']
        elif value > param['max']:
          value = param['max']
    value = math.floor(value + random.uniform(0,1))
    return value

  def f3(param, increment):
    inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
    if inc > 0:
        value = param['theta'] + inc * increment / abs(increment)
    else: #revert to old behavior to bounce off boundary
        value = param['theta'] + increment
        if value < param['min']:
          value = param['min']
        elif value > param['max']:
          value = param['max']
    return value

  if clipping == 'old':
    if rounding == 'randomized':
      return f0
    else:
      return f1
  else: #clipping == 'careful':
    if rounding == 'randomized':
      return f2
    else:
      return f3

def spsa_param_clip_round(param, increment, clipping, rounding):
  if clipping == 'old':
    value = param['theta'] + increment
    if value < param['min']:
      value = param['min']
    elif value > param['max']:
      value = param['max']
  else: #clipping == 'careful':
    inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
    if inc > 0:
        value = param['theta'] + inc * increment / abs(increment)
    else: #revert to old behavior to bounce off boundary
        value = param['theta'] + increment
        if value < param['min']:
          value = param['min']
        elif value > param['max']:
          value = param['max']

  #'deterministic' rounding calls round() inside the worker.
  #'randomized' says 4.p should be 5 with probability p, 4 with probability 1-p,
  #  and is continuous (albeit after expectation) unlike round().
  if rounding == 'randomized':
      value = math.floor(value + random.uniform(0,1))

  return value

def main():
  a=[[random.random() for i in range(3)] for j in range(1000000)]
  a=[i if i[0]<=i[2] else i[::-1] for i in a]
#  a=[[1.0, 2.0, 3.0] for j in range(1000000)]

  a=[{'min': i[0], 'theta': i[1], 'max': i[2]} for i in a]

  increment = 0.001

  for clip_type in ('old', 'careful'):
    for round_type in ('deterministic', 'randomized'):
      print()
      print("clip: %s  round: %s" % (clip_type, round_type))
      t0=time.time()
      for param in a:
        value = spsa_param_clip_round(param, increment, clip_type, round_type)
      t1=time.time()
      print("%s - master" % (t1-t0))

      t0=time.time()
      spsa_param_clip_round_f = spsa_param_clip_round_maker(clip_type, round_type)
      for param in a:
        value = spsa_param_clip_round_f(param, increment)
      t1=time.time()
      print("%s - closure" % (t1-t0))

main()
```

PR #453, opened 2019-11-29
Script to create indexes.

xoto10:
First go at a script to recreate the indexes in the fishtest db.

PR #457, opened 2019-12-01
Retire unused 'old_user_cache' MongoDb collection

tomtor:
See #453 discussion


ppigazzini:
@tomtor dropped the collection from DEV mongo
```
> use fishtest_new
switched to db fishtest_new
> show collections
actions
clop
deltas
flag_cache
old_runs
old_user_cache
pgns
regression_data
runs
top_month
user_cache
users
> db.old_user_cache.drop()
true
> show collections
actions
clop
deltas
flag_cache
old_runs
pgns
regression_data
runs
top_month
user_cache
users
>

```

PR #458, opened 2019-12-02
Keep some of the gain when a test gets ahead of expected throughput.

xoto10:
If fishtest is quiet, the tests that are running can get well ahead of where the throughput calculation expects them to be. If a new test is then submitted the new test gets most/all of the available cores because it is neutral and the older tests are ahead of expected throughput. e.g. a few days ago I noticed one of my STC tests get more than half of the available cores when it started. In some cases, repeated creation of new tests seems to starve older tests of resources.

This patch attempts to reduce/remove this effect by adjusting the start time used in the throughput calculation. A copy of the actual start time of the job is made, and if the job gets ahead of its expected throughput this copy is adjusted to an earlier time so that the job can keep some of its "extra" gains. The calculation used only keeps half of the gain, but in practice I expect repeated application of this adjustment to allocate most of the gain to the job. This should mean that new tests only get a modest allocation of cores instead of the current situation where they sometimes get a large proportion of the available cores. It's possible this will also help when many extra cores are added to fishtest.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -286,7 +286,7 @@ class RunDb:
     # A copy of the test start date is used. This is adjusted if the test gets ahead of its expected schedule
     # so that it can keep some of those gains.
     if (run['args']['throughput'] != None and run['args']['throughput'] != 0):
-      if (run['args']['ip_start_time'] == None):
+      if ('ip_start_time' not in run['args']):
         run['args']['ip_start_time'] = run['start_time']
       new_internal_prio = - time.mktime(run['args']['ip_start_time'].timetuple()) - \
         task_id * 3600 * self.chunk_size * run['args']['threads'] / run['args']['throughput']
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -89,8 +89,8 @@ class RunDb:
       'auto_purge': auto_purge,
       'throughput': throughput,
       'priority': priority,
-      'ip_start_time': - time.mktime(start_time.timetuple())
-      'internal_priority': - time.mktime(start_time.timetuple()),
+      'ip_start_time': - time.mktime(start_time.timetuple()),
+      'internal_priority': - time.mktime(start_time.timetuple())
     }
 
     if sprt != None:
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -286,7 +286,8 @@ class RunDb:
     # A copy of the test start date is used. This is adjusted if the test gets ahead of its expected schedule
     # so that it can keep some of those gains.
     if (run['args']['throughput'] != None and run['args']['throughput'] != 0):
-      run['args']['ip_start_time'] = run['start_time'] if (run['args']['ip_start_time'] == None)
+      if (run['args']['ip_start_time'] == None):
+        run['args']['ip_start_time'] = run['start_time']
       new_internal_prio = - time.mktime(run['args']['ip_start_time'].timetuple()) - \
         task_id * 3600 * self.chunk_size * run['args']['threads'] / run['args']['throughput']
       time_gain = -time.time() - new_internal_prio
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -89,6 +89,7 @@ class RunDb:
       'auto_purge': auto_purge,
       'throughput': throughput,
       'priority': priority,
+      'ip_start_time': - time.mktime(start_time.timetuple())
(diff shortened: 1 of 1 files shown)


xoto10:
This is currently untested by me as I am having trouble getting workers to connect to my fishtest server. I have created the PR so we can discuss it - no point in me continuing if the consensus is that this isn't a problem anyway.

tomtor:
I think a similar effect is possible with less code:

Do not adjust the priority of a test if the effective througput is higher than the requested througput and only one test is runnable.

xoto10:
I don't think that does anything? If only one test is runnable then all available cores will be allocated to that test whatever the priority. Once another test becomes runnable the priorities will be set in the same way as currently happens.

To clarify what is happening, how about we display the internal priority as a number (as we do now) but also as a time in words ` time.ctime(-internal_prio) `? If the system is running at the expected throughput this time should roughly equal the current UTC time, but generally it will be ahead or behind and the date/time displayed (and how it changes) will give us a better idea of how the internal priority system is working?
e.g. it is now Dec 8 16:09 and a job currently running has internal prio at Dec 9 03:58 because only a few jobs have been running in the last few hours.

Ideally I'd like to do something about tests that are parked at lower priorities, but that seems hard to deal with. This seems not too complicated, i.e. something we can fix. We adjust the base time used for the throughput calc instead of always using the start_time - easy to do for jobs that get ahead.

tomtor:
> I don't think that does anything? If only one test is runnable then all available cores will be allocated to that test whatever the priority. Once another test becomes runnable the priorities will be set in the same way as currently happens.

The difference is that the single running test its priority remains unchanged, it gets the task slots for free (as if they never were received by this test). So it will be sooner on equal priority with newer tests.

Anyway, I do not think there is really a need to change the current time allocation...

vondele:
personally, I would prefer to see a system where each running test gets a fraction of available resources, independent of how many games a test has played so far. The base would be that every test gets the same fraction, but one could adjust that e.g. by LLR / test type / TC etc.

xoto10:
@tomtor 
But the priority updates are not increments, each update recalculates the internal_prio from the number of tasks used so far. The first update that is made will include all the "missing" task slots.

@vondele 
Yes, I think my change will broadly do what you are looking for. Not 100%, but much, much closer than now. Longer running tests should not be sidelined by more recently started tests like they often are now.

Vizvezdenec:
We had 1000 - problem was that tests basically refused to finish so we had enormous queue of tests that played like 10k STC games and didn't converge.
Then we had 10000 - problem was the opposite, even newly submitted tests got nothing, so workers were recycling between the same 2-3 tests till they finish so tests could've got stuck with 0 games played for half a day.
Then we had 3000 - some middle ground, but it seems that since change of SPRT bounds we have again some showings of 1st problem - my LTC for example is stuck with basically 0 workers with LLR 1.5 for quite some time. And it's pretty usual case since after new SPRT bounds we run much more LTCs (because STCs are quite easy to pass now).
So probably we should lower default TP a bit to 2000/2500 or make LTC tests distribute workers more linearly and not have some tests grabbing 500 workers while leaving others with 10 because they are 8 hours younger.
STC tests seem to be doing pretty fine currently for me, LTCs though have problems of not converging if they need > 30k games to do so.

xoto10:
@Vizvezdenec 
Exactly. The problem is that the number of games / hour varies depending on how many tests there are, so the number built into the throughput calculation can only ever be a very rough approximation. I think this change will make the calculation more reliable over a variable range of test numbers. It won't be any help when the system is overloaded, but mostly we have the opposite problem at the moment, with it not scheduling small numbers of tests well. Also, it should help if fishtest gets busy again and noob adds a load of cores, since we get the same problem then - new tests tend to freeze out old ones.

xoto10:
As an example of the kind of thing I hope this will improve, a few moments ago on fishtest, this interesting core split between 2 LTC tests: https://imgur.com/a/fuA2Vgf
It seems a bit unfair that my test has 10x the number of cores that Viz test has.

tomtor:
> personally, I would prefer to see a system where each running test gets a fraction of available resources, independent of how many games a test has played so far. The base would be that every test gets the same fraction, but one could adjust that e.g. by LLR / test type / TC etc.

@vondele That would also deal with changes in priorities/throughput in a less disruptive way, which is IMO the only real issue with the current implementation. 
The worker allocator can allocate the workers to the run with the lowest total active Mnps, that should not be too hard to implement.

Throughput should be changed to a percentage, defaulting to 100% for STC, so users can still lower their test priority.  LTC should default to 600%.

If most fishtest users think this would be a substantial improvement over the current implementation, then I will have a look at implementing this.

PR #459, opened 2019-12-04
Small speedup of SPSA by removing unneeded code

tomtor:
Suggested by @ppigazzini 

See https://github.com/glinscott/fishtest/pull/455


ppigazzini:
@tomtor DEV updated, fortunately I found a SPSA w/ still the branch, rescheduled with the careful algorithm (the SPSA page don't show the options selected)

PR #461, opened 2019-12-06
Raise worker version to 70 (server side)

ppigazzini:
Force the workers update in fishtest


ppigazzini:
@tomtor I wrote the post on fishcooking, PR merge scheduled at 21.00 CET

PR #462, opened 2019-12-06
Small SPSA speedup with several code improvements

ppigazzini:
Clip a value in a range using an if statement instead that a double function calls.
Drop some if statements and rewrote the rounding code as an oneliner.
The new rounding code has the same probabilistic behavior of the old code, results are equal if mirroring the random variable (but the code in that case will be a little slower).
Code simplification removing unnecessary variables.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -518,23 +518,27 @@ class RunDb:
 
   def spsa_param_clip_round(self, param, increment, clipping, rounding):
     if clipping == 'old':
-      value = min(max(param['theta'] + increment, param['min']), param['max'])
+      value = param['theta'] + increment
+      if value < param['min']:
+        value = param['min']
+      elif value > param['max']:
+        value = param['max']
     else: #clipping == 'careful':
       inc = min(abs(increment), abs(param['theta'] - param['min']) / 2, abs(param['theta'] - param['max']) / 2)
       if inc > 0:
           value = param['theta'] + inc * increment / abs(increment)
       else: #revert to old behavior to bounce off boundary
-          value = min(max(param['theta'] + increment, param['min']), param['max'])
+          value = param['theta'] + increment
+          if value < param['min']:
+            value = param['min']
+          elif value > param['max']:
+            value = param['max']
 
     #'deterministic' rounding calls round() inside the worker.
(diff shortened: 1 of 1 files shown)


ppigazzini:
```
$ python3 test_rand.py
master:
0.030910253524780273
pr fast:
0.020459413528442383
pr fastest:
0.01808762550354004
```

```python
import time
import random
import math

a=[[20 * (random.random() - 0.5), random.uniform(0,1)] for i in range(100000)]

t=time.time()
for value, rand_u in a:
  fl = math.floor(value) #greatest integer <= value, thus works for negative.
  if rand_u < value - fl:
    value = fl + 1
  else:
    value = fl
print("master:\n%s" % (time.time()-t))

t=time.time()
for value, rand_u in a:
  value = math.floor(value + 1 - rand_u) # same results like master, a bit slower
print("pr fast:\n%s" % (time.time()-t))

t=time.time()
for value, rand_u in a:
  value = math.floor(value + rand_u) # same probabilistic effect but different results wrt master, fastest
print("pr fastest:\n%s" % (time.time()-t))
```

ppigazzini:
@tomtor PR testing on DEV.
By the way, SPSA is CPU taxing or it add delay to the workers? Just to hit the right balance between performance/code readability.

tomtor:
> SPSA is CPU taxing or it add delay to the workers

If we max out to 100% CPU it will obviously add delay to the workers, but in general some delay is no issue because cutechess will continue playing while the worker waits for a fishtest response.

The main CPU-usage used to be parsing the JSON SPSA parameter list received from the worker (we now store that in the server), so computing the new SPSA parameters (the code you working on now) uses a variable amount of CPU depending on the number of SPSA parameters.

For the current number of workers and number of SPSA parameters tuned there is no performance issue, so it is better to prefer readability over slightly faster code.

ppigazzini:
@tomtor I never controlled the CPU load on PROD during SPSA, so I was asking.
About this PR, IMO :
- rounding now is very easy to understand, easy for my eyes and faster
- flip dropping bool() is easy to my eye (and faster too)
- variables `a` and `R` were never used
- the biggest speedup is the clipping with `if`, but it add code, min(max()) was very easy to my eyes

Let me know if you want some changes applied :)

tomtor:
I am OK with the `if` and the other code changes 👍

PR #463, opened 2019-12-07
Drop another MongoDb index

tomtor:
See https://github.com/glinscott/fishtest/pull/444#issuecomment-562833654


ppigazzini:
@tomtor PROD updated :)

PR #464, opened 2019-12-08
Red color when pending users

tomtor:
Clear pending cache on update


ppigazzini:
Tested ok on DEV, PROD updated. Thank you @tomtor

PR #466, opened 2019-12-08
Update reCAPTCHA public key

ppigazzini:
Update the public key after the owner change.
Private key updated by hand.


ppigazzini:
@tomtor I registered the PROD and DEVs domains (mine and yours) under the email address created for fishtest.

PR #467, opened 2019-12-09
Remove parameter tweak option for test bounds

xoto10:
In the past, parameter tweak tests were allowed to use easier bounds than normal tests. Now that normal test bounds have been changed they are easier to pass than the old parameter tweak bounds so the [0,4] bounds are redundant.

This patch removes the ` Parameter Tweak [0,4] ` option from the SPRT bounds drop-down in the New Test web page.


tomtor:
See also code at line 190

tomtor:
See also code at line 209

ppigazzini:
Tested OK on DEV, merged and PROD updated.
Thank you @xoto10 !

tomtor:
Updated wiki:

https://github.com/glinscott/fishtest/wiki/Creating-my-first-test

PR #468, opened 2019-12-10
Calculate priority with number of active cores

tomtor:
See https://github.com/glinscott/fishtest/pull/458

@ppigazzini @vdv @xoto10 I have tested this locally and it is ready for some more testing and feedback.

This change allocates cores round-robin. It is not completely fair now, because all cores are treated equally, so some workers might get faster cores than others. This is easy to change to Mnps calculation, but because we show allocated cores on the main page this is both simple to code and least confusing for users.

Note that old existing tests will have 3000 as throughput, so until they are finished they will get a large amount of the cores. When merging we should change prio to 100 for old tests

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -13,6 +13,7 @@ from bson.binary import Binary
 from pymongo import MongoClient, ASCENDING, DESCENDING
 from userdb import UserDb
 from actiondb import ActionDb
+from views import format_results
 
 import fishtest.stat_util
 
@@ -62,7 +63,7 @@ class RunDb:
               username=None,
               tests_repo=None,
               auto_purge=True,
-              throughput=3000,
+              throughput=100,
               priority=0):
     if start_time == None:
       start_time = datetime.utcnow()
@@ -88,8 +89,8 @@ class RunDb:
       'tests_repo': tests_repo,
       'auto_purge': auto_purge,
       'throughput': throughput,
+      'itp': 100,  # internal throughput
       'priority': priority,
-      'internal_priority': - time.mktime(start_time.timetuple()),
     }
 
     if sprt != None:
@@ -271,20 +272,30 @@ class RunDb:
 
     return results
 
-  def recalc_prio(self, run, task_id=None):
-    if task_id is None:
-      task_id = -1
-      for task in run['tasks']:
-        task_id = task_id + 1
-        if not task['active'] and task['pending']:
-          break
-    # Recalculate internal priority based on task start date and throughput
-    # Formula: - second_since_epoch - played_and_allocated_tasks * 3600 * chunk_size / games_throughput
-    # With default value 'throughput = 3000', this means that the priority is unchanged as long as
-    # we play at rate '3000 games / hour'.
-    if (run['args']['throughput'] != None and run['args']['throughput'] != 0):
-      run['args']['internal_priority'] = - time.mktime(run['start_time'].timetuple()) - \
-        task_id * 3600 * self.chunk_size * run['args']['threads'] / run['args']['throughput']
+  def calc_itp(self, run):
+    itp = run['args']['throughput']
+    if itp < 1:
+      itp = 1
+    elif itp > 500:
+      itp = 500
+    itp *= math.sqrt(float(run['args']['tc'].split('+')[0]) / 10)
+    itp *= math.sqrt(run['args']['threads'])
+    if 'sprt' not in run['args']:
+      itp *= 0.5
+    else:
+      results = self.get_results(run)
+      run['results_info'] = format_results(results, run)
+      if 'llr' in run['results_info']:
+        llr = run['results_info']['llr']
+        itp *= (5 + llr) / 5
+    run['args']['itp'] = itp
+
+  def sum_cores(self, run):
+    cores = 0
+    for task in run['tasks']:
+      if task['active']:
+        cores += int(task['worker_info']['concurrency'])
+    run['cores'] = cores
 
   # Limit concurrent request_task
   task_lock = threading.Lock()
@@ -304,12 +315,17 @@ class RunDb:
       return {'task_waiting': False}
 
   def sync_request_task(self, worker_info):
-
     if time.time() > self.task_time + 60:
       self.task_runs = []
       for r in self.get_unfinished_runs():
+        run = self.get_run(r['_id'])
+        self.sum_cores(run)
+        r['cores'] = run['cores']
+        self.calc_itp(run)
+        r['args']['itp'] = run['args']['itp']
         self.task_runs.append(r)
-      self.task_runs.sort(key=lambda r: (-r['args']['priority'], -r['args']['internal_priority'], r['_id']))
+      self.task_runs.sort(key=lambda r: (-r['args']['priority'],
+        r['cores'] / r['args']['itp'] * 100.0, -r['args']['itp'], r['_id']))
       self.task_time = time.time()
 
     max_threads = int(worker_info['concurrency'])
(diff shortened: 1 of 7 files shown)


ppigazzini:
@tomtor DEV updated, feel free to suggest how to test :)

ppigazzini:
DEV updated

tomtor:
@ppigazzini Thanks, I have allocated my 2 workers to DEV. Could you do the same?

All we can do is fiddle with some tests and look for strange events?

tomtor:
@ppigazzini Looks good, the run with 25% throughput get about 25% ...

ppigazzini:
@tomtor added 8 workers each w/ 3 cores.

tomtor:
@ppigazzini @vondele Looks good to me. It is obviously less accurate than the old method. A worker with lower priority could get eg a 32-core worker, when it is on average entitled to 3 cores, but it should all average out with a lot of workers over a longer time....

ppigazzini:
@tomtor I suddenly caught all the workers raising the throughput, cheating is so easy :)

tomtor:
I squashed....

vondele:
interesting... I wouldn't worry about the precise NPS of workers, or 'fractions of a worker'... this is just a detail.

I would do the adjustment for the TC internally... i.e. leave the throughput at 100% also for LTC on the submission website, but correct for TC internally, e.g. by giving all jobs a  (TC / 10s) boost. 

In fact, I would scale with
- sqrt(TC / 10s) 
- sqrt(cores)
- 0.5 for numgames type runs
- 0.5 for SPSA runs
- (5 + LLR)/5 for SPRT.

The rationale is to give more weight to SPRT tests (relative to numgames and spsa), and even more so if they are close to pass. I would sqrt-scale wrt to cores and TC, just because these jobs need more resources, but if we put too much weight on those, they will drain the framework for testing.

xoto10:
There's a function somewhere that calculates the scaling factor for the
time control, you can just call that to allocate more codes to longer tc
tests. I would do that linearly, so ltc gets 6x cores of stc, at least for
now. And leave the other adjustments vdv suggests as well. They are
interesting to consider, but I think the initial implementation should be
simpler and closer to the current method. We can experiment with e.g. more
cores for jobs with higher LLR later.

My change is a lot less code :-)

tomtor:
@ppigazzini @vondele I implemented the suggestions of Joost and tested locally...

ppigazzini:
@tomtor DEV updated, I wrote some tests with different throughput and connected 10 workers.

vondele:
nice! The real load-balancing won't be visible unless there are many workers.

As soon as you feel the code is robust, this can be tested on the production server IMO, as an experiment. If this is not performing as expected it can be reverted or refined.

ppigazzini:
@tomtor @vondele started 30 mini workers.

tomtor:
> @tomtor @vondele started 30 mini workers.

@ppigazzini They all do 0.00 Mnps?

ppigazzini:
@tomtor 30 workers w/ 1 core put too much pressure on the VPS storage, I rebooted the server and started 10 workers w/ 3 cores.

tomtor:
@ppigazzini @vondele Note that you can see the `itp` (internal throughput) on the individual test pages from 
https://dfts-2.pigazzini.it/tests

ppigazzini:
@tomtor DEV updated

xoto10:
> I would do the adjustment for the TC internally... i.e. leave the throughput at 100% also for LTC on the submission website, but correct for TC internally

@tomtor 
Not sure if you're doing this bit or not, I didn't see tc mentioned in your code but I only looked briefly. The function I was thinking of earlier is ` parse_tc() ` in views.py, very useful if you do want to automatically adjust for tc.

xoto10:
> As soon as you feel the code is robust, this can be tested on the production server IMO, as an experiment. If this is not performing as expected it can be reverted or refined.

Agreed. I was hoping that would happen with my suggestion, but we're going for the rewrite approach instead. Whatever.

vondele:
My first impression is that this looks good ...

ppigazzini:
@tomtor some notes:
* I was able to set throughput <= 0: this makes sense?
* in test page perhaps I suggest to change "Adjust throughput:" in "Adjust Throughput%:"
* rescheduling an old test I got "Throughput%: 3000"

tomtor:
> > I would do the adjustment for the TC internally... i.e. leave the throughput at 100% also for LTC on the submission website, but correct for TC internally
> 
> @tomtor
> Not sure if you're doing this bit or not, I didn't see tc mentioned in your code but I only looked briefly. The function I was thinking of earlier is `parse_tc()` in views.py, very useful if you do want to automatically adjust for tc.

parse_tc does something else. It computes an expected game duration

xoto10:
>  It computes an expected game duration

Yes, exactly. I thought that might be useful if you want to adjust core allocation based on expected game duration since it covers base time and increment.

vondele:
>     * I was able to set throughput <= 0: this makes sense?

No, we should keep this input >0 

>     * in test page perhaps I suggest to change "Adjust throughput:" in "Adjust Throughput%:"

agreed

>     * rescheduling an old test I got "Throughput%: 3000", a bit dangerous during the transition phase

I wonder if we should reject Throughput > e.g. 500. If needed we still have priority 1 for that.

ppigazzini:
@vondele @tomtor I'm wondering if we should have quantized values (e.g 10, 20,..., 100, default 60)

tomtor:
> No, we should keep this input >0

It is mapped to 1 when < 1, I will map > 500 to 500. I will change throughput of old tests by hand to 100.

I missed `Throughput%` for adjusting, changed that too and squashed.

@ppigazzini For me it's ok to merge and test in prod

tomtor:
> I'm wondering if we should have quantized values (e.g 10, 20,..., 100, default 60)

Why 60 and not 100? I think it's fine for now.

ppigazzini:
@tomtor just started commuting. At the moment there is a quantized set 1 - 500 with default 100, just wondering if we need so much headroom above the default.

tomtor:
@ppigazzini The old implementation had no checking or limits on throughput at all. Users can also raise priority which has always allocated all cores. So the 1-500 range is now more limiting than the original code.

ppigazzini:
@tomtor @vondele Before merging and updating the PROD server IMO we should write a post on fishcooking to inform about the change.

tomtor:
> @vondele @tomtor I'm wondering if we should have quantized values (e.g 10, 20,..., 100, default 60)

@ppigazzini @vondele I do agree that this could be an improvement when implemented with a popup selector with values:

10%, 25%, 50%, 100% (default),  200% (in red)

tomtor:
I added the selector for NEW tests. Updating is still with typed input so a range of 1-500 remains possible. Best of both worlds.

tomtor:
> @tomtor @vondele Before merging and updating the PROD server IMO we should write a post on fishcooking to inform about the change.

I will write one...

ppigazzini:
Nice patch @tomtor , thank you to @xoto10 to starting the improvement.

tomtor:
@ppigazzini Merged, restarted and changed old runs to prio 100 (and 50/25)

ppigazzini:
@tomtor I'm viewing on the web. Finally the throughput value makes sense.

xoto10:
Nice work @tomtor !

xoto10:
Looks ok as far as I can tell so far. :-)
I would allocate the full 6x cores for LTC, but only half of what you're allocating in the threaded cases, but no doubt there will be many different opinions on that kind of detail, we can canvas opinion on that once it's been running for a while ...

vondele:
Yes, it seems to be working rather well:

image

rather balanced distribution of cores to tasks, no pending tasks, more cores in the more promising tests.

tomtor:
One thing that could be changed now is creating smaller batches (eg 100) for ltc/ multi core because they can occupy workers a long time...

vondele:
@tomtor that would be really a nice addition as well. Certainly worker side, since we cache the SF builds, I think the overhead of having smaller batches must be relatively small.

You'll be better aware of the possible tricky issues server-side (performance of the server, or 'accounting' of node hours to users donating time).

xoto10:
Odd distribution of workers just now:

image

Two long-running tunes with itp 12 and 30, seems odd one would get almost all the workers like this.

vondele:
Actually, right now looks a bit surprising:

image

I would expect to see the two tunes have roughly the same amount of cores (corrected by the user-set throughput), as they were assigned all available cores. The internal throughput seems right, but the number of cores assigned is vastly different, as if all cores were assigned to the test with the highest internal throughput.

tomtor:
@vondele But `ps_pawnscoring6` has prio `-1` so it SHOULD get no new workers !

ppigazzini:
@tomtor enhancements:
* expose the priority/throughput values, e.g. `p: 0 t: 100 c: 993`
* compute p-values and residuals to get some real time indexes about the fair cores allocation

This should help to keep a low fishtest's developers stress level.

Just kidding :)

vondele:
@tomtor thanks... I overlooked the priority -1 ... so more evidence it is working well :-)

PR #470, opened 2019-12-12
Minor fix for restarting SPSA

tomtor:


fishtest/fishtest/rundb.py
```diff
@@ -572,7 +572,7 @@ def store_params(self, run_id, worker, params):
 
   def get_params(self, run_id, worker):
     run_id = str(run_id)
-    if not run_id in self.spsa_params:
+    if not run_id in self.spsa_params or not worker in self.spsa_params[run_id]:
```
ppigazzini: @tomtor I think that we should use the `not in` operator :)
`if run_id not in self.spsa_params or worker not in self.spsa_params[run_id]:`

tomtor: Yes, you are right!

PR #473, opened 2019-12-17
Create script for adding test users

tomtor:
See https://github.com/glinscott/fishtest/pull/373


ppigazzini:
@tomtor I don't like to have this script on PROD, we could add for accident an approver with public credential. IMO better to write this script in the wiki.

tomtor:
Ok, no problem...

PR #474, opened 2019-12-17
favicon.ico and robots.txt independent from CWD

tomtor:
See https://github.com/glinscott/fishtest/pull/373


vdbergh:
Thanks. This appears to fix the issue.

ppigazzini:
Tested working on DEV and with a clean setup.
PROD updated, crontab fixed.
Thank you @tomtor, patch of the year :)

PR #475, opened 2019-12-19
Raise worker version to 71 server side

ppigazzini:


ppigazzini:
@d3vv I will force the auto update in 20 minutes, my test with some ubuntu workers was successful. Hopefully your fleet should update without problem: finger crossed :)

PR #479, opened 2019-12-24
Use logistic elo in fishtest.

vdbergh:
Following up on 

https://github.com/official-stockfish/Stockfish/issues/2459

The BayesElo bounds are a historical artifact and serve no purpose anymore. In fact internally they are now converted into logistic Elo bounds. This PR converts fishtest to use logistic Elo also externally.

To avoid confusion and to indicate that logistic bounds are used, various brackets [] have been replaced by braces {}.

We use the following default bounds
```
  STC     {-1,3}
  LTC     {0,2}
  Non-regression and simplification  {-1.5,0.5}
```
For comparison the BayesElo bounds were
```
  STC     [-1.5,4.5]
  LTC     [0,3.5]
  Non-regression and simplification  [-3,1]
```
We obtained the logistic bounds by inspecting the audit page of some tests (which lists the, previously internal, logistic Elo bounds) and doing some sensible rounding.


ppigazzini:
@vdbergh DEV updated. The code seems to have broken the automatic selection of the bounds when clicking STC/LTC buttons, view https://github.com/glinscott/fishtest/pull/406

vdbergh:
Thanks. I will check it out.

vdbergh:
Should be fixed now. I updated the PR.

ppigazzini:
@vdbergh DEV updated, fixed :)

ppigazzini:
@vdbergh I'm wondering if we should switch the order bayes/logistic Elo to logistic/bayes in raw statistic. If I'm not wrong Elo should be written with the first letter capitalized (named upon Arpad Elo).

vdbergh:
Do you mean writing the header as 

Logistic/BayesElo ?

ppigazzini:
Yes, logistic/bayes Elo in the header (and Elo wherever is written) :)

vdbergh:
Ok I changed it.

ppigazzini:
DEV updated

Merry Xmas :) @tomtor @vdbergh @vondele @snicolet @noobpwnftw

vondele:
Thanks! Same to you/all.

On the raw stats page, it is not yet clear if the 'basic statics Elo' are logistic or Bayes Elo, same for RMS bias. Maybe those could get an additional label as well.

vdbergh:
@vondele I put in a comment to clarify that all Elo is logistic, except when otherwise specified.

ppigazzini:
DEV updated

Merry Xmax to @glinscott !!!

vdbergh:
Does anyone know who is responsible for this script?

http://chess-sprt-calc.azurewebsites.net/

It should be updated for logistic (I can do that).

Probably it would be better to link to this script from the Wiki. It always takes me forever to find it.

ppigazzini:
@vdbergh it's already linked in this wiki page https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#testing-methodology-or-when-should-i-change-the-standard-parameters-for-tests

ppigazzini:
@vdbergh it should be written by @hwiechers (in the github page there is an email address). This is the forum thread https://groups.google.com/d/msg/fishcooking/7lUCJho7dLI/1AaMVJQbf48J

ppigazzini:
@vdbergh this should be the repo https://github.com/hwiechers/chess-sprt-calc

vdbergh:
Thanks!

vdbergh:
Yes I eventually found the link https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#testing-methodology-or-when-should-i-change-the-standard-parameters-for-tests you mention.

However "Creating my first test" would not be the first place I would go to look for an SPRT calculator... I would expect it to be in the side bar. Unfortunately the side bar is getting overcrowded. Maybe this is another "Issue"...

vondele:
ideally the expected number of games is part of the stats page, based on the used bounds and the estimated Elo :-)

ppigazzini:
@vdbergh let me know if I can merge this PR

vdbergh:
@ppigazzini I just did a few more checks on DEV on the Logictic/BayesElo values and everything seems to be consistent. So as far as I am concerned the PR is mergeable.

vdbergh:
@vondele Well we definitely also need something which is independent of any particular test.

I agree that it would also be useful to put some kind of ETA information in the stats page. But then I would like to put something which has a well defined statistical meaning so that for example it can be verified by simulation. It is tempting to try use somehow the current LLR or Elo estimate as part of an ETA estimate but right now I don't know what exactly the result would represent.

In any case it is an interesting problem to think about!

vdbergh:
@ppigazzini I squashed the PR to a single commit.

ppigazzini:
By the way we need to updated the wiki page with the new bounds.

vdbergh:
Thanks!

I changed the bounds in the wiki and put in a note that the resource table (which used BayesElo) is obsolete. I will recompute it.

vdbergh:
Wiki updated with new resource consumption tables.

PR #480, opened 2019-12-27
Fix Time Control Format.

vondele:
Looking into a number of timeouts in fishtest run using TC 0.5+0.005,
the workers with problems had an effective TC without increment, while others had one of 0.01:

5e0654a4c13ac2425c4a9e14-74.pgn:[TimeControl "0.55+0.01"]
5e0654a4c13ac2425c4a9e14-76.pgn:[TimeControl "0.36"]

This is due to rounding, with the given format string.

Cutechess allows for 3 digits, use that internally as well.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -182,10 +182,10 @@ def adjust_tc(tc, base_nps, concurrency):
     time_tc = float(chunks[0])
 
   # Rebuild scaled_tc now
-  scaled_tc = '%.2f' % (time_tc * factor)
+  scaled_tc = '%.3f' % (time_tc * factor)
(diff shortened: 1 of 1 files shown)


ppigazzini:
@vondele sprt @ 0.5+0.005 on DEV with a couple of workers updated with this PR (6 threads to filter out the other workers)
https://dfts-2.pigazzini.it/tests/view/5e0747ba7f5d2239cf1cd6cd

Info from a random pgn: `[TimeControl "0.602+0.006"]` looks good to me.

vondele:
yes, LGTM, feel free to merge whenever practical.

PR #483, opened 2019-12-29
Import the javascript SPRT calculator by Henri Wiechers in Fishtest.

vdbergh:
In this way it becomes easier to maintain.

The SPRT calculator has been updated to use logistic Elo. Moreover some input validation was added.

### DIFF
--- a/fishtest/fishtest/static/html/SPRTcalculator.html
+++ b/fishtest/fishtest/static/html/SPRTcalculator.html
@@ -1,0 +1,102 @@
+<!doctype html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <title>Chess SPRT Calculator</title>
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
+        <style>
+            td {
+                text-align: right;
+            }
+            .axis text {
+              font: 10px sans-serif;
+            }
+
+            .axis path, .axis line {
+              fill: none;
+              stroke: #000;
+              shape-rendering: crispEdges;
+            }
+
+            .line {
+              fill: none;
+              stroke: #000;
+              stroke-width: 2px;
+            }
+
+            .elo {
+              fill: none;
+              stroke: #000;
+              shape-rendering: crispEdges;
+            }
+
+            .grid-background {
+              fill: #ddd;
+            }
+
+            .grid line {
+              stroke: #fff;
+              stroke-width: 1px;
+              shape-rendering: crispEdges;
+            }
+
+            .grid .minor {
+              stroke-opacity: 0.25;
+            }
+
+            .grid text {
+              display: none;
+            }
+
+            .grid path,
+            .axis path {
+              display: none;
+            }
+
+            .form-control.number {
+              width: 4em;
+              text-align: right;
+              margin-right: 0.5em;
+            }
+
+            .chart-tooltip circle {
+                fill: white;
+                stroke: black;
+                stroke-width: 2px;
+            }
+
+        </style>
+    </head>
+    <body>
+        <div class=container>
+            <h1 class=page-header>Chess SPRT Calculator</h1>
+            <form id=parameters class=form-inline>
+                <div class=form-group>
+                    <label for=elo-0>Elo 0</label>
+                    <input id=elo-0 name=elo-0 value=-1 class='form-control number'>
+                </div>
+                <div class=form-group>
+                    <label for=elo-1>Elo 1</label>
+                    <input id=elo-1 name=elo-1 value=3 class='form-control number'>
+                </div>
+                <div class=form-group>
+                    <label for=draw-ratio>Draw ratio</label>
+                    <input id=draw-ratio name=draw-ratio value=0.56 class='form-control number'>
+                </div>
+                <div class=form-group>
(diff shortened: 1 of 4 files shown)


vdbergh:
This is a "non-functional change" :)

I have written to Henri Wiechers if he wants to continue to maintain this. If so then I will also make a PR on his github repository. He has not replied yet.

But in any case it is more convenient to have this script hosted on Fishtest so that it can be updated quickly. Once the script is imported I will adapt the wiki reference as well.

vdbergh:
I got a reply from Henri Wiechers. He says he has left computer chess. He is no longer interested in maintaining this and proposes that we take it over.

vdbergh:
I am reworking the script a bit to further simplify the formulas. I will update the PR soon.

vdbergh:
I have updated the PR. The script is now extremely simple.

http://hardy.uhasselt.be/Fishtest/sprt.js.

fishtest/fishtest/static/html/SPRTcalculator.html
```diff
@@ -0,0 +1,102 @@
...
+
+        </style>
+    </head>
+    <body>
+        <div class=container>
+            <h1 class=page-header>Chess SPRT Calculator</h1>
+            <form id=parameters class=form-inline>
+                <div class=form-group>
+                    <label for=elo-0>Elo 0</label>
+                    <input id=elo-0 name=bayes-elo-0 value=-1 class='form-control number'>
```
vondele: @vdbergh is the name (bayes-elo-0) still correct?

vdbergh: Thanks. The script uses getElementById it seems so the name of the field is not actually important. But of course it is not nice. I will change it (later during the day).

vdbergh: PR updated.

vondele: git push ?

vdbergh: Pushed to the wrong branch. Fixed now.

vdbergh:
I made the default value for the RMS bias zero now. The noob_3moves book appears to be much more balanced (the test showed a RMS bias of around 30) so this is a better default.

ppigazzini:
@vdbergh DEV updated
https://dfts-2.pigazzini.it/html/SPRTcalculator.html

I suggest to add a link in the main page:

https://github.com/glinscott/fishtest/blob/3d116f5349da4aeeb76079df483908a0f9d4ec17/fishtest/fishtest/templates/base.mak#L51-L84

vdbergh:
I put a link in the side bar under "Misc". If you feel it would be better under "Links" then I will change it.

vdbergh:
@ppigazzini As far as I am concerned the PR can be merged.

vdbergh:
After the merge I will update the links in the wiki (now they point to a version on my personal server).

ppigazzini:

vdbergh:
Thanks!

snicolet:
Comments about the SPRT calculator page, now that we have a version that we can improve on:
http://tests.stockfishchess.org/html/SPRTcalculator.html

1) Would it be possible to add the definitions of "Draw ratio" and "RMS bias" at the bottom of the page, and examples of use? The definition may be obvious to you, but not to me, for instance...

2) It would be nice to put credits on the page to Henri Wiechers, and maybe also a reference section with links to some PDF files explaining the theory of the calculations (your PDFs and/or other sources)

Thanks for the nice work!

vdbergh:
@snicolet I more or less agree. I am currently traveling but I will try to make the changes.

vdbergh:
Here is a version that adds some explanations http://hardy.uhasselt.be/Toga/SPRTcalculator.html. If it is ok then I will make a PR.

vondele:
Certainly nice & useful, but 'RMS bias' is still defined around 'the biases in the opening book', which isn't clear to me. The opening book is biased towards.... or deviates from the expected ... what?

vdbergh:
Ok I tried to add a definition of "bias of an opening". Let me know if it is clearer. To avoid confusion I am repeating the link (the above link is the same).

vondele:
Ah, yes, so 'white advantage' in Elo for the opening book. Makes sense.

Also, the tool and the explanation, of course, assume our alpha = beta = 0.05, mentioning this could be the first bullet on the list.

vdbergh:
One more :)

vondele:
nice, ready for an update.

kiudee:
@vdbergh would it be possible to allow users to pass larger absolute Elo values than 10 or are there numeric problems?

vdbergh:
@kiudee The bound 10 is arbitrary. The formulas used in the SPRT calculator are approximations which are extremely good for small Elo differences. For larger Elo differences the formulas become somewhat less accurate. It is also influenced by the batch size (I think a larger batch size is better for the accuracy). One can check this using simulation. 

I am curious. What is your use case?

kiudee:
I think for mature projects like Stockfish these tight bounds make sense. For new engine authors or during prototyping of new search algorithms one often wants to quickly test a few algorithmic variants. There it can often happen that they differ by more than 30 Elo points and it would be wasteful to run a [-10, 10] test. It would be very useful to plug in something like [-10, 30] into the calculator and see the rough expected samplesize.

PR #486, opened 2020-01-03
Use awscli v1.x package for backup management on aws s3

ppigazzini:
Switch from the outdated boto v2 package to the awscli v1.x package.
https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html

Document the aws cli usage and simplify the script.


ppigazzini:
@tomtor PROD updated, after some tests IMO aws cli is the best tool for our simple backup tasks.

PR #494, opened 2020-01-08
Percentage of interval display

Chess13234:
Add percentage interval display: its elo*100/abs(upperbound-lowerbound).
 Its a useful indicator of progress of tests, comparing current elo with interval between bounds - a higher % indicates a test closer to completion and bigger elo gain(elo takes more of interval).


ppigazzini:
`j.ci_upper-j.ci_lower` is always >= 0, so `Math.abs()` should be dropped (and you should avoid the divide by zero, e.g. adding an epsilon. At the start of a test this code could breaks or pollute the log).
Anyway:
* the name "Interval" IMO is misleading, that value should be a figure of Elo (gain/loss) per Elo unity of confidence
* that value theoretically span (-inf, +inf) so I don't like the use of the %
* I found strange to compute a relative figure for "normalized" value as  Elo

I ask to @vdbergh to review this PR

ppigazzini:
@Chess13234 PR applied on [DEV](https://dfts-2.pigazzini.it/).
I found very difficult to understand the second half of this information:
```
Elo  -1.62 [-5.90,2.80] (95%) Interval:-18.595%
```

Chess13234:
@ppigazzini the interval is negative for negative elo, its shows that the test is failing:
e.g.
-50% certainly losing, elo occupies 50% of interval
-10% slightly failing, elo occupies 10% of interval
+10% slightly winning, elo occupies 10% of interval.
+50% certainly winning , elo occupies 50% of interval.
Its lost of like LOS% but applies to interval confidence.

Chess13234:
Closed and added a cleaned up version here:
https://github.com/glinscott/fishtest/pull/495

ppigazzini:
@Chess13234 you are right about the division in javascript

PR #495, opened 2020-01-09
Add Elo/Bounds 

Chess13234:
Adds a display of Elo divided by bounds(Elo*100/(upper-lower)), which indicates how far the test is going towards resolution  (how much elo relative to confidence bounds its gaining).
(division by zero in JavaScript returns infinity).
Examples:
-50% certainly losing, elo occupies 50% of interval
-40% almost losing
-30% probably losing
-10% slightly failing, elo occupies 10% of interval
+10% slightly winning, elo occupies 10% of interval.
+30% likely winning,
+40% almost winning,
+50% certainly winning , elo occupies 50% of interval.
Its kind of like LOS% but applies to elo/interval confidence.


ppigazzini:
@Chess13234 DEV updated with this PR, now is more readable.
https://dfts-2.pigazzini.it/html/live_elo.html?5e0ee87ae97ea42ea89da9e1

I wait a couple of days before merging this, @vdbergh could suggest a more statistically sound measure.

vdbergh:
Hmm... I am sorry but I am not in favor of this change. 

I think it duplicates in a rather obscure (and a bit ugly) way information that is already available in a more standard way. 

- The LOS-value (a misnomer, for historical reasons, defined as 1-p here) gives an indication for how likely a patch is to be an Elo gain.  

- Similar information can be obtained from the green/red division in the Elo dial (the one on the right).

Both the p-value, the Elo confidence interval and the Elo point estimate represent standard concepts in statistics. The information that the PR would like to add does not.

Chess13234:
Closing. I think most people wouldn't need this, but ELO dial isn't that detailed(vs percentage display).

PR #493, opened 2020-01-08
Remove statistics updates

Chess13234:
Remove the display of "Update user statistics" in action log.
They just waste space and provide nothing useful.

PR #496, opened 2020-01-09
Hide update stat events by default

tomtor:
See #493


tomtor:
@ppigazzini We might add in the future add other events, so I would like to keep them in the database but hide them from the default view.

ppigazzini:
Thank you @tomtor

PR #497, opened 2020-01-10
Add explanations to the SPRT calculator.

vdbergh:
The new version of the SPRT calculator is here. 

Feel free to comment.

### DIFF
--- a/fishtest/fishtest/static/html/SPRTcalculator.html
+++ b/fishtest/fishtest/static/html/SPRTcalculator.html
@@ -64,7 +64,6 @@
                 stroke: black;
                 stroke-width: 2px;
             }
-
         </style>
     </head>
     <body>
@@ -72,11 +71,11 @@
             <h1 class=page-header>Chess SPRT Calculator</h1>
             <form id=parameters class=form-inline>
                 <div class=form-group>
-                    <label for=elo-0>Elo 0</label>
+                    <label for=elo-0>Elo0</label>
                     <input id=elo-0 name=elo-0 value=-1 class='form-control number'>
                 </div>
                 <div class=form-group>
-                    <label for=elo-1>Elo 1</label>
+                    <label for=elo-1>Elo1</label>
                     <input id=elo-1 name=elo-1 value=3 class='form-control number'>
                 </div>
                 <div class=form-group>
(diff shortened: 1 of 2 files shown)


vdbergh:
PS. This PR implements some requests by @snicolet .

ppigazzini:
DEV updated

snicolet:
You should add a `<br><hr>` before the start of the explanations, otherwise on some browsers the first paragraph is cut, see http://cassio.free.fr/divers/screenshot.png

Also concerning the first paragraph: 

- it is strange to start a sentence with a variable name, it could read for instance "The parameters Elo0 and Elo1 represent the bounds [...]"

- I would rename the variables everywhere in the document
```
   Elo 0      --->      Elo0
   Elo 1      --->      Elo1
```

More feedback later, I have to run for now. Thanks for the page :-)

vdbergh:
@snicolet I posted a new version. I put in <br><hr> . Can you check if it fixes the rendering error?

snicolet:
> I put in <br><hr> . Can you check if it fixes the rendering error?

No it does not, it seems that the `<br>` and `<hr>` has been styled in CSS.
However this is only a cosmetic problem.

I propose that we commit the current state of the PR and we can improve/simplify the text and the graphical layout later in other PR, if necessary.

vdbergh:
@snicolet Ok. I must say I am a bit puzzled by this issue. The html is trivial and the page renders as intended in Firefox and Chrome.

As you say it is probably the CSS styling applied by bootstrap. But right now I don't have much time to investigate.

vdbergh:
Actually now it seems that it may not actually be a bug.  It has to do with the floating behavior of the graphs. I looked on my cell phone and there the behavior is sensible. The text floats to the right of the graphs, depending on the availability of screen real estate. However the non-breakability of the URLs in the second bullet, forces most of the text to be below the graphs anyway.

Later I will try to make the graphs non-floating.

vdbergh:
I think making only the first graph float left has fixed the issue. The second graph will be either to the right of below the second graph, but the text (contained in an UL environment) will always be below the second graph. So it is should always be ok.

ppigazzini:
I wait the test by @snicolet to merge this.

snicolet:
It now works fine in Safari!

ppigazzini:
PROD updated. Thanks to @vdbergh for the patch and to @vondele @snicolet for the test/review :)

vdbergh:
I guess we should now add Henri Wiechers to the list of authors also.

ppigazzini:
@vdbergh correct. The list must be update with all the contributors:
```
$ git shortlog -sn
   728  Gary Linscott
   301  Marco Costalba
   134  Tom Vijlbrief
    53  ppigazzini
    24  Kamyar Kaviani
    15  Joost VandeVondele
    15  xoto10
    11  ianfab
     9  Michel Van den Bergh
     9  Werner Fenchel
     8  Stefano Cardanobile
     8  lucasart
     6  Jean-Francois Romang
     4  Vince Negri
     3  David Zar
     2  Chess13234
     2  Joona Kiiski
     2  Lucas Braesch
     2  Stéphane Nicolet
     2  mcostalba
     2  noobpwnftw
     1  FieryDragonLord
     1  Henri Wiechers
     1  Michael An
     1  Michał Robaczyk
     1  Noé Costa
     1  Stefano80
     1  Your Name
     1  fishtest
     1  hxim
     1  joergoster
     1  theo77186
```

ppigazzini:
@vdbergh opened the #500, Henri Wiechers was already an authors (but not credited).

vdbergh:
The SPRT calculator uses the d3 framework which is a javascript layer above svg (in other words the graphs are essentially created by hand).

As an experiment I made a version using Google Charts. 

It is functionally more or less equivalent and the javascript code is much shorter and also much easier to understand.

ppigazzini:
Feel free to open a new PR :)

ppigazzini:
@vdbergh minor glitch of the Google Charts version: after inserting an invalid value the charts disappear.

snicolet:
a) The master version seems a little bit smoother, maybe because it works even when the mouse is not touching the curves -- in fact I always use it by moving the mouse horizontally anywhere in the grey areas, never by following the curves.

b) since you are at it, can you make the Elo steps 0.05 or 0.02 instead of 0.1 ? this way it would feel even smoother

snicolet:
Btw, the link in the left menu (in the MISC section) could read "Simulator", as it would be fit on one line.

vdbergh:
@ppigazzini This was intentional, but not a good idea in retrospect. I was trying to think what to do if the user supplies illegal parameters via the URL (then there are no previous parameters to fall back on). But I guess the solution is easy. Just draw the graph with the default parameters. 

@snicolet Google Charts do not generate mouseover events outside the data points. So this behaviour of master cannot be duplicated. I guess it would be a choice between more easily maintainable code and more flexibility.

vdbergh:
@snicolet I experimented with setting the Elo increment to 0.01. While then the behavior is a  bit smoother (to my surprise), it also becomes harder to position the cursor accurately. So I left it as is, but made the increment easy to change in the source.

vdbergh:
It appears master's behaviour can be duplicated in a not too hacky way by attaching eventlisterners directly to the enclosing div.

ppigazzini:
> Btw, the link in the left menu (in the MISC section) could read "Simulator", as it would be fit on one line

"SPRT simul" ?

vdbergh:
But it is not a simulator... Perhaps SPRT calc?

ppigazzini:
I agree for "SPRT calc"

vdbergh:
Here is a new version that more or less emulates the behavior of master 

Occasionally there is a small dead zone next to the curve which has to do with GC eating events.

Issue #498, opened 2020-01-10
Fix the geo location resolve by using another API endpoint


NoTeefy:
I was wondering why some subnets/ip addresses don't receive proper flags in the testing framework. Turns out the API you guys and gals use is considered deprecated and no longer operates. 
Here's an example GET request to that API endpoint:
```json
curl -L http://geoip.nekudo.com/api/

{
  "0": "#################################################################################################################################",
  "1": "#                                                                                                                               #",
  "2": "# IMPORTANT - PLEASE UPDATE YOUR API ENDPOINT                                                                                   #",
  "3": "#                                                                                                                               #",
  "4": "# This API endpoint is deprecated and has now been shut down. To keep using the Free Geolocation API, please update             #",
  "5": "# your integration to use the new ipapi API endpoint, designed as a simple drop-in replacement.                                 #",
  "6": "# You will be required to create an account at https://ipapi.com and obtain an API access key.                                  #",
  "7": "#                                                                                                                               #",
  "8": "# For more information on how to upgrade please visit our Github page at: https://github.com/nekudocom/shiny_geoip#readme       #",
  "9": "#                                                                                                                               #",
  "a": "#################################################################################################################################",
  "country": ""
}
```

https://github.com/glinscott/fishtest/blob/481591d6aa28fed433c6d773b6c10f0d3150c266/fishtest/fishtest/api.py#L8 would be the place where a logic rehaul is needed.

I know that it's just a style fix and nothing major but I think it would look better if all runners would be represented by their locations. I could provide an own patch for it if nobody feels interested in it, but I am not too familiar with your framework yet and want to respect all of your coding/contribution guidelines before providing patches.

ppigazzini:
@NoTeefy feel free to open a PR, every contribution is welcomed :)

NoTeefy:
Gonna give it a shot then. I need to know certain things though: Is there any service mail I could register such an account on? Like a technical user for that token we need with the new API?
And how would you guys provide such a token. Are you using environment variables/files for your deploys on the official Fishtest server?
I wouldn't put a token into public code 😄

ppigazzini:
@NoTeefy we use 2 files with secrets (recaptha, github) and a couple of variables (aws s3). You can develop and test locally the patch using your email and then providing the instruction when opening the PR.

[...]

PR #499, opened 2020-01-10
changed geo provider for country lookups

NoTeefy:
As stated in #498 the old API endpoint has been declared as outdated and thus been shut down about 2 years ago.
We can use a similar, free provider (it only limits at around 15'000 requests/hour, which we never exceed because of caching) to circumvent that problem.
Please share your thouights about using a third-party service instead of shipping the fishtest server with a prebuilt IP database file. We don't have control of what they want to do with the data we sent them (same goes for the old provider). Both of them have their source-codes publicly available but you never know how they implemented it on their side...

Feel free to change/adjust code if I misinterpreted your contribution guidelines :)

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -13,11 +13,11 @@ def get_flag(request):
 
   # Get country flag ip
   try:
-    FLAG_HOST = 'http://geoip.nekudo.com/api/'
+    FLAG_HOST = 'https://freegeoip.app/json/'
 
     r = requests.get(FLAG_HOST + request.remote_addr, timeout=1.0)
     if r.status_code == 200:
-      country_code = r.json()['country']['code']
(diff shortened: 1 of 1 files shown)


ppigazzini:
@NoTeefy DEV server updated with the PR, my workers got the flag, feel free to join a worker to test the patch 
https://dfts-2.pigazzini.it

NoTeefy:
@ppigazzini looks fine 👌

ppigazzini:
@NoTeefy I applied the patch by hand on the PROD server to check if the no-flag workers will take a flag in the next hours.

ppigazzini:
We call only once the external service for a no cached worker IP address; this PR doesn't change the privacy level, so I'm fine to merge it.

ppigazzini:
PROD updated, thank you @NoTeefy :)

PR #502, opened 2020-01-12
Lower p-value to 0.001

ppigazzini:
Lower the p-value to limit the numner of autopurge with the high throughput of the framework.


ppigazzini:
PROD updated :)

Issue #501, opened 2020-01-12
The strange case of the half-finished Outpost2 test


snicolet:
There is something strange in the test "Outpost2 (take 2)" by @SFIsGOD, as it appeared finished and passed (green) in some pages, and not finished (at LLR 2.03) in some other pages.

First, it is at LLR 2.03 in the pending list of the main page, and also in the list of tests for user SFisGD (I put a screenshot of the display under each link, to get a copy of the strange state in case the database autocleans somehow):
https://tests.stockfishchess.org/tests/user/SFisGOD
http://cassio.free.fr/divers/shot3.png

However, when clicking on the test it opens green at LLR 2.93:
https://tests.stockfishchess.org/tests/view/5e1b4ecdd12216a2857e636b
http://cassio.free.fr/divers/shot2.png

This is the raw statistics page so that interested people can sneak in the numbers:
https://tests.stockfishchess.org/tests/stats/5e1b4ecdd12216a2857e636b
http://cassio.free.fr/divers/shot1.png

An auto-purge problem?

ppigazzini:
@snicolet @SFisGOD some Idx (17, 29-32) are missing, it should be the result of autopurge.

ppigazzini:
Should be fixed now.

PR #504, opened 2020-01-13
Write database after purge

tomtor:
See #501


ppigazzini:
I started a test on DEV raising the p-value to trigger the autopurge.

ps: I reinstalled DEV to test the python3 server setup, new domain name https://dfts-0.pigazzini.it

ppigazzini:
PROD updated, thank you @tomtor

snicolet:
Thanks! :-)

PR #505, opened 2020-01-13
Rewrite the SPRT Calculator using Google Charts.

vdbergh:
Google Charts are already used at other places in Fishtest and people are familiar with it.

The latest version can be found here.


vdbergh:
I discovered the cause of the dead zone and the flicker in the neighborhood of the curve. Will update.

vdbergh:
Ok updated.

ppigazzini:
@vdbergh https://dfts-0.pigazzini.it updated with this PR, it's looks good to me.
I will merge it as soon as you will shorten this:
https://github.com/glinscott/fishtest/blob/591a401dc6a144d7c4dd4324adf3e6909a9030fd/fishtest/fishtest/templates/base.mak#L65

vdbergh:
Done :)

ppigazzini:

PR #507, opened 2020-01-14
Various internal/external cleanups for the SPRT calculator.

vdbergh:
As usual the new version is here.

**Changes**

There are now minor and major vertical grid lines.

The Elo increment used for graph drawing is now a bit adaptive. This gives smoother tooltip behaviour for small elo differences.

All mouse events are now captured by a div specially created for that purpose. The GC default event handlers are thus completely disabled. This removes some prior artifacts, e.g. when the user clicked on the graphs.

Events are now disabled until the charts are fully drawn.

In general event handling is greatly simplified internally.

Some html cleanup. In particular the name attribute of the input fields is not used and thus was removed. Most inline styling was removed. A spurious ```</div>``` was removed.

### DIFF
--- a/fishtest/fishtest/static/html/SPRTcalculator.html
+++ b/fishtest/fishtest/static/html/SPRTcalculator.html
@@ -10,6 +10,11 @@
               text-align: right;
               margin-right: 0.5em;
             }
+            .chart-div {
+              display:inline-block;
+              width:560px;
+              height:500px;
+            }
         </style>
     </head>
     <body>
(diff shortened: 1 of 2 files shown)


ppigazzini:
@vdbergh just noted now:
* if "draw ratio" > 1 the message is "The draw ratio and the RMS bias are not compatible." instead that "The draw ratio must be strictly between 0.0 and 1.0."
* if "Elo0" < -10 the message is "Elo values cannot be larger than 10." perhaps better to use "Elo values must be between -10.0 and 10.0." (math correct, positive form, and similar to draw ratio message)

vdbergh:
I think it is fixed now. The first issue was a genuine bug (a forgotten else statement).

vdbergh:
Oops not yet it seems. I did something wrong with git.

ppigazzini:
@vdbergh on [DEV](https://dfts-0.pigazzini.it) looks good to me, I wait your check.

vdbergh:
The PR should be ok.

ppigazzini:
PROD updated, thanks @vdbergh :)

PR #508, opened 2020-01-15
Fix bad formatting for multiple modifications

tomtor:
```
modify: priority changed from -1 to 0,throughput changed from 100 to 50
```

should have a space after ','


ppigazzini:
Tested OK on DEV, PROD updated. Hawkeye @tomtor :)

PR #513, opened 2020-01-18
Return 404 for non existant run view

tomtor:


ppigazzini:
@tomtor it works with a proper test-code length e.g.
https://dfts-0.pigazzini.it/tests/view/5322443a109637b31634b345
it doesn't work wit a random test-code length e.g.
https://dfts-0.pigazzini.it/tests/view/5322443a109637b31634b34

tomtor:
@ppigazzini I discovered that too at the same time as you did 😉

Should be fixed now

ppigazzini:
PROD updated, thank you @tomtor !

PR #517, opened 2020-01-22
A few minor cleanups in the SPRT calculator.

vdbergh:
Fixes an html error (apparently inconsequential) and some warnings found by https://validator.w3.org/ .

Fixes an inconsistency. I was using both float:left and display:inline-block for the char divs. They serve the same purpose. So now I only use display:inline-block which is the more modern solution.

Fixes a small, probably untriggerable, race condition in event handling. 

As usual the latest version is here 

Externally it is indistinguishable from the previous version (or should be).

Future plans: when I find the time I will add a curve with the 95% quantile...


ppigazzini:

PR #519, opened 2020-01-22
Various stats tweaks.

vdbergh:
Cleanups of SPRT handling in stat_util.py.

Remove a bit of code duplication.

Use """ """ style doc strings.

Document the LLR approximations we use in the raw_stats page.

### DIFF
--- a/fishtest/fishtest/stats/LLRcalc.py
+++ b/fishtest/fishtest/stats/LLRcalc.py
@@ -16,7 +16,7 @@ that the ai are strictly ascending, a1<s<aN and p1>0, pN>0.
 The theory behind this function can be found in the online
 document
 
-http://hardy.uhasselt.be/Toga/computeLLR.pdf
+http://hardy.uhasselt.be/Fishtest/support_MLE_multinomial.pdf
 
 (see Proposition 1.1).
 
@@ -76,7 +76,7 @@ for s=s1 versus s=s0 where pdf is an empirical distribution and
 s is the expectation value of the true distribution.
 pdf is a list of pairs (value,probability). See
 
-http://hardy.uhasselt.be/Toga/computeLLR.pdf
+http://hardy.uhasselt.be/Fishtest/support_MLE_multinomial.pdf
 """
     r0,r1=[sum([prob*(value-s)**2 for value,prob in pdf]) for s in (s0,s1)]
     return 1/2*math.log(r0/r1)
@@ -88,7 +88,7 @@ for s=s1 versus s=s0 where pdf is an empirical distribution and
 s is the expectation value of the true distribution.
 pdf is a list of pairs (value,probability). See
 
-http://hardy.uhasselt.be/Toga/GSPRT_approximation.pdf
+http://hardy.uhasselt.be/Fishtest/GSPRT_approximation.pdf
 """
     s,var=stats(pdf)
     return (s1-s0)*(2*s-s0-s1)/var/2.0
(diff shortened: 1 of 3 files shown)


vdbergh:
I did some benchmarking. On Python3 on my laptop the LLR approximation we use is 20 times faster than the exact one. On the other hand, one call to compute the exact LLR takes about 80 microseconds (it requires numerically solving a secular equation). I wonder how many SPRT updates there are per second on fishtest... 

```
import time
from LLRcalc import LLR, LLR_alt, L_,results_to_pdf

elo0=-1
elo1=3
results=[726, 5007, 10065, 5021, 734]

s0,s1=[L_(elo) for elo in (elo0,elo1)]
N,pdf=results_to_pdf(results)

E=100000
t0=time.time()
for i in range(0,E):
    X=N*LLR(pdf,s0,s1)
t1=time.time()
A=(t1-t0)/E
t2=time.time()
for i in range(0,E):
    X=N*LLR_alt(pdf,s0,s1)
t3=time.time()
B=(t3-t2)/E
print(A,B,A/B)

```

vdbergh:
Using scipy's brentq is a bit faster but not much. 60 microseconds.

vdbergh:
Note that there is no clear theoretical reason why the "exact" LLR would be the best (the only thing that is known is that it behaves correctly asymptotically). With my testing resolution I could not establish any difference in behavior between the approximations. I would have to write a much faster simulator (C++ multi-threaded) to be able to detect any difference I guess.

ppigazzini:
@vdbergh I my work prefer to use standard library and simplify the code base. IMO a new implementation makes sense only if the standard library is very slow or lacks precision.

vdbergh:
@ppigazzini We can discuss this, but this is not what I was addressing in my comment. I was just wondering if it would be feasible to replace the LLR approximation I use with the exact value (which is about 15 times slower to compute, even if using the scipy brentq).

vdbergh:
To clarify: **the LLR computation itself is not part of scipy.** But it needs a numerical solver and brentq is the best.

ppigazzini:
I was talking about using scipy brentq and dropping the implementation used now.
About the impact of the slower LLR exact implementation, after some tests on DEV I can update by hand the files on PROD and test for sensible slowdown. The bottleneck on fishtest is MongoDB not the python code.

vdbergh:
I will update the PR so that the exact LLR is always used.

But let's postpone changing brentq (and using the normal cdf from scipy) to a different PR. The scipy brentq has a slightly different interface so the assertions need to be adapted. I would prefer to test this first.

vdbergh:
I updated to PR to always use the exact LLR (in a separate commit so that it is easier to revert).

ppigazzini:
[DEV](https://dfts-0.pigazzini.it) updated after https://github.com/glinscott/fishtest/pull/519/commits/30705eab773cee607bc506db5e1bd74f37f4f4eb

vdbergh:
I guess it should not be a problem. ```1868 games/minute```. 

```1868/60*80e-6=0.0025=0.25%```

A negligible increase in load.

ppigazzini:
@vdbergh PROD update by hand copying the 3 files, CPU load unchanged (and very very low).

If I'm not wrong "secular equation" should be the "characteristic equation", in this case I suggest to use the latter term. At the moment the references doc "computeLLR.pdf" is unreachable and I'm not able to check by myself.

vdbergh:
"Secular equation" has two meanings: a general one and a specific one. I am using it it in the specific sense of 
https://en.wikipedia.org/wiki/Divide-and-conquer_eigenvalue_algorithm (see the section conquer).

However I agree it is better that I change the wording as it is confusing.

vdbergh:
Updated and squashed PR to a single commit.

vdbergh:
Oops. Note rebased on current master.

vdbergh:
Ok now.

ppigazzini:

ps: interesting algorithm :)

Issue #521, opened 2020-01-23
Import issues


vdbergh:
I think that a package in python generally should be self contained (I may be wrong, I have not checked other projects). In any case this used to be the case with the stats package.

But now for example if I do

``` fishtest@pt:~/fishtest/fishtest/fishtest/stats$ python stat_util.py ```

I get

```Traceback (most recent call last):
  File "stat_util.py", line 5, in <module>
    from fishtest.stats import LLRcalc
  File "/home/fishtest/fishtest/fishtest/fishtest/__init__.py", line 8, in <module>
    from fishtest.rundb import RunDb
  File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 17, in <module>
    from fishtest.views import format_results
  File "/home/fishtest/fishtest/fishtest/fishtest/views.py", line 12, in <module>
    import html
ImportError: No module named html
```

This is quite baffling to me since all the dependencies for properly executing the command are in the stats package itself. Is this a bug? Otherwise what is the rationale for the new behavior?

tomtor:
@vdberg You need python >= 3.2 for the html package. Nothing has changed otherwise, except that views.py has this new dependency.

__init__.py is read, because it is in your PYTHONPATH 
or `from fishtest.stats ...` would fail

vdbergh:
But I do not understand why the fishtest.stats... imports are necessary inside the stats package.  This wasn't like this before.

vdbergh:
BTW I tried to run pserve under python3 but it did not work. I got the error

```Traceback (most recent call last):
  File "/home/fishtest/.local/bin/pserve", line 11, in <module>
    sys.exit(main())
  File "/home/fishtest/.local/lib/python3.5/site-packages/pyramid/scripts/pserve.py", line 34, in main
    return command.run()
  File "/home/fishtest/.local/lib/python3.5/site-packages/pyramid/scripts/pserve.py", line 275, in run
    app = loader.get_wsgi_app(app_name, config_vars)
  File "/home/fishtest/.local/lib/python3.5/site-packages/plaster_pastedeploy/__init__.py", line 129, in get_wsgi_app
    global_conf=defaults,
  File "/home/fishtest/.local/lib/python3.5/site-packages/paste/deploy/loadwsgi.py", line 253, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/home/fishtest/.local/lib/python3.5/site-packages/paste/deploy/loadwsgi.py", line 277, in loadobj
[...]

PR #522, opened 2020-01-24
Change the stats package to use relative imports.

vdbergh:
Adresses #521.

Currently when doing ```import stats```  the whole fishtest framework is imported (causing a noticeable delay) even though ```stats``` does not depend on this. I feel this is not right. I feel that ```stats``` should be a self contained relocatable package (except for using system libraries).


vdbergh:
Well it seems this is a controversial issue. Then let's things be as they were. Closing.

Issue #524, opened 2020-01-25
python3.2


d3vv:
```
 python3.2 fishtest/worker/worker.py
Traceback (most recent call last):
  File "fishtest/worker/worker.py", line 12, in <module>
    import requests
  File "/home/dew/fishtest/worker/requests/__init__.py", line 64, in <module>
    from .models import Request, Response, PreparedRequest
  File "/home/dew/fishtest/worker/requests/models.py", line 856
    http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
                                                      ^
SyntaxError: invalid syntax
```

d3vv:
```
python3.1 fishtest/worker/worker.py
Traceback (most recent call last):
  File "fishtest/worker/worker.py", line 12, in <module>
    import requests
  File "/home/dew/fishtest/worker/requests/__init__.py", line 64, in <module>
    from .models import Request, Response, PreparedRequest
  File "/home/dew/fishtest/worker/requests/models.py", line 856
    http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
                                                      ^
SyntaxError: invalid syntax
```

ppigazzini:
@d3vv I suppose that you are using linux, please try to install the `requests` package (e.g. `${VENV}/bin/pip install requests`) and rename the folder `fishtest/worker/requests`.

I fear that python < 3.4 could have some incompatibility, so check if your servers are able to run:
- docker (or a linux container technology like lxd/lxc): this would be the best solution
- a different python distribution, e.g. :
  - pypy: https://pypy.org/download.html
  - miniconda https://docs.conda.io/en/latest/miniconda.html

PR #527, opened 2020-01-26
Set book default to noob_3moves.pgn

snicolet:
Old versions of cutechess-cli have problems with opening books in EPD format, so switch to book in PGN format as default in fishtest.

See: https://groups.google.com/forum/?fromgroups=#!topic/fishcooking/hcGWL4mXFGw

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -146,7 +146,7 @@ Cowardice,150,0,200,10,0.0020"""})['raw_params']}</textarea>
   <div class="control-group">
     <label class="control-label">Book:</label>
     <div class="controls">
-      <input name="book" value="${args.get('book', 'noob_3moves.epd')}">
+      <input name="book" value="${args.get('book', 'noob_3moves.pgn')}">
     </div>
   </div>
   <div class="control-group">


snicolet:
I have uploaded all the (zipped) PGN versions of the books in the book directory https://github.com/official-stockfish/books , so now it seems that we can switch to noob_3moves.pgn as default book. This may help workers with old cutechess-cli versions.

snicolet:
Also in the https://github.com/official-stockfish/books directory is the utility to convert EPD files to PGN.

ppigazzini:
@snicolet the old cutechess-cli must be dropped. And in the next weeks we will should update cutechess-cli binaries with this https://github.com/cutechess/cutechess/pull/551

vondele:
I also think this is a bit of a hack. Eventually we must be able to use an updated cutechess-cli.

ppigazzini:
At the moment there are 19 dew workers running fine in fishtest, so IMO the ["Farewell"](https://groups.google.com/forum/#!topic/fishcooking/WDwbBcrRyO4) post was not about cutechess-cli but python3, view #524 .
I hope that @d3vv will clarify a bit the tech details of the roadblock. It's not easy to try to help with so little information.

PR #528, opened 2020-01-28
Fix the worker update process

ppigazzini:
Update only the worker files so that the update process will work
also with a custom worker folder name.
Enforce the download of the lastest version of cutechess-cli, books etc.
renaming the "testing" folder with a timestamp. Keep the lastest 3 backups
of the "testing" folder to save eventual custom user files.
Thanks to @vondele for some ideas and code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ worker/custom_make.txt
 worker/env
 worker/fish.exit
 worker/fishtest.cfg
+worker/_testing_*
 worker/testing
--- a/worker/test_worker.py
+++ b/worker/test_worker.py
@@ -48,7 +48,7 @@ class workerTest(unittest.TestCase):
   def test_updater(self):
     file_list = updater.update(restart=False, test=True)
     print(file_list)
-    self.assertTrue('worker' in file_list)
+    self.assertTrue('worker.py' in file_list)
 
 if __name__ == "__main__":
   unittest.main()
--- a/worker/updater.py
+++ b/worker/updater.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import datetime
+import glob
 import os
 import shutil
 import sys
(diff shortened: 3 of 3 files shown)


ppigazzini:
Tested OK w/ Windows, WSL, Linux workers.
@tomtor @vondele please review the code.

vondele:
at first glance this looks good to me (but I would leave the final OK to you or @tomtor, I'm not that used to reading Python).

However, I would actually not copy the engines. Maybe the new worker changes the process to build them (e.g. optimization, native, etc), and in that case the binaries that are copied don't get recompiled, whereas they should. Not copying them might introduce a small overhead rebuilding them, but a worker update is infrequent enough not to care.

ppigazzini:
@vondele good point. Now I delete the old engines. Another options it's to keep the engines in the backup folders and simplify a bit the code.

vondele:
I'd go for simplify, and deleting as little as possible. Who knows what a user stores in stockfish_* ;-)

ppigazzini:
> Who knows what a user stores in stockfish_* ;-)

The safest way to save something! :D

https://github.com/glinscott/fishtest/blob/8546c4988fbc417e8b3fb064e9f608ad8953e6cd/worker/games.py#L455-L464

vondele:
ah, right. so yes, in light of this we could just delete as well.

ppigazzini:
OT @vondele I'm wondering why SPSA tunes aren't run with Ultra STC to minimize/nullify the search effect and speedup the tune. This SPSA tune seems to behave the same way on PROD @ 20.0+0.2 and on DEV @ 1+0.01
https://tests.stockfishchess.org/tests/view/5e2dade6ab2d69d58394fb5e
https://dfts-0.pigazzini.it/tests/view/5e306ac276f6b3d227dee9d5

vondele:
I've actually been playing with this recently, but unfortunately, TC effects are clearly visible, even for evaluation terms. For reference, I tuned a couple of eval parameters locally at UltraSTC (0.5+0.005), which could be checked to gain roughly +4 Elo at that TC, but lost >10 at STC:
```
TC                                score               LOS                     Elo
 0.5+0.005 (124'000 games) -> (0.5063870967741936, 0.9999993335989763)     ->  +4 Elo (fishtest: 1.54 [-1.45,4.25], http://tests.stockfishchess.org/tests/view/5e0eee7be97ea42ea89da9e7)
 2.5+0.025 (124'000 games) -> (0.48921370967741934, 3.351508116488963e-26) ->  -8 Elo
10.0+0.100 ( 31'000 games) -> (0.47469354838709676, 4.233848506761162e-49) -> -18 Elo (fishtest: -25.92 [-38.26,-13.95], http://tests.stockfishchess.org/tests/view/5e0e0f6887585b1706b683cd)
```

ppigazzini:
For the fine tuning step it makes sense the TC dependency, but for the first tune when some parameters can have a big change of value IMO the USTC could be useful.

ppigazzini:
@vondele same behaviour. 
20+0.2
https://tests.stockfishchess.org/tests/view/5e2dade6ab2d69d58394fb5e

1+0.01
http://dfts-0.pigazzini.it/tests/view/5e30c9548122a318a9332b18

0.2+0.01
http://dfts-0.pigazzini.it/tests/view/5e30b34f8122a318a9332b13

IMO the tests show clearly that only factor(8) is tunable, either for factors sensibility and/or starting values.

tomtor:
Looks good to me

ppigazzini:
OT
@vondele I think that some/many SPSA ran at LTC and with wrong coefficients.
Perhaps the user should use an Ultra STC to test the SPSA coefficients and to initialize the values, then fine tuning a STC/LTC.

20+0.2 (IMO wrong SPSA coefficients, all factors but one are static)
https://tests.stockfishchess.org/tests/view/5e2dade6ab2d69d58394fb5e

0.2+0.01 (IMO right SPSA coefficients, 2 runs to check the result):
https://dfts-0.pigazzini.it/tests/view/5e3193d48122a318a9332b3b
https://dfts-0.pigazzini.it/tests/view/5e32e1308122a318a9332b47

vondele:
@ppigazzini that's certainly a strategy... The SPSA coefficients, are they:
SPSA A:
SPSA Gamma:
SPSA Alpha:
Do you have guidelines on how to set them ? I've always used defaults.

ppigazzini:
@vondele there is a section on the wiki that states "A good tune run exhibits a significant change in the tuned values while minimizing random change" and this is true for any optimization algorithm
https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#spsa-tests

This means:
- find the right SPSA coefficients. I started 3 SPSA with different values, it's easy to view when the behavior is wrong:
  - no variation between iteration
  - too much variation between iteration
- start N SPSA with the same coefficients and starting point
- keep a factor only if converge at the same ending point for the N SPSA

Keep in mind that all the optimization algorithms are designed to refine an already good starting point locate in the concavity of the local minimum.
But that algorithms typically are used with a bad starting point, so the coefficient at the start must be set at a big value in order to explore several point of the function and find a better local minimum.

One of these days I will check the SPSA implementation in fishtest.

ppigazzini:
@vondele I updated https://github.com/glinscott/fishtest/pull/528#issuecomment-580485762 with other 2 shorter USTC SPSA runs. The final values are fairly consistent (factors 6 and 8 = 30; other factors = 60) also with a crazy rate of time losses. IMO the USTC SPSA concept (used to find good SPSA coefficients and good starting point values for the SPSA@LTC) should be tested by some developers SPSA addicted :)

vondele:
@ppigazzini feel free to advertise :-)

ppigazzini:
@vondele I'm nearly finished the tests to show some evidence of the proposed process, then I will advertise the finding on the forum. By the way I didn't changed "A", "Alpha", "Gamma" (ie the algorithm coefficients), but "ck", "rk", the variables coefficients. With N variables you should find 2N good "ck_n", "rk_n" in order to compute stochastic gradients that make sense for any variable.

PR #529, opened 2020-01-28
Replace homebrew by scipy functions.

vdbergh:
As the subjects says.


vdbergh:
Sorry. The raw statistics pages gives an exception now (I only ran the unit tests in stat_util.py and they worked fine). I'll check it out.

vdbergh:
Should be fixed now. I changed phi to Phi as it is more standard, but I forgot to do it in one of the templates.

ppigazzini:
@vdbergh DEV updated, it's fixed :)

ppigazzini:
PROD updated, thank you @vdbergh !

PR #530, opened 2020-01-30
Reduce queueing when MongoDb is under load and flush buffers on signals

tomtor:
Flush buffers on signals and

Reduce log warnings:

Jan 30 17:03:07 tests pserve[31906]: 2020-01-30 17:03:07,583 WARNI
[waitress.queue][MainThread] Task queue depth is 10
Jan 30 17:03:07 tests pserve[31906]: 2020-01-30 17:03:07,612 WARNI
[waitress.queue][MainThread] Task queue depth is 10
Jan 30 17:03:07 tests pserve[31906]: 2020-01-30 17:03:07,660 WARNI
[waitress.queue][MainThread] Task queue depth is 1
Jan 30 17:03:07 tests pserve[31906]: 2020-01-30 17:03:07,660 WARNI
[waitress.queue][MainThread] Task queue depth is 2
Jan 30 17:03:07 tests pserve[31906]: 2020-01-30 17:03:07,677 WARNI
[waitress.queue][MainThread] Task queue depth is 2

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -1,3 +1,5 @@
+import sys
+import signal
 import copy
 import os
 import random
@@ -18,6 +20,7 @@ from fishtest.views import format_results
 
 import fishtest.stats.stat_util
 
+last_rundb = None
 
 class RunDb:
   def __init__(self, db_name='fishtest_new'):
@@ -33,6 +36,9 @@ class RunDb:
     self.deltas = self.db['deltas']
 
     self.chunk_size = 250
+
+    global last_rundb
+    last_rundb = self
 
   def build_indices(self):
     self.runs.ensure_index([('finished', ASCENDING),
@@ -150,6 +156,16 @@ class RunDb:
 
   timer = None
 
+  # handle termination
+  def exit_run(signum, frame):
+      global last_rundb
+      if last_rundb:
+        last_rundb.flush_all()
+      sys.exit(0)
+
+  signal.signal(signal.SIGINT, exit_run)
+  signal.signal(signal.SIGTERM, exit_run)
+
   def get_run(self, r_id):
     with self.run_cache_lock:
       r_id = str(r_id)
@@ -190,9 +206,20 @@ class RunDb:
                               'ftime': ftime, 'run': run}
 
   def stop(self):
+    self.flush_all()
     with self.run_cache_lock:
       self.timer = None
     time.sleep(1.1)
+
+  def flush_all(self):
+    print("flush")
+    # Note that we do not grab locks because this method is
+    # called from a signal handler and grabbing locks might deadlock
+    for r_id in list(self.run_cache):
+      if self.run_cache[r_id]['dirty']:
+        self.runs.save(self.run_cache[r_id]['run'])
+        print(".", end='')
+    print("done")
 
   def flush_buffers(self):
     if self.timer is None:
@@ -214,7 +241,7 @@ class RunDb:
       self.run_cache[oldest]['dirty'] = False
       self.run_cache[oldest]['ftime'] = time.time()
       self.run_cache_lock.release()  # Release the lock while writing
-      print("SYNC")
+      # print("SYNC")
       with self.run_cache_write_lock:
         self.runs.save(self.run_cache[oldest]['run'])
       # start the timer when writing is done
@@ -491,7 +518,7 @@ class RunDb:
       task['pending'] = False  # Make pending False before making active false
                                # to prevent race in request_task
       task['active'] = False
-      flush = True
+      # flush = True
 
     update_time = datetime.utcnow()
     task['last_updated'] = update_time
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -167,11 +167,14 @@ class RunDb:
       except:
         return None
 
+  def start_timer(self):
+    self.timer = threading.Timer(1.0, self.flush_buffers)
+    self.timer.start()
+
   def buffer(self, run, flush):
     with self.run_cache_lock:
       if self.timer is None:
-        self.timer = threading.Timer(1.0, self.flush_buffers)
-        self.timer.start()
+        self.start_timer()
       r_id = str(run['_id'])
       if flush:
         self.run_cache[r_id] = {'dirty': False, 'rtime': time.time(),
@@ -192,28 +195,34 @@ class RunDb:
     time.sleep(1.1)
 
   def flush_buffers(self):
-    with self.run_cache_lock:
-      if self.timer is None:
-        return
-      now = time.time()
-      old = now + 1
-      oldest = None
-      for r_id in list(self.run_cache):
-        if not self.run_cache[r_id]['dirty']:
-          if self.run_cache[r_id]['rtime'] < now - 60:
-            del self.run_cache[r_id]
-        elif self.run_cache[r_id]['ftime'] < old:
-          old = self.run_cache[r_id]['ftime']
-          oldest = r_id
-      if oldest is not None:
-        if int(now) % 60 == 0:
-          self.scavenge(self.run_cache[oldest]['run'])
-        with self.run_cache_write_lock:
-          self.runs.save(self.run_cache[oldest]['run'])
-        self.run_cache[oldest]['dirty'] = False
-        self.run_cache[oldest]['ftime'] = time.time()
-      self.timer = threading.Timer(1.0, self.flush_buffers)
-      self.timer.start()
+    if self.timer is None:
+      return
+    self.run_cache_lock.acquire()
+    now = time.time()
+    old = now + 1
+    oldest = None
+    for r_id in list(self.run_cache):
+      if not self.run_cache[r_id]['dirty']:
+        if self.run_cache[r_id]['rtime'] < now - 60:
+          del self.run_cache[r_id]
+      elif self.run_cache[r_id]['ftime'] < old:
+        old = self.run_cache[r_id]['ftime']
+        oldest = r_id
+    if oldest is not None:
+      if int(now) % 60 == 0:
+        self.scavenge(self.run_cache[oldest]['run'])
+      self.run_cache[oldest]['dirty'] = False
+      self.run_cache[oldest]['ftime'] = time.time()
+      self.run_cache_lock.release()  # Release the lock while writing
+      print("SYNC")
+      with self.run_cache_write_lock:
+        self.runs.save(self.run_cache[oldest]['run'])
+      # start the timer when writing is done
+      self.start_timer()
+      return
+    # Nothing to flush, start timer:
+    self.start_timer()
+    self.run_cache_lock.release()
 
   def scavenge(self, run):
     old = datetime.utcnow() - timedelta(minutes=30)


vondele:
btw, is fishtest otherwise doing fine under the current 'worker' load ? Seems like it is dealing with this OK.

ppigazzini:
@tomtor We could already using a too much high value for waitress threads, because the bottleneck should be python global interpreter lock
https://groups.google.com/forum/#!topic/pylons-discuss/kGNO3ifiacY

This thread suggest that the warnings should due to requests of static assets.
https://stackoverflow.com/questions/55857058/how-to-find-the-cause-of-task-queue-depth-warnings-from-waitress/55861495#55861495

If true we could use nginx to serve the static assets:
https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/deployment/nginx.html#step-3-serving-static-files-with-nginx-optional

DEV is working fine with this config, static is server by nginx (the static content is not found changing the root path)

```
...
server {
    listen 80;
    listen [::]:80;

    server_name dfts-0.pigazzini.it;

    location / {
        proxy_pass http://backend_all;
...
        location /tests {
            proxy_pass http://backend_tests;
        }
    }

    location /static {
        root                    /home/usr00/fishtest/fishtest/fishtest;
        expires                 30d;
        add_header              Cache-Control public;
        access_log              off;
    }
```

tomtor:
> We could already using a too much high value for waitress threads, because the bottleneck should be python global interpreter lock

@ppigazzini We have 30-50% average cpu load for pserv in the current high worker situation, so the processor is in general not busy spinning on the GIL. If you really hammer pserve, it will max out at about 200% CPU (I did some synthetic load testing in the past), because serving IO is done in the C-library parts, which are not impacted by the GIL.

I do not think that there is a real issue with the current 16 threads, but the current CPU load and the logging hint that a minor improvement in the throughput is possible by raising the limit. I suggest to test changing it at a time when the number of active workers is low.

tomtor:
> btw, is fishtest otherwise doing fine under the current 'worker' load ? Seems like it is dealing with this OK.

@vondele Yes, it does fine!

tomtor:
> This thread suggest that the warnings should due to requests of static assets.
@ppigazzini I do not think that static assets serving is an issue here. It is probably the locking on the update function which consumes the active threads from time to time (about twice a minute)

tomtor:
@ppigazzini I did some tests. Requesting the main page triggers the log errors. Generating that page with that long list of workers uses a lot of cpu.

tomtor:
@ppigazzini While mongodb dump is running a LOT of log lines are generated with queue depths over 100. Raising the threads will not fix that, but I still think that raising to 32 is good to test.

tomtor:
@ppigazzini I changed it by hand to 32 on prod to test...

ppigazzini:
@tomtor to get rid of those queue depth warnings some months ago I played unsuccessfully with the "threads" value, but the server is now so optimized that this time the result could be different. I suggest to try also a very low value (4) to view if the warnings get worse. BTW I meant simply that a too much high value is not useful, not that that hurts the performance.

Serving static assets with nginx could help?

ppigazzini:
@tomtor I found a way to serve static content with nginx.
* `~/fishtest/fishtest/fishtest/__init__.py`:
```
...
  config.add_subscriber(add_rundb, NewRequest)

#  config.add_static_view('html', 'static/html', cache_max_age=3600)
#  config.add_static_view('css', 'static/css', cache_max_age=3600)
#  config.add_static_view('js', 'static/js', cache_max_age=3600)
#  config.add_static_view('img', 'static/img', cache_max_age=3600)
  config.add_route('home', '/')
...
```
* `nginx.conf`:
```
...
server {
    listen 80;
    listen [::]:80;

    server_name dfts-0.pigazzini.it;

    location / {
...
        location /tests {
            proxy_pass http://backend_tests;
        }
    }

    location /css {
        root                    /home/usr00/fishtest/fishtest/fishtest/static;
        expires                 30d;
        add_header              Cache-Control public;
        access_log              off;
    }

    location /html {
        root                    /home/usr00/fishtest/fishtest/fishtest/static;
        expires                 30d;
        add_header              Cache-Control public;
        access_log              off;
    }

    location /img {
        root                    /home/usr00/fishtest/fishtest/fishtest/static;
        expires                 30d;
        add_header              Cache-Control public;
        access_log              off;
    }

    location /js {
        root                    /home/usr00/fishtest/fishtest/fishtest/static;
        expires                 30d;
        add_header              Cache-Control public;
        access_log              off;
    }
...
```

tomtor:
The root cause is that writing MongoDb can take a while and that I use a lock when a write is active. The error messages are in short bursts of < 0.5s and only happen when MongoDb has a high cpu usage.

I will have a look at this, but raising threads is not a fix.  I could skip some redundant syncs or do some less locking on writes or do non-blocking writes, but that is all a bit tricky.

vdbergh:
@tomtor Probably a naive question but does Fishtest use "transactions"? I do not know much about this but it could be an alternative to doing one own locking.

tomtor:
> Probably a naive question but does Fishtest use "transactions"? I do not know much about this but it could be an alternative to doing one own locking.

No, there is no transaction concept in FishTest or in MongoDb.

tomtor:
@ppigazzini This could be an easy fix. Currently every second the oldest run in the cache is written to MongoDb.

With this patch when we cannot acquire the lock we simply skip the update and log a message. 
I would expect that we see this log message once in a while instead of the original warning when under heavy load.

Testing this is a bit problematic (we need to generate either a large number of workers or slow down mongodb), but if the patch runs OK on dev we could try it on prod and monitor the log when under load.

tomtor:
> I found a way to serve static content with nginx.

@ppigazzini You could apply the change only to the nginx config and leave the Fishtest python handler unchanged? So Fishtest developers do not need to install nginx...

vdbergh:
@tomtor Well there appear to be transactions in MongoDb https://docs.mongodb.com/manual/core/transactions/ .

But I do not understand it when they say that operations on a single document are "atomic". This is not what transactions are supposed to solve. They are for multiple operations that have to be treated as a unit.

tomtor:
> But I do not understand it when they say that operations on a single document are "atomic".

They mean that an document is always updated as a whole. You cannot update a part of the bson document with the API.

Regarding the transaction description, that looks like decent transactions in later versions of MongoDb, but we cannot use them anyway because maintaining the state of every worker update in MongoDb is much to slow. Fishtest DID do that but that was changed two years ago because it could not handle the larger number of workers.

It IS slow, because of the atomic update. You cannot update the slot of a single worker with new W/L/D data in the task. You have to write the whole run data structure with all slots (of which many are not used). It is big and the data has to be passed to the MongoDb server over a local network connection, written to disk and acknowledged to the Fishtest server. This cannot be done every few milliseconds.

Transactions will also not solve the reason why locks are needed. We want to apply worker updates sequentially to the LLR computation. Originally there was no locking and updates overwrote the results of other threads. Tests ended but the final LLR result written to MongoDb was eg 2.91 and the test did not color green.

ppigazzini:
@tomtor I modified `__init__.py` to be sure that the content was served only by nginx :)

tomtor:
@ppigazzini I hope the last commit is OK. The change is to release the lock just before the (slow) MongoDb update is started. The timer for the periodic flush is restarted after the write completes, so we have never more than one update queued.

MongoDb is written on more places, but those events are rare, eg when new tests are created, tests are ended, etc.

vdbergh:
> Transactions will also not solve the reason why locks are needed. We want to apply worker updates sequentially to the LLR computation. Originally there was no locking and updates overwrote the results of other threads. Tests ended but the final LLR result written to MongoDb was eg 2.91 and the test did not color green.

I realize this discussion is a bit theoretical but the above issue can be handled with transactions. If two transactions (typically (1) read data (2) do something with it (3) write data) are not serializable one of them will fail to commit. The idea is that you then restart the transaction (in a loop).

tomtor:
> I realize this discussion is a bit theoretical but the above issue can be handled with transactions. If two transactions (typically (1) read data (2) do something with it (3) write data) are not serializable one of them will fail to commit. The idea is that you then restart the transaction (in a loop).

That is indeed a bit theoretical, because implementing it would be more complicated/less efficient than simply using a lock for the critical sections of code? Unless there is a Python framework to implement this, but I am not aware of it.

ppigazzini:
@tomtor PROD nginx configured:
```
...
    server_name tests.stockfishchess.org;

    location ~ ^/(css|html|img|js|favicon.ico|robots.txt) {
        root        /home/fishtest/fishtest/fishtest/fishtest/static;
        expires     30d;
        add_header  Cache-Control public;
        access_log  off;
    }

    location / {
        proxy_pass http://backend_all;
...

```

vdbergh:
@tomtor That part is not theoretical. It is a common design pattern https://en.wikipedia.org/wiki/Optimistic_concurrency_control
The idea is to let the database figure out for itself how to serialize transactions. It is also a common pattern in lockless programming (it requires an atomic "compare and swap").

ppigazzini:
@tomtor cleaned up PROD nginx.conf using regex (and improved formatting), updated wiki setup script.

ppigazzini:
@tomtor DEV updated with this PR, but I don't know how to test it. Do we must check only for clean logs?

tomtor:
> but I don't know how to test it.

@ppigazzini Yes, hard to test... The logs should stay clean, but for a real test we should first be able to simulate the old errors with master. 

On dev I can stop/continue mongodb on dfts-0 without errors:

    kill -STOP 323; sleep 5; kill -CONT 323

Not sure how to restore master on DEV (git status shows already master? First time I am fiddling on DEV)

ppigazzini:
@tomtor check "update_fishtest_p3.sh", to setup the master simply comment out the pull of PRs and start the script (view the comment at the start of the script).

ppigazzini:
@tomtor if you need to overload with request is should work something like this:
```python
  
 try: 
   print('Fetch task...') 
   t0 = datetime.utcnow()
   while True:
     try: 
        req = requests.post(remote + '/api/request_version', data=json.dumps(payload), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT) 
     except:
       print("request failed")
       time.sleep(0.1) # adjust according the result
   req = json.loads(req.text) 
 
```

https://github.com/glinscott/fishtest/blob/cc84d9c57e5224144b5219605f20f5b82c66c642/worker/worker.py#L90-L96

tomtor:
> if you need to overload with request is should work something like this:

Yes, but we need to slow down mongodb, Python fishtest handles the load fine.

I changed threads to `4` and get the queue warnings on master when I stop mongodb for 5 seconds.

With this PR no error messages.

The downside of this PR is that mongodb is less often updated.

Every second a single run (the last recently written) is written to MongoDb.

Master also writes when a worker completes its task, that is no longer done due to the second commit in this PR (https://github.com/glinscott/fishtest/pull/530/commits/0212d877e8ef4958c3376a7cb326d025b9827c6c).

It is now also possible that a partially updated Run is written to MongoDb (because we release the lock on the Run just before writing to MongoDb), eg the W/L/D results of a task are updated, but the sprt stats are not.

This will only make a difference for active tests when fishtest is restarted, but the correct WLD data will be rewritten by the workers within seconds.

ppigazzini:
I wrote this script on DEV if can help:
```
#!/bin/bash
while : ; do
  if [[ -f stop.loop ]] ; then
    exit
  else
    cd ${HOME}/backup
    mongodump
  fi
done
```

tomtor:
> I wrote this script on DEV if can help:

Thanks, with your looping MongoDb and threads = 2 I get some errors, but threads = 4 is fine.

tomtor:
@ppigazzini It's OK for me. The real test is when we have a really large worker pool.
Feel free to test or simulate with some more synthetic load. 

I would like to leave the two commits in this PR, because they can be applied (reverted) independently.

ppigazzini:
@tomtor the 30 January we had over 5000 workers connected
https://groups.google.com/d/msg/fishcooking/lebEmG5vgng/WdpuHKcOFgAJ

Perhaps this triggered the Warnings. Tonight I will check the log for the frequency of the Warnings in the last month.

tomtor:
> Perhaps this triggered the Warnings.

I am sure it did :-) I do not expect these often when we have less workers, but I remember seeing it on occasion. We do not really NEED this PR, but the PR make fishtest a bit more robust when we have a large worker pool.

tomtor:
@ppigazzini Did some refactoring (set timer in function)

tomtor:
@ppigazzini I squashed into two independent commits...

ppigazzini:
PROD updated, thank you @tomtor !

PR #532, opened 2020-01-31
Adjust bounds

vondele:
as per discussion in https://github.com/official-stockfish/Stockfish/issues/2531

introduce new bounds:

standard STC {-0.5, 1.5} -> 50% pass rate at 0.5 Elo (100k games), 1% pass rate at -1.0 Elo
standard LTC {0.25, 1.75} -> 50% pass rate at 1.0 Elo (137k games), 0.3% pass rate at -0.5 Elo
simplification STC/LTC {-1.5, 0.5} unmodified.

### DIFF
--- a/fishtest/fishtest/templates/base.mak
+++ b/fishtest/fishtest/templates/base.mak
@@ -62,7 +62,7 @@
           <li><a href="/users/monthly">Top Month</a></li>
           <li><a href="/actions">Actions</a></li>
           <li><a href="https://github.com/glinscott/fishtest/wiki" target="_blank">Wiki</a></li>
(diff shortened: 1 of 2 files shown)


vondele:
@ppigazzini @tomtor this needs a careful look to see if I updated all relevant places in the code.

ppigazzini:
@vondele DEV updated, looks good.

I'm viewing that you had already updated the wiki page, please update the STC bound values for the SPRT Calc too:
https://github.com/glinscott/fishtest/blob/59e046aa4d37a56d2119cde387778394e70b2b02/fishtest/fishtest/templates/base.mak#L65

ppigazzini:

vondele:
thank you!

PR #536, opened 2020-02-09
Better overshoot correction.

vdbergh:
The theoretical properties of an SPRT are established under the
    assumption that updates are continuous.
    
    To have accurate agreement with the predictions, when the updates are
    discrete, one has to estimate the expected "overshoot" and terminate
    the SPRT a bit early, accordingly.  Unfortunately, to the best of our
    knowledge, overshoot estimation is (surprisingly) computationally expensive.
    
    Moreover in the case of a GSPRT there is the added complication of
    unknown parameters and these have to be continuously estimated from
    the sample. This more or less means that the expected overshoot
    estimation has to be performed after every update, which is
    prohibitively expensive (IMHO, doing it only when we are "close" to
    one of the boundaries would lead to messy and possibly fragile code).
    
    A theoretical formula for the expected overshoot is: Siegmund -
    Sequential Analysis - Corollary 8.33. Like other such formulas, it is
    difficult to evaluate theoretically, but the result can be very easily
    estimated, with only a few lines of code, from the ongoing test itself.
    This "dynamic overshoot estimation" is implemented in this commit.
    
    That this works very well can be verified by simulation. A multi-threaded
    simulator written in C can be found here.
    
    https://github.com/vdbergh/simul
    
    Example:
    
    $ ./simul --elo0 0 --elo1 10 --elo 0
    
    yields
    
    ...
    sims=20001904 pass=0.050322[0.050175,0.050469] ...
    ...
    
    whereas
    
    $ ./simul --elo0 0 --elo1 10 --elo 0 --noovcor
    
    yields
    
    ...
    sims=20000943 pass=0.048530[0.048385,0.048674] ...
    ...
    
    (the confidence interval is 3 sigma).
    
    Obviously the version with overshoot correction is much more accurate.
    
    Currently Fishtest uses an approximate formula for overshoot
    correction which is actually not even asymptotically correct, although
    it works well for small Elo differences. However for larger Elo
    differences it starts to break down. This is undesirable as people may
    copy the Fishtest code and apply it in a different context.
    
    In the implementation the SPRT() function in stat_util.py now takes an
    sprt dictionary as argument. This dictionary may optionally contain a
    subdictionary with dynamic overshoot data which is updated in-place.
    It is assumed that SPRT() is called after each finished game (trinomial)
    or game pair (pentanomial). But calling SPRT() multiple times with
    the same parameters is safe and skipped updates are also handled sensibly.
    After a purge the overshoot data is deleted. Deleting the overshoot data
    is always safe as it simply means that the test is carried out with error
    probabilities that are slightly smaller than the assigned ones.
    
    The actual overshoot estimates can be found on the raw statistics
    page. Note that the estimate corresponding to the test outcome (H0 or H1)
    is typically the most accurate.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -498,11 +498,7 @@ class RunDb:
       sprt = run['args']['sprt']
       sprt_stats = fishtest.stats.stat_util.SPRT(
             self.get_results(run, False),
-            elo0=sprt['elo0'],
-            alpha=sprt['alpha'],
-            elo1=sprt['elo1'],
-            beta=sprt['beta'],
-            elo_model=sprt.get('elo_model', 'BayesElo')
+            sprt
             )
       if sprt_stats['finished']:
         run['args']['sprt']['state'] = sprt_stats['state']
--- a/fishtest/fishtest/stats/LLRcalc.py
+++ b/fishtest/fishtest/stats/LLRcalc.py
@@ -164,10 +164,6 @@ elo0,elo1 are in logistic elo.
 """
     s0,s1=[L_(elo) for elo in (elo0,elo1)]
     N,pdf=results_to_pdf(results)
-    s,var=stats(pdf)
-    # The well-known universal constant 0.583 is for normal increments.
-    # In practice it appears to work well in general.
-    overshoot=0.583*(s1-s0)/math.sqrt(var)
-    return N*LLR(pdf,s0,s1),overshoot
+    return N*LLR(pdf,s0,s1)
 
 
--- a/fishtest/fishtest/stats/stat_util.py
+++ b/fishtest/fishtest/stats/stat_util.py
@@ -144,7 +144,7 @@ Calculate an elo estimate from an SPRT test.
   a=sp.analytics(p)
 
   # Override the LLR approximation with the exact one
-  a['LLR']=LLRcalc.LLR_logistic(lelo0,lelo1,R_)[0]
+  a['LLR']=LLRcalc.LLR_logistic(lelo0,lelo1,R_)
   del a['clamped']
   # Now return the estimates
   return a
@@ -160,9 +160,23 @@ drawelo is estimated "out of sample".
   P1=bayeselo_to_proba(belo1,drawelo)
   return sum([results[i]*math.log(P1[i]/P0[i]) for i in range(0,3)])
 
-def SPRT(R, elo0, alpha, elo1, beta, elo_model=None):
-  """
-Sequential Probability Ratio Test
+def SPRT(R, sprt):
+  """Sequential Probability Ratio Test
+
+sprt is a dictionary with fields
+elo0, alpha, elo1, beta, elo_model
+
+sprt may optionally contain a dictionary with data for dynamic
+overshoot estimation. This will be updated in-place.  The theoretical
+basis for this is: Siegmund - Sequential Analysis - Corollary 8.33.
+The correctness can be verified by simulation
+
+https://github.com/vdbergh/simul
+
+Normally this function should be called after each finished game (trinomial) or
+game pair (pentanomial) but it is safe to call it multiple times with the same parameters.
+Skipped updates are also handled sensibly.
+
 H0: elo = elo0
 H1: elo = elo1
 alpha = max typeI error (reached on elo = elo0)
@@ -176,7 +190,15 @@ finished - bool, True means test is finished, False means continue sampling
 state - string, 'accepted', 'rejected' or ''
 llr - Log-likelihood ratio
 lower_bound/upper_bound - SPRT bounds
-"""
+
+  """
+
+  elo0=sprt['elo0']
+  alpha=sprt['alpha']
+  elo1=sprt['elo1']
+  beta=sprt['beta']
+  elo_model=sprt.get('elo_model', 'BayesElo')
+
   assert(elo_model in ['BayesElo','logistic'])
 
   result = {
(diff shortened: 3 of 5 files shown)


vdbergh:
It seems that it is indeed true that when the server is down (or communication fails temporarily for other reasons), the worker does not queue the updates. The culprit is run_game() in games.py. The worker should only read a new line from the queue if the previous one has been successfully processed (this could be accomplished by setting a flag `line_processed=False` while processing the failed update exception).

This is of course only a very minor issue but it is interesting to know.

ppigazzini:
@vdbergh tests on DEV looks good.
I hope to enforce the worker update before April 2020, so there is some time for write and test the proposed worker PR.

vdbergh:
Ok I'll give it a try!

vdbergh:
#539

vdbergh:
@ppigazzini  This version records some debug info and shows it in the raw_stats page. 

Perhaps the call to SPRT() in "format_results" is the culprit? That would only be a cosmetic issue.

ppigazzini:
@vdbergh problem triggered at the second try of `sudo systemctl stop fishtest@{6543..6544}; sleep 5; sudo systemctl start fishtest@{6543..6544}`
```
Expected overshoot [H0,H1] | [0.00000, 0.00000]

Overshoot debug | {'overshoot': {'last_update': 1086, 'minLLR': -0.04268204147865852, 'sq0': 0.0006003698494055341, 'maxLLR': 0.4493842715271147, 'sq1': 0.0037554184003874705}, 'num_samples': 1088, 'R_': [142, 213, 370, 215, 148], 'origin': 'sync_update_task', 'state': ''}
```

https://dfts-0.pigazzini.it/tests/stats/5e431a5ee27e4b719edf4a69

vdbergh:
Thanks for the test! So format_results is not the culprit (now). 

Since I think the worker with #539 cannot skip updates it appears it must be the server somehow. 

@tomtor do you see any mechanism by which this might happen? Perhaps if the server goes down after processing an update but before it has been written to the database?

ppigazzini:
With this test the problem triggered at the first try
https://dfts-0.pigazzini.it/tests/stats/5e431e44274b5d016e900c3e

vdbergh:
@ppigazzini  I hope @tomtor can comment but it now seems likely to me that the issue is simply that the server does not get a chance to update the database while going down. There was never any real need to do so since the workers would simply send new data superseding the old data.

 #539 fixes the issue of the worker skipping updates while the server is down (I think). But the above issue appears not so easily fixable. One may have to live with it.

I think I see a solution to deal with skipped updates without discarding the overshoot data. I will think about it.

ppigazzini:
@tomtor I keep connected my workers (updated with all the workers PRs) to process other tests.

tomtor:
> @ppigazzini I hope @tomtor can comment but it now seems likely to me that the issue is simply that the server does not get a chance to update the database while going down. There was never any real need to do so since the workers would simply send new data superseding the old data.
> 
> #539 fixes the issue of the worker skipping updates while the server is down (I think). But the above issue appears not so easily fixable. One may have to live with it.
> 
> I think I see a solution to deal with skipped updates without discarding the overshoot data. I will think about it.

Yes, updates are written only once per second to MongoDb, not on every update, for obvious performance reasons.

We would have to signal fishtest before restart and implement a flush if this is considered a real issue.

tomtor:
https://stackoverflow.com/questions/42978358/how-systemd-stop-command-actually-works/42979206

So it will not be difficult to add.  I will create a PR.

vdbergh:
@tomtor  Thanks. I just implemented a method to deal in a correct way with lost updates but it would obviously be better if there were no lost updates! So I'll wait for the PR to see how it fares.

If you will be catching SIGTERM perhaps you can also catch SIGINT (^C)? This is how I stop the server during development.

tomtor:
I have added proper flushing to #530 so that and this PR should be tested in combination

vdbergh:
I decided to make the overshoot data robust against skipped updates anyway so that it will work with the current worker. I have disabled the logging since I hope it is no longer needed. One can check if updates were skipped using `/api/get_run/<testId>` (the overshoot dictionary contains a field "skipped_updates"). 

The format of the overshoot data has changed so consulting tests with the old format will give an exception.

Things must be getting confusing:

#530: Guard against skipping updates on the server side during a server restart.
#539: Guard against skipping updates on the worker side during a server restart (or communication failure).

This PR (#536): Make overshoot data robust against skipped updates anyway. 

The overshoot data is still deleted after a purge (there does not seem to be a good way to recover from this).

ppigazzini:
@vdbergh @tomtor DEV updated with #530 #539.
https://dfts-0.pigazzini.it/tests/stats/5e43d0a075c6cbab61e90676
The overshoot data are still there after 15 minutes of running this script:
```bash
#!/bin/bash
t0=5
t1=25
while : ; do
  if [[ -f ${HOME}/stop.loop ]] ; then
    echo "stop ${0}"
    exit
  else
    sudo systemctl stop fishtest@{6543..6544}
    sleep ${t0}
    sudo systemctl start fishtest@{6543..6544}
    sleep ${t1}
  fi
done
```

Let me know if you want different t0 t1

vdbergh:
@ppigazzini Thanks! It is good to know that the server now seems to be completely unaffected by restarts!

ppigazzini:
The server seems running fine even stopping mongod for 20 seconds every 30 seconds :)

tomtor:
@ppigazzini I pushed a commit to #530 so could you redeploy/retest on DEV on a convenient time?

ppigazzini:
@tomtor done, started this script:
```
#!/bin/bash
t0=15
t1=45
while : ; do
  for u_s in "fishtest@6543 fishtest@6544" mongod nginx; do
    if [[ -f ${HOME}/stop.loop ]] ; then
      echo "stop ${0}"
      exit
    else
      sudo systemctl stop ${u_s}
      sleep ${t0}
      sudo systemctl start ${u_s}
      sleep ${t1}
    fi
  done
done
```

ppigazzini:
@tomtor @vdbergh server and workers are working fine despite the caos created with the stop-start script. Only SPSA test makes some ghost workers that are cleared after 60 minutes.

vdbergh:
This one appears to have lost its overshoot data in the end http://dfts-0.pigazzini.it/tests/stats/5e4407ba1074de5529736b2e. This is currently only possible if the number of games _goes down_. I thought this was only possible with purging.

Indeed the sprt dictionary (see http://dfts-0.pigazzini.it/api/get_run/5e4407ba1074de5529736b2e) contains "purged": True  but the test has not been purged (as we can now check because of the "audit trail" patch). 

Obviously this is a very exceptional situation and it is not a big deal.

ppigazzini:
I started that test with the services stopped for 20 seconds (t0) and running for 10 seconds (t1). The workers were mostly not able to get tasks from the server and that test lost very soon the overshoot data. So I changed t0  to 15 seconds and t1 to 45 seconds and created the other tests.

ppigazzini:
@vdbergh the other 3 SPRT have the overshoot data. I just started the loop script with t0=20 t1=10.

ppigazzini:
After 2 hours the overshoot data are still there, but alongside the 8 real workers there are 72 ghost workers waiting to be cleared after 60 minutes of inactivity.

vdbergh:
@ppigazzini Do you perhaps think #539 might be the culprit for the ghost workers? I don't really see by which mechanism this could happen...

ppigazzini:
No, always happened. Sometimes I purposely created a big fleet of ghost workers to test some PRs.

ppigazzini:
PROD updated, thank you @vdbergh !

vdbergh:
So it seems the number of games can go down as well...

http://tests.stockfishchess.org/tests/stats/5e4804a043e6cbcdc92d7874

According to 

http://tests.stockfishchess.org/api/get_run/5e4804a043e6cbcdc92d7874

one sample (2 games) was lost.

ppigazzini:
@vdbergh I just enforced the update to v72 for all workers.

vdbergh:
Thanks! However the fact that the number of games can go down (I saw more instances) must be a race condition on the server I think. It is benign of course but it would be interesting to understand it anyway.t

vdbergh:
I think it has to do with users bombarding the server with requests for the main page. Such requests call the SPRT function to get the LLR value. Perhaps there is a race condition with such requests being interleaved with updates from the workers.

This can be easily fixed by restricting such requests to only _consulting_ the sprt dictionary (the LLR value could be stored in the sprt dictionary itself, like the state field currently already is).

The idea is that the sprt dictionary would become more like a true object.

tomtor:
> I think it has to do with users bombarding the server with requests for the main page. Such requests call the SPRT function to get the LLR value. Perhaps there is a race condition with such requests being interleaved with updates from the workers.

The main page shows outdated info retrieved from a MongoDb query, not the internal fishtest server data. It does not call a SPRT function.

vdbergh:
The SPRT function is called in two places. In "rundb.py" in "update_task()" and in views.py in "format_results()". I assume the latter can be called when formatting the main page?

Anyway with my latest PR I eliminated the second call and enhanced the first call to store the necessary data for format_results() inside the sprt object. In this way format_results() will get the most recent data.

tomtor:
> The SPRT function is called in two places. In "rundb.py" in "update_task()" and in views.py in "format_results()". I assume the latter can be called when formatting the main page?

Yes, it is called, but it operates on a run copy which is retrieved from a MongoDb query. My guess would be that users viewing individual test pages trigger a call of the SPRT function on live data in the Python server.

tomtor:
> In this way format_results() will get the most recent data.

@vdbergh Not when it operates on old data retrieved from a MongoDb query

PR #538, opened 2020-02-10
Fix #537: cutechess warning for opening repetitions

### LINKED ISSUE #537: specifying 250 games
small issue, just so we don't forget for the eventual worker update.

with the new cutechess, we should specify -games 2 -rounds 125 -repeat to avoid a warning at cutechess startup. Also mentioned:

https://github.com/official-stockfish/Stockfish/issues/2538#issuecomment-583007717

ppigazzini:
Fix the "Warning: 2 opening repetitions vs 1 games per encounter"
Also raise the worker version to 72 (worker side).

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -550,7 +550,7 @@ def run_games(worker_info, password, remote, run, task_id):
 
   while games_remaining > 0:
     # Run cutechess-cli binary
-    cmd = [ cutechess, '-repeat', '-rounds', str(int(games_to_play)), '-tournament', 'gauntlet'] + pgnout + \
+    cmd = [ cutechess, '-repeat', '-rounds', str(int(games_to_play/2)), '-games', ' 2', '-tournament', 'gauntlet'] + pgnout + \
           ['-site', 'https://tests.stockfishchess.org/tests/view/' + run['_id']] + \
           ['-event', 'Batch %d: %s vs %s' % (task_id, make_player('new_tag'), make_player('base_tag'))] + \
           ['-srand', "%d" % struct.unpack("<L", os.urandom(struct.calcsize("<L")))] + \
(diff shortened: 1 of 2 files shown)


ppigazzini:
Tested with linux/windows/wsl workers.

ppigazzini:
@noobpwnftw I just merged 5 worker PRs and raised the worker version to 72 (only worker side), feel free to update your workers fleet at convenient time. Please mind that #528 fix the worker updated for custom folder names, perhaps this could help to automate the future updates of your workers fleet.

ppigazzini:
@tomtor worker PRs merged :)

ppigazzini:
@d3vv feel free to update your workers at convenient time :)

ppigazzini:
@vondele 
I'm confused, the pgns on fishtest seem to lack opening repetition: e.g. see rounds 7 and 9 from
https://tests.stockfishchess.org/api/pgn/605a08052433018de7a392d6-0.pgn

I was reviewing this commit. IMO it's working only thanks to the pair of games of the pentanomial model. With the old trinomial model if a worker was stopped after an odd number of games the `--rounds- , str(int(games_to_play/2))` should be wrong.

BTW that cutechess-cli warnings seems to disappear when replacing `'-rounds', str(int(games_to_play))` with `'-games', str(int(games_to_play))`, but I don't know cutechess-cli to state that the change is correct: in my tests the pngs written with either options lack opening repetition... :(

ppigazzini:
@vondele I got it: FEN is the starting position so the opening repetition is working fine.

From my tests the correct setting seems to be `'-games', str(int(games_to_play))`

vondele:
yes, it might be better to use games and not rounds.

ppigazzini:
#899 done, thanks.

PR #539, opened 2020-02-11
Queue worker updates if the server is down (worker update)

vdbergh:
Dropping worker updates violates the SPRT contract which says
that updates should occur after every game pair.

Also: for connection related exceptions print the exception instead
of the full backtrace.

This issue came up while testing. The backtraces are too verbose
as they refer to the inner workings of things like urllib etc... The
exceptions themselves seem to contain exactly the right amount
of information.

The effect of this PR can be seen using #536. With this PR the
overshoot data should no longer be deleted after a server restart.
This was verified with some local testing.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -272,8 +272,6 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
   global old_stats
   rounds={}
 
-  failed_updates = 0
-
   q = Queue()
   t = threading.Thread(target=enqueue_output, args=(p.stdout, q))
   t.daemon = True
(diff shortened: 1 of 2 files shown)


tomtor:
@vdbergh The loop was originally present, but I removed it because it prevented the fishtest server from recovering under heavy load. The server is much faster nowadays, so restoring it should be OK now.

Note that in general retrying will have no effect, because in 90% of the situations the server will not be reachable for quite a long time in case of a system failure. A fishtest restart because of a code update takes about a single second, so this PR will recover from "connection refused" during those short periods. That is certainly an improvement.

vdbergh:
@tomtor Ok I see. Although the loop in the PR isn't all that different from the loop in the current code (which also tries 5 times, but with different updates).

If ever there is a problem with the server being unable to cope with a barrage of queued worker output then the solution would be to reduce the number of retries or the sleep between retries. If a server restart indeed only takes one second then a single retry after a couple of seconds would be enough.

ppigazzini:
@vdbergh tested ok w/ linux/wsl/windows workers, overshoot data after some fishtest server restarts (vps reboot, fishtest setup, fishtest unit restart):
 - still there for:
https://dfts-0.pigazzini.it/tests/stats/5e428221b798c3119f9850ce
 - lost for:
https://dfts-0.pigazzini.it/tests/stats/5e4281a1b798c3119f9850c8


fishtest server restart times measured on DEV
- 1 second: after a git pull
- 38 seconds: after a fishtest server setup (routinely used on DEV, used on PROD to change python version)
- 100 seconds: after a vps reboot

The latter could be lower on PROD (748 MB/s disk throughput) wrt DEV (134 MB/s): same provider but different datacenters (US vs NL). I have in my todo the datacenter switch for DEV.

vdbergh:
Hmmm. I think the biggest cause for skipped updates is dealt with in this PR. Apparently there are some more possibilities but I would not really know where to look. Is a race condition on the server possible? But I guess it is not that important as it is really about rather rare events.

ppigazzini:
Overshooting data lost also for this new test after a manual service stop for 5 seconds
https://dfts-0.pigazzini.it/tests/stats/5e429d92f39defcc150d2335

The data persisted w/ a previous fishtest server setup (40 seconds server stop).

vdbergh:
Thanks! Is there a reasonable way to find out what goes on? Is there a way to log things? 

I could not reproduce it locally but on my laptop I can start at most 4 workers (the servers at work are CentOS 6 and I still haven't upgraded them).

ppigazzini:
@vdbergh some months ago I ran successfully the fishtest server on CentOS6 with up to date SW in this way:
 ```
# add some bash variables
sudo -i -u fishtest << EOF
cat << EOF0 >> .bashrc

export VENV=/home/fishtest/fishtest/fishtest/env
source /etc/bash_completion.d/git
EOF0
EOF

# install mongodb community edition
cat << EOF > /etc/yum.repos.d/mongodb-org-4.2.repo
[mongodb-org-4.2]
name=MongoDB Repository
baseurl=https://repo.mongodb.org/yum/redhat/\$releasever/mongodb-org/4.2/x86_64/
gpgcheck=1
enabled=1
gpgkey=https://www.mongodb.org/static/pgp/server-4.2.asc
EOF
yum install -y mongodb-org

# install python from software collections
yum install centos-release-scl
yum install rh-python36

# setup fishtest
sudo -i -u fishtest << EOF
source scl_source enable rh-python36
python36 -m venv \$VENV
\$VENV/bin/pip install --upgrade pip setuptools
cd /home/fishtest/fishtest/fishtest
\$VENV/bin/pip install -e .
EOF
```

The new cutechess-cli requires glibc2.14 so it doesn't work w/ CentOS6. I never tried to use a parallel glibc2.14:
https://unix.stackexchange.com/questions/176489/how-to-update-glibc-to-2-14-in-centos-6-5

**Comment by ppigazzini** [comment]
@vdbergh suggest way/code to log what happen (server, worker etc.)

**Comment by ppigazzini** [comment]
@tomtor 

> fishtest server restart times measured on DEV:
> 100 seconds: after a vps reboot

Solved changing `/etc/systemd/system/fishtest@service` from:
```
[Unit]
Description=Fishtest Server port %i
After=multi-user.target mongodb.service
```
to:
```
[Unit]
Description=Fishtest Server port %i
After=mongodb.service
```

At least one of the services in multi-user.target group is taking much time to start on DEV, no tried to reboot on PROD.

pserve seems to run fine also with mongod stopped for 20 seconds every 30 seconds so the new service unit should be ok.

ppigazzini:
@tomtor PROD service units fishtest@ and fishtest_dbg updated with `After=network.target mongod.service`. All reboot tests on DEV were fine. In case of errors logged after a PROD VPS reboot, we should or revert the change or apply one of this solutions:
https://unix.stackexchange.com/questions/531086/create-a-systemd-startup-script-that-delays-30-minutes
https://superuser.com/questions/1115426/systemd-autostart-service-after-mongodb

PR #541, opened 2020-02-16
Make it so that the sprt object is only updated in a single location.

vdbergh:
This will hopefully fix the occasional loss of overshoot data.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -532,12 +532,8 @@ class RunDb:
     # Check if SPRT stopping is enabled
     if 'sprt' in run['args']:
       sprt = run['args']['sprt']
-      sprt_stats = fishtest.stats.stat_util.SPRT(
-            self.get_results(run, False),
-            sprt
-            )
-      if sprt_stats['finished']:
-        run['args']['sprt']['state'] = sprt_stats['state']
+      fishtest.stats.stat_util.update_SPRT(self.get_results(run, False),sprt)
+      if sprt['state']!='':
         self.stop_run(run_id, run)
         flush = True
 
--- a/fishtest/fishtest/stats/stat_util.py
+++ b/fishtest/fishtest/stats/stat_util.py
@@ -160,54 +160,71 @@ drawelo is estimated "out of sample".
   P1=bayeselo_to_proba(belo1,drawelo)
   return sum([results[i]*math.log(P1[i]/P0[i]) for i in range(0,3)])
 
-def SPRT(R, sprt):
+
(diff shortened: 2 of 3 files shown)


ppigazzini:
Tests started on DEV.

vdbergh:
Thanks! It seems the code causes an exception when looking at older tests. This is not the case on my installation. Are such exceptions logged?

vdbergh:
I made an extra commit with recomputes the upper/lower bound fields in the sprt object to deal with earlier tests.

ppigazzini:
DEV updated, hoping in a @tomtor review, my server side knowledge is too low to help.

vdbergh:
@tomtor I am continuing the discussion here. It seems that the sprt object seen by `update_task` is at least sometimes the same as the one seen by `format_results`. I printed out the id's.

```
format_results 140694687141640
update_task 140694687141640
update_task 140694687141640
format_results 140694677086728
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
format_results 140694687141640
update_task 140694687141640
update_task 140694687141640
update_task 140694687141640
```

With the current PR `format_results` will no longer call `SPRT` (renamed as `update_SPRT`). It will just faithfully display what is already in the `sprt` object that it is given.

tomtor:
@vdbergh This PR is ok, but you should be aware that the main view shows outdated info from run-objects retrieved from MongoDb. These objects have the same id as those used in update_task, but they contain older data. So number of played games shown on the main page can be lower than shown on the individual test page.

vdbergh:
@ppigazzini  With the additional commit there are no longer exceptions when looking at older tests, it seems.

ppigazzini:
@vdbergh I will test a bit tonight before merging.

tomtor:
> Updating the sprt object while building the main page should not lead to a race condition as the object that would be updated is discarded afterwards (I assume).

Yes, that is correct

ppigazzini:
@vdbergh please squash :)

vdbergh:
Squashed.

Overshoots of random walks is really fascinating stuff. There is some deep math involved. There is for example an important connection with the zeta function.

https://projecteuclid.org/download/pdf_1/euclid.aop/1024404419

(Theorem 1.1).

ppigazzini:
PROD updated, thank you @vdbergh!

ps: learned a bit about random walk and martingale

PR #543, opened 2020-02-27
Remove some remnants of the old sprt handling.

vdbergh:
The commit 926b8853c0a8c1fc7 appears to have cured the problem of disappearing overshoot data. So it seems to be the right approach to keeping state across updates. In fact I now realize it is similar to how spsa works (Instead of an sprt object there is an spsa object, both represented by dictionaries of course, for easy conversion to json). It could be the model for new types of tests.

This commit removes some remnants of the old sprt handling. In particular it fixes a small bug. Currently a ghost (empty) sprt item is left in the results dictionary after the test is finished. This is innocent of course, but not nice.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -660,7 +660,8 @@ def purge_run(rundb, run):
 
     run['finished'] = False
     if 'sprt' in run['args'] and 'state' in run['args']['sprt']:
-      del run['args']['sprt']['state']
+      fishtest.stats.stat_util.update_SPRT(results,run['args']['sprt'])
+      run['args']['sprt']['state']=''
 
     rundb.buffer(run, True)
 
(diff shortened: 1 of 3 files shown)


fishtest/fishtest/views.py
```diff
@@ -660,7 +660,8 @@ def purge_run(rundb, run):
 
     run['finished'] = False
     if 'sprt' in run['args'] and 'state' in run['args']['sprt']:
-      del run['args']['sprt']['state']
+      fishtest.stats.stat_util.update_SPRT(results,run['args']['sprt'])
+      run['args']['sprt']['state']=''
```
tomtor: @vdbergh You replace the `del` with assignment of an empty string. That is less code but also less clean up of objects?

vdbergh: From my point of view the `'state'` field can have three values `''` (test isn't finished), `'accepted'` and `'rejected'`. 

So `run['args']['sprt']['state']='' ` seems somewhat more correct to me than deleting the state field and then elsewhere interpreting a missing state field as `''`.

ppigazzini:
@vdbergh test started on [DEV](https://dfts-0.pigazzini.it/tests)

ppigazzini:

PR #548, opened 2020-03-21
Add index on username to actions

tomtor:
To speed up queries, especially for non existing users


tomtor:
@ppigazzini Because this script drops/recreates the indices which might cause issues in production, I editted ~/x/rebuild_indices on prod to create just this index. I did not run it....

ppigazzini:
@tomtor script ran, PROD updated.

Issue #384, opened 2019-08-31
SMP tests: add "Standard STC SMP" and "Standard LTC SMP" buttons


Vizvezdenec:
Currently we have 2 buttons - standart STC and standart LTC which set TC and hash accordingly.
But we have no standart STC SMP and standart LTC SMP which are actually more or less standart but not having them can create quite a mess.
I propose to create this 2 buttons (at test submitting page), stc should set TC to 5+0.05, threads to 8 and hash to 32, LTC should set TC to 20+0.2, threads to 8 and hash to 128.
@vondele what do you think about it? :)

tomtor:
I agree that adding those options is a good suggestion.

vondele:
yes, that's a good idea, as long as the 8 core machines dominate fishtest.

Vizvezdenec:
So, will this be done?
Whenever I try to do SMP tests I ALWAYS fuck up TC/hash/whatever at least once because I click to standart buttons and forget either to set threads/hash/bounds/tc at least once in series test.
Like there.
https://tests.stockfishchess.org/tests/view/5e87a3694411759d9d0986ca
:) Hash for STC should be 64 now (I think) and for LTC probably 256 or smth like this.

ppigazzini:
@Vizvezdenec merged, I will update PROD as soon as it will be on line.

PR #555, opened 2020-04-05
Single click SMP tests

vondele:
add single click SMP options.
8 threads, 5+0.05, 64Mb
8 threads, 20+0.2, 256Mb

fixes https://github.com/glinscott/fishtest/issues/384

Untested.... copy&paste from similar code around, would need a careful check.

No so urgent either.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -12,6 +12,12 @@
     <label class="control-label">Test branch:</label>
     <div class="controls">
       <input name="test-branch" value="${args.get('new_tag', '')}" ${'readonly' if re_run else ''}>
+      <div class="btn-group">
+        <div class="btn" id="fast_test">short (STC)</div>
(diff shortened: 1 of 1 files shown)


ppigazzini:
@vondele [DEV](https://dfts-0.pigazzini.it/tests) updated, it seems work fine.
Please make a double check.
user01:user01 (normal)
user00:user00 (approver)

vondele:
@ppigazzini thanks .. functionality seem to be OK. I've added one more commit to move this up, closer to the branch. It sets more than just TC, and it is convenient to have it around. It seems a trivial change, but maybe better have a second look.

vondele:
let me squash it first...

ppigazzini:
DEV updated, it seems OK, please double check.

vondele:
looks good to me.

ppigazzini:
I will updated PROD as soon as it will be on line, thank you @vondele :)

Issue #340, opened 2018-12-14
Put the "test repo" field next to the "test branch" field


Alayan-stk-2:
I lost time in trial and error before finding how to correctly submit a test, because the branch is only indicated at the beginning while the base of the repo, a closely related inormation, is put as the last information to fill out. I think these two are logically related and so should be placed close to each other.

PR #556, opened 2020-04-05
Move "Test repo" field right under "Test branch"

linrock:
The test repo and test branch fields are closely related. As reported in a separate issue, I also think this is a UI problem for first-time test submitters. Fixes https://github.com/glinscott/fishtest/issues/340

---
This is how the top of the "Create new test" page will look:

image

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -21,6 +21,12 @@
     </div>
   </div>
   <div class="control-group">
+    <label class="control-label">Test repo:</label>
+    <div class="controls">
+      <input name="tests-repo" value="${args.get('tests_repo', tests_repo)}" ${'readonly' if re_run else ''}>
+    </div>
+  </div>
+  <div class="control-group">
     <label class="control-label">Test options:</label>
     <div class="controls">
     <input name="new-options" value="${args.get('new_options', 'Hash=16')}">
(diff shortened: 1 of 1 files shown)


tomtor:
Looks good to me

ppigazzini:

xoto10:
As a user comment, I would say the repo would be better as the 4th Test field, then the Test and Base sections each start with Branch, Options, Signature.
Just my 2p.

If you could push the 4 Test fields a bit closer together, and the same for the 3 base fields, that would be good for me as well. I always feel the New Test page is a bit long, just a bit more white space than required between the fields. (I guess if you have a higher res screen than me, and better eyesight, then maybe this isn't a problem!)

These are only cosmetic suggestions, so I have no problem if you disagree.

linrock:
@xoto10 to confirm, by grouping them closer together, do you mean something like this?



I agree that the UI is very worth improving, so i'm happy to make these sorts of changes :)

ppigazzini:
@linrock you are very skilled about web programming. Please review the libraries (bootstrap etc.) and the way we are using them and open an issue if you think that something needs an update and an improvement :)

linrock:
@ppigazzini sounds good, will do :)

xoto10:
@linrock 
Yes, exactly. :)

linrock:
@xoto10 sounds good, i'll open a PR to update the form with your suggestions

Issue #332, opened 2018-11-14
embed github diff as an iframe ?


vondele:
One thing that would potentially make the workflow of approving tests easier is if the github diff would also be embedded on the test's webpage. (E.g. after the summary and before the list of tasks). 

Obviously if this is a lot of work, that's not needed, but if that's a quick hack...

PR #558, opened 2020-04-06
Show github diff above tasks section on test pages

linrock:
This shows the Github diff on the test page so you don't have to open the diff in a separate tab or window. The browser fetches the diff from Github through their API. If for some reason the request fails, then the diff section just won't show up. Fixes https://github.com/glinscott/fishtest/issues/332

---
For example, this is how this test page would look with the diff rendered:
https://tests.stockfishchess.org/tests/view/5e8a751164ea67c1c449bb7e

image

---
Github API docs for commit ranges:
https://developer.github.com/v3/repos/commits/#compare-two-commits

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -138,6 +138,14 @@ Gaussian Kernel Smoother&nbsp;&nbsp;<div class="btn-group"><button id="btn_smoot
 <div id="div_spsa_history_plot"></div>
 %endif
 
+<section class="diff" style="display: none">
+  <h3>
+    Diff contents
+    <button id="diff-toggle" class="btn">Show</button>
+  </h3>
+  <pre class="diff-contents"></pre>
+</section>
+
 <h3>Tasks ${totals}</h3>
 <table class='table table-striped table-condensed'>
  <thead>
@@ -209,3 +217,38 @@ Gaussian Kernel Smoother&nbsp;&nbsp;<div class="btn-group"><button id="btn_smoot
   %endfor
  </tbody>
 </table>
+
+<script>
+  $(function() {
+    var apiUrlBase = "${base.repo(run)}".replace("//github.com/", "//api.github.com/repos/");
+    $.ajax({
+      url: apiUrlBase + "/compare/${run['args']['resolved_base'][:7]}...${run['args']['resolved_new'][:7]}",
+      headers: {
+        Accept: "application/vnd.github.v3.diff"
+      },
+      success: function(response) {
+        var numLines = response.split("\n").length;
+        var $toggleBtn = $("#diff-toggle");
+        var $diffContents = $(".diff-contents");
+        $diffContents.text(response);
+        $toggleBtn.on("click", function() {
+          $diffContents.toggle();
+          if ($toggleBtn.text() === "Hide") {
+            $toggleBtn.text("Show");
+          } else {
+            $toggleBtn.text("Hide");
+          }
+        });
+        // Hide large diffs by default
+        if (numLines < 50) {
+          $diffContents.show();
+          $toggleBtn.text("Hide");
+        } else {
+          $diffContents.hide();
+          $toggleBtn.text("Show");
+        }
+        $(".diff").show();
+      }
+    });
+  });
+</script>


vondele:
Cool! (Without being able to review the patch). 

Is there something in place to avoid filling too much screen space if the patch is 100s of lines? (e.g. scroll or truncate or expand...). I'd be happy to have a look at this on the dev server once it is ready.

ppigazzini:
DEV updated.
Please insert a "show/hide" button to deal with very long diff, view here the result
https://dfts-0.pigazzini.it/tests/view/511f04d18f0c3e301ced53e6 

```
<section class="diff" style="display: none">
<h3>Diff contents
<button id="diff-button" class="btn" data-toggle="collapse" data-target="#diff-box">Show</button>
</h3>
<div class="collapse" id="diff-box">
  <pre class="diff-contents"></pre>
</div>
</section>
```

ppigazzini:
@linrock please squash the commit when finished :)

vondele:
additionally, if this is merged the AUTHORS file should be updated.

linrock:
@vondele got it, i'll open a separate PR where the "Master diff" link shows some visual hint if the base branch doesn't match official master.

@ppigazzini i've added a "show/hide" button as you suggested. the bootstrap button toggling felt laggy since it animates the collapse which feels unnecessary for large diffs so i made the diff simply show/hide instantly. i've squashed the changes into one commit.

the current behavior of the diff section now is:

- the diff is shown by default if it's small (<= 50 lines total)
- the diff is hidden by default if it's large
- click the "show/hide" button to toggle the diff

I'm happy to make more changes if you all feel this could be better :)

vondele:
The described diff behavior sounds very good.

ppigazzini:
DEV updated, It looks good to me :)

Long diff
https://dfts-0.pigazzini.it/tests/view/5e790d12e42a5c3b3ca2e9b7

Short diff
https://dfts-0.pigazzini.it/tests/view/5e8b500f64ea67c1c449bc24

ppigazzini:

jhellis3:
This is great. :)

vondele:
@linrock I found a first disadvantage of this patch... previously, looking at the diff on github, one would also see if there were github comments on the patch. Now, having the diff inline in the page, they are overlooked. Is it possible to extract the fact that there are comments on a commit / branch from the github api, and if so could that be shown one way or another?

linrock:
@vondele i looked at the Github API and there indeed is a way to tell the # of comments on the commit range. For example summing up the `comment_count` values in this API response will give you the number of comments on the diff:
https://api.github.com/repos/protonspring/Stockfish/compare/195a4fe...f43c53e

Maybe showing the # of comments on the top of test pages would help?
https://tests.stockfishchess.org/tests/view/5e8cbcd75450ceed16295142

What do you think of something like this?
image

vondele:
yes, that would work well. 

I'm wondering what the best location of that number would be, or if we even should change the layout of the page a little. Basically, I think I'll be looking most of the time at the new `diff contents` section before approving, and might oversee the `diff (9 comments)` link unless that info is closer. What about moving the `diff (9 comments)` link closer to the caption `diff contents`?  Alternatively/ additionally, would it make sense to have the `diff contents` located on the page where we have the `details` now and move details down ?

linrock:
i think moving the # comments to the diff section makes sense. maybe something like this?
image

---

For now, i don't think we can have the diff contents on the top of the page yet because the diff data is fetched by the browser after the page loads, and it would be jarring to insert it above the details section after it's already on the page. alternatively, if the diff were fetched from the server, this would not be a problem. i just realized that the github API has an unauthenticated rate limit of 60 requests/hour per user, so that might be an issue if you look at enough test pages.

vondele:
that screenshot look good to me. Location is fine as well.

Let's see if the API limit becomes a problem...

ppigazzini:
On PROD/DEV servers we use .netrc basic authentication to get 5000 request/hour, view this
https://github.com/glinscott/fishtest/issues/385#issuecomment-528313637

linrock:
@ppigazzini thanks, good to know. i'll see about moving github API requests to the server then. perhaps starting with the "Master diff" check

ppigazzini:
Just a remainder.

Hypotheses:
 - this feature stay unmodified (one api call after opening a test page with short diff)
 - some users start complaining about the 60/h rate limit wasted opening not interesting test pages

Solution:
I think that should be possible to start the test page w/o the diff box and make the api call only after pushing the button "Show". In this way an user should blame only himself to have hit the hourly rate limit :)

By the way we could show also some rate limit info if any api call is already returning the `X-RateLimit-Limit` and the `X-RateLimit-Remaining`

```bash
curl -I https://api.github.com/users/github --netrc -v
...
< Server: GitHub.com
Server: GitHub.com
< Status: 200 OK
Status: 200 OK
< X-RateLimit-Limit: 5000
X-RateLimit-Limit: 5000
< X-RateLimit-Remaining: 4996
X-RateLimit-Remaining: 4996
< X-RateLimit-Reset: 1586366162
X-RateLimit-Reset: 1586366162
...
```

vondele:
It is nice the diffs are open upon visiting the page (on the way to 1-click approvals ;-). 

I think, practically, it wouldn't matter much if the only thing that happens upon exceeding the 60/h (which is generous anyway) is the diff not showing, and having to go back to github. However, if storing server side diffs would enable searches (of the diffs and their commit messages) I could really see a benefit. Searching old patches with a certain idea is something that happens frequently, but is rather tedious right now.

ppigazzini:
> on the way to 1-click approvals ;-)

mmm, we should add a chess captcha like lichess ;)

ppigazzini:
@vondele @linrock github has also api for search, with a different rate limit

https://developer.github.com/v3/search

PR #559, opened 2020-04-07
Show result of comparing base branch with master on pending tests

linrock:
@vondele this is an update to the "Master diff" link in the actions sidebar section for pending tests as you suggested here: https://github.com/glinscott/fishtest/pull/558#issuecomment-610022442

The client will check the diff between the base branch of the submitted test and official Stockfish master. If there's no difference, it shows a green success message. Otherwise, a red error message.

---
If there's no difference between the base branch and official Stockfish master:

---
If the base branch doesn't match official Stockfish master:

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -58,10 +58,9 @@ var spsa_history_url = '${run_args[0][1]}/spsa_history';
   </form>
 %if not run.get('approved', False):
   <span>
-    <a href="https://github.com/official-stockfish/Stockfish/compare/master...${run['args']['resolved_base'][:7]}" target="_blank">Master diff</a>
     <form action="/tests/approve" method="POST" style="display:inline">
       <input type="hidden" name="run-id" value="${run['_id']}">
-      <button type="submit" class="btn btn-success">
+      <button type="submit" id="approve-btn" class="btn btn-success">
         Approve
       </button>
     </form>
@@ -78,6 +77,15 @@ var spsa_history_url = '${run_args[0][1]}/spsa_history';
   <a href="/tests/run?id=${run['_id']}">
     <button class="btn">Reschedule</button>
   </a>
+
+%if not run.get('approved', False):
+  <br/>
+  <br/>
+  <div id="master-diff" class="alert">
+    Comparing base branch with master...
+  </div>
+  <a href="https://github.com/official-stockfish/Stockfish/compare/master...${run['args']['resolved_base'][:7]}" target="_blank">Master diff</a>
+%endif
 
   <hr>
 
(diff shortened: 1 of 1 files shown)


vondele:
Looks excellent!

(not a review, I notice one strange piece of syntax).

fishtest/fishtest/templates/tests_view.mak
```diff
@@ -209,3 +217,24 @@ Gaussian Kernel Smoother&nbsp;&nbsp;<div class="btn-group"><button id="btn_smoot
...
+  $(function() {
+    var apiUrlBaseMaster = "https://api.github.com/repos/official-stockfish/Stockfish";
+    $.ajax({
+      url: apiUrlBaseMaster + "/compare/master...${run['args']['resolved_base'][:7]}",
+      headers: {
+        Accept: "application/vnd.github.v3.diff"
+      },
+      success: function(response) {
+        var $masterDiff = $("#master-diff");
+        if (response.length === 0) {
```
vondele, line 231: is  `response.length === 0` (3x =) correct syntax ?

tomtor: Yes, it's equal without type coercion.

ppigazzini:
@linrock please rebase on new master

tomtor:
Hi @linrock Can you also add your name to AUTHORS in this commit?

ppigazzini:
@tomtor @linrock #560 AUTHORS already updated :)

ppigazzini:
@vondele @tomtor CI checks stopped working?

tomtor:
> @vondele @tomtor CI checks stopped working?

Yes and no: You can see them here:

https://travis-ci.org/github/glinscott/fishtest/builds

Not sure what to change to recreate the green/red markers in the overview

vondele:
Several projects have had this problem, it is unclear how to solve it. Disabling and reabling the github integration worked for some, not for all.

ppigazzini:
@vondele the SF Readme.md references the @mcostalba repo: is it correct? 
https://raw.githubusercontent.com/official-stockfish/Stockfish/master/Readme.md
```
## Overview

[![Build Status](https://travis-ci.org/official-stockfish/Stockfish.svg?branch=master)](https://travis-ci.org/official-stockfish/Stockfish)
[![Build Status](https://ci.appveyor.com/api/projects/status/github/official-stockfish/Stockfish?branch=master&svg=true)](https://ci.appveyor.com/project/mcostalba/stockfish/branch/master)

```

vondele:
hmm seems like it works, but it is possibly not the best setup.

linrock:
@ppigazzini i've rebased on master and squashed into one commit. ready for testing!

ppigazzini:
DEV updated

ppigazzini:
Looks good to me.
@tomtor perhaps we should write a different message in "Actions" when approving with the Orange button (base branch different from official master)

vondele:
Dev seems to work as expected... not sure it is very important to adjust Actions. If it is trivial, why not, otherwise I would leave as is. The most important part if to give an easy indication to the approver, all info is there somehow anyway.

linrock:
Since it's unsafe to click "Approve" when the base branch is different from official master, in these cases I suggest that the "Approve" button is disabled until the approver clicks on the "Master diff" link. Adding that extra step ensures the approver looks at the diff before they can click the button.

If this behavior sounds good, i can open a separate PR for it.

ppigazzini:
PROD updated.
@linrock you proposal looks good to me :)

vondele:
I'm not sure a requirement is needed. In some cases, e.g. long term devs that trigger a regtest, looking at the diff makes no sense (1000s of lines), and we kind of need to trust the dev. 

Is it possible to check if the selected sha is the same as an existing one if official-stockfish first, and skip the requirement if the sha corresponds to a commit in official stockfish? If they are the same, it should not be a requirement to check (the code won't be malicious). Regression tests, and devs that are a commit or so behind are relatively common.

The other case where large diffs can exist is for the tunes, as people have to merge the tuning branch. These diffs are practically also un-checkable. Not sure if we have a good solution (one option would be to integrate the tuning branch in master).

ppigazzini:
@vondele using a previous commit as base makes a green "Approve" button, tested with this
https://dfts-0.pigazzini.it/tests/view/5e8c9f1918ffe8a1c9794f27

To get an orange "Approve" button I merged a new commit (new txt file) in the master.

ppigazzini:
@linrock mmm, it seems that we have a problem with old tests, view at example https://tests.stockfishchess.org/tests/view/5214a8390ebc59160a0182f6
this seems the last test OK
https://tests.stockfishchess.org/tests/view/5214ce9a0ebc59160a0182f8

linrock:
@ppigazzini got it, i'll look into fixing it. seems like for older tests we will have to check if both the test is not finished + not approved to decide whether to show that message.

PR #561, opened 2020-04-07
Add colored syntax highlighting to diffs

linrock:
This should make the diffs more readable and look more similar to the diffs on Github.

---
Example of how a colored diff looks:

image

Uses https://highlightjs.org/


ppigazzini:
PROD updated thank you @linrock :)

Issue #450, opened 2019-11-29
Missing number of active tests


mrobaczyk:
On the main page we have:
Pending - X tests
Finished - X tests
but Active has no such counter - it displays machines and cores info though.

I looked in the code and there is something that I dont understand:
```
%else:
 <h3>Active - ${len(runs['active'])} tests</h3>
%endif
```
It should be displayed when 'show_machines' is false but it doesn't change when I hide machines list.

@tomtor if you think that adding active tests counter is a nice improvement, could you please look at the code?
Thanks :)

tomtor:
@mrobaczyk That code is executed when you show tests by a user by selecting a user name (click on a blue username on the main page).

The active line is already a bit long, but adding the tests counter is ok for me. Perhaps we should remove the nps info or show it somewhere else.

Can you create a PR?

mrobaczyk:
Yes maybe we should remove the average nps, but leave cores and total nps. But I don't think I'm the right person to make such decisions. :)

xoto10:
I guess this is a separate suggestion:
I think the games/minute is a theoretical calculation rather than actually counting how many games are being played each minute. It would be nice to display the actual throughput rather than a theoretical throughput.

PR #562, opened 2020-04-07
Show # of active tests on homepage

linrock:
So you can easily tell how many tests are running. The homepage will also look more consistent with user pages this way. Fixes https://github.com/glinscott/fishtest/issues/450

---
The homepage will look something like this:

image

### DIFF
--- a/fishtest/fishtest/templates/tests.mak
+++ b/fishtest/fishtest/templates/tests.mak
@@ -6,22 +6,8 @@
 
 %if page_idx == 0:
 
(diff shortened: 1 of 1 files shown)


vondele:
result looks good! 

it is unrelated... but does anybody know what the pending_hours really mean... actually, looking a bit at the code, this really is a very very hacky way to estimate runtime till completion. (e.g. hardcoded expected number of games = 16000). I'll file a separate issue.

ppigazzini:
DEV updated

ppigazzini:

Issue #531, opened 2020-01-30
Using impossible book stalls workers


snicolet:
Hi all,

In the following test:
https://tests.stockfishchess.org/tests/view/5e330750ec661e2e6a34253b

I made a copy-cut error for the book field, and the resulting book name was 
"**book 2moves_v2.pgn**" instead of just "**2moves_v2.pgn**"

==> The result was that this test sucked hundreds of workers, which never started (I manually stopped the faulty test at the end when I understood).

EDIT: in fact this test had two errors at the same time, the faulty book name and a faulty bench number for the base.

PR #565, opened 2020-04-07
Validate book is in official books repo when submitting tests

linrock:
As reported in a separate issue, submitting a test with an invalid book name will cause workers to hang until the test is manually stopped. This PR prevents that from happening by making sure the book exists in https://github.com/official-stockfish/books. If the supplied book name is invalid, the page will display an error.

Fixes https://github.com/glinscott/fishtest/issues/531

---
Example of the error message when using an invalid book name:


ppigazzini:
PROD updated, thank @linrock :)

PR #566, opened 2020-04-07
Show number of github comments on the diff when viewing tests

linrock:
@vondele this adds the # of comments to the diff section as you suggested here https://github.com/glinscott/fishtest/pull/558#issuecomment-610596557

---
Example of how this page would look:
https://tests.stockfishchess.org/tests/view/5e8cab330ffd2be7f15e560f

image


vondele:
thanks!

ppigazzini:
Thanks @linrock :)

Issue #339, opened 2018-12-12
Propositions in improving fishtest.


Vizvezdenec:
https://github.com/official-stockfish/Stockfish/issues/1865

linrock:
Quoting the text from the linked issue here for convenience.

FauziAkram said:
> Can we implement a system of filters in the main page of fishtest, so we can play around with the filters, and find exactly the tests that we need.
> 
> Criteria that might be used for the filters:
> -Type of test (Sprt/Spsa/Num of games)
> -Color (White, Red, Yellow, Green)
> -Time Control (Stc/Ltc/Other)
> -SPRT bounds ( [-3,1]/[0,4]/[0,5]/other)
> -Number of games (from: X / to: Y)

Vizvezdenec said:
> This will be useful to surf for for example "good" parameter tweaks to make big combo patches of them.
> Once in 2/3 months there is always some combo patch passing but it requires quite a lot of deep digging into fishtest and a lot of effort. With this functionality it will get 50 times easier.
> Also if we will soon have SPRT bounds changes it will be really useful for searching for "almost passed" patches to see if they will pass new SPRT (see discussion there basically approved by both maintainers #1859 ).

vondele:
some discussion in fischcooking on the usefulness of search also there https://groups.google.com/d/msg/fishcooking/rQGALpXDZew/boJ2UPQVAQAJ

vondele:
So, some ideas about searching, not yet very organized or firm. 

Ideally we can use the search to answer questions like 'what are all LTC evaluation patches by Viz he ran last month, changed something about pawns and finished yellow'

certain aspects seem 'easy'
* restrict date to a range
* find the author (github and fishtest handles might be different), maybe fuzzy search (have Viz match Vizvezdenec)
* test type (sprt), outcome (yellow, green, or Elo range), parameters (ltc, stc, smp, ...)

the most tricky, but most valuable questions are trying to infer the topic. This could be things we have in the database
* branch name (fuzzy match e.g. pawns match PawnTweak12)
* info field

others might require additional interaction with github
* commit messages
* files changed (e.g. evaluate.cpp hints at and eval patch)
* actual diff content ? 
* github comments

mstembera:
Sometimes the test note is not the same as the github commit message.  Being able to search the notes and the name of the test branch using keyword partial match would be really helpful.

vondele:
[...]

PR #568, opened 2020-04-07
Server-side check for whether master diff matches base branch

linrock:
Since the client-side unauthenticated Github API limit is only 60 requests/hour, users will run into issues with the master diff check if they look at a bunch of tests.

This PR fixes that by moving this API request to the server, which has a limit of 5,000 API requests/hour (source: https://github.com/glinscott/fishtest/issues/385#issuecomment-528313637)

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -67,6 +67,7 @@ class RunDb:
               msg_new='',
               base_signature='',
               new_signature='',
+              base_same_as_master=None,
               start_time=None,
               sprt=None,
               spsa=None,
@@ -118,6 +119,7 @@ class RunDb:
       # Aggregated results
       'results': {'wins': 0, 'losses': 0, 'draws': 0},
       'results_stale': False,
+      'base_same_as_master': base_same_as_master,
       'finished': False,
       'approved': False,
       'approver': '',
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -60,7 +60,8 @@ var spsa_history_url = '${run_args[0][1]}/spsa_history';
   <span>
     <form action="/tests/approve" method="POST" style="display:inline">
       <input type="hidden" name="run-id" value="${run['_id']}">
-      <button type="submit" id="approve-btn" class="btn btn-success">
+      <button type="submit" id="approve-btn"
+              class="btn ${'btn-success' if run['base_same_as_master'] else 'btn-warning'}">
         Approve
       </button>
     </form>
@@ -78,11 +79,16 @@ var spsa_history_url = '${run_args[0][1]}/spsa_history';
     <button class="btn">Reschedule</button>
   </a>
 
-%if not run['finished'] and not run.get('approved', False):
+%if run.get('base_same_as_master') is not None:
   <br/>
   <br/>
-  <div id="master-diff" class="alert">
-    Comparing base branch with master...
+  <div id="master-diff"
+       class="alert ${'alert-success' if run['base_same_as_master'] else 'alert-error'}">
+    %if run['base_same_as_master']:
+      Base branch same as Stockfish master
+    %else:
+      Base branch not same as Stockfish master
+    %endif
   </div>
   <a href="https://github.com/official-stockfish/Stockfish/compare/master...${run['args']['resolved_base'][:7]}" target="_blank">Master diff</a>
 %endif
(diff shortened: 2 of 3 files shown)


ppigazzini:
DEV updated, after opening more than 60 test pages the code stopped working.
```bash
$ git log --oneline -n 5
6d85f19 (HEAD -> master) Check if master diff matches base branch from server
788d36b (origin/master, origin/HEAD) Don't show base branch comparison message for old finished tests
52b37f0 Make sure book exists in official repo when validating tests
e04e8e4 Show # active tests on homepage. Move active machines section
e91c12a Merge pull request #561 from linrock/colored-diffs
```

ppigazzini:
@linrock if can help we introduced the github api in this PR https://github.com/glinscott/fishtest/pull/239

linrock:
hmm, sounds like the server-side github API requests aren't using `.netrc`.. i'll look into this

ppigazzini:
I redone the test to be sure to don't have cached the old js code. After hitting with chrome the 60 page limit, edge doesn't show the diff box too.

linrock:
@ppigazzini do you mean this diff box is going away after 60 page refreshes? if so, this is expected behavior for now. i can fix that separately



---
This PR only makes server-side Github API requests for this section:


i double-checked on a dev server and python requests do correctly use the credentials in `.netrc` so the 5,000 Github API request limit on the server should be respected

xoto10:
Not sure if this belongs here:
Note that the message about not matching the base branch should not be displayed for SPSA (tuning) runs, since these test the tuning branch against itself.

vondele:
I think it is good to have it shown.... it is a warning that one should, in principle, check the branch for malicious code... this is right now almost not possible for SPSA, so people just skip it. We should have a better way to do that. The only option I see right now is to merge the tune branch in master, but maybe there is an alternative.

vondele:
BTW, I'm not sure a server side check is a good idea... if this is a counted access for each page view (of tests), the 5000/h could be exhausted easily and certainly as a DDOS and would result in being unable to submit new tests. Not sure how many page views fishtest has /h, but it could be quite many. On the other hand limiting a user to 60 page views / h is safe and probably relatively reasonable. I could imagine reaching it, but then it is presumably better to take a break anyway ;-)

ppigazzini:
@linrock not able to understand all the js code (and a bit lazy) I was testing the wrong way, sorry :(

@vondele If I'm not wrong a skilled js developer is able already now to forge a simple attack exhausting our API call limit. So we should simply avoid to hit naively our 5000 API rate limit. At example I have some test scripts that loop on several pages, so we should keep the 60 limit rate for the diff box to avoid a disaster.

If I understand correctly the code in this PR the API is called only for pending test (fenced by `if not run['finished'] and not run['approved']:`), so we could merge it safely. An obsessive/compulsive approver could hit his 60/h limit lurking test pages and then he will not be able to approve tests or, worse, he will be able to approve some test without the master diff check.

tomtor:
The api is only called when a new test is created (quick check from my phone) so hitting the limit is no issue.

PR looks good to me.

tomtor:
Just had a closer look. Api is called when viewing. Why can't it be called on test creation?

linrock:
@tomtor good call, the master diff can never change after creating a test? so i think making one API request when creating a new test and then saving the result in the run attributes makes sense

@vondele what do you think? this would prevent server DoS from exhausting the API rate limit by people viewing pending test pages. then there's only possible DoS during test creation, which is already a risk

side note - the tune branch is really just this one commit, right?
https://github.com/official-stockfish/Stockfish/commit/4a7bda

if it's an important part of testing, perhaps it is worth considering working that into master.

vondele:
@linrock yes that seems like a good idea, but I'm happy to pass on the technical aspects and let @ppigazzini and @tomtor judge.

I would be most concerned about the viewing not so much about new tests. (New tests are behind the required login, which won't stop anybody who really wants to break it, but makes it less likely to happen by accident).

Merging the tune in master is an option to consider, this was also discussed in the past.

linrock:
@tomtor @ppigazzini could you guys review this again? i changed it so that the API call is only made once upon test creation, as tomtor suggested.

vondele:
Another remark (sorry always after the fact :-). If we store diffs server size do we need possibly be concerned about very large diffs e.g. when people add data files?
https://tests.stockfishchess.org/tests/view/5e6508f8e42a5c3b3ca2e2d2
I see the diff being returned by github is actually reasonable, but we might want to limit what we store server side.

linrock:
right now in this PR, for checking the master diff, it only fetches the diff and then stores True/False depending on whether the diff is empty or not.

would it be useful to store the master diff at all? if so, a max size limit would be good.

vondele:
OK, thanks for explaining... I was thinking about the diff of the test wrt to master that is now shown on the test page.

linrock:
ah i see, sounds like it'd be useful to store those diffs. i imagine if the diffs for all tests ever were searchable, that would be helpful for planning tests. see https://github.com/glinscott/fishtest/issues/339

vondele:
searchable diffs would be one thing, but the git commit comments (not github comments) on these diffs would be almost more important to be searchable. That would potentially be a really useful feature BTW (assuming people start writing better commit comments, which is a bit a chicken and egg thing).

ppigazzini:
DEV updated, I added/approved some tests w/ and w/o the official master, it seems to work fine.
Please double check, as usual you can use (user01/user01) to add a new test and use (user00/user00) to approve a test.

tomtor:
@linrock Thanks for moving the API call to the test creation!

Looks good, but I do not see the message in https://dfts-0.pigazzini.it/tests/view/5e8ded5e4420e2e736e2aae6

I think you could remove:

    %if not run['finished'] and not run.get('approved', False):

so it is also shown after approving

linrock:
@tomtor cool, i've updated this PR so that the message is shown for approved test runs as well. it will only show for test runs created after this PR, since old test runs won't have that `base_same_as_master` key set on the run

tomtor:
@linrock Nice job! @ppigazzini merging is OK!

ppigazzini:
PROD updated, great job @linrock :)

vondele:
@linrock transforming the fishtest experience... much appreciated!

Issue #563, opened 2020-04-07
Incorrect # finished tests on user pages


linrock:
If you go to a user's page, it incorrectly shows a large number of finished tests. However, if you go to a high page number, it shows that the actual number of tests is a lot smaller.

---
https://tests.stockfishchess.org/tests/user/protonspring
<img src="https://user-images.githubusercontent.com/208617/78704890-9a96d000-78da-11ea-86bf-5dee0a3f71ad.png" width="300"/>

---
https://tests.stockfishchess.org/tests/user/protonspring?page=1457
<img src="https://user-images.githubusercontent.com/208617/78704920-a8e4ec00-78da-11ea-8c67-734fa5b936f6.png" width="300"/>

tomtor:
@linrock Yes, this is a known issue, introduced by various performance optimizations a few months ago when fishtest experienced performance issues.

The issue is caused (and performance is increased) by having an `old_runs` collection and `runs` collection, see eg 

https://github.com/glinscott/fishtest/pull/434

It probably can be fixed with a little bit of additional (MongoDb query) code, at the cost of a slowdown for access to page 1.

linrock:
thanks for the info! i find it very surprising that there are performance problems for these pages with only 70k documents between `runs` and `old_runs` since each page only shows a limited number of runs. i'd expect a single `runs` collection to handle millions of documents without issue and without needing to move anything to `old_runs`.

i see you mentioned in https://github.com/glinscott/fishtest/pull/433 that there was a situation where the entire `runs` collection is read, which would definitely be bad. i'll look further into this

tomtor:
@linrock  `old_runs` was introduced a few years ago by Gary because fishtest got suddenly slower. I think the working_set of the tables/indices exceeded physical memory at that time. 

Now MongoDb is a much newer version and machine is bigger. `old_runs` should not be needed when queries and indices are all OK.

ppigazzini:
@tomtor @linrock should we try to move all data from `old_runs` to `runs` on DEV and run some benches?

ppigazzini:
@linrock PROD ran until November 2019 on CentOS 6.10, Python 2.6.6, MongoDB 2.4.4 (not so stable with many data...)

linrock:
@ppigazzini yea that sounds good. i'll prepare a PR that switches to using the `runs` collection exclusively. then while using that PR, you could try migrating the data from `old_runs` to `runs` in dev and see how well it works.

PR #570, opened 2020-04-08
Fix slow queries due to missing MongoDB indexes

linrock:
Some MongoDB queries are really slow because they don't use indexes, so they have to scan an entire collection (runs, old_runs, actions). Slow queries also degrade performance for other people using the site at the same time.

@tomtor this adds indexes to `old_runs` as you suggest here https://github.com/glinscott/fishtest/issues/563#issuecomment-610868360. It looks like there were no indexes on that collection at all, so queries being sent to `old_runs` would be slow.

@ppigazzini if you check out this branch in DEV, you'll have to use `fishtest/utils/create_indexes.py` to set up the indexes. My dev server has almost no data on it, so I don't have a good way of verifying the performance improvement.

Examples of really slow pages that should be fixed:
- https://tests.stockfishchess.org/actions?action=modify_run&user=Vizvezdenec
- https://tests.stockfishchess.org/tests?page=50&success_only=1
- https://tests.stockfishchess.org/tests?page=50&ltc_only=1

### DIFF
--- a/fishtest/utils/create_indexes.py
+++ b/fishtest/utils/create_indexes.py
@@ -23,18 +23,52 @@ runs = db['runs']
 pgns = db['pgns']
 
 
-indexList = { 'runs' : [ [(u'args.username', ASCENDING)],
-                         [(u'finished', ASCENDING), (u'last_updated', DESCENDING), (u'start_time', DESCENDING)] ],
-                         # [(u'finished', ASCENDING), (u'last_updated', DESCENDING)],
-                         # [(u'deleted', ASCENDING)],
-                         # [(u'deleted', ASCENDING), (u'finished', ASCENDING)],
-                         # [(u'last_updated', DESCENDING), (u'start_time', DESCENDING)],
-                         # [(u'tasks.pending', ASCENDING)],
-              'pgns' : [ [('run_id', ASCENDING)] ],
-              'users' : [ [('username', ASCENDING)] ],
-              'actions' : [ [('username', ASCENDING)] ],
-            }
-uniqueList = [ 'username' ]
+index_list = {
+  'runs' : [
+    [
+      # Index used for querying combos of username / Greens / LTC
+      ('finished', DESCENDING),
+      ('args.username', ASCENDING),
+      ('args.tc', ASCENDING),
+      ('results_info.style', ASCENDING),
+      ('last_updated', DESCENDING),
+    ],
+    [
+      # Index used for get_unfinished_runs()
+      ('finished', ASCENDING),
+      ('last_updated', DESCENDING),
+      ('start_time', DESCENDING),
+    ],
+    # [(u'deleted', ASCENDING)],
+    # [(u'deleted', ASCENDING), (u'finished', ASCENDING)],
+    # [(u'last_updated', DESCENDING), (u'start_time', DESCENDING)],
+    # [(u'tasks.pending', ASCENDING)],
+  ],
+  'old_runs' : [
+    [
+      # Index used for querying combos of username / Greens / LTC only
+      ('finished', DESCENDING),
+      ('args.username', ASCENDING),
+      ('args.tc', ASCENDING),
+      ('results_info.style', ASCENDING),
+      ('_id', DESCENDING),
+    ],
+  ],
+  'pgns' : [
+    [('run_id', ASCENDING)]
+  ],
+  'users' : [
+    [('username', ASCENDING)]
+  ],
+  'actions' : [
+    [
+      ('action', ASCENDING),
+      ('username', ASCENDING),
+      ('_id', DESCENDING)
+    ]
+  ],
+}
+unique_list = [ 'username' ]
 
 
 def printout(s):
@@ -42,9 +76,9 @@ def printout(s):
   sys.stdout.flush()
 
 
-# Loop through all collections and recreate indexes if given in indexList
+# Loop through all collections and recreate indexes if given in index_list
 for cn in db.list_collection_names():
-  if cn in indexList:
+  if cn in index_list:
     c = db[cn]
 
     # Drop all indexes on collections cn except _id_
@@ -54,12 +88,12 @@ for cn in db.list_collection_names():
         printout("Dropping " + cn + " index " + idx + " ...")
         c.drop_index(idx)
 
-    # Create indexes in indexList
+    # Create indexes in index_list
     printout("")
-    for flds in indexList[cn]:
+    for flds in index_list[cn]:
       printout("Creating " + cn + " index " + str(flds) + " ...")
       uniq = False
-      if cn != 'actions' and flds[0][0] in uniqueList:
+      if cn != 'actions' and flds[0][0] in unique_list:
           uniq = True
       c.create_index(flds, unique=uniq)
 
@@ -71,4 +105,3 @@ for cn in db.list_collection_names():
   pprint.pprint(c.index_information(), stream=None, indent=2, width=110, depth=None)
 
 printout("")
-
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -16,7 +16,6 @@ from pymongo import MongoClient, ASCENDING, DESCENDING
 
 from fishtest.userdb import UserDb
 from fishtest.actiondb import ActionDb
-from fishtest.views import format_results
 
 import fishtest.stats.stat_util
 


fishtest/utils/create_indexes.py
```diff
@@ -23,28 +23,62 @@
...
+    [('run_id', ASCENDING)]
+  ],
+  'users' : [
+    [('username', ASCENDING)]
+  ],
+  'actions' : [
+    [
+      ('action', ASCENDING),
+      ('username', ASCENDING),
+      ('_id', DESCENDING)
```
tomtor, line 67: Index on `_id` is always present by  default

fishtest/utils/create_indexes.py
```diff
@@ -23,28 +23,62 @@
...
+    # [(u'tasks.pending', ASCENDING)],
+  ],
+  'old_runs' : [
+    [
+      # Index used for querying combos of username / Greens / LTC only
+      ('finished', DESCENDING),
+      ('args.username', ASCENDING),
+      ('args.ltc', ASCENDING),
+      ('results_info.style', ASCENDING),
+      ('_id', DESCENDING),
```
tomtor, line 54: `_id` index is default

ppigazzini:
DEV updated, LTC pages are still slow (no pun intended), other pages got a good speedup

* https://tests.stockfishchess.org/tests?page=50&success_only=1
* https://dfts-0.pigazzini.it/tests?page=50&success_only=1
* https://tests.stockfishchess.org/tests?page=50&ltc_only=1
* https://dfts-0.pigazzini.it/tests?page=50&ltc_only=1
* https://tests.stockfishchess.org/actions?action=modify_run&user=Vizvezdenec
* https://dfts-0.pigazzini.it/actions?action=modify_run&user=Vizvezdenec

Here some benches `$ time bash curl_tests.sh` with this script:
```bash
#!/bin/bash
url_base="https://dfts-0.pigazzini.it/tests?page="
n=55 # max 142
for ((i=50; i<${n}; ++i)); do
  echo $((i))
  for ((j=0; j<10; ++j)); do
    curl -L "${url_base}$((i*10+j))" &> /dev/null &
    # curl -L "${url_base}$((i+j))&ltc_only=1" &> /dev/null &
    # curl -L "${url_base}$((i+j))&success_only=1" &> /dev/null &
  done
  wait
done
```

* bench master
```bash
# ${url_base}$((i*10+j))
real    0m27.602s
user    0m1.047s
sys     0m2.719s

# ${url_base}$((i+j))&ltc_only=1
real    0m44.249s
user    0m1.344s
sys     0m2.625s

# ${url_base}$((i+j))&success_only=1
real    0m43.586s
user    0m1.234s
sys     0m2.375s
```

* bench PR
```bash
# ${url_base}$((i*10+j))
real    0m17.010s
user    0m0.922s
sys     0m2.688s

# ${url_base}$((i+j))&ltc_only=1
real    1m15.270s
user    0m1.109s
sys     0m2.828s

# ${url_base}$((i+j))&success_only=1
real    0m6.359s
user    0m0.984s
sys     0m2.703s
```

tomtor:
I noticed on dev that the finished list does no longer have 50 entries?

fishtest/utils/create_indexes.py
ppigazzini: @tomtor my bad but I don't understand the type of correction that you are suggesting:
 1. delete `('_id', DESCENDING),`
 2. write `('_id', DEFAULT)`
 3. something else

ppigazzini:
@tomtor I restored the original DB to check a bug in another PR, see https://github.com/glinscott/fishtest/pull/571#issuecomment-612003751

fishtest/utils/create_indexes.py
tomtor: I meant 1.

linrock:
@ppigazzini thanks for the info! my bad, i had a typo in the index used for LTC pages which is why that query was slow. Try now and you should see a significant speedup.

fishtest/utils/create_indexes.py
ppigazzini: @linrock please delete `('_id', DESCENDING),`

linrock: there's a single-field `_id` index on collections by default yes, but it can't be used here.

`_id` needs to be on this compound index since the old_runs query filters by a few fields, then sorts by `_id`.

Try running this example code with and without `_id` in the index and you will see the performance difference:

```python
from pprint import pprint
from pymongo import MongoClient, DESCENDING

def print_query_plan(mongo_cursor):
    explanation = mongo_cursor.explain()
    winning_plan = explanation['queryPlanner']['winningPlan'];
    if winning_plan:
        # Query starts with an index scan
        pprint(winning_plan)
    else:
        # Slow query - collection scan
        pprint(explanation['queryPlanner'])

conn = MongoClient()
old_runs_col = conn['fishtest_new']['old_runs']
print_query_plan(old_runs_col.find({
  'finished': True,
  'args.tc': {'$regex': '^([4-9][0-9])|([1-9][0-9][0-9])'}
}, sort=[('_id', DESCENDING)]))
```

With `_id` in the index, the query is fast (entries fetched already-sorted from index):
`IXSCAN -> SORT_KEY_GENERATOR -> SORT -> FETCH`

Without `_id` in the index, the input stages in the query plan is slow (fetched, then sorted):
`IXSCAN -> FETCH -> SORT_KEY_GENERATOR -> SORT`

linrock: @ppigazzini you can try benchmarking with and without `_id` on this index, and you will find that with `_id` on the index, you will see a significant performance boost 😉

fishtest/utils/create_indexes.py
linrock: same as https://github.com/glinscott/fishtest/pull/570#discussion_r406834968. `_id` needs to be in this compound index since the actions queries all sort by `_id` after filtering.

Example code:
```python
from pprint import pprint
from pymongo import MongoClient, DESCENDING

def print_query_plan(mongo_cursor):
    explanation = mongo_cursor.explain()
    winning_plan = explanation['queryPlanner']['winningPlan'];
    if winning_plan:
        # Query starts with an index scan
        pprint(winning_plan)
    else:
        # Slow query - collection scan
        pprint(explanation['queryPlanner'])

conn = MongoClient()
print_query_plan(conn['fishtest_new']['actions'].find({
  'action': 'updated_stats',
  'username': 'username'
}, sort=[('_id', DESCENDING)]))
```

With `_id` in the index this fetches entries already-sorted using an index scan (fast):
```
{'inputStage': {'direction': 'forward',
                'indexBounds': {'_id': ['[MaxKey, MinKey]'],
                                'action': ['["updated_stats", '
                                           '"updated_stats"]'],
                                'username': ['["username", "username"]']},
                'indexName': 'action_1_username_1__id_-1',
                'indexVersion': 2,
                'isMultiKey': False,
                'isPartial': False,
                'isSparse': False,
                'isUnique': False,
                'keyPattern': {'_id': -1, 'action': 1, 'username': 1},
                'multiKeyPaths': {'_id': [], 'action': [], 'username': []},
                'stage': 'IXSCAN'},
 'stage': 'FETCH'}
```

Without `_id` in the index this fetches entries first, then sorts them separately (slow):
```
{'inputStage': {'inputStage': {'inputStage': {'direction': 'forward',
                                              'indexBounds': {'action': ['["updated_stats", '
                                                                         '"updated_stats"]'],
                                                              'username': ['["username", '
                                                                           '"username"]']},
                                              'indexName': 'action_1_username_1',
                                              'indexVersion': 2,
                                              'isMultiKey': False,
                                              'isPartial': False,
                                              'isSparse': False,
                                              'isUnique': False,
                                              'keyPattern': {'action': 1,
                                                             'username': 1},
                                              'multiKeyPaths': {'action': [],
                                                                'username': []},
                                              'stage': 'IXSCAN'},
                               'stage': 'FETCH'},
                'stage': 'SORT_KEY_GENERATOR'},
 'sortPattern': {'_id': -1},
 'stage': 'SORT'}
```

fishtest/utils/create_indexes.py
tomtor: Thanks for looking into this. I am a bit allergic to indices which are used infrequently, because they need to be updated and use space (disk/memory), but these are probably fine.

ppigazzini:
@linrock 
Speedup for LTC:
* master:
```
real    1m0.239s
user    0m0.938s
sys     0m2.500s
```
* pr:
```
real    0m29.528s
user    0m0.922s
sys     0m2.563s
```

Speedup for actions:
* master:
```
real    0m26.771s
user    0m1.125s
sys     0m2.938s
```
* pr:
```
real    0m20.550s
user    0m1.109s
sys     0m3.156s
```

vondele:
@ppigazzini server crash?

ppigazzini:
@vondele I stopped fishtest to apply the new indexed to mongodb. On DEV the script ran for 15 seconds, pgns collection on PROD is taking very much time.

vondele:
OK... thanks.

ppigazzini:
a watched pot never boils :(

vondele:
eventually it does :-)

ppigazzini:
I closed the eyes and the pot boiled instantly :)

ppigazzini:
PROD updated, thank you @linrock (also for the fast indexing tutorial :) )

PR #572, opened 2020-04-09
Update layout of 'create new test' page

linrock:
@xoto10 suggested ways to improve the [create new test page](https://tests.stockfishchess.org/tests/run)  (see https://github.com/glinscott/fishtest/pull/556#issuecomment-610750254)

Here is how the page would look with this PR:

image

---

This is how the page currently looks in master:

image

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -1,187 +1,249 @@
 <%inherit file="base.mak"/>
 
-<% re_run = len(args) > 0 %>
+<style>
+  input[type=number].no-arrows::-webkit-inner-spin-button,
+  input[type=number].no-arrows::-webkit-outer-spin-button {
+      -webkit-appearance: none;
+      -moz-appearance: none;
+      appearance: none;
+      margin: 0;
+  }
+
+  input[type=number] {
+    -moz-appearance: textfield;
+  }
+
+  .flex-row {
+    display: flex;
+    align-items: center;
+    margin: 10px 0;
+  }
+
+  .field-label {
+    font-size: 12px;
+    margin: 0;
+    text-align: right;
+    padding-right: 15px;
+    width: 100px;
+  }
+
+  .field-label.leftmost {
+    width: 75px;
+    flex-shrink: 0;
+  }
+
+  .rightmost {
+    margin-left: auto;
+  }
+
+  .third-size {
+    width: 107px;
+    flex-shrink: 0;
+  }
+
+  input.quarter-size {
+    margin-right: 10px;
+    width: 70px;
+    flex-shrink: 0;
+  }
+
+  #create-new-test {
+    width: 700px;
+    margin: 7px auto;
+    padding-right: 30px;
+  }
+
+  #create-new-test input, #create-new-test select {
+    margin: 0;
+  }
+
+  .quarter-size {
+    width: 80px;
+    flex-shrink: 0;
+  }
+
+  .choose-test-type .btn {
+    width: 75px;
+  }
+
+  #create-new-test label:hover {
+    cursor: text;
+  }
+
+  #create-new-test textarea {
+    width: 100%;
+    min-height: 40px;
+    margin: 0;
+  }
+
+  section.test-settings input {
+    width: 235px;
+  }
+</style>
 
 <legend>Create New Test</legend>
-<div>
+
+<section class="instructions" style="margin-bottom: 35px">
   Please read the
   <a href="https://github.com/glinscott/fishtest/wiki/Creating-my-first-test">Testing Guidelines</a>
   before creating your test.
-</div>
-<br><br>
+</section>
 
-<form class="form-horizontal" action="${request.url}" method="POST">
-  <div class="control-group">
-    <label class="control-label">Test branch:</label>
-    <div class="controls">
-      <input name="test-branch" value="${args.get('new_tag', '')}" ${'readonly' if re_run else ''}>
+<form id="create-new-test" action="${request.url}" method="POST">
+  <section class="test-settings" style="margin-bottom: 35px">
+    <div class="flex-row">
+      <label class="field-label leftmost">Test type</label>
+      <div class="btn-group choose-test-type">
+        <div class="btn" id="fast_test"
+             data-options='{"tc": "10+0.1", "threads": 1, "options": "Hash=16", "bounds": "standard STC"}'>
+          short (STC)
+        </div>
+        <div class="btn" id="slow_test"
+             data-options='{"tc": "60+0.6", "threads": 1, "options": "Hash=64", "bounds": "standard LTC"}'>
+          long (LTC)
+        </div>
+        <div class="btn" id="fast_smp_test"
+             data-options='{"tc": "5+0.05", "threads": 8, "options": "Hash=64", "bounds": "standard STC"}'>
+          SMP (STC)
+        </div>
+        <div class="btn" id="slow_smp_test"
+             data-options='{"tc": "20+0.2", "threads": 8, "options": "Hash=256", "bounds": "standard LTC"}'>
+          SMP (LTC)
+        </div>
+      </div>
+      <button type="submit" class="btn btn-primary rightmost" style="width: 180px">Submit test</button>
+    </div>
+
+    <div class="flex-row">
+      <label class="field-label leftmost">Test branch</label>
+      <input type="text" name="test-branch" id="test-branch"
+             value="${args.get('new_tag', '')}" ${'readonly' if is_rerun else ''}>
+
+      <label class="field-label">Base branch</label>
+      <input type="text" name="base-branch"
+             value="${args.get('base_tag', 'master')}" ${'readonly' if is_rerun else ''}>
+    </div>
+
+    <div class="flex-row">
+      <label class="field-label leftmost">Test signature</label>
+      <input type="number" name="test-signature" class="no-arrows"
+             placeholder="Defaults to last commit message"
+             value="${args.get('new_signature', '')}" ${'readonly' if is_rerun else ''}>
+
+      <label class="field-label">Base signature</label>
+      <input type="number" name="base-signature" class="no-arrows"
+             value="${args.get('base_signature', bench)}" ${'readonly' if is_rerun else ''}>
+    </div>
+
+    <div class="flex-row">
+      <label class="field-label leftmost">Test options</label>
+      <input type="text" name="new-options"
+             value="${args.get('new_options', 'Hash=16')}">
+
+      <label class="field-label">Base options</label>
+      <input type="text" name="base-options"
+             value="${args.get('base_options', 'Hash=16')}">
+    </div>
+
+    <div class="flex-row">
+      <label class="field-label leftmost">Notes</label>
+      <textarea name="run-info" placeholder="Defaults to commit message"
+                rows="3">${args.get('info', '')}</textarea>
+    </div>
+
+    <div class="flex-row">
+      <label class="field-label leftmost">Test repo</label>
+      <input type="text" name="tests-repo" style="width: 100%;"
+             value="${args.get('tests_repo', tests_repo)}" ${'readonly' if is_rerun else ''}>
+    </div>
+  </section>
+
+  <section id="stop-rule" style="min-height: 130px">
+    <div class="flex-row">
+      <label class="field-label leftmost">Stop rule</label>
       <div class="btn-group">
-        <div class="btn" id="fast_test">short (STC)</div>
-        <div class="btn" id="slow_test">long (LTC)</div>
-        <div class="btn" id="fast_smp_test">SMP (STC)</div>
-        <div class="btn" id="slow_smp_test">SMP (LTC)</div>
+        <div class="btn btn-info" data-stop-rule="sprt" style="width: 94px">SPRT</div>
+        <div class="btn" data-stop-rule="numgames" style="width: 100px">Num games</div>
+        <div class="btn" data-stop-rule="spsa" style="width: 94px">SPSA</div>
+      </div>
+      <input type="hidden" name="stop_rule" id="stop_rule_field" value="sprt" />
+
+      <div class="rightmost" style="display: flex; align-items: center">
+        <label class="field-label stop_rule spsa numgames"
+               style="${'display: none' if (args.get('sprt') or not is_rerun) else ''}"># games</label>
+        <input type="number" name="num-games" min="1" class="stop_rule spsa numgames third-size"
+               value="${args.get('num_games', 60000)}"
+               style="${'display: none' if (args.get('sprt') or not is_rerun) else ''}" />
       </div>
     </div>
-  </div>
-  <div class="control-group">
-    <label class="control-label">Test repo:</label>
-    <div class="controls">
-      <input name="tests-repo" value="${args.get('tests_repo', tests_repo)}" ${'readonly' if re_run else ''}>
-    </div>
-  </div>
-  <div class="control-group">
-    <label class="control-label">Test options:</label>
-    <div class="controls">
-    <input name="new-options" value="${args.get('new_options', 'Hash=16')}">
-    </div>
-  </div>
-  <div class="control-group">
-    <label class="control-label">Test signature:</label>
-    <div class="controls">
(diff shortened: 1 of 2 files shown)


vondele:
looks nice, but I want to suggest some changes based on the workflow we typically have. 

Let me try if I can modify/annotate the image...

vondele:
suggest

would be my suggestion (but feel free to to adjust). It puts the 4 common fields (submit, test branch, long LTC and Notes) near the upper half of the page.

vondele:
maybe the four buttons could be under the repo name...

linrock:
@vondele here are some mockups i made after thinking about your suggestions. any thoughts? i suppose submit could be near the top so that people can submit quickly without having to go through the entire form for options they may not need. i realized that SPSA has a lot of fields so i tried to figure out a way to fit that in.

i'll go look through more actual tests later to get better ideas of how to improve this. fitting the design with the typical workflows would be great.

---

image

---

image

vondele:
@linrock thanks! 

I think the most common workflow is for an established dev. He only need to fill 'Test branch' and press enter (no need for submit button, test  repo is prefilled, all other fields have the right default). That's why I think the 'Test branch' field should be easiest to find, in these variants it is 'somewhat hidden'. 

'Test repo' is really something one hardly ever touches. A new user should find the field once, after that, it remains prefilled forever, and essentially there is no need to touch it (very rare exception, test branches of somebody else, not using the 'reschedule' button on individual test page'. It could probably be moved under 'Notes'.

worker threads, and time control, as well as Test options Hash change frequently, but with well defined rules. These fields should presumably almost never be touched manually, but only changed with the 4 buttons. At least for SPRT. So actually, the 'Worker' row could have the four boxes 'Time control' (or TC if you need a shortcut), 'Threads', 'Priority', and 'Throughput'. There could be a Book row, with 'file' and 'Book depth', both need changing very rarely. 

Let me, based on this description and your suggestion, try to make a new iteration of it.

For SPSA, this looks good to me, I would not anymore prefill the 'SPSA parameters' box, this doesn't help, first thing one needs to do is delete that content, before filling it without something new. I would also swap 'SPSA clipping' and 'SPSA Alpha' boxes. As far as I know clipping and rounding are really left-overs and could be removed from the form, fixed to their defaults. @ppigazzini might have an opinion on this.

vondele:
suggest_2

a copy & paste update with suggestions of new locations...

ppigazzini:
> As far as I know clipping and rounding are really left-overs and could be removed from the form, fixed to their defaults. @ppigazzini might have an opinion on this.

Those are (somehow interesting https://github.com/glinscott/fishtest/commit/7eebda7e6d1f47f2672aefe46db35baee7cb5b1f, https://github.com/glinscott/fishtest/commit/5f63500db3f40569ea406a8b8b4b987f054ee79f) theoretical features that nobody use. I will try a couple of fast SPSA to verify if they are able to change the behavior of a SPSA run. In case of not measurable effects IMO we could simplify the SPSA code keeping only one way to make things.

https://github.com/glinscott/fishtest/blob/5b07986dab3e638292cd04d6cf95d89d9959faeb/fishtest/fishtest/rundb.py#L606-L623

> 'Test repo' is really something one hardly ever touches.
> These fields should presumably almost never be touched manually

@vondele take on responsibility as fishtest tester for a couple of hours before making so bold statements ;)

vondele:
@ppigazzini of course, your expert use is quite different (and much appreciated), but you'll agree that we should optimize for the user here ;-)

vondele:
My suggestion for these two SPSA params actually comes from https://github.com/glinscott/fishtest/wiki/Creating-my-first-test
In particular:
```
Leave "SPSA clipping" and "SPSA rounding" at their defaults ("old" and "deterministic") unless you have tried the defaults already and seen poor convergence. These are experimental options which should be used with care; for more details please see https://groups.google.com/forum/#!topic/fishcooking/LT55ExR0m9U
```
I can't really recall anybody using anything else but the defaults in recent times, but I might have overlooked things.

ppigazzini:
> @ppigazzini of course, your expert use is quite different (and much appreciated), but you'll agree that we should optimize for the user here ;-)

I contested only the statements correctness, they should be typed in this way:
 - 'Test repo' is really something only one person touches more that once
 - These fields are surely manually touched (over, over, over...) only by one person
:D

ppigazzini:
@vondele I'm to drop now the "clipping" and "rounding" fields in the UI.
I will make some test on DEV (modifying the code by hand) and then I will simplify the SPSA code keeping only one method for "clipping" and one method for "rounding".

OT: IMO the know how is, unfortunately, in the code/commit, most parts of the wiki were written ex post, sometimes with no so big knowledge of the issue.

snicolet:
The general idea behind this pull request is great, thanks @linrock for opening it !

One detail: I much prefer to have the triplets "branch/signature/options" in columns (for "Tested branch" and "Base branch") in rather than in row. In other words, I prefer the layout of these six fields in https://github.com/glinscott/fishtest/pull/572#issuecomment-611655658 over https://github.com/glinscott/fishtest/pull/572#issuecomment-611903907 .

No opinion about the rest of the discussion, I'm fine with everything that has been said :-)

vondele:
the following could be along the lines of @snicolet suggestion, would be OK with me as well:

newtest_design_3

(possible improvemnents: 'worker' line could be two columns wide?)

linrock:
based on the info so far, i think these are some priorities for the design of the form:
1. it should be fast for established devs to create a test
2. new devs should find the form easy to use
3. the layout is simple enough to work on mobile

for (1), the workflow could be:
- when the dev lands on the page, the `test branch` field is focused immediately
  - since this is the only field they may need, it'd be the first field they see (in top-left)
- the dev then just needs to type/paste the branch name and then press enter to submit

for (2), i think the important thing is for the fields to be organized in logical groups
- the top section is for the test vs. base params that they'd have to fill out at least once
- the worker section, they might not need to edit at all

for (3), maybe just sticking to mobile design principles
- the submit button sticks to the top right of the page since the top right is a very common location for a CTA on mobile forms
- if the user scrolls down the page, they can still conveniently press submit

---
here's another design concept. also, i was wondering whether the tests options and base options are always the same. if so, maybe that field could be a single field in the worker section:



---
this is how this first section would look with the test/base fields in columns as @snicolet suggested:

vondele:
@linrock test options and base options have to be modified only exceptionally, but they do not need to be the same. So, we need both input fields.

xoto10:
Maybe just "Keep It Simple" and stick with the current layout, tall and thin should work fine on even small mobile screens? Just make related fields a bit closer together as in my original request and move the Submit button to the top of the page?
Just my 2p.

On a PC many/most STC tests can be submitted by simply entering the test branch and pressing return, so the button not even needed, but I guess it is required for mobile users.

If we could make the buttons keyboard navigable that would be a nice option from my point of view. No downside to that as far as I can see.

vondele:
I think we're kind of converging, the last screenshot (second in https://github.com/glinscott/fishtest/pull/572#issuecomment-612637267) would work for me. I would put the '4 button bar' closer to 'submit' since those two are needed most commonly for a reschedule/resubmit.... but I think that these are small improvements on the edge of bike-shedding.

I'd guess few people will submit new tests via mobile, as few people will be coding on their phone, possibly with the exception of rescheduling a test at ltc after passing stc. But obviously it wouldn't harm if it looks nice on a phone as well. Having the main page look good on a phone would be nice.

linrock:
yup, sounds like we're almost there. here's how it looks with the '4 button bar' up closer to the submit button. i'm thinking if the user clicks the '4 button bar', the 'test branch' field can remain in focus if it's blank so users can still just type the branch name and press enter to submit the test.

@xoto10 what do you mean by making the buttons keyboard navigable? as in certain keys would be keyboard shortcuts for activating buttons?

---

vondele:
that looks good to me... SPSA parameters should not be prefilled IMO.

linrock:
sounds good, i'll remove the pre-filled SPSA parameters and work with this direction then 👍

snicolet:
Would it be possible to label the text field on the left instead of on the top?

For instance this version you proposed had labels on the left: https://github.com/glinscott/fishtest/pull/572#issue-401527106

The reason I ask this is because on a laptop screen it is very inconvenient to have to scroll down the page to access the fields at the bottom of the page, so it is better to keep the page as compact as possible.

I understand that maybe you put the labels above the text fields to save width for the phone version, but as Joost said, the phone version will be accessed much less frequently than the laptop version so the phone version would have lower priority, IMHO. And phone usage will be mainly resubmitting a passed test at LTC.

Thanks for all the hard work :-)

linrock:
that makes sense. so more like this? this fits on my laptop screen without scrolling (1440 x 900)


image

vondele:
to me that looks very nice

xoto10:
@linrock 
Not hot-keys. I just meant so that you can tab through the fields _and_ the buttons. At the moment I can tab through the fields to enter / change them, but the buttons don't receive focus, e.g. I'd like to be able to tab to the LTC button. I think that is something that can be altered?

linrock:
@xoto10 i see, yes that can be altered so those buttons receive focus when pressing tab.

so a workflow could be:
- `test branch` field is focused when you land on the page
- fill out `test branch`
- press tab twice to select `LTC`
- press space to chooose `LTC`
- press enter to submit

ppigazzini:
@linrock IMHO the title "Worker" should be dropped in your latest draft: "Priority" and "Throughput" are queue properties and "Hash" is a worker property.

linrock:
i've updated this PR with the code for the latest design changes. this is how it currently looks:

image

snicolet:
Looks good (and great!) to me.

vondele:
yep looks great.  Let's see on DEV if feels equally good.

vondele:
playing on DEV, the layout is really nice to use IMO.
Two minor nits:
* the error message popup (`Cannot find branch in developer repository`, there might be others) is green (like master), but would be nice to have it red or so.
* the default for 'number of games' in num games and SPSA tests should be 60000 not 500000. The 500000 was mentioned but this should only apply to SPRT tests. However, these seem to be on their old default of 250000 (this number can only be seen after submitting the test on the test page, which is OK, no need for the submitter to set this).

vondele:
and a third comment, have you considered having the form centered on the page rather than left aligned ? On a wide screen (useful for seeing the homepage or the test individual page), there is a lot of which space on the right (which needs to be crossed after pressing reschedule on the test individual page). I have no strong opinion on left aligned of centered being better ... IMO up to @linrock.

ppigazzini:
DEV updated.

To be very nitpicking I would like the static content block ("Threads", "TC", "Priority" etc) just under the "Stop Rule" in order to minimize the moving content when switching between "SPRT" "Num games" "SPSA"

vondele:
@ppigazzini alternatively one could consider to  keep those last lines where they are, but dynamically adjust the whitespace under the stop rule buttons (i.e. a bit of whitespace for SPRT, that fills once one presses the SPSA), so that the 'static content block' doesn't move as one presses the buttons.

ppigazzini:
@vondele I'm fine with any solution that keep in fixed spots the "always showed content": as tester I frantically push every possible button to catch the bugs :)

ppigazzini:
@linrock on DEV it seems that a SPSA test is created as SPRT test, please double check the code.

linrock:
thanks for the feedback! i've made some changes/fixes in this PR:

- the "stop rule" section now has a fixed min height so fields don't move when stop rule changes
- the default # games for num games and SPSA is now 60,000
- the form section is now centered on the page for wide browser windows
- fixed the bug where the form only submits SPRT test types

the error messages showing up as green is a site-wide problem, so i'll plan to fix that in another PR. i agree it'd be better if they showed up as red while the success messages show up as green.

ppigazzini:
DEV updated :)

vondele:
seems like there is a problem: if I go to the main page, select a finished test, click reschedule, click SMP LTC, click submit ... submit has no effect. If I click the normal STC button, submit works.

ppigazzini:
@linrock rescheduling a SPSA test open the new test page with the SPSA fields already written but with the button "SPRT" selected. Pushing "Save" start a SPRT test

linrock:
thanks for catching those. i updated this PR
- the correct stop-rule field (SPRT, SPSA, # games) should focused for re-runs now
- fixed the bug where LTC stopped working (i broke that when switching some text fields to number fields)
- when a time control in a re-run matches a preset test type, that test type will be highlighted

ppigazzini:
DEV updated.

ppigazzini:
- [ ] "Priority" reject negative values
- [ ] "Test/Base signature" fields have the arrows for +1 -1 (nitpicking here :)

vondele:
I would actually remove the arrows for signature, threads, book depth and priority. These are all fields that should really be changed very rarely, and an accidental click could easily happen.

another change could be to only show book depth field if the book name ends in pgn and not for epd. book depth will be ignored by cutechess for epd (but should be passed as >0). Actually, I see book depth allows 0, it should be >= 1

linrock:
i just pushed to this PR with these changes:
- removed the arrows for test/base signature, threads, book depth and priority
  - these are still numeric fields, so only numbers can be input into them
- book depth field will hide if the user inputs a .pgn book
- book depth min value is now 1
- priority accepts negative values now

vondele:
@linrock book depth should hide for an .epd and be there for a .pgn

ppigazzini:
DEV updated

linrock:
oops, updated to only show book depth field for .pgn books

ppigazzini:
DEV updated for review.

vondele:
> removed the arrows for test/base signature, threads, book depth and priority
> 
>     * these are still numeric fields, so only numbers can be input into them

@linrock this in not removed on DEV right now.... is there a different version of the patch on DEV?

ppigazzini:
@vondele I don't view arrows for those fields.

vondele:
image

ppigazzini:
Untitled

ppigazzini:
@vondele chrome and edge are fine.

vondele:
yes, seems to be a firefox issue.

ppigazzini:
@vondele yes, chrome, edge, opera are fine but I confirm that firefox shows the arrows :(

vondele:
unless @linrock has an easy fix, this shouldn't be a show stopper, I think.

ppigazzini:
WOW opera! Always known never tried :)

linrock:
those arrows should no longer show up on firefox now :)

ppigazzini:
Yes, gone :)

vondele:
yes, looks good.

ppigazzini:
@vondele numeric spinners alert! View the "SPRT Elo0" and "SPRT Elo1" fields after selecting "Custom bound" in "SPRT bounds".
@linrock Terminate that little arrows!

linrock:
ok, the SPRT Elo arrows are gone now too

ppigazzini:
Gone!

ppigazzini:

snicolet:
Congrats :-)

Would it be possible in another pull request to try a version with the title and the sentence "Please read..." centered on the page above the form?

                             Create New Test
      Please read the Testing Guidelines before creating your test.
                      
                              [the form]

linrock:
did you want to be able to try the form in dev, or just see how it looks? i can open a PR if you want. otherwise, here's what it would look like:

---
image

snicolet:
Thanks, this would looks nicer on a large screen, IMHO :-)

linrock:
sounds good, i've opened a PR for the centered text here: https://github.com/glinscott/fishtest/pull/606

ppigazzini:
@vondele I made a quick test on SPSA "careful clipping" and "randomized rounding", IMO: they are theoretically fine (I like the randomized rounding) but they don't have measurable effects wrt other SPSA parameters.
We could simplify the server SPSA code keeping only one way to make those processes. See
https://github.com/glinscott/fishtest/issues/535#issuecomment-616117869

ppigazzini:
> and a third comment, have you considered having the form centered on the page rather than left aligned ? On a wide screen (useful for seeing the homepage or the test individual page), there is a lot of which space on the right (which needs to be crossed after pressing reschedule on the test individual page). I have no strong opinion on left aligned of centered being better ... IMO up to @linrock.

IMHO the centered new_test_page is a design regression.
From a design POV should be left aligned to be consistent with other pages.
From a functional POV on a widescreen add mouse movement both for "New" link and "Reschedule" button.

Issue #564, opened 2020-04-07
estimated time (pending_hours) is wrong


vondele:
Looking in the code, the pending_hours calculation is really hacky (see https://github.com/glinscott/fishtest/pull/562#issuecomment-610547902). Uses hardcoded number 16000 games, which is way off, and depends on the test type etc. We might need a better model if we want to keep it.

PR #573, opened 2020-04-09
Improve pending_hours estimate

vondele:
this should improve the pending hours estimate, by reflecting current average games / test, and not counting tests run so far for sprt tests.
More accurately estimating the remaining number of games for a running sprt test is not obvious.

move this estimate away from the pending tests line, to the top status line, since this is an estimate of all tests, related to the current estimated fishtest performance.

fixes https://github.com/glinscott/fishtest/issues/564

### DIFF
--- a/fishtest/fishtest/templates/tests.mak
+++ b/fishtest/fishtest/templates/tests.mak
@@ -12,6 +12,7 @@
         cores ${'%.2fM' % (nps / (cores * 1000000.0 + 1))} nps
         (${'%.2fM' % (nps / (1000000.0 + 1))} total nps)
         ${games_per_minute} games/minute
+        ${pending_hours} hours remaining
       </span>
       <button id="machines-button" class="btn">
         ${'Hide' if machines_shown else 'Show'}
@@ -62,7 +63,7 @@
   %endif
 
   <h3>
-    Pending - ${len(runs['pending'])} tests ${pending_hours} hrs
+    Pending - ${len(runs['pending'])} tests
     <button id="pending-button" class="btn">
       ${'Hide' if pending_shown else 'Show'}
     </button>
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -1167,12 +1167,16 @@ def tests(request):
 
       def remaining_hours(run):
         r = run['results']
-        expected_games = run['args']['num_games']
         if 'sprt' in run['args']:
-          expected_games = 16000
-        remaining_games = max(0,
-                              expected_games
-                              - r['wins'] - r['losses'] - r['draws'])
+          # current average number of games. Regularly update / have server guess?
+          expected_games = 53000
+          # checking randomly, half the expected games needs still to be done
+          remaining_games = expected_games / 2
+        else:
+          expected_games = run['args']['num_games']
+          remaining_games = max(0,
+                                expected_games
+                                - r['wins'] - r['losses'] - r['draws'])
         game_secs = parse_tc(run['args']['tc'])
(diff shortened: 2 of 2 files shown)


vondele:
As usual, a review and a test run on DEV would be appreciated.

tomtor:
`expected_games` could be computed when page 1 of finished tests on the main `/tests` page is assembled? This would use the most recently finished SPRT games for computing an average.

Here:

https://github.com/glinscott/fishtest/blob/f17dd787c1127c44d577cedf906394883f6c7cc9/fishtest/fishtest/views.py#L1220

we could add in the loop:

        # calculate expected_games average for SPRT runs

and after the loop:

      if do_cache:
        # set the global expected_games

vondele:
I'm a bit unsure if this 'on-the-fly' averaging is a good thing to do right now, I think the largest shortcoming is actually that it is difficult to estimate the remaining games for a SPRT test anyway (depends on current LLR, numbers of games played, etc. I think that was something vdberg wanted to think about). In principle the hardcoded average will only change significantly if we change the SPRT bounds, I used the past 2000 tests to average it.

The current version fixes what was mostly a bug in the old version, namely subtracting tests run so far for SPRT tests. Basically, any SPRT test with more than 16k games was considered done. Now it uses the more reasonable approach that any running SPRT test still needs about 26k games.... that's really how patch authors feel :-)

ppigazzini:
@vondele I will be able to test this on DEV after having iron out some bugs from another PR. There are some conflicts loading also your PR.

vondele:
thanks. Yes, tests.mak conflicts (rather trivially), this can wait till after the other PR is merged, and I can rebase.

ppigazzini:
@vondele rebase please :)

vondele:
@ppigazzini done

ppigazzini:
DEV updated, joined some workers.

ppigazzini:
> any running SPRT test still needs about 26k games.... that's really how patch authors feel :-)

@vondele you are suggesting that (for patch authors) the SPRT time to finish has the memorylessness property like an exponential distribution: I'm very curious to know the opinion of @vdbergh @lantonov

vdbergh:
@ppigazzini  If we knew the true Elo of the test then we could at any time compute an exact ETA. But of course we do not know the true Elo. So we should use an Elo estimate instead. 

The cleanest solution would be to use an Elo prior for the Elo estimate. If we use N(-1,1) then it roughly means that for Elo estimation we preload the test with about 40000 games with a core of 0.49856 and the appropriate draw ratio (for d=0.71 we would have W,D,L=5742, 28400, 5857) and then we calculate the estimated Elo in a naive way (adding the actual trinomial state of the test to preload).

Of course for LTC the Elo prior should be more optimistic.

vondele:
@ppigazzini did you revert again on dev, I don't see the change I was expecting (using https://dfts-0.pigazzini.it/tests )?

I was more referring to `a watched pot never boils` saying, rather than a true mathematical statement.

ppigazzini:
@vondele yes, I want to test the PR in this way:
 - reload db, master, workers fleet, keep the number
 - reload db, pr, workers fleet, keep the number

ppigazzini:
* master
```
7 machines 33 cores 1.07M nps (35.34M total nps) 70 games/minute
Pending - 9 tests 8.4 hrs
```
* PR
```
7 machines 33 cores 1.07M nps (35.29M total nps) 70 games/minute 112.9 hours remaining
```

vondele:
looks good to me. I think that's a more realistic estimate.

ppigazzini:

PR #575, opened 2020-04-10
Protect links to cross-origin destinations

ppigazzini:
Linking another site using `target="_blank"` attribute exposes the site
to performance and security issues:
 - the linked page may run on the same process as the start page
 - the linked page can access the `window` object of the start page

Adding `rel="noopener"` to `target="_blank"` link avoids these issues.

https://web.dev/external-anchors-use-rel-noopener/


ppigazzini:
And now Google Lighthouse rates 100% Best Practice for all our pages :)

PR #576, opened 2020-04-10
Minor sidebar cosmetic update

linrock:
- All sidebar links now show up on one line
- Hide register and login links when already logged-in
- Add Github section for fishtest and stockfish repos

---
Before and after:

### DIFF
--- a/fishtest/fishtest/static/css/application.css
+++ b/fishtest/fishtest/static/css/application.css
@@ -1,3 +1,11 @@
+
+/* Bootstrap responsive override */
+@media (max-width: 767px) {
+  body {
+    padding: 0;
+  }
+}
+
 .elo-results {
   font-size: 12px;
   margin: 2px;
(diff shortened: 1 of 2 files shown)


ppigazzini:
@linrock I found this strange behavior:
- register a new user
- pending (1) in red 
- login as approver
- click on pending (1)
- unblock new user
- always pending (1) in red
- click on pending (1): no waiting user
- pending in black
Until here it works like the master (there is this little glitch that pending requires a second visit to change from red to black) but with the PR:
- login as other user
- pending (1) in red (???)

ppigazzini:
@linrock errata corrige: also master has the same strange behavior

ppigazzini:
PROD updated, thank you @linrock

linrock:
@ppigazzini cool no prob, maybe worth filing an issue for the `pending (1)` bug

vondele:
I've filled an issue for the pending bug. I had seen this before, it is there since quite a long time.

tomtor:
It is caused (introduced) by serving the main page from a different pserv instance.

PR #577, opened 2020-04-11
Enable CSRF protection for POST requests on website

linrock:
Auto-attaches CSRF token fields to POST forms and adds CSRF protection for logout link.

These routes are now protected. Normal users should see no difference.

- /login
- /signup [erratum corrige: /register]
- /logout
- /tests/run
- /tests/modify
- /tests/delete
- /tests/stop
- /tests/approve
- /tests/purge

For a simple example of a CSRF attack in master, put this in an html file locally (`csrf.html`):

```html
<form action="https://tests.stockfishchess.org/logout"></form>
<script>document.querySelector("form").submit()</script>
```

While logged-in to https://tests.stockfishchess.org/ open `csrf.html`. You will be taken to the fishtest website and you'll be logged out. When CSRF protection is enabled, you'll see an error message instead.

Fixes https://github.com/glinscott/fishtest/issues/386

PR #580, opened 2020-04-12
Fix failing tests, use different db for tests

linrock:
this fixes the failing travis tests as @tomtor pointed out in https://github.com/glinscott/fishtest/pull/577.

it also uses a `fishtest_tests` db so running tests in dev/prod will use a different db.


tomtor:
@linrock Thanks!

PR #581, opened 2020-04-12
Handle Show/Hide cookies for cached content

### LINKED ISSUE #579: main page hide/show pending
@linrock one of the recent changes to the main page (couldn't find it back which one exactly), stored the state of the hide/show button for the pending tests. I recall it is supposed to store the state in a cookie, but in my browser (where I use an tab with an autoreload extension), it sometimes flips from one state to the other, hiding/showing these tests upon reload. I haven't been able to figure out a pattern, why/when this happens, it seems more or less random.

tomtor:
Should fix #579 

@linrock You moved the setting of the button states from client-side cookie inspection to server side.

This PR patches up the server cached state, but what is/was the reason to moving this server side?

EDIT: @linrock If you think it's better to restore the old client-side handling instead of this PR or have other suggestions, please let me know.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -1087,18 +1087,23 @@ def tests(request):
 
   full_info = (len(username) != 0 or do_cache)
 
+  def set_show_hide(t):
+    t['machines_shown'] = request.cookies.get('machines_state') == 'Hide'
+    t['pending_shown'] = request.cookies.get('pending_state') == 'Hide'
+    return t
+
   global last_tests, last_time
   if do_cache:
     if time.time() - last_time < cache_time:
-      return last_tests
+      return set_show_hide(last_tests)
     if not building.acquire(last_tests is None):
       # We have a current cache and another thread is rebuilding,
       # so return the current cache
-      return last_tests
+      return set_show_hide(last_tests)
     elif time.time() - last_time < cache_time:
       # Another thread has built the cache for us, so we are done
       building.release()
-      return last_tests
+      return set_show_hide(last_tests)
 
   runs = {'pending': [], 'failed': [], 'active': [], 'finished': []}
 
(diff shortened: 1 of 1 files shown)


linrock:
nice, the reasons for moving to server were:
1. application JS script loading and execution start is async now. before, an inline script would run right after the machines and pending sections are rendered in the DOM to show/hide those sections. with async script execution, those sections might annoyingly flicker from being shown in one state, then hidden some time later. when this show/hide logic is handled from the server, those sections can be rendered right away in the correct state.
2. a possible performance optimization. if the server knows whether or not a section is hidden based on the cookie value sent by the client, the server can do less work and let the client lazy-load info for faster initial pageloads. for example, if the machines section is hidden by a user, the server doesn't need to calculate it at all for the homepage. instead, the first click on the "Show" button could fetch the machines info instead

tomtor:
> nice, the reasons for moving to server were:

@linrock Thanks for the info. Perhaps we could indeed load the machines info on demand in the future...

@ppigazzini It's OK for me to test/merge

ppigazzini:
### Chrome
PROD:
- open the home and push "Show" only for machines (or only for pending tests); opening home in new tabs I got randomly "Show" for machines or for pending tests.
- open the home and push "Show" for machines and for pending tests; opening home in new tabs I got randomly "Show" for machines and pending tests or "Hide" for machines and pending tests.

DEV:
- open the home and push "Show" only for machines (or only for pending tests); opening home in new tabs I got always "Show" only for machines (or only for pending tests).
- open the home and push "Show" for machines and for pending tests; opening home in new tabs I got always "Show" for machines and pending tests.

### Edge
PROD and DEV are like DEV on Chrome

Let me know other things to test.

vondele:
dev seems to work for me.

ppigazzini:
PROD updated, thank you @vondele for the test, thank you @tomtor for the PR :)

PR #586, opened 2020-04-13
Limit max number of worker cores for a SPSA tuning job

tomtor:
EDIT: SPSA tuning jobs with many parameters require additional server resources compared to SPRT tests, due to the recalculation of the parameters after each mini batch of (2 * concurrency) games.

This PR limits the number of cores that are allocated to these jobs to (10000 / sqrt(nr-of-tuning-parameters)).

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -420,7 +420,16 @@ class RunDb:
          and run['args']['threads'] >= min_threads \
          and need_tt <= max_memory:
         task_id = -1
+        cores = 0
+        if 'spsa' in run['args']:
+          limit = 40000 / math.sqrt(len(run['args']['spsa']['params']))
+        else:
+          limit = 1000000  # No limit for SPRT
         for task in run['tasks']:
+          if task['active']:
+            cores += task['worker_info']['concurrency']
+            if cores > limit:
+              break
           task_id = task_id + 1
           if not task['active'] and task['pending']:
             task['worker_info'] = worker_info


vondele:
if this is merged I'm fine with increasing the limit of games to 500k. I believe the main reason to limit to number of games is the load on the server. I vaguely recall there was a performance reason other than the number of cores, but I'm not sure I recall correctly.

BTW, is it the number of cores that is the issue, or is it the number of workers? If it is the number of cores, would it make sense to e.g. batch their results together so that there are fewer api requests to the server (e.g. having an 8 core worker report back every 8 played games would still be reasonable).

tomtor:
Raising number of games increases memory usage of the pserv instances, it is not really related to number of workers/cores.

> BTW, is it the number of cores that is the issue, or is it the number of workers? If it is the number of cores, would it make sense to e.g. batch their results together so that there are fewer api requests to the server (e.g. having an 8 core worker report back every 8 played games would still be reasonable).

It is the number of updates/second which impacts performance. Batching API requests (this is already done for SPSA updates which bundle the result of a number of games) would indeed improve scalability...

vondele:
but the memory increase is not a problem right now, since we have more, and the memory is only allocated if the tests actually has that many games, right?

also, since the introduction of the pentanomial frequencies, the batching makes even more sense, since we need at least 2 games for a single core worker for a first result. One could consider to use batches of size 2 * ncores. If that reduces the load on the server by ~10x or so, we have made good progress, and no doubt find the next bottleneck.

tomtor:
> but the memory increase is not a problem right now, since we have more, and the memory is only allocated if the tests actually has that many games, right?

Yes, not a problem with the current memory,
but also No: all slots are preallocated at the start:
https://github.com/glinscott/fishtest/blob/8f52a54fb8ae7809e5132666ced46666cae94d63/fishtest/fishtest/rundb.py#L47

But removed when the test finishes:
https://github.com/glinscott/fishtest/blob/8f52a54fb8ae7809e5132666ced46666cae94d63/fishtest/fishtest/rundb.py#L577




> One could consider to use batches of size 2 * ncores

If 
https://github.com/glinscott/fishtest/blob/8f52a54fb8ae7809e5132666ced46666cae94d63/fishtest/fishtest/stats/stat_util.py#L185
handles an adjustment of eg 16 played games at once then that would be fine.
If we have to call that function with the individual match results (2 * ncores) times then the gain would be much smaller. I am sure @vdbergh can answer that question.

tomtor:
> and no doubt find the next bottleneck

Many parameter SPSA tests still are an issue, perhaps we should limit the number of cores for these runs related to the number of tuned parameters: eg 10000 / nr-parameters

vondele:
> https://github.com/glinscott/fishtest/blob/8f52a54fb8ae7809e5132666ced46666cae94d63/fishtest/fishtest/stats/stat_util.py#L185
> 
> handles an adjustment of eg 16 played games at once then that would be fine.
> If we have to call that function with the individual match results (2 * ncores) times then the gain would be much smaller. I am sure @vdbergh can answer that question.

@vdbergh should confirm, but to me the code looks like it will just deal fine with it, update_SPRT just takes the match statistics so far. Checking every 16th game will essentially yield the same results as checking every game, especially in our case where 10000s of games are typically played.

vdbergh:
Can somebody explain why update_SPRT could be a bottleneck? I would just like to understand the issue. I was under the impression that the main bottleneck is the database.

ppigazzini:
@vdbergh with very high number of workers the python instance is at 100% CPU, view the `top` in this comment https://github.com/glinscott/fishtest/issues/553#issuecomment-609080979 . You find other info in that issue. I haven't the knowledge to say that update_SPRT is the culprit, though :)

vdbergh:
_If_ update_SPRT is the culprit then it could be related to the switch to using the exact LLR. See #519.

vondele:
@vdbergh but back to the question before. You agree that one could batch the replies of the workers without significant error wrt. to SPRT? It is not just the SPRT CPU load (if it is there), IMO one could remove a lot of overhead by batching things elsewhere.

vdbergh:
@vondele 
I assume by batching you mean calculating the LLR only every 16 games or so.

Of course the expected overshoot will be much higher (currently it is about 0.009). But update_SPRT already does overshoot correction. It is easy to adapt it to batching. 

To measure the exact effect on pass probabilities I would have to adapt my simulator https://github.com/vdbergh/simul . I will try to do it tonight.

Simulation is necessary since the guarantees of the GSPRT are only asymptotic.

vondele:
yes, calculate every N games (where N might depend on the worker, e.g. 16 games for an 8 core worker). Is the overshoot correction ~0.009 on the LLR value ? 

I'd be curious to know the result of the sims.... I'd be very suprised if it has a sizeable effect.

vdbergh:
> yes, calculate every N games (where N might depend on the worker, e.g. 16 games for an 8 core worker). Is the overshoot correction ~0.009 on the LLR value ?

Yes. The test stops when the absolute value of the LLR exceeds log(19)-expected overshoot.
The expected overshoot is dynamically measured using this algorithm which I invented.

http://hardy.uhasselt.be/Fishtest/dynamic_overshoot_correction.pdf

The expected overshoot is extremely difficult to calculate theoretically. However batching might actually make it easier since then the increments will be closer to being normally distributed and for normal increments there is a simple closed formula. 
> 
> I'd be curious to know the result of the sims.... I'd be very suprised if it has a sizeable effect.

It depends what you call sizable :) With the current algorithm the pass fail probabilities are 0.0500
as measured by simulation... I am rather proud of that (as a mathematician). But of course something like 0.051 would be good enough for practical applications.

vondele:
I agree that a sound theoretical basis is important that's what is needed to understand the impact of design decisions that might be needed for practical applications. I'd be fine with 0.051, but highly critical about 0.075.

tomtor:
> Can somebody explain why update_SPRT could be a bottleneck? I would just like to understand the issue. I was under the impression that the main bottleneck is the database.

@vdbergh The issue is that the API call `/api/update_task` is the only API that is frequently called by the workers for SPRT tests. Note that `update_SPRT` need not be slow, it is the whole stack for handling the API call which also includes parsing the HTTP-header and JSON REST parameters and returning a JSON result which consumes these CPU cycles.

Reducing the number of API calls by combining more worker game results is a way to reduce CPU load. The data for MongoDb is cached and only written once per second. Without that caching MongoDb was indeed a bottleneck.

vdbergh:
@tomtor Ok thanks. Now I understand. We are discussing a worker change. I had missed this.

vdbergh:
A quick update. I updated the simulator https://github.com/vdbergh/simul to allow batching (argument --batch).

I am running some sims with batches of size 64. With overshoot correction the error probabilities appear to be still 0.0500 for the LTC bounds currently used on Fishtest. Moreover the average length of the test agrees perfectly with the prediction of the SPRT Calculator. I will also run the test for the STC bounds and then post the results here (probably tomorrow).

tomtor:
@linrock Could you have a look at: https://github.com/glinscott/fishtest/issues/553#issuecomment-614500983

I know you have not worked on the Python worker code yet, but I have a feeling that you might do that soon ;-)

vondele:
@tomtor, I agree with your statement that batching on the worker side would also reduce the traffic and the number of api calls, and is thus a good thing.

tomtor:
Closing this for now... See #601

tomtor:
Should adapt this to only act on SPSA runs.

ppigazzini:
@tomtor DEV updated, here is the log:
```
Apr 18 14:36:20 vps1572612706 pserve[6138]: During handling of the above exception, another exception occurred:
Apr 18 14:36:20 vps1572612706 pserve[6138]: Traceback (most recent call last):
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/waitress/channel.py", line 349, in service
Apr 18 14:36:20 vps1572612706 pserve[6138]:     task.service()
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/waitress/task.py", line 169, in service
Apr 18 14:36:20 vps1572612706 pserve[6138]:     self.execute()
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/waitress/task.py", line 439, in execute
Apr 18 14:36:20 vps1572612706 pserve[6138]:     app_iter = self.channel.server.application(environ, start_response)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/waitress/proxy_headers.py", line 65, in translate_proxy_headers
Apr 18 14:36:20 vps1572612706 pserve[6138]:     return app(environ, start_response)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/router.py", line 270, in __call__
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = self.execution_policy(environ, self)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/router.py", line 279, in default_execution_policy
Apr 18 14:36:20 vps1572612706 pserve[6138]:     return request.invoke_exception_view(reraise=True)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/view.py", line 778, in invoke_exception_view
Apr 18 14:36:20 vps1572612706 pserve[6138]:     reraise_(*exc_info)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/compat.py", line 179, in reraise
Apr 18 14:36:20 vps1572612706 pserve[6138]:     raise value
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/router.py", line 277, in default_execution_policy
Apr 18 14:36:20 vps1572612706 pserve[6138]:     return router.invoke_request(request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/router.py", line 249, in invoke_request
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = handle_request(request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/tweens.py", line 43, in excview_tween
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = _error_handler(request, exc)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/tweens.py", line 17, in _error_handler
Apr 18 14:36:20 vps1572612706 pserve[6138]:     reraise(*exc_info)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/compat.py", line 179, in reraise
Apr 18 14:36:20 vps1572612706 pserve[6138]:     raise value
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/tweens.py", line 41, in excview_tween
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = handler(request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/router.py", line 148, in handle_request
Apr 18 14:36:20 vps1572612706 pserve[6138]:     registry, request, context, context_iface, view_name
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/view.py", line 667, in _call_view
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = view_callable(context, request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/viewderivers.py", line 436, in rendered_view
Apr 18 14:36:20 vps1572612706 pserve[6138]:     result = view(context, request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid/viewderivers.py", line 144, in _requestonly_view
Apr 18 14:36:20 vps1572612706 pserve[6138]:     response = view(request)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/fishtest/api.py", line 108, in request_task
Apr 18 14:36:20 vps1572612706 pserve[6138]:     result = request.rundb.request_task(worker_info)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/fishtest/rundb.py", line 362, in request_task
Apr 18 14:36:20 vps1572612706 pserve[6138]:     return self.sync_request_task(worker_info)
Apr 18 14:36:20 vps1572612706 pserve[6138]:   File "/home/usr00/fishtest/fishtest/fishtest/rundb.py", line 425, in sync_request_task
Apr 18 14:36:20 vps1572612706 pserve[6138]:     limit = 40000 / math.sqrt(len(spsa['params']))
Apr 18 14:36:20 vps1572612706 pserve[6138]: NameError: name 'spsa' is not defined
```

tomtor:
@ppigazzini Oops, sloppy, pushed fix.

ppigazzini:

Issue #585, opened 2020-04-13
Sorting on machine headers fails when lazy-loaded


tomtor:
@linrock #584 has a minor issue. When machines are hidden, pressing Show shows them, but you cannot sort on the headers. After a page reload all is fine.

vondele:
and maybe the column on which is being sorted should be added to a cookie or so, since if one shows, and sorts, and afterwards reloads, the sorted column is forgotten.

ppigazzini:
Funny, I discovered only now the possibility to sort on the headers :)

ppigazzini:
@vondele the sorted column is forgotten for these pages:
 - "Home"
 - "Users"
 - "Top Month"
 - "Actions"

tomtor:
> and maybe the column on which is being sorted should be added to a cookie or so, since if one shows, and sorts, and afterwards reloads, the sorted column is forgotten.

I generally only use sort to find a specific user worker(s) and reload to restore default ordering for tables (@vondele when sorting is remembered we also need a restore to default feature?). Perhaps it should sort on worker names by default on the main page? On individual test pages I also use sort for this, but on those pages the default order is probably the best.

Showing the machines for the logged-in user shown in a separate table (above the current machines table) on the main page would be nice.

vondele:
maybe it is not so important on recall the sort order (I just need to switch off the autoreload feature for the main page if I like to look at the machines in sorted order, which I essentially never do).

tomtor:
> Funny, I discovered only now the possibility to sort on the headers :)
 
@ppigazzini  #295 was merged by you ;-)

ppigazzini:
> > Funny, I discovered only now the possibility to sort on the headers :)
> 
> @ppigazzini #295 was merged by you ;-)

@tomtor yes, and that conversation is the proof that I didn't understood what that PR did: I didn't test anything and I simply asked to you if the PR was ready to be merged :)

linrock:
ah i'll fix this.. i forgot the machines table was sortable

PR #587, opened 2020-04-13
Fix sorting of lazy-loaded tables

linrock:
Fixes https://github.com/glinscott/fishtest/issues/585

General fix so that all lazy-loaded tables will be sortable. Does not yet track the default sort key


ppigazzini:

PR #574, opened 2020-04-10
Stop using old_runs collection for runs queries

linrock:
@ppigazzini this PR only uses the `runs` collection for queries, so you can use it to see how the site is when everything is moved from `old_runs` to `runs`.

Fixes https://github.com/glinscott/fishtest/issues/569
Fixes https://github.com/glinscott/fishtest/issues/563

PR #588, opened 2020-04-13
Use different titles for finished test pages

linrock:
So the title of the page tells you which page you're on:

image

---
- Use a separate view for finished test pages
- Simplify tests view function for better readability
- Show Greens and LTC (Slow) titles on finished test pages
- Handle table pagination in one place

### DIFF
--- a/fishtest/fishtest/__init__.py
+++ b/fishtest/fishtest/__init__.py
@@ -61,16 +61,19 @@ def main(global_config, **settings):
 
   config.add_route('tests', '/tests')
   config.add_route('tests_machines', '/tests/machines')
+  config.add_route('tests_finished', '/tests/finished')
   config.add_route('tests_run', '/tests/run')
-  config.add_route('tests_modify', '/tests/modify')
   config.add_route('tests_view', '/tests/view/{id}')
   config.add_route('tests_view_spsa_history', '/tests/view/{id}/spsa_history')
+  config.add_route('tests_user', '/tests/user/{username}')
+  config.add_route('tests_stats', '/tests/stats/{id}')
+
+  # Tests - actions
+  config.add_route('tests_modify', '/tests/modify')
   config.add_route('tests_delete', '/tests/delete')
   config.add_route('tests_stop', '/tests/stop')
   config.add_route('tests_approve', '/tests/approve')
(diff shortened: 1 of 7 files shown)


vondele:
LGTM

vondele:
BTW, a request that comes up often, and is reasonable IMO, is to have a page where the yellow tests are collected. Maybe that can be integrated here? 

If we would have a good search function (see https://github.com/glinscott/fishtest/issues/339) that might be somewhat redundant

linrock:
yea that sounds good. we can start with a "Yellows" section in the sidebar, then build up to a multi-faceted search

tomtor:
Nice code cleanup!

ppigazzini:
DEV updated.
@tomtor does it work fine adding this to nginx?
```
        location /tests/finished {
            proxy_pass http://backend_all;
        }
```

vondele:
Minor nit. The sidebar says `Slow (LTC)` while the title says `LTC (Slow)`. I wouldn't mind having just `LTC`

vondele:
Minor nit 2. On the main page we have `Finished - 77477` consistent would be `Finished - 77477 tests`

ppigazzini:
Minor minor nit. There is a double pages bar when selecting a (wrong counted) high page number. This shouldn't be a problem when Green and LTC pages will be counted correctly after merging #574.

ppigazzini:
**Pro memoria**: test this also with python3 before merging and updating PROD :)

tomtor:
> does it work fine adding this to nginx?
> 
> ```
>         location /tests/finished {
>             proxy_pass http://backend_all;
>         }
> ```

@ppigazzini Looks OK

ppigazzini:
DEV nginx updated, after some test I will updated PROD and setup scripts.
```
        location ~ ^/tests/(finished|user/) {
            proxy_pass http://backend_all;
        }
```

linrock:
fixed so that:
- both the sidebar and the page title say `LTC`
- homepage also says `Finished - 77477 tests` for consistency

ppigazzini:
DEV updated.

ppigazzini:
PROD updated.
Thank you @linrock :)

PR #591, opened 2020-04-14
Fix unnecessary expensive work on tests pages past page 1

linrock:
Previously, expensive calculations for unfinished runs (active, pending, failed) were being done on ALL `/tests` and `/tests/user` pages when you visit them. They're calculated when you visit any page number, but the results of these calculations are only shown on page 1. Only the homepage (page 1 of `/tests`) caches the results.

This PR should speed up page 2 and beyond of all `/tests` and `/tests/user` pages while reducing CPU usage.

Summary:
- Don't calculate unfinished runs on tests pages after page 1
- Separate view for user tests, show username at top of user tests
- Don't calculate unfinished runs on user pages after page 1
- Simplify tests view logic
- favicon.ico and robots.txt views are served by nginx
- Add a todo for calculating results_info when the run is finished, not when viewing

PR #593, opened 2020-04-14
Fix finding and setting failed runs

linrock:
@ppigazzini this should fix https://github.com/glinscott/fishtest/pull/591#issuecomment-613667898

### DIFF
--- a/fishtest/fishtest/templates/run_table.mak
+++ b/fishtest/fishtest/templates/run_table.mak
@@ -42,7 +42,13 @@
       <span class="pagination pagination-small">
         <ul>
           %for page in pages:
-            <li class="${page['state']}"><a href="${page['url']}">${page['idx']}</a></li>
+            <li class="${page['state']}">
+              %if page['state'] not in ['disabled', 'active']:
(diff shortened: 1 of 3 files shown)


ppigazzini:
@linrock the failed runs bug is fixed.

I found another bug: when at page 1, the "prev" button is grayed out but still clickable, page=0 triggers an Internal Server Error, see:
https://dfts-0.pigazzini.it/tests/user/xoto?page=0
https://dfts-0.pigazzini.it/tests?page=0

Green and LTC go to page=0 but without error
https://dfts-0.pigazzini.it/tests/finished?page=0&success_only=1

linrock:
thanks for finding that.. working on a fix now

linrock:
@ppigazzini that page=0 crash is fixed. i can also fix those greyed-out links too since they should not be clickable

linrock:
@ppigazzini ok, those disabled pagination links are no longer clickable now, so users should no longer be able to go to page=0 pages from clicking

ppigazzini:

PR #589, opened 2020-04-14
Add initial tests for fishtest API

linrock:
if we want to change the server-side behavior of `update_task` or `request_task` for the 40k cores problem, having good test coverage would be useful for stability

PR #594, opened 2020-04-15
Fix some pymongo deprecation warnings

linrock:
Fixes most of those pymongo warnings when running tests, as seen here https://github.com/glinscott/fishtest/pull/589#issuecomment-613717674

a few of the warnings are fixed in https://github.com/glinscott/fishtest/pull/574 so i won't fix them here to avoid a merge conflict

### DIFF
--- a/fishtest/fishtest/actiondb.py
+++ b/fishtest/fishtest/actiondb.py
@@ -43,7 +43,7 @@ class ActionDb:
     self._new_action(username, 'block_user', data)
 
   def _new_action(self, username, action, data):
-    self.actions.insert({
+    self.actions.insert_one({
       'username': username,
       'action': action,
       'data': data,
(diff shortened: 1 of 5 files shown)


ppigazzini:
@linrock please rebase :)

linrock:
no prob! merge conflict fixed and rebased

ppigazzini:
@linrock bugs:
 - after approving a test in the test page, the test is still showed in the pending test table with the yellow icon
 - after stopping a test in the test page, the test is still showed in the pending/active test table
 - after deleting a test in the pending table, the test is still showed in the pending table

ppigazzini:
@linrock I leave DEV with this PR in order to verify the strange behavior (master was fine). Bed is calling here :)

linrock:
thanks for catching that. the problem was that the run updates were failing because the `replace_one` calls require an `ObjectId` instead of a string when querying by `_id`

ppigazzini:

PR #595, opened 2020-04-15
Iterate through runs using range queries in delta_update_users.py

linrock:
`delta_update_users.py` runs every 15 minutes. Using `skip` for mongodb collections is inefficient and CPU-intensive for iterating through large collections. It's much more efficient to use range queries on an indexed field.

See https://docs.mongodb.com/manual/reference/method/cursor.skip/#using-range-queries

A quick benchmark on a $5 VPS:

Before:
```
Num runs processed: 74887
Successfully updated 1200 users

real	1m46.332s
user	0m22.695s
sys	0m0.744s
```

After:
```
Num runs processed: 74887
Successfully updated 1200 users

real	0m19.514s
user	0m15.732s
sys	0m0.397s
```

### DIFF
--- a/fishtest/utils/delta_update_users.py
+++ b/fishtest/utils/delta_update_users.py
@@ -2,8 +2,9 @@
 
 import os
 import sys
-
 from datetime import datetime, timedelta
+
+from pymongo import DESCENDING
 
 # For tasks
 sys.path.append(os.path.expanduser('~/fishtest/fishtest'))
@@ -115,13 +116,16 @@ def update_users():
       print("Exception on run: ", run)
 
   # Step through these in small batches (step size 100) to save RAM
-  current = 0
   step_size = 100
 
   now = datetime.utcnow()
   more_days = True
+  last_run_id = None
   while more_days:
-    runs = rundb.get_finished_runs(skip=current, limit=step_size)[0]
+    q = { 'finished': True }
+    if last_run_id:
+      q['_id'] = { '$lt': last_run_id }
+    runs = list(rundb.runs.find(q, sort=[('_id', DESCENDING)], limit=step_size))
     if len(runs) == 0:
       break
     for run in runs:
@@ -136,7 +140,7 @@ def update_users():
           print("Exception on run: ", run['_id'])
       elif not clear_stats:
         more_days = False
-    current += step_size
+    last_run_id = runs[-1]['_id']
 
   if new_deltas:
     new_deltas.update(deltas)
--- a/fishtest/utils/update_users.py
+++ b/fishtest/utils/update_users.py
@@ -2,6 +2,8 @@
 
 import datetime, os, sys
 
+from pymongo import DESCENDING
+
 # For tasks
 sys.path.append(os.path.expanduser('~/fishtest/fishtest'))
 from fishtest.rundb import RunDb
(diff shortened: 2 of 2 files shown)


tomtor:
On prod the script currently takes about 17 seconds:

https://tests.stockfishchess.org/actions?action=update_stats&user=

@linrock Note that about a year ago this was many minutes (no delta updates), but additional speedup is always welcomed!

ppigazzini:
mmm, my benches on DEV (I made 3 tries and taken the median, after `sudo systemctl stop cron`) don't show any speedup, and pypy3 is a little slowdown :(

#### python3
- master
```
Successfully updated 1360 users

real    0m13.002s
user    0m10.183s
sys     0m1.353s
```

- pr
```
Successfully updated 1360 users

real    0m13.150s
user    0m10.307s
sys     0m1.453s
```

#### pypy3
- master
```
Successfully updated 1360 users

real    0m17.716s
user    0m15.393s
sys     0m1.085s
```

- pr
```
Successfully updated 1360 users

real    0m16.308s
user    0m14.063s
sys     0m1.015s
```

tomtor:
> and pypy3 is a little slowdown :(

That is expected. It is Mongodb intensive code, no gain can be expected from the pypy compiler.

ppigazzini:
> > and pypy3 is a little slowdown :(
> 
> That is expected. It is Mongodb intensive code, no gain can be expected from the pypy compiler.

Sure, but I hadn't expecting a 23% slowdown too.

ppigazzini:
@tomtor @linrock to return in topic, I'm fine merging this PR like a mongodb best practice, but I would like a change of the commit, something like this should work.

**Use mongodb best practices for delta_update_users.py**

delta_update_users.py runs every 15 minutes. Using skip for mongodb collections is inefficient and CPU-intensive for iterating through large collections. It's much more efficient to use range queries on an indexed field.

See https://docs.mongodb.com/manual/reference/method/cursor.skip/#using-range-queries

In some conditions this can result in a speedup, see this quick benchmark on a $5 VPS:

Before:
```
Num runs processed: 74887
Successfully updated 1200 users

real	1m46.332s
user	0m22.695s
sys	0m0.744s
```
After:
```
Num runs processed: 74887
Successfully updated 1200 users

real	0m19.514s
user	0m15.732s
sys	0m0.397s
```

linrock:
hmm, in prod, the # of documents being iterated through is probably not large enough to see a performance improvement.

i've updated the commit and PR title to reflect a code change instead of a performance improvement since it looks to have no speedup in prod.

linrock:
i updated the `update_users.py` script too. it's not being used currently, but in case someone does use it, it should now have a large speedup in dev and prod since it iterates through the entire runs collection.

ppigazzini:
@tomtor do we could drop `update_users.py`?

ppigazzini:
@linrock big speedup (tried 2 times) scanning all users ;)
* master:
```
time ${VENV}/bin/python fishtest/fishtest/utils/delta_update_users.py all
scan all
Successfully updated 1360 users

real    26m33.732s
user    2m21.219s
sys     0m4.885s
```
* pr:
```
time ${VENV}/bin/python fishtest/fishtest/utils/delta_update_users.py all
scan all
Successfully updated 1360 users

real    2m30.753s
user    2m19.093s
sys     0m4.499s
```

linrock:
nice, that's a significant speedup if anyone has to go through the entire collection with the script

tomtor:
> @tomtor do we could drop `update_users.py`?

Yes, the delta script when called with `all` flag should do the same.

ppigazzini:

PR #601, opened 2020-04-16
Scale down update_task request frequency by 2x

linrock:
Instead of calling `/api/update_task` after each game finishes, the worker now calls it either:
- once per N games, where N is worker concurrency (min of concurrency flag and # cpus - 1)
- or after all the games in the task have been completed

This should ease the load on the server by generating less network traffic (see https://github.com/glinscott/fishtest/issues/553#issuecomment-614500983) and allow prod to handle a higher # of worker cores.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -50,6 +50,11 @@ if IS_WINDOWS:
   EXE_SUFFIX = '.exe'
   MAKE_CMD = 'make COMP=mingw ' + ARCH
 
+def send_api_post_request(api_url, payload):
+  return requests.post(api_url, data=json.dumps(payload), headers={
+    'Content-Type': 'application/json'
+  }, timeout=HTTP_TIMEOUT)
+
 def github_api(repo):
   """ Convert from https://github.com/<user>/<repo>
       To https://api.github.com/repos/<user>/<repo> """
@@ -83,7 +88,7 @@ def verify_signature(engine, signature, remote, payload, concurrency):
     if int(bench_sig) != int(signature):
       message = 'Wrong bench in %s Expected: %s Got: %s' % (os.path.basename(engine), signature, bench_sig)
       payload['message'] = message
-      requests.post(remote + '/api/stop_run', data=json.dumps(payload), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT)
+      send_api_post_request(remote + '/api/stop_run', payload)
       raise Exception(message)
 
   finally:
@@ -257,20 +262,20 @@ def update_pentanomial(line,rounds):
   pentanomial=[rounds['pentanomial'][i]+old_stats['pentanomial'][i] for i in range(0,5)]
   return pentanomial
 
-def validate_pentanomial(wld,rounds):
+def validate_pentanomial(wld, rounds):
   def results_to_score(results):
-    return sum([results[i]*(i/2.0) for i in range(0,len(results))])
-  LDW=[wld[1],wld[2],wld[0]]
-  s3=results_to_score(LDW)
-  s5=results_to_score(rounds['pentanomial'])+results_to_score(rounds['trinomial'])
-  assert(sum(LDW)==2*sum(rounds['pentanomial'])+sum(rounds['trinomial']))
-  epsilon=1e-4
-  assert(abs(s5-s3)<epsilon)
+    return sum([results[i] * (i / 2.0) for i in range(0, len(results))])
+  LDW = [wld[1], wld[2], wld[0]]
+  s3 = results_to_score(LDW)
+  s5 = results_to_score(rounds['pentanomial']) + results_to_score(rounds['trinomial'])
+  assert(sum(LDW) == 2 * sum(rounds['pentanomial']) + sum(rounds['trinomial']))
+  epsilon = 1e-4
+  assert(abs(s5 - s3) < epsilon)
 
 
 def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
   global old_stats
-  rounds={}
+  rounds = {}
 
   q = Queue()
   t = threading.Thread(target=enqueue_output, args=(p.stdout, q))
@@ -278,12 +283,14 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
   t.start()
 
   end_time = datetime.datetime.now() + datetime.timedelta(seconds=tc_limit)
-  print('TC limit '+str(tc_limit)+' End time: '+str(end_time))
+  print('TC limit {} End time: {}'.format(tc_limit, end_time))
 
+  num_game_pairs_updated = 0
   while datetime.datetime.now() < end_time:
-    try: line = q.get_nowait()
+    try:
+      line = q.get_nowait()
     except Empty:
-      if p.poll() != None:
+      if p.poll():
         break
       time.sleep(1)
       continue
@@ -312,7 +319,7 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
       chunks = chunks[1].split()
       wld = [int(chunks[0]), int(chunks[2]), int(chunks[4])]
 
-      validate_pentanomial(wld,rounds)
+      validate_pentanomial(wld, rounds)
 
       if not spsa_tuning:
         result['stats']['wins']   = wld[0] - rounds['trinomial'][2] + old_stats['wins']
@@ -326,38 +333,41 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
         spsa['losses'] = wld[1]
         spsa['draws'] = wld[2]
 
-      update_succeeded=False
-      for _ in range(0,5):
-        try:
-          t0 = datetime.datetime.utcnow()
-          req = requests.post(remote + '/api/update_task', data=json.dumps(result), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT).json()
-          print("Task updated successfully in %ss" % ((datetime.datetime.utcnow() - t0).total_seconds()))
-          if not req['task_alive']:
-            # This task is no longer neccesary
-            print('Server told us task is no longer needed')
-            kill_process(p)
-            return req
-          update_succeeded=True
+      num_game_pairs_finished = sum(rounds['pentanomial'])
+      # Send an update_task request after a game pair finishes, not after every game
+      if num_game_pairs_finished > num_game_pairs_updated:
+        # Attempt to send game results to the server. Retry a few times upon error
+        update_succeeded = False
+        for _ in range(0, 5):
+          try:
+            t0 = datetime.datetime.utcnow()
+            response = send_api_post_request(remote + '/api/update_task', result).json()
+            print("  Task updated successfully in %ss" % ((datetime.datetime.utcnow() - t0).total_seconds()))
+            if not response['task_alive']:
+              # This task is no longer neccesary
+              print('Server told us task is no longer needed')
+              kill_process(p)
+              return response
+            update_succeeded = True
+            break
+          except Exception as e:
+            sys.stderr.write('Exception from calling update_task:\n')
+            print(e)
+            # traceback.print_exc(file=sys.stderr)
+          time.sleep(HTTP_TIMEOUT)
+        if not update_succeeded:
+          print('Too many failed update attempts')
+          kill_process(p)
           break
-        except Exception as e:
-          sys.stderr.write('Exception from calling update_task:\n')
-          print(e)
-#          traceback.print_exc(file=sys.stderr)
-        time.sleep(HTTP_TIMEOUT)
+        num_game_pairs_updated = num_game_pairs_finished
 
-      if not update_succeeded:
-        print('Too many failed update attempts')
-        kill_process(p)
-        break
+    pentanomial = update_pentanomial(line, rounds)
+    result['stats']['pentanomial'] = pentanomial
 
-    pentanomial=update_pentanomial(line,rounds)
-    result['stats']['pentanomial']=pentanomial
-
-  if datetime.datetime.now() >= end_time:
-    print(str(datetime.datetime.now()) + ' is past end time ' + str(end_time))
-
+  now = datetime.datetime.now()
+  if now >= end_time:
+    print('{} is past end time {}'.format(now, end_time))
   kill_process(p)
-
   return { 'task_alive': True }
 
 def launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play, tc_limit):
@@ -370,7 +380,7 @@ def launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play, tc_limit):
   if spsa_tuning:
     # Request parameters for next game
     t0 = datetime.datetime.utcnow()
-    req = requests.post(remote + '/api/request_spsa', data=json.dumps(result), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT).json()
+    req = send_api_post_request(remote + '/api/request_spsa', result).json()
     print("Fetched SPSA parameters successfully in %ss" % ((datetime.datetime.utcnow() - t0).total_seconds()))
 
     global w_params, b_params
@@ -563,7 +573,8 @@ def run_games(worker_info, password, remote, run, task_id):
           ['-engine', 'name=Base-'+run['args']['resolved_base'][:7], 'cmd=%s' % (base_engine_name)] + base_options + ['_spsa_'] + \
           ['-each', 'proto=uci', 'tc=%s' % (scaled_tc)] + nodestime_cmd + threads_cmd + book_cmd
 
-    task_status = launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play, tc_limit * games_to_play / min(games_to_play, games_concurrency))
+    task_status = launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play,
+                                   tc_limit * games_to_play / min(games_to_play, games_concurrency))
     if not task_status.get('task_alive', False):
       break
 
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -7,7 +7,7 @@ from pyramid.view import view_config
 
 import fishtest.stats.sprt
 
-WORKER_VERSION = 72
+WORKER_VERSION = 73
 
 
 def get_flag(request):
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -27,7 +27,7 @@ from updater import update
 from datetime import datetime
 from os import path
 
-WORKER_VERSION = 72
+WORKER_VERSION = 73
 ALIVE = True
 HTTP_TIMEOUT = 15.0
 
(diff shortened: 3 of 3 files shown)


vdbergh:
For overshoot correction to work correctly all batches would have to have the same size...

vondele:
this looks good to me, but maybe @vdbergh can give a thumbs up as well. In particular if any more changes would be needed to the SPRT code now that we batch on the worker. I see he just commented already. If that's a strict requirement we can make the batch size equal to 8, IMO.

It also answers the question I had additionally, if we could deal with http timeouts in a lazy way. I.e. if there is a timeout, we don't try 5 times, just compute the next games, and try again, unless we're done with the task.

@vdbergh any way to relax that requirement, or does it get tricky?

vondele:
hmm, would require the batch size to be a divisor of 250 as well.

vdbergh:
It is not really necessary to do overshoot correction. It just means that the characteristics of the test differ slightly from the theoretical ones, which assume continuous time.

However it is safe. The error probabilities will always be smaller than the theoretical ones (the price is slightly more games of course).

It is probably possible to do theoretically correct overshoot correction for varying batch sizes but currently I don’t know how. The area is called “group sequential tests”.

ppigazzini:
DEV + workers updated.

vondele:
@vdbergh does it make sense to have the batch size a 2 * ncores, since the pentanomial model requires both games anyway?

In any case, I'd be fine with the current PR proposal, but somebody else needs to do proper review.

vdbergh:
What would be a typical value of ncores these days? I can check what the error probabilities will be without overshoot correction.

vondele:
ncores would be ~8. So batch sizes 16. 

I'm more asking because I vaguely recall that the condition for stopping will be computed with the pentanomial model, so will require that both sides of the opening have been played. I.e. if we update after each core played a single game, there will be not enough info anyway.

vdbergh:
The changing of the pentanomial frequencies is a bit unpredictable. So I think one should just send an update if sum(pentanomial)>=previous_sum+batch_size where batch_size may be derived from the number of cores.

(( The current implementation is a bit inefficient as the worker will send an update after every game. So it will send several times the same pentanomial frequencies. )j

vdbergh:
NB: The unpredictability comes from the fact that the pentanomial frequencies change every time a game pair is completed. But these game pairs are distributed randomly over the threads and their outcomes are interleaved and may even be out of order.

vdbergh:
An alternative model to batching could be queuing. By this I mean that the worker would just create a stack of N pentanomial frequencies (N depending on the number of cores) and send these together to the server.

vdbergh:
So the result sent to server would be a list of pentanomial frequencies through which the server would loop. 

Update_SPRT could then still decide to process every 32 requests or so, if it would be a bottleneck.

vdbergh:
Hmm it would be much better if the worker would send a list of queued outcomes (a list of numbers 0-4) besides the current pentanomial frequencies. This would then not require any immediate server changes.

Update_SPRT can then to use these outcomes.

tomtor:
I would suggest adapting this PR in a minimal way to simply call the API every 4th time and changing chunksize to 400, this:

- Reduces load by a factor 4
- Does not impact overshoot correction if I understand @vdbergh correctly
- Changes default number of games to 400000 as requested in another topic

vondele:
@tomtor increasing chunksize (batch size, right?) to 400 has the disadvantage that single core workers can be busy for a very long time on a batch of LTC games. (about 22h, if we use the measured average of 200s per game), and even a 8core worker will stick to the task for 3h. I would rather keep chunksize at 250 or even reduce it to keep the assignment of workers to tasks dynamic, and to make sure numgames type tests (that finish with the slowest worker assigned) finish quickly.

tomtor:
@vondele Changing batch/chunksize to 200 or 240 is also possible. Leaving it at 250 and calling the api every second time is also an option. My preference would be to start with the last option.

vondele:
250 and 2 ? Right now, fishtest seems to be running stable at about 10k cores, that should move it to 20k cores. It would be interesting to understand what the next bottleneck is. However, 250 and 2 or 200 and 4 are also fine with me. Once we have the knob we can easily adjust as needed.

tomtor:
@ppigazzini Can you choose between 200/4 or 250/2, I am fine with both. To test it's effectiveness perhaps it is better to start with 200/4.

ppigazzini:
@tomtor @vondele I read propositions for factor value in range 2-32, I'm wondering if we could use a batch size = 256 to have more choices in factor values w/o remainder.

At the moment DEV/worker seem working fine with 256/4 (changes by hand).

vondele:
OK, let's try 200 / 4 for starters.

ppigazzini:
DEV updated to use batch size 200.
My workers are already set to use factor 4 (games finished, not 2 pairs: so in the test page the games finished sometimes have increment of 2 and not 4)

vdbergh:
But doing updates every N games does not help for the overshoot correction as currently implemented. There should be an update whenever sum(pentanomial) has increased by N/2 (meaning N/2 game pairs have completed). This is different.

vondele:
Thanks! The patch should thus be updated to use the check on sum(pentanomial). I assume that we might have to be careful that the update condition works from sprt, spsa and numgames tests.

vdbergh:
@vondele numgames I think is not a problem (it also uses the pentanomial frequencies). I currently don't really know how spsa works unfortunately. But it principle it could have a different update policy.

vondele:
I think we can leave the update strategy of spsa, for the time being, to what it was. Just being careful it is not broken by the changes we make.

ppigazzini:
@vdbergh This seems to work fine:

```python
  num_pairs_prev = 0

  while datetime.datetime.now() < end_time:
    try:
      line = q.get_nowait()

...

      num_games_finished = sum(wld)
      num_pairs_finished = sum(pentanomial)
      pairs_factor = 4
      if (num_pairs_finished > num_pairs_prev and num_pairs_finished % pairs_factor == 0) or num_games_finished == games_to_play:
        # Send game results after 1 game per worker CPU is played or after the last game is played
        num_pairs_prev = num_pairs_finished
        print("pentanomial: {} pairs: {} games: {}".format(pentanomial, num_pairs_finished, num_games_finished))
        update_succeeded = False
        # Attempt to send game results to the server. Retry a few times upon error
        for _ in range(0, 5):

```

vdbergh:
@ppigazzini Yes this seems fine provided games_to_play is divisible by 2*pairs_factor. This should probably be an assert.

I need to make a small corrensponding server change (update_SPRT does some sanity checking on its inputs, and currently does not like updates of size>1). I will do this tonight.

vdbergh:
Perhaps the worker should be told by the server what pairs_factor is?

ppigazzini:
> Perhaps the worker should be told by the server what pairs_factor is?

IMO yes if there are some requirements from server side.

vdbergh:
One practical reason is that otherwise changing pairs_factor would require a worker update. 

But it is also true that it would be convenient to have it.

worker/games.py
```diff
@@ -326,41 +332,47 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
...
+              return response
+            update_succeeded = True
+            break
+          except Exception as e:
+            sys.stderr.write('Exception from calling update_task:\n')
+            print(e)
+            # traceback.print_exc(file=sys.stderr)
+          time.sleep(HTTP_TIMEOUT)
+
+          if not update_succeeded:
```
ppigazzini, line 359: If I'm reading right this `if` seems to be in the `for` cycle, in this way at the first exception the process will be terminated.

tomtor: No it will print the exception and continue the loop at the sleep...

ppigazzini: After the sleep it should execute `if not update_succeeded:` and kill the process.
In master `if not update_succeeded:` is out that `for` cycle.

vdbergh: Hmm in my version of games.py ``` if not update_succeeded ``` is outside the loop.

tomtor: Ah, yes it should be outside the loop...

ppigazzini:
If I followed correctly the issue now we are unnecessarily updating the tasks at any new game finished.
In this case we could start to lower the CPU load with a simple PR that updated the task only when a new pair finish, something like this https://github.com/ppigazzini/fishtest/tree/updated_on_new_pairs

```
index e2a64e1..cc88e7d 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -279,6 +279,7 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):

   end_time = datetime.datetime.now() + datetime.timedelta(seconds=tc_limit)
   print('TC limit '+str(tc_limit)+' End time: '+str(end_time))
+  num_pairs_updated = 0

   while datetime.datetime.now() < end_time:
     try: line = q.get_nowait()
@@ -326,29 +327,36 @@ def run_game(p, remote, result, spsa, spsa_tuning, tc_limit):
         spsa['losses'] = wld[1]
         spsa['draws'] = wld[2]

-      update_succeeded=False
-      for _ in range(0,5):
-        try:
-          t0 = datetime.datetime.utcnow()
-          req = requests.post(remote + '/api/update_task', data=json.dumps(result), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT).json()
-          print("Task updated successfully in %ss" % ((datetime.datetime.utcnow() - t0).total_seconds()))
-          if not req['task_alive']:
-            # This task is no longer neccesary
-            print('Server told us task is no longer needed')
-            kill_process(p)
-            return req
-          update_succeeded=True
+      num_games_finished = sum(wld)
+      num_pairs_finished = sum(pentanomial)
+      if num_pairs_finished > num_pairs_updated or num_games_finished == games_to_play:
+        # Send game results after every new pair completed
+        num_pairs_updated = num_pairs_finished
+        # print(pairs: {}; games: {}".format(num_pairs_finished, num_games_finished))
+
+        update_succeeded=False
+        for _ in range(0,5):
+          try:
+            t0 = datetime.datetime.utcnow()
+            req = requests.post(remote + '/api/update_task', data=json.dumps(result), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT).json()
+            print("Task updated successfully in %ss" % ((datetime.datetime.utcnow() - t0).total_seconds()))
+            if not req['task_alive']:
+              # This task is no longer neccesary
+              print('Server told us task is no longer needed')
+              kill_process(p)
+              return req
+            update_succeeded=True
+            break
+          except Exception as e:
+            sys.stderr.write('Exception from calling update_task:\n')
+            print(e)
+#            traceback.print_exc(file=sys.stderr)
+          time.sleep(HTTP_TIMEOUT)
+
+        if not update_succeeded:
+          print('Too many failed update attempts')
+          kill_process(p)
           break
-        except Exception as e:
-          sys.stderr.write('Exception from calling update_task:\n')
-          print(e)
-#          traceback.print_exc(file=sys.stderr)
-        time.sleep(HTTP_TIMEOUT)
-
-      if not update_succeeded:
-        print('Too many failed update attempts')
-        kill_process(p)
-        break

```

vdbergh:
@ppigazzini Yes you are correct. I already pointed this out above. Currently update_task is called too often. I was aware of this but since it did not appear to be an issue I did not change it.

vondele:
@ppigazzini that's the minimally invasive patch that should improve the load on the server 2x. That's obviously a good thing. 

But it does remain a bottleneck, for >20k cores, which is arguably rare. To address that bottleneck, one could update_task for batches of completed game pairs. In my understanding, if we make those batches fixed size, the simulations suggest we're playing safe, and my feeling says that, if we keep the batch small (e.g. 4 game pairs), the user experience of fishtest will remain essentially the same, and the efficiency of the framework is also roughly the same. 

Since it involves a worker change, my preference would be to have the code generic, so we can decide to use the equivalent of your minimally invasive patch, or batch games a bit more with a server change.

ppigazzini:
@vondele we could ask to @linrock to updated this PR to reflect the minimal changes (keeping the code improvements :) ), push the PR to the workers and measure the real gain on PROD with that sampling factor 2. Then we can start designing the batch games PR with more information and more time to test on DEV if the SPRT is working fine.

vondele:
that would be a reasonable strategy as well... definitely up to you guys to decide what the most meaningful implementation and path is.

worker/games.py
linrock: oops, fixed

linrock:
i've updated this PR to send `update_task` after every game pair finishes like in the patch @ppigazzini posted. i think starting off with a simple 2x improvement is a great start and we can make further improvements after.

i'm guessing we can assume the number of games a worker will play for a task is always an even number?

ppigazzini:
@linrock already updated my workers :)
I think that we could enforce an even number of games (due the games paired) for the size of the runs.

vondele:
@ppigazzini the batch size is even (by our choice), I wonder what happens when a worker quits and a new one is assigned to the same batch. How is the number of games of the reassigned worker computed? 

Note that we call cutechess with the following syntax: `'-rounds', str(int(games_to_play/2)), '-games', ' 2'` so cutechess wants to play an even number of games for sure before it is finished. If games_to_play is not even, we are probably in trouble...

tomtor:
@ppigazzini at dev the individual test page is updating but the main page is not?

ppigazzini:
@tomtor one hour ago I reinstalled fishtest on DEV to drop the PR https://github.com/glinscott/fishtest/pull/572#issuecomment-615402747 because I wan't able to create a SPSA test.

ppigazzini:
@noobpwnftw with this worker PR we should be able to scale at 20k cores on PROD.
Let me know your preference to update your workers:
1. pull the PR with:
```bash
git clone --single-branch --branch master https://github.com/glinscott/fishtest.git
cd fishtest
git config user.email "you@example.com"
git config user.name "your_name"
git pull --no-edit origin pull/601/head
```
2. pull master after merging the PR
3. force the workers auto update merging the PR and updating PROD

noobpwnftw:
@ppigazzini 
I'd like to use method 2, method 1 is not feasible with many workers and method 3 I will easily hit github throttling and unable to download anything for an hour or so.

ppigazzini:
@noobpwnftw OK, merged :)

noobpwnftw:
Can you raise my machine limit? :)

ppigazzini:
@noobpwnftw raised to 10k :)

ppigazzini:
5032 machines 22222 cores 1.07M nps (23834.74M total nps) 24314 games/minute 0.3 hours remaining.

50% < CPU core < 85%

```
top - 23:21:30 up 12 days,  6:48,  2 users,  load average: 1.02, 1.36, 1.72
Tasks:  33 total,   2 running,  31 sleeping,   0 stopped,   0 zombie
%Cpu0  :  27.0/1.7    29[|||||||||||||||||||||||||||||                                                                       ]
%Cpu1  :  22.6/1.3    24[||||||||||||||||||||||||                                                                            ]
%Cpu2  :  12.6/1.0    14[||||||||||||||                                                                                      ]
%Cpu3  :  13.3/0.7    14[||||||||||||||                                                                                      ]
KiB Mem : 65.9/5242880  [||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||                                  ]
KiB Swap:  0.0/0        [                                                                                                    ]

    1 root      20      224908   2868   1552 S        0.1   0:07.27 init -z
    2 root      20                           S                       `- [kthreadd/1988]
    3 root      20                           S                           `- [khelper]
   65 root      20      604284  24508  17224 S        0.5   9:46.94  `- /lib/systemd/systemd-journald
  167 root      20       42092    428    376 S        0.0   0:00.91  `- /lib/systemd/systemd-udevd
  170 systemd+  20       71848    412    348 S        0.0   0:01.07  `- /lib/systemd/systemd-networkd
  295 root      20       62100    844    672 S        0.0   0:01.44  `- /lib/systemd/systemd-logind
  296 message+  20       47724    852    524 S        0.0   0:00.58  `- /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
  300 syslog    20      189016   1584    356 S        0.0   3:11.97  `- /usr/sbin/rsyslogd -n
  301 root      20      184296    308    272 S        0.0   0:00.11  `- /usr/bin/python3 /usr/share/unattended-upgrades/unattended-upgrade-shutdown --wait-for-signal
  305 mongodb   20     3448480 1.775g   5332 S   2.7 35.5   1407:07  `- /usr/bin/mongod --config /etc/mongod.conf
  314 root      20       72288    484    372 S        0.0   0:00.12  `- /usr/sbin/sshd -D
13714 root      20       97180    568    416 S        0.0                `- sshd: fishtest [priv]
13729 fishtest  20       97180    476    224 S        0.0   0:00.52          `- sshd: fishtest@pts/0
13730 fishtest  20       20480   1220    820 S        0.0   0:00.10              `- -bash
25680 root      20       97180   2096   1156 S        0.0   0:00.02      `- sshd: fishtest [priv]
25707 fishtest  20       97180   1480    540 S        0.0   0:00.18          `- sshd: fishtest@pts/1
25708 fishtest  20       19660   2352    764 S        0.0   0:00.04              `- -bash
30111 fishtest  20       36752   1644   1036 R   0.3  0.0   0:01.41                  `- top
  581 root      20       24176    288    284 S        0.0            `- /usr/sbin/xinetd -pidfile /run/xinetd.pid -stayalive -inetd_compat -inetd_ipv6
  594 Debian-+  20       59184    496    384 S        0.0   0:01.02  `- /usr/sbin/exim4 -bd -q30m
  597 root      20      100972    272    244 S        0.0            `- /usr/sbin/saslauthd -a pam -c -m /var/run/saslauthd -n 2
  598 root      20      100972     28        S        0.0                `- /usr/sbin/saslauthd -a pam -c -m /var/run/saslauthd -n 2
  619 root      20      142520   2576   1104 S        0.0   0:00.03  `- nginx: master process /usr/sbin/nginx -g daemon on; master_process on;
31264 www-data  20      147840   8716   2396 R   6.3  0.2  68:54.20      `- nginx: worker process
31265 www-data  20      143716   4464   2248 S        0.1   3:20.06      `- nginx: worker process
31266 www-data  20      143320   4204   2132 S        0.1   0:01.77      `- nginx: worker process
31267 www-data  20      142864   3448   1652 S        0.1   0:00.06      `- nginx: worker process
16839 root      20       28344    744    460 S        0.0   0:00.44  `- /usr/sbin/cron -f
 2705 fishtest  20     2040928 568016   4552 S       10.8   7:53.59  `- /home/fishtest/fishtest/fishtest/env/bin/python3 /home/fishtest/fishtest/fishtest/env/bin/pserve production.ini http_port=6544
 2709 fishtest  20     2348960 981144   4700 S  73.8 18.7 276:37.51  `- /home/fishtest/fishtest/fishtest/env/bin/python3 /home/fishtest/fishtest/fishtest/env/bin/pserve production.ini http_port=6543
32003 root      20       13008    840    712 S        0.0            `- /sbin/agetty -o -p -- \u --noclear tty2 linux
32005 root      20       13008    848    716 S        0.0            `- /sbin/agetty -o -p -- \u --noclear --keep-baud console 115200,38400,9600 linux
```

ppigazzini:
PROD updated, started the forced update of the workers.
Thank you @linrock @vdbergh @vondele @tomtor @noobpwnftw :)

vdbergh:
If the idea of batching comes back then I think ik makes sense to make the batch size dependent on the TC. A large batch size will change the user experience less at STC than at LTC.

PR #606, opened 2020-04-19
Center header text on 'create new test' page

linrock:
As suggested by @snicolet in https://github.com/glinscott/fishtest/pull/572#issuecomment-615969657

---
image


ppigazzini:

MJZ1977:
Very nice looking, even if we are disturbed by the submit button on top of the page

PR #608, opened 2020-04-19
Show error messages in red instead of green

linrock:
error messages across the site will show up as red as suggested by @vondele in https://github.com/glinscott/fishtest/pull/572#issuecomment-615124753

---
image


vondele:
It might be you already did this, but if not, can you consider adding a little white space above the message boxes:
Screenshot from 2020-04-19 09-32-37

ppigazzini:
DEV updated

linrock:
yep, i'll plan to add some whitespace above those messages. the design of the site will slowly improve with time :)

Issue #605, opened 2020-04-18
statistics on CPU hours ?


vondele:
seems not to be updated anymore according to:
https://groups.google.com/d/msg/fishcooking/pGFcmUsNmAk/vIyTDRuiAAAJ

ppigazzini:
Something is wrong, I loaded yesterday a fresh DB on DEV, and "user01" (used in my tests) should have the same information
- page "Users"

| Username | Last active | Games/Hour | CPU Hours | Games played | Tests submitted | Tests repository |
| - | - | - | - | - | - | - |
| user01 | 13 minutes ago | 30493 | 9 | 3292 | 2 | https://github.com/official-stockfish/Stockfish |

- page "Top Month"

| Username | Last active | Games/Hour | CPU Hours | Games played | Tests submitted | Tests repository |
| - | - | - | - | - | - | - |
| user01 | seconds ago | 30493 | 105 | 115970 | 14 | https://github.com/official-stockfish/Stockfish |

tomtor:
@ppigazzini The MongoDb databases were processed on prod to patch fields. Could you try fully rebuilding the stats on prod?

ppigazzini:
@tomtor I hope to have understood correctly, I ran `${VENV}/bin/python fishtest/fishtest/utils/delta_update_users.py all` on PROD and DEV.
On DEV pages "Users" and "Top Month" are still very different.

ppigazzini:
@tomtor could be that "Users" is updated only after that a test is finished and "Top Month" is updated in real time?

tomtor:
> could be that "Users" is updated only after that a test is finished and "Top Month" is updated in real time?
> By the way this doesn't explain the different numbers of tests submitted.

Yes, users is only updated when tests are finished and monthly on every update cron job. 

Monthly tests submitted is tests submitted last month. That is 0 for me and correct. Everything looks fine to me? I have not checked if numbers increase yet, but according to https://groups.google.com/d/msg/fishcooking/pGFcmUsNmAk/vIyTDRuiAAAJ they do?

JojoM-sf:
Following Pasquale‘s measures, the numbers got updated picking up on the recent „delay“, but now they appear to be stuck again.

tomtor:
@JojoM-sf I just checked, they incremented for me (tvijlbrief) compared to about 2 hours ago....

JojoM-sf:
No, the value is constant: 312471

tomtor:
[...]

PR #611, opened 2020-04-19
Cache flag lookups

tomtor:
Flag look ups kill performance when the queue is empty.

See https://github.com/glinscott/fishtest/issues/605#issuecomment-616095195

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -10,10 +10,15 @@ import fishtest.stats.sprt
 WORKER_VERSION = 73
 
 
+flag_cache = {}
+
 def get_flag(request):
   ip = request.remote_addr
+  if ip in flag_cache:
+    return flag_cache[ip]
   result = request.userdb.flag_cache.find_one({'ip': ip})
   if result is not None:
+    flag_cache[ip] = result['country_code']
     return result['country_code']
 
   # Get country flag ip
@@ -26,8 +31,10 @@ def get_flag(request):
         'ip': ip,
         'country_code': country_code
       })
+      flag_cache[ip] = country_code
       return country_code
   except:
+    flag_cache[ip] = None
     return None
 
 


tomtor:
@ppigazzini @linrock I wonder if the current caching of the flag in MongoDb makes sense.
If we only look up flags when (re)starting fishtest the load on the flag server will be low?

Anyway, shouldn't we periodicly flush the MongoDb flag cache?

ppigazzini:
@tomtor we restart fishtest when updating PROD with a merged PR. As alternative we could schedule a daily script (with `cpulimit` to lower the load if necessary).

ppigazzini:
@tomtor DEV updated (but I don't know how test it), can I merge this and update PROD?

tomtor:
@ppigazzini I mean if we stop using the cache in MongoDb then we will only query `https://freegeoip.app/json/` on a restart and flags will remain in the Python cache till the next PR restart.

If we keep using the cache in MongoDb then load on `https://freegeoip.app/json/` will be minimal, but if a flag changes we will never notice because the mongodb data is never refreshed.

tomtor:
> DEV updated (but I don't know how test it), can I merge this and update PROD?

@ppigazzini If DEV does not crash then its OK to merge for me... We can remove the MongoDb cache later if we think it serves no real purpose.

ppigazzini:
@tomtor I'm not able to login on PROD to apply this change :(

vondele:
server seems down indeed.

ppigazzini:
I wrote to @glinscott asking to restart the vps.

tomtor:
Note regarding mongodb flag cache. If we decide to keep it then we should either

- Empty it eg once a day or 
- retire entries eg older than 7 days

Adding an index on ip is also an option but the table should be small if cleaned up at regular intervals.

tomtor:
@linrock @ppigazzini https://freegeoip.app/ allows 15000 queries/hour so I suggest dropping the mongodb cache table and just caching the results in Python.

linrock:
i don't see an index on `ip` in `flag_cache`.. that means terrible performance at high load especially if that collection never gets cleaned up.

i agree it should be cleared periodically since geo IP info can change over time.

i'd suggest adding an index with a TTL as a one-line fix, so the queries are efficient and the entries auto-expire after some time (7 days sounds fine).

tomtor:
> i'd suggest adding an index with a TTL as a one-line fix, so the queries are efficient and the entries auto-expire after some time (7 days sounds fine).

Sounds good to me @linrock

glinscott:
Rebooted the server, it was shutdown due to high load:
image

We may want to put some preemptive measures in place for high load - the VPS operator reserves the right to remove us if we keep exceeding the CPU limits.  (Ideally they would limit us to the limits, but that seems not technically possible).

vondele:
@glinscott thanks for the reboot and the graph.

The good thing, looks like all the time before the peak, the load was pretty low and smooth.

ppigazzini:
Thank you @glinscott !
We already use `cpulimit` to keep in control the CPU load during the backup, unfortunately today the flag code went crazy :(

vondele:
can we put all of fishtest under cpulimit ?  Or use linux cgroups ? I'm actually somewhat surprised that a VPS is not limited automatically, but I don't know much about that topic.

ppigazzini:
The VPS is an OpenVZ7 container, perhaps there is a bit of overselling (OpenVZ is cheaper than KVM ISP offering) with rules that stop a VPS hogging the host resources.

fishtest process is limited to one core by python GIL, It was mongo that used over 300% CPU.

The simpler solution should be to tweak the mongod systemd configuration
https://unix.stackexchange.com/questions/494843/how-to-limit-a-systemd-service-to-play-nice-with-the-cpu

Issue #610, opened 2020-04-19
reschedule and new without being logged in


vondele:
Right now, pressing new or reschedule without being logged in brings one to the login window (good), but afterwards back to the main page. Would be neat if one could continue with the new/reschedule, without being sent back to main. This came up before, but seemingly needs some development.

An alternative (or additional feature) could be to keep a user logged in over browser sessions (a 'stay logged in' tick box on the login window or so), if there are no security implications.

PR #613, opened 2020-04-19
Persistent logins and http-only auth cookies

linrock:
Login sessions now persist between browser sessions (for a year) instead of until the browser window is closed if the user checks the 'stay logged in' checkbox during login.

The auth cookie is now http-only to prevent XSS attacks from stealing user sessions. For example, if you try `alert(document.cookie)` in prod, the auth cookie currently shows up there.

Partial fix for https://github.com/glinscott/fishtest/issues/610.
Still would be nice to redirect to the page you wanted to visit before the login prompt.

---


ppigazzini:

PR #618, opened 2020-04-21
Fix cpu stats calculation without a slowdown

linrock:
Should fix https://github.com/glinscott/fishtest/issues/605
See https://github.com/glinscott/fishtest/issues/605#issuecomment-617153470


ppigazzini:

PR #619, opened 2020-04-21
Bring back default values for SPSA params

linrock:
These seem like reasonable defaults:
- alpha = 0.602
- gamma = 0.101
- A = 10% expected # iterations ~ 5% number of games (2 games per iteration)

See https://github.com/glinscott/fishtest/issues/535

---
image


ppigazzini:

PR #621, opened 2020-04-21
Simplify displaying params in SPSA table

linrock:
- Use integers for numbers that start off as ints
- Right-align numbers to line up digits and decimal points
- Show less sig figs when not needed

---
Before
image

---
After
image

### DIFF
--- a/fishtest/fishtest/static/css/application.css
+++ b/fishtest/fishtest/static/css/application.css
@@ -21,3 +21,13 @@
 th {
   cursor: pointer;
 }
+
+/* tests_view page */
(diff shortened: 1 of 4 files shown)


ppigazzini:
@linrock mmm, glitch with narrowing the browser:

Untitled

previous was (unreadable) fine:

Untitled2

linrock:
thanks, the sidebar overlapping the info area is fixed now. the sidebar will go under the test info area if the browser width becomes too narrow. it would go under before too, just now it'll go under at a higher width than before.

ppigazzini:

PR #622, opened 2020-04-22
Go to /tests/run after login if logged-out user tried going there

linrock:
If a logged-out user tries making a new run or re-scheduling an existing run, after logging in they'll be redirected to the page they tried to visit before logging in. Fixes https://github.com/glinscott/fishtest/issues/610


ppigazzini:

PR #623, opened 2020-04-22
Add some space above flash (success/error) messages

linrock:
Add some space above it so it isn't displayed as stuck to the very top of the page.


ppigazzini:

PR #624, opened 2020-04-22
Show short new/base commit hashes on tests_view pages

linrock:
Edit: i changed this PR to show short hashes on tests_view pages since when copying the hashes, it's more common to use a short prefix than the entire hash. the links are removed since the `base_tag` and `new_tag` fields link to the commit already. maybe there's a better way to present this info more succinctly?

---

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -60,6 +60,8 @@
                 </tbody>
               </table>
             </td>
+          %elif arg[0] in ['resolved_new', 'resolved_base']:
+            <td>${arg[1][:7]}</td>
           %else:
             <td>${str(markupsafe.Markup(arg[1])).replace('\n', '<br>') | n}</td>
           %endif


vondele:
@linrock that's another nice improvement. 

I wonder what the best process would be to suggest other improvements (without putting any pressure on you to solve them). Should we file separate issues, or would you like to start an issue to collect 'ideas for improvement [RFC] ?'

linrock:
oops, i just realized that the `base_tag` and `new_tag` already link to the commit. maybe just showing a short hash would for `resolved_new` and `resolved_base` would be better? i assumed they linked to the branches this whole time, since they show the branch name.

linrock:
@vondele i think starting an issue with misc ideas would be good. that may also let people suggest ideas or brainstorm more freely without feeling like they need to create separate issues that may or may not get worked on.

vondele:
@linrock I agree that that might be the best. Let me start it.

linrock:
i've opened an "ideas for improvement [RFC]" issue here: https://github.com/glinscott/fishtest/issues/625

ppigazzini:
Tested ok on DEV.

PR #626, opened 2020-04-22
Disable base branch/signature fields for SPSA tests

linrock:
The base branch/signature should always match the test branch/signature for SPSA tests.

In this PR, when a user selects an SPSA test type, the base branch/signature fields are disabled and match whatever the user puts into the test branch/signature fields. This change prevents people from submitting invalid SPSA tests through the UI. Users can also safely do less work by not having to copy/paste the same info into two sets of fields.

Fixes https://github.com/linrock/Stockfish/commit/a548e8e66f544fba3b174c27d27a9a939e5a5759#commitcomment-38631870

---

image

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -122,18 +122,18 @@
              value="${args.get('new_tag', '')}" ${'readonly' if is_rerun else ''}>
 
       <label class="field-label">Base branch</label>
-      <input type="text" name="base-branch"
+      <input type="text" name="base-branch" id="base-branch"
              value="${args.get('base_tag', 'master')}" ${'readonly' if is_rerun else ''}>
     </div>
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
Tested ok on DEV.

vondele:
just noticed a small nit on prod, after clicking spsa and going back to sprt, the base branch doesn't revert back to master and master bench. Maybe that's not ideal.

linrock:
i can fix setting the master branch and bench back to what it was originally at when switching back to SPRT. should the base almost always be master branch/bench for SPRT?

vondele:
> i can fix setting the master branch and bench back to what it was originally at when switching back to SPRT. 

thanks!

> should the base almost always be master branch/bench for SPRT?

no, not always, exceptionally people will tests against a different branch (like a pull request branch of a passed patch). However, master is a good default.

Issue #347, opened 2019-01-02
Accidentally submitting multiple of the same test


Kurtbusch:
So this issue seems to be exclusive to me, but I've noticed that occasionally my tests get submitted twice or three times when I click submit. I'm not sure if this is caused by me somehow double/triple-clicking the submit button (one theory is that I subconsciously reclick the submit button if the page hasn't loaded after a few seconds - I haven't caught myself doing this though), or possibly by my bad internet. 

Usually I catch these double-submissions right away and delete the duplicate but twice I haven't, and although me or someone else has caught it before running to completion they've still drained resources that could've been used on another distinct patch.

The simplest solution I can think of is a simple 3-second minimum between test submissions from the same user (or maybe preventing successive submissions with the exact same patch name and notes from same user). Not sure how easy one of those would be to implement - this isn't really a big issue so we shouldn't go out of our way to "fix" it but if it's an easy fix then why not.

PR #629, opened 2020-04-22
Prevent 'create new test' form creating duplicate tests

linrock:
If you click the "Submit test" button more than once or hit enter to submit more than once before the test submission goes through, you might create duplicate tests. This PR prevents that from happening through the UI by disabling the form and the submit button after the form is submitted the first time.

Fixes https://github.com/glinscott/fishtest/issues/347

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -113,7 +113,10 @@
           SMP (LTC)
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


ppigazzini:
Tested ok on DEV.

snicolet:
I think there might be an issue with the "Submitting test" button, see https://github.com/glinscott/fishtest/issues/647

PR #630, opened 2020-04-22
When switching away from SPSA test type, use initial base values

linrock:
When switching from SPSA to another test type, this sets the base branch and signature back to their initial values when the page was loaded. Usually, this will be master and master bench.

Fixes https://github.com/glinscott/fishtest/pull/626#issuecomment-618026692


linrock:
fixed so that the test/base fields stay the same if you change test types for a re-run.
see https://github.com/glinscott/fishtest/pull/631#issuecomment-618064510

ppigazzini:
PROD updated, thank you @linrock ;)

Issue #625, opened 2020-04-22
Ideas for improvement [RFC]


linrock:
This issue is for misc ideas for improving the site. Feel free to suggest anything, big or small, or brainstorm ideas here.

Some ideas for starters:

- Real-time updates on the homepage. I find I sometimes refresh the homepage a lot when checking up on the status of tests. Would be nice to have the option of seeing changes show up in real-time without manually refreshing the page.
- Link to "My tests" in the sidebar. If you're on a tests_view page, there's no direct way to get back to the list of your tests.
- Show the status of a test prominently on tests_view. There's no indicator that a test has finished other than the "Purge" button on the right. If a test is running, there's a "Stop" button there. It'd be nice to not have to think, and see right away the status of a test.
- Unit tests for statistical calculations. The only guarantee that these stats are correct is the fact that Stockfish's ELO goes up with patches from green LTC tests. We have no other way of guaranteeing that the stats calculations are correct afaik.
- Save entered fields on tests_run if the test submission fails. Right now if a test submission fails, ALL fields are cleared and you have to start over. Better would be if all fields were left as is.
- Prevent submitting tests_run form if fields are obviously missing or invalid. Catch mistakes right away instead of letting obviously invalid data get sent to the server.

vondele:
all good ideas :-) 

One slightly more complicated: to make things more searchable, can we make edges between tests ? E.g. a reschedule could result in an new field 'rescheduled as: ....' and 'rescheduled from: ...' in the detailed info of a test. In this way one could easily find the STC result of an LTC tests or the other way around. Alternatively one could search on the git SHA which would provide similar functionality.

linrock:
@vondele that makes sense. maybe adding a `rescheduled from` row to start with?

---
![image](https://user-images.githubusercontent.com/208617/80026341-fdc65c00-84af-11ea-9c32-e1ec3f011a3b.png)

![image](https://user-images.githubusercontent.com/208617/80026649-77f6e080-84b0-11ea-83a9-7cc0e689cfda.png)

---

search would generally be useful. once we have enough ideas for what search should be able to do, we can work towards making it happen

vondele:
yes, we can start with a `rescheduled from` field only (that can be empty of course). The examples are possible layouts?  I have no strong preference.

Should we add search ideas to the open search issue?

linrock:
[...]

PR #631, opened 2020-04-22
Hide master diff link if base branch is same as stockfish master

linrock:
The master diff link now only shows up if:

- the base branch is not the same as stockfish master
- we don't know if the base branch is the same (old runs)

See https://github.com/glinscott/fishtest/issues/625#issuecomment-618030661

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -123,8 +123,11 @@
       %endif
     </div>
   %endif
-  <a href="https://github.com/official-stockfish/Stockfish/compare/master...${run['args']['resolved_base'][:7]}"
-     target="_blank" rel="noopener">Master diff</a>
+
+  %if not run.get('base_same_as_master'):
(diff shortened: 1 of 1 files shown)


ppigazzini:
When rescheduling the fields are blocked, but after pushing another test type button the master's fields are reset to default and writable.

ppigazzini:
@linrock for very old tests it doesn't work perfectly see PROD and DEV (updated w/ with this PR) 
https://tests.stockfishchess.org/tests/view/515e5f488f0c3e17002f7d64
https://dfts-0.pigazzini.it/tests/view/515e5f488f0c3e17002f7d64

linrock:
@ppigazzini yep, that's expected behavior for the link to show up there, isn't it? old tests are from before we started checking the master diff server-side so there's no way to know if the base and master branch are different for those. that's why there's no message, and it's safer to leave the link there if someone wants to manually check the diff.

ppigazzini:

PR #632, opened 2020-04-22
Homepage visits no longer cause writes to runs collection

linrock:
This is a code cleanup that changes the behavior of the homepage so that it doesn't write to runs when a user visits the page. This enables:

- routing requests by `run_id` to multiple pserves. this allows scaling server load over multiple cores when routing to pserves by `run_id` since each pserve would need to consistently write to the same subset of runs. this isn't possible if the homepage writes to runs since homepage requests are not routable by `run_id`. (see https://github.com/glinscott/fishtest/issues/133#issuecomment-526931028)
- speeding up the homepage and lowering overall CPU usage when combined with future code changes where homepage visits will no longer need to loop through all unfinished tasks.

Currently, visiting the homepage writes to the runs collection by:

- saving whether tests are finishes by looping through all tasks of all unfinished runs.
- running auto-purge for tests that have finished.

This also fixes a TODO in the code comments by running auto-purge when the run is finished, and not when it's first viewed on the homepage after it's finished:

https://github.com/glinscott/fishtest/blob/master/fishtest/fishtest/views.py#L1167-L1169
> Auto-purge runs here (this is hacky, ideally we would do it when the run was finished, not when it is first viewed)

this PR also adds more tests to confirm the new behavior.

one way to manually confirm if this is working is to note that a run will now be marked as finished when the last task is complete, and not when someone visits the homepage after the last task is complete.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -18,8 +18,15 @@ from fishtest.userdb import UserDb
 from fishtest.actiondb import ActionDb
 
 import fishtest.stats.stat_util
+from fishtest.util import (
+  calculate_residuals,
+  format_results,
+  post_in_fishcooking_results,
+  remaining_hours
+)
 
 last_rundb = None
+
 
 class RunDb:
   def __init__(self, db_name='fishtest_new'):
@@ -259,10 +266,58 @@ class RunDb:
       if task['active'] and task['last_updated'] < old:
         task['active'] = False
 
-  def get_unfinished_runs(self):
+  def get_unfinished_runs(self, username=None):
     with self.run_cache_write_lock:
-      return self.runs.find({'finished': False},
-                            sort=[('last_updated', DESCENDING)])
+      unfinished_runs = self.runs.find({'finished': False},
+                                       sort=[('last_updated', DESCENDING)])
+      if username:
+        unfinished_runs = [r for r in unfinished_runs if r['args'].get('username') == username]
+      return unfinished_runs
+
+  def aggregate_unfinished_runs(self, username=None):
+    unfinished_runs = self.get_unfinished_runs(username)
+    runs = {'pending': [], 'active': []}
+    for run in unfinished_runs:
+      state = 'active' if any(task['active'] for task in run['tasks']) else 'pending'
+      runs[state].append(run)
+    runs['pending'].sort(key=lambda run: (run['args']['priority'],
+                                          run['args']['itp']
+                                          if 'itp' in run['args'] else 100))
+    runs['active'].sort(reverse=True, key=lambda run: (
+        'sprt' in run['args'],
+        run['args'].get('sprt',{}).get('llr',0),
+        'spsa' not in run['args'],
+        run['results']['wins'] + run['results']['draws']
+        + run['results']['losses']))
+
+    # Calculate but don't save results_info on runs using info on current machines
+    cores = 0
+    nps = 0
+    for m in self.get_machines():
+      concurrency = int(m['concurrency'])
+      cores += concurrency
+      nps += concurrency * m['nps']
+    pending_hours = 0
+    for run in runs['pending'] + runs['active']:
+      if cores > 0:
+        eta = remaining_hours(run) / cores
+        pending_hours += eta
+      results = self.get_results(run, False)
+      run['results_info'] = format_results(results, run)
+      if 'Pending...' in run['results_info']['info']:
+        if cores > 0:
+          run['results_info']['info'][0] += ' (%.1f hrs)' % (eta)
+        if 'sprt' in run['args']:
+          sprt = run['args']['sprt']
+          elo_model = sprt.get('elo_model', 'BayesElo')
+          if elo_model == 'BayesElo':
+            run['results_info']['info'].append(('[%.2f,%.2f]')
+                                              % (sprt['elo0'], sprt['elo1']))
+          else:
+            run['results_info']['info'].append(('{%.2f,%.2f}')
+                                              % (sprt['elo0'], sprt['elo1']))
+    return (runs, pending_hours, cores, nps)
+
 
   def get_finished_runs(self, skip=0, limit=0, username='',
                         success_only=False, yellow_only=False, ltc_only=False):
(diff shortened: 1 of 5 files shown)


tomtor:
This moves a lot of essential code, making it harder to view the git history of that moved code.
I don't think that's an issue, but this might be also a good time to auto-format all code or at least change the indentation of the code from 2 to the standard 4 spaces?  That should be done in a separate PR.

vdbergh:
I cannot claim I fully understand this PR, but I get that currently viewing the homepage would change the database. Avoiding this seems like a good idea for the maintainebility of the code. A view should be a view and nothing else.

ppigazzini:
I opened #634 to collect idea about the fishtest style guide :)

ppigazzini:
@linrock DEV updated, home page doesn't reflect the tests results.

linrock:
@ppigazzini thanks, i fixed the bug where the results weren't updating. i tested auto-purge locally a few times and it kicks in for me and purges bad tasks when the last task finishes.

this is what i did to test auto-purge:

1. create an approved run, run a worker against dev and let it finish a task
2. manually edit the task from the console with some bad stats (250 wins, 0 losses, 0 draws)
3. restart the webserver because it's caching the run from before the edit, since each python process has its own local run cache
4. after the restart, you can see the task with bad stats
5. finish a few tasks with the same worker, then switch to a different worker to finish the run
6. when the last task finishes, the first task with bad stats gets purged

linrock:
@tomtor auto-formatting the code later sounds good. i've used https://github.com/psf/black at work for auto-formatting python code before. it's not bad, but maybe there are other options worth considering

ppigazzini:
@linrock to trigger auto purge I simply create some tests with ultra STC (eg 02+0.008), windows workers have many time losses and the p-value is always under the threshold. 
A SPRT test is finished just now, the auto purge is not kicked in, the test is in "pending" state, see
https://dfts-0.pigazzini.it/tests/view/5ea3191dce12ddb4027881cc

```
LLR: -2.97 (-2.94,2.94) {-0.50,1.50}
Total: 67752 W: 28582 L: 28917 D: 10253
Ptnml(0-2): 6058, 4417, 13197, 4210, 5994
```

linrock:
ah thanks for the info. i was only testing the # games test type. i think i've found the bug. i'll test the fix out locally and let you know when it's fixed

linrock:
ok, the bug where SPRT tests don't finish should be fixed now

ppigazzini:
@linrock auto purge for SPRT works now, test continuing.

ppigazzini:
@linrock bug, it could be unrelated to this PR, though.
I created a Fixed test with 60k games, after approving it I lowered the games to 10k, now the test is at 15k games finished and counting.

linrock:
yea the task slots are created when the test is created. i don't know if lowering the # games after the test is running will delete the extra slots. if not, then it probably should.

ppigazzini:
Only the framework tester can try to lower the # games :)
Perhaps should be  better to accept only a new value that raises the # games to avoid a strange behavior.

ppigazzini:
@linrock Fixed test auto purged at 60k games finished but now is still waiting in "pending" state, see
https://dfts-0.pigazzini.it/tests/view/5ea356f8973d7136fa5d8be9

linrock:
lol nice, negative numbers of games 😆

i looked through the code (in master) and found that modifying the number of games on the run to be lower than the number of played games will break the run if auto-purge finds tasks to purge after the last task on the run is finished.

no tasks are deleted if you lower the number of games. also, no new tasks will be created in the run if auto-purge successfully purges a task and the number of played games in the run is more than the `num_games` value set on the run.

when this happens, the run is stuck in limbo. the run doesn't get marked as finished because auto-purge found tasks to purge, but since it doesn't create any new pending tasks for workers to pick up, the run will no longer get updated and will be stuck forever.

ppigazzini:
@linrock the new Fixed games test was correctly auto purged and now is completing the deleted games :)
https://dfts-0.pigazzini.it/tests/view/5ea36a6c973d7136fa5d8bff

ppigazzini:
@linrock Fixed test finished with a strange behavior
https://dfts-0.pigazzini.it/tests/view/5ea36a6c973d7136fa5d8bff
```
ELO: 1.18 +-3.5 (95%) LOS: 74.8%
Total: 32400 W: 13681 L: 13571 D: 5148
Ptnml(0-2): 2852, 2110, 6197, 2158, 2883
```

Valid=162 games
Purged=36 games
Played after purge=48 games (why not 36???)

I saw live that after auto-purge were scheduled more games than expected (by me).

linrock:
hmm strange.. i'll try to reproduce this bug locally and hunt it down. i wonder if a similar bug happens in master when a lot of tasks on a run are consistently bad and get auto-purged.

ppigazzini:
DEV on master, rescheduled the same Fixed 30k @ 0.1+0.008
https://dfts-0.pigazzini.it/tests/view/5ea3784e87951ccc8383440a

linrock:
in dev, i triggered auto-purge by visiting the homepage a couple times when i saw the games had reached the 30k limit. as of right now, the games count is 34k and going up. since this bug is also happening in master, there's probably a bug in the worker purging code that's unrelated to this PR

ppigazzini:
@linrock bug confirmed also in master, scheduled 16 tasks more than necessary. The first task purged was moved to number 137 instead than 121
https://dfts-0.pigazzini.it/tests/view/5ea3784e87951ccc8383440a
```
ELO: -0.59 +-3.4 (95%) LOS: 36.6%
Total: 34250 W: 14430 L: 14488 D: 5332
Ptnml(0-2): 3009, 2230, 6704, 2174, 3008
```

linrock:
thanks, i'll look further into this bug and fix it

ppigazzini:
@linrock DEV rebuilt with this PR, I loaded other tests at USTC (SPRT, SPSA, Fixed)

ppigazzini:
@linrock Fixed just auto purged, 14 tasks purged, the first task purged was moved to n. 133 (119+14) instead that n. 120 (119+1)

linrock:
@ppigazzini auto-purge bug is now fixed. it was indeed a problem with purge_run where if auto-purge successfully purges tasks multiple times in a row, it will create more tasks than you expect.

linrock:
the expected behavior for a running `# games` test now is that:

1. when the last task completes, it runs auto-purge without you having to visit the homepage
2. auto-purge should create just the right number of tasks now
3. repeat 1 and 2 until auto-purge doesn't purge any more tasks
4. the run is marked as finished without you visiting the homepage

vondele:
> Only the framework tester can try to lower the # games :)
> Perhaps should be better to accept only a new value that raises the # games to avoid a strange behavior.

These observations might be unrelated to this PR, but given the above remark:

* I think right now the only thing that 'works' is increasing the number of games for SPRT tests. Or, said differently, that is the only case I have seen so far.
* Presumably increasing works for numgames as well, but I've never seen anybody doing that. I could imagine that decreasing a numgames test would be something one could do if it is clear a large number of games is not needed to get a meaningful result, or if the number of games is excessive.
* I think with SPSA increasing/decreasing is more tricky, as the A parameter is not modified, and maybe should not be allowed at all.

ppigazzini:
@linrock tests results on DEV:
- Fixed games test went in "finished" state without triggering auto-purge (66% tasks with red residuals, p-value=0.00), see
https://dfts-0.pigazzini.it/tests/view/5ea40d53a804a3855ce143fe
- SPRT test triggered auto-purge, see
https://dfts-0.pigazzini.it/tests/view/5ea40d67a804a3855ce14400

I started new tests to double check the behavior.

linrock:
@ppigazzini no worries, in the code auto-purge is set to only run for runs with one thread. i was wondering why that is.

vdbergh:
@linrock The chi2 tests if all machines behave the same. Long ago SMP tests were influenced too much by HW differences so that the chi2 test was useless in that case. Now I think it would be safe to enable purging also for SMP tests.

vdbergh:
@vondele I wonder if for SMP tests, tasks cannot be simply created on the fly so that setting a maximum number of games is not necessary.

ppigazzini:
@linrock Fixed test at 1 thread auto-purged correctly :)
https://dfts-0.pigazzini.it/tests/view/5ea437c0a804a3855ce14409

Please, enable the auto purge also for SMP :)

vondele:
@vdbergh you mean SPRT tests?

vdbergh:
@vondele Yes... I am editing my comment.

linrock:
@ppigazzini ok, i've removed the 1-thread check for auto-purge, so auto-purge should now run for SMP tests.

vondele:
@vdbergh yes, if it is technically feasible, auto-increasing the maxgames for SPRT tests would be perfectly fine with me. I'll add it toe the RFC issue as a feature. Otherwise this PR becomes hard to follow.

ppigazzini:
@linrock started a couple of fast SPM tests.

ppigazzini:
PROD updated, great job @linrock :)

PR #633, opened 2020-04-23
Set page title on test pages to test type+branches

linrock:
If you have a lot of stockfish test browser tabs open, they all say `Stockfish Testing Framework` which makes it hard to look for a particular test you might have open.

On tests pages, depending on the test type, the browser tab titles will be more descriptive:

- SPRT testBranchName vs master
- SPSA tuningBranchName
- 5000 games - anotherTestBranch


ppigazzini:

PR #635, opened 2020-04-23
Adjust time estimate

vondele:
* rename parse_tc to estimate_game_duration: as that is what it does
* in this routine, use the current average number of moves per game (68),
  as well as the percentage of time used (92).
* Use the correct reference nps to adjust tc in views, this must match the value in the worker 1.6Mnps
* take threading of a test into account to estimate game rate.

These changes should result in more accurate estimates for games/minute and duration.

It will also affect the displayed CPU hours contribution attributed to a user,
as that is estimated from the number of games played. The latter quantity is unaffected.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -148,20 +148,12 @@ def delta_date(date):
     delta = 'Never'
   return delta
 
-
-def parse_tc(tc):
+def estimate_game_duration(tc):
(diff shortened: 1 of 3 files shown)


ppigazzini:

vondele:
Thanks @ppigazzini ! You're doing fantastic work in this whole fishtest overhaul effort!

PR #639, opened 2020-04-24
Update page titles to match browser tab names

linrock:
So when viewing a test page, the page title shows up like this:

- SPRT testBranchName vs master
- SPSA tuningBranchName
- 5000 games - anotherTestBranch

Instead of always like this regardless of test type:

- testBranchName vs testBranchName


ppigazzini:

PR #640, opened 2020-04-25
Link to your tests in the sidebar for logged-in users

linrock:
So you can get back to the list of your tests from any page on the site. Right now, most pages have no direct way of getting back to your tests.

See https://github.com/glinscott/fishtest/issues/625#issuecomment-619228174


ppigazzini:

PR #641, opened 2020-04-26
Enable modifying auto-purge on tests_view pages

linrock:
Adds a checkbox for auto-purge when you go to a test page so you can change auto-purge on running tests instead of just upon test submission. See https://github.com/glinscott/fishtest/issues/625#issuecomment-619376772 for original suggestion.

---


ppigazzini:

PR #645, opened 2020-04-27
Limit concurrent flag queries

tomtor:
Guard against slow or unavailable flag server

See https://github.com/glinscott/fishtest/issues/553#issuecomment-619777152

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -10,6 +10,8 @@ from fishtest.stats.stat_util import SPRT_elo
 
 WORKER_VERSION = 73
 
+flag_cache = {}
+
 
 def strip_run(run):
   run = run.copy()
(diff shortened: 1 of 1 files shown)


noobpwnftw:
I think one more thing can be done regarding to flags: IPv4 address blocks has to be no less than /24, so you can assume they have the same origin, thus reducing the number of duplicated queries/caches.

tomtor:
> I think one more thing can be done regarding to flags: IPv4 address blocks has to be no less than /24 in order to advertise on the Internet, so you can assume any IP under the same class C network is from the same origin, thus reducing the number of duplicated queries/caches.

Yes, that's true. It would however add considerable code for a limited gain?

ppigazzini:
@tomtor on DEV with the first PR version the server dropped the flags, with latest version the flags are still there (reloaded a DB backup), the log is clean, the geoip service is still down.
Can I merge?

tomtor:
@ppigazzini yes ok to merge. Thanks.

ppigazzini:

ppigazzini:
GeoIP service restarted working
https://freegeoip.app/json/

Issue #647, opened 2020-04-27
"Submit" button on the new test page


snicolet:
When there is an error in a test, we get an error message in red, which is fine. But even after correcting the error, if we hit the "back" button on the browser to get back to the new test page with to correct one field in the form, the "Submit" button is still showing "Submitting test" and can not be used.

Steps to reproduce:

1. Go to "new test" page
2. Fill a stupid branch name in the form, for instance "dfhfdhdfh"
3. Click on "Submit"
4. On the "Cannot find branch in developer repository" error page, use the back buttun in the browser to get back on the "new test" page

=> expected behavior: I can re-use the fields in the form, for instance correcting the branch name, and click on "Submit" again
=> observed behavior: I can re-use the fields in the form, but the "Submit" button is disabled.

This may be a consequence of https://github.com/glinscott/fishtest/commit/469fcddf6d62c6a55bd54eea841e355c2d5d6cd9
and https://github.com/glinscott/fishtest/pull/629

EDIT : more generally, the change in https://github.com/glinscott/fishtest/commit/469fcddf6d62c6a55bd54eea841e355c2d5d6cd9 prevents us to use the following workload for submitting a series of similar tests on the same branch:

a. hit "Submit" to submit a test on branch A
b. push a code variant to branch A
c. use back button in browser
d. hit "Submit" to submit a test on branch A
e. push a code variant to branch A
f. use back button in browser
g. hit "Submit" to submit a test on branch A
h. push a code variant to branch A
i. use back button in browser
...

ppigazzini:
@snicolet the bug is fixed in my test; your workload is still disrupted, though.

snicolet:
@ppigazzini @linrock 

Thanks! :)

PR #649, opened 2020-04-27
Enable submit test button if arriving from browser back button

linrock:
When using the browser back button to get to the 'create new test' form, the submit button might be stuck in the 'Submitting test...' state.

This PR should fix that issue in Firefox. Chrome didn't seem to have the issue. Fixes https://github.com/glinscott/fishtest/issues/647


ppigazzini:

PR #628, opened 2020-04-22
Server changes to set the batch size for sprt tests.

vdbergh:
Server changes to set the batch size for sprt tests.
    
In this PR the batch size is set to 8 which seems like a reasonable value. As the batch size has to be a divisor of the chunk size, the latter was changed to 200 (formerly 250). To compensate  for the smaller chunk size the initial maximum number of games for an sprt test is changed to 2000 times the chunk size (formerly 1000 times the chunk size).
    
 One unittest is updated.

PR #650, opened 2020-04-28
Fix TC parsing for itp

vondele:
fixes the 40/10 use case, using the estimate_game_duration call.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -20,6 +20,7 @@ from fishtest.actiondb import ActionDb
 import fishtest.stats.stat_util
 from fishtest.util import (
   calculate_residuals,
+  estimate_game_duration,
   format_results,
   post_in_fishcooking_results,
   remaining_hours
@@ -387,7 +388,7 @@ class RunDb:
       itp = 1
     elif itp > 500:
       itp = 500
-    itp *= math.sqrt(float(run['args']['tc'].split('+')[0]) / 10)
+    itp *= math.sqrt(estimate_game_duration(run['args']['tc'])/estimate_game_duration('10+0.1'))
     itp *= math.sqrt(run['args']['threads'])
     if 'sprt' not in run['args']:
       itp *= 0.5


vondele:
supposed to fix https://github.com/glinscott/fishtest/pull/628#issuecomment-620275030 

there might be other cases in fishtest where that TC (40/10 or 40/10+0.1) could cause trouble.

ppigazzini:
@vondele DEV updated, with TC=40/2 server and worker code don't crash, I will do other tests:
```
CPU factor : 0.818557 - tc adjusted to 40/1.637
Running master vs master
['/home/usr00/_git/__test_folder00/fishtest/worker/testing/cutechess-cli', '-repeat', '-rounds', '100', '-games', ' 2', '-tournament', 'gauntlet', '-pgnout', 'results-d9722fa9-6019-4164-a832-9ba7b24ebe01.pgn', '-site', 'https://tests.stockfishchess.org/tests/view/5ea7e21fbf51299e10a01e9d', '-event', 'Batch 0: master vs master', '-srand', '1099351279', '-resign', 'movecount=3', 'score=400', '-draw', 'movenumber=34', 'movecount=8', 'score=20', '-concurrency', '3', '-openings', 'file=noob_3moves.epd', 'format=epd', 'order=random', 'plies=16', '-engine', 'name=New-bb5589b', 'cmd=stockfish_bb5589b829b79d7c60a820d4b1634dccbc4bbb3f', 'option.Hash=16', '-engine', 'name=Base-bb5589b', 'cmd=stockfish_bb5589b829b79d7c60a820d4b1634dccbc4bbb3f', 'option.Hash=16', '-each', 'proto=uci', 'tc=40/1.637', 'option.Threads=1']
TC limit 818.5570987062196 End time: 2020-04-28 10:16:08.816085
```

@vdbergh @tomtor do you view any problems adding this TC format?

vondele:
@ppigazzini this format has been used in the past https://tests.stockfishchess.org/tests/view/59e1322d0ebc5916ff64c095 so it is actually a regression. We have used it very rarely, basically only when testing the TC code.

vdbergh:
@ppigazzini Since you are CC'ing me... I don't see any issues. From the point of view of the stats code the TC is irrelevant.

tomtor:
I am not really familiar with this part of the code. I see no issues.

ppigazzini:

ppigazzini:
@vondele please add a little section with the TC formats accepted by fishtest
https://github.com/glinscott/fishtest/wiki/Creating-my-first-test

I was unaware of TC 40/10 (I thought a typo for +) or TC 40/10+0.1

vondele:
I added a note in the wiki, not a section.

ppigazzini:
Perfect :)

PR #655, opened 2020-04-29
Enforce even sized worker updates.

vdbergh:
The current worker cannot handle tasks with an odd number of games.
On the other hand it will (should...) never send odd sized updates
itself. Enforce this contract server side.

Some unit tests updated.

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -553,12 +553,14 @@ class RunDb:
     # Guard against incorrect results
     count_games = lambda d: d['wins'] + d['losses'] + d['draws']
     num_games = count_games(stats)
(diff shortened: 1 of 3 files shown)


ppigazzini:
DEV updated and started some workers.
I will return online in a couple of hours, so feel free to submit/approve tests on DEV
(user01/user01 to submit; user00/user00 to approve)

ppigazzini:
DEV rebuilt.

vdbergh:
@ppigazzini I needed to push again since there was still a failing unit test. No effect on the code.

BTW I think the unit tests are great!

ppigazzini:
@vdbergh DEV is running this PR for 2 days without problems. Can I merge it?

vdbergh:
@ppigazzini Sure. I think it is non-controversial.

ppigazzini:

Issue #652, opened 2020-04-28
"Missing required option when submitting spsa tests"


vdbergh:
Steps to reproduce:

Goto "Create New Test page"
Click SPSA button
Fill out tune options
Set "Test branch" to master
Click submit

Probably not all these steps are necessary but I have no time to debug further now.

vondele:
seems to require explicitly the bench number...

vdbergh:
@vondele The field is grayed out however.

vondele:
not for the test branch ?

vdbergh:
No. The usual "Defaults to last commit message".

I did some debugging earlier and it is really `base_signature` that is blank. I guess for spsa tests it should always be set to 'new_signature'.

vondele:
I think the 'Defaults to last commit message'  is not working, I appear to need to put an explicit number there. The greyed out fields are on purpose so that base_signature will be new_signature automatically. 

https://github.com/glinscott/fishtest/pull/626
there was a fixup
https://github.com/glinscott/fishtest/pull/630

vdbergh:
'Defaults to last commit message' works perfectly for sprt tests. 

If it is not intended to work for spsa tests then I think the field should be blank.

vondele:
I think it should work :-) just buggy, and filling the field is a workaround.

PR #656, opened 2020-04-30
Fixes some issues with test submission. Including #652.

vdbergh:


ppigazzini:
DEV updated

- this works:
https://github.com/glinscott/fishtest/blob/a4bd1fc89118c9ab12fa71c031a0e9a1ad2cd69c/fishtest/fishtest/views.py#L343-L344
- I don't know how to trigger this, please try yourself and let me know if it works:
https://github.com/glinscott/fishtest/blob/a4bd1fc89118c9ab12fa71c031a0e9a1ad2cd69c/fishtest/fishtest/views.py#L361-L363

vdbergh:
@ppigazzini You can trigger it by for example blanking out the tc field.

ppigazzini:
@vdbergh thank, it works.
However the true problem it's that the signature auto complete isn't working anymore in master for all test types (SPRT, Fixed, SPSA).

vdbergh:
Actually I am getting 

`Invalid URL '/commits/master': No schema supplied. Perhaps you meant http:///commits/master?`

I have no idea where this comes from.

vdbergh:
@ppigazzini  The latest master has no signature. So the error message you are getting is correct.

ppigazzini:
> Actually I am getting
> 
> `Invalid URL '/commits/master': No schema supplied. Perhaps you meant http:///commits/master?`
> 
> I have no idea where this comes from.

@vdbergh I switched DEV back and forth from master and PR, perhaps this disrupted your test, try now (DEV is running the PR).
I thought that the signature was mandatory in Stockfish commits. I will check if fishtest is able to work fine without a base/test signature.

vdbergh:
I had to logout and login again to get rid of the message 

`Invalid URL '/commits/master': No schema supplied. Perhaps you meant http:///commits/master?`

This points to another bug (but not this PR since it didn't change anything in the way the Github api is invoked).

vdbergh:
Ah messages crossed. Perhaps it was not logout/login then that fixed it. It should still be understood though.

tomtor:
> Ah messages crossed. Perhaps it was not logout/login then that fixed it. It should still be understood though.

You will get that message in `get_sha` when `data['tests_repo']` is empty:
https://github.com/glinscott/fishtest/blob/a4bd1fc89118c9ab12fa71c031a0e9a1ad2cd69c/fishtest/fishtest/views.py#L375

vdbergh:
Thanks. I now catch the exception and give slightly less cryptic message.

ppigazzini:
DEV updated.

ppigazzini:
@vondele for any SF commit without signature a fishtest tester takes a scare. If possible please make the signature mandatory :)

Issue #552, opened 2020-04-03
Stats shallow copy in games.py


shooreek:
We save stats to `old_stats` variable like this:

https://github.com/glinscott/fishtest/blob/0c076f225bd6f0e15f2d0ddd2de80e1dd5e206f4/worker/games.py#L570

`stats` defined like this:

https://github.com/glinscott/fishtest/blob/0c076f225bd6f0e15f2d0ddd2de80e1dd5e206f4/worker/games.py#L413

`result['stats']` won't be fully copied using `copy()` method because it does a shallow copy, list under `pentanomial` key will be just referenced. If the intention was to create a full/deep copy, then another copying methods should be used.

I'm not so familiar with the code and cannot tell if this really causes some problems, probably it is harmless due to the slicing:

https://github.com/glinscott/fishtest/blob/0c076f225bd6f0e15f2d0ddd2de80e1dd5e206f4/worker/games.py#L257

linrock:
The shallow copy here doesn't look like it'll be an issue since `old_stats['pentanomial']` is only ever overwritten or read and its individual list elements are never modified.

vdbergh:
Likewise the individual entries of result[‘pentanomial’] are never changed so that old_stats[‘pentanomial’] will not change inadvertently. Still, this code seems very fragile and deepcopy instead of copy would be more robust.

vdbergh:
Using a global variable old_stats seems rather dangerous anyway. It would be nice to eliminate it altogether.

vdbergh:
I am now slowly cleaning up the worker code. I have eliminated the ugly, dangerous and confusingly named global old_stats variable.

The code states that old_stats is global "because it is shared across all threads". But AFAICS the worker is not multi-threaded. All concurrency is handled by cutechess-cli. So I suspect the old_stats variable was a left over from times when the worker handled concurrency itself. @tomtor ?

tomtor:
> So I suspect the old_stats variable was a left over from times when the worker handled concurrency itself.

I do not think that the worker has ever been multithreaded, but I and @ppigazzini did not work on it in the early fishtest years...

vondele:
The comment was added long ago in 517559f0 and at that point there was threading in the worker it seems:
```
  th = []
  for idx in range(threads):
    th.append(threading.Thread(target=launch_cutechess, args=payload))
    th[idx].start()
```
this seem gone

vdbergh:
[...]

PR #657, opened 2020-05-01
Some code cleanup in the worker

vdbergh:
The worker code is serial and hence in principle
quite simple. Unfortunately it has grown organically
and currently it is not so easy too follow anymore.
This makes the idea of adding new types of tests
rather unattractive.

This is PR is the first timid attempt at cleaning
up some of the worker code.

This version of the worker should be functionally
equivalent to version 75. 

The most notable changes are

- The global variable old_stats is eliminated. The fact
that such a global variable exists needlessly complicates the
logic of the worker code.

- Trinomial and pentanomial frequencies are
now kept in sync for all types of tests. This simplifies
the code a lot.

- More asserts have been added.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -16,6 +16,7 @@ import traceback
 import platform
 import struct
 import requests
+import copy
 from base64 import b64decode
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


ppigazzini:
Workers updated on DEV.

vondele:
This will fix https://github.com/glinscott/fishtest/issues/552

ppigazzini:
@vdbergh workers have bee fine all day long.

ppigazzini:
@noobpwnftw code clean up. Please update your fleet at your convenience.

noobpwnftw:
@ppigazzini All workers updated to 76.

ppigazzini:
@noobpwnftw thank you, I'm starting the auto update for all workers.

PR #658, opened 2020-05-03
A quick attempt at fixing https://github.com/glinscott/fishtest/issue…

vdbergh:
https://github.com/glinscott/fishtest/issues/625#issuecomment-622998366


ppigazzini:
Tested ok on DEV, PROD updated, thnak you @vdbergh :)

Issue #659, opened 2020-05-05
Make task queue FIFO


noobpwnftw:
Factor in (last_update - start_time) of a test in calculating internal priority so that workers that just finished a batch of running tests don't get allocated to a fresh test. This also helps reducing burst downloads from Github due to task switch and prevent spending too many resources on "filler" tests.

vondele:
we had the equivalent of FIFO based on itp adjustment using start_time for a while, but people complained (rightfully so?) that new tests were not starting. A long running LTC would get all resources till it finished. 

However, if I understand correctly, you would like some scheme  in which workers switch tasks as little as possible. Some kind of affinity of workers to tests. Maybe something like... unless there is a task with an itp that is 2x of what I'm doing now, I'm going to stick with my current one (if that still needs compute). Maybe that's something that is not too difficult to implement, don't know if the server has an idea about what workers did in the past.

ppigazzini:
@noobpwnftw If I'm not wrong the Github API rate limit is the underlying problem, in that case we could try to raise the limit with an optional user authentication.

https://developer.github.com/apps/building-github-apps/understanding-rate-limits-for-github-apps/

noobpwnftw:
@ppigazzini When I run a few thousand workers(they use .netrc authentication), it is still easy to hit the limit.

tomtor:
@vondele Not sure how to add this. Currently the allocation of workers is completely fair, a new worker gets its share of workers and is not punished because it is fresh.
I agree that affinity to tests for workers can be a nice property. I'll have to think about how to code that in an elegant way, without changing the fair allocation, might take a while... ;-)

vondele:
I do like our current itp scheme, works really from the user point of view. However, I think we can violate the 'fairness' by a little bit, because it comes at a cost (many recompiles needed). Also, all tests will be treated the same, so it is still fair, it just will ramp up the number of workers assigned to a test a little more slowly. To give an example To give an example 
https://tests.stockfishchess.org/tests/view/5eb050eda460c0e4b8b9bf14   
195 workers jumped on this task, but after about 9000 games it was out. On average 30 games per worker. If we assume about 15sec compilation time (+ extra time for bench etc), that is about 2-3% overhead.

Note that even the itp changes quite quickly over time (as the LLR changes, poor tests have more workers assigned than the fairness would suggest).

noobpwnftw:
[...]

PR #660, opened 2020-05-05
Remove unused list

vondele:
seems to be a leftover from the CLOP implementation, and is not used anymore.


tomtor:
Nice catch! Thanks @vondele

ppigazzini:
PROD updated :)

vondele:
@tomtor trying to understand the code to see if it is possible to preferentially hand a task from the same test/run a worker worked on to that worker requesting a new task... (https://github.com/glinscott/fishtest/issues/659) 

admittedly, I'm not able to follow the code..

PR #662, opened 2020-05-06
Limit worker API usage

tomtor:
See #659

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 from __future__ import print_function
 
+import math
 import json
 import multiprocessing
 import os
@@ -27,7 +28,7 @@ from updater import update
 from datetime import datetime
 from os import path
 
-WORKER_VERSION = 76
+WORKER_VERSION = 77
 ALIVE = True
 HTTP_TIMEOUT = 15.0
 
@@ -80,6 +81,15 @@ def on_sigint(signal, frame):
   ALIVE = False
   raise Exception('Terminated by signal')
 
+rate = None
+
+def get_rate():
+  global rate
+  rate = requests.get('https://api.github.com/rate_limit', timeout=HTTP_TIMEOUT).json()['rate']
+  remaining = rate['remaining']
+  print("API call rate limits:", rate)
+  return remaining >= math.sqrt(rate['limit'])
+
 def worker(worker_info, password, remote):
   global ALIVE
 
@@ -90,6 +100,9 @@ def worker(worker_info, password, remote):
 
   try:
     print('Fetch task...')
+    if not get_rate():
+      raise Exception('Near API limit')
+
     t0 = datetime.utcnow()
     req = requests.post(remote + '/api/request_version', data=json.dumps(payload), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT)
     req = json.loads(req.text)
@@ -105,7 +118,11 @@ def worker(worker_info, password, remote):
     print("Worker version checked successfully in %ss" % ((datetime.utcnow() - t0).total_seconds()))
 
     t0 = datetime.utcnow()
-    req = requests.post(remote + '/api/request_task', data=json.dumps(payload), headers={'Content-type': 'application/json'}, timeout=HTTP_TIMEOUT)
+    worker_info['rate'] = rate
+    req = requests.post(remote + '/api/request_task',
+                        data=json.dumps(payload),
+                        headers={'Content-type': 'application/json'},
+                        timeout=HTTP_TIMEOUT)
     req = json.loads(req.text)
   except Exception as e:
     sys.stderr.write('Exception accessing host:\n')
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -411,6 +411,8 @@ class RunDb:
   task_time = 0
   task_runs = None
 
+  worker_runs = {}
+
   def request_task(self, worker_info):
     if self.task_semaphore.acquire(False):
       try:
@@ -454,6 +456,14 @@ class RunDb:
     if connections >= self.userdb.get_machine_limit(worker_info['username']):
       return {'task_waiting': False, 'hit_machine_limit': True}
 
+    # Limit worker Github API calls
+    if 'rate' in worker_info:
+      rate = worker_info['rate']
+      limit = rate['remaining'] <= 2 * math.sqrt(rate['limit'])
+    else:
+      limit = False
+    worker_key = worker_info['unique_key']
+
     # Get a new task that matches the worker requirements
     run_found = False
     for runt in self.task_runs:
@@ -471,6 +481,8 @@ class RunDb:
         need_tt *= max_threads // run['args']['threads']
 
       if run['approved'] \
+         and (not limit or (worker_key in self.worker_runs
+                            and run['_id'] in self.worker_runs[worker_key])) \
          and run['args']['threads'] <= max_threads \
          and run['args']['threads'] >= min_threads \
          and need_tt <= max_memory:
@@ -508,6 +520,12 @@ class RunDb:
         break
 
     self.buffer(run, False)
+
+    # Update worker_runs (compiled tests)
+    if worker_key not in self.worker_runs:
+      self.worker_runs[worker_key] = {}
+    if run['_id'] not in self.worker_runs[worker_key]:
+      self.worker_runs[worker_key][run['_id']] = True
 
     return {'run': run, 'task_id': task_id}
 


ppigazzini:
@tomtor DEV and some workers updated, perhaps should be better to test with a lower limit on DEV and raise just before merging on PROD.

ppigazzini:
DEV updated, limit = 2

ppigazzini:
@tomtor 
```
May 06 11:54:38 vps1572612706 pserve[14488]:     if worker_info not in self.worker_runs:
May 06 11:54:38 vps1572612706 pserve[14488]: TypeError: unhashable type: 'dict'

```

tomtor:
> May 06 11:54:38 vps1572612706 pserve[14488]: TypeError: unhashable type: 'dict'

@ppigazzini Was also caught by Travis. Should be ok now....

ppigazzini:
DEV up and running with limit=2

vondele:
@tomtor can you explain a little? If noob connects 4000 workers, we don't want this process to start them take 40 min right? Or is this per running test, and connecting 4000 workers with more than 10 tests in queue will be smooth?

ppigazzini:
@vondele API rate is 5000/hour (for one account).
@noobpwnftw are you using only one account?

vondele:
@ppigazzini yes, but I'm trying to say we should be careful to design the limiting such that it doesn't interfere in a bad way. I'm not yet sure what the right way of limiting would be. It actually seems hard to find a hard limit which isn't 'wrong' in some case, the only hard limit should be the github api rate. 

If we have the list of ~50 most recent tasks a worker has worked on, I think scaling the itp for each task with a simple factor depending on that task being in the recent list for each worker would just do the job (and if we want it fancy, we have this factor depend on the number of workers per user, i.e. give a users with >100 workers a strong task affinitiy).

ppigazzini:
@vondele "If noob connects 4000 workers we don't want this process to start them take 40 min".
We can speedup the fleet startup process only monitoring worker side X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, with a safety margin for the race condition.
`Note: Making a conditional request and receiving a 304 response does not count against your Rate Limit, so we encourage you to use it whenever possible.`
https://developer.github.com/v3/#conditional-requests

tomtor:
@vondele @noobpwnftw I could implement user API-credits, starting at 5000 and each worker which gets a new test decrements the credit counter. Each 60s you get 80 new credits (60*80 ~= 5000). When credits is 0 the workers get only tests they have already worked on.

Fiddling with ITP does not solve the problem when there is only 1 new runnable test or after a server restart.

vondele:
I think that's a the best solution. 

(you might need to cap the credits and lower the rate if you want to make sure the user can't use more than 5000 credits/h, since if he accumulates 5000, starts running, and gets 5000/h he can issue 10k calls in the first 1h. So probably limit to 2500, and give out at 40/min)

ppigazzini:
@tomtor please rebase with latest master

tomtor:
@ppigazzini Hmm, travis build fails but no log?

vondele:
one more question. Is a new task just 1 github api call ? 

Could it be 3 in the worst case (i.e. new worker might need master, test, and book) ?

ppigazzini:
@tomtor I create a token for the user used by PROD and DEV, I don't know if this can have caused a problem with Travis. I deleted the token before starting the commute, I'll be back at the keyboard in 90 minutes.

ppigazzini:
@vondele cutechess-cli

vondele:
@ppigazzini I assume @noobpwnftw puts that on the worker image. But maybe I'm wrong.

tomtor:
> one more question. Is a new task just 1 github api call ?

In most cases it would only fetch test?

ppigazzini:
@vondele I'm talking about general use case. I wait always that noob has finished the fleet update before starting the auto update.

ppigazzini:
@tomtor yes

vondele:
Yes, 4 might be most general. 1 most common. But I think when noob starts his fleet it will be 2 (provided some of the books are in the image, and master is new).

tomtor:
> Yes, 4 might be most general. 1 most common. But I think when noob starts his fleet it will be 2 (provided some of the books are in the image, and master is new).

The 2500 start value (5000 real limit) should handle that?

tomtor:
> I create a token for the user used by PROD and DEV, I don't know if this can have caused a problem with Travis. I deleted the token before starting the commute, I'll be back at the keyboard in 90 minutes.

I had an error in the code, I assume it was an intermittent error that it showed no log.

ppigazzini:
@vondele noob could skip the update by hand if the fishtest API credit system will work fine. Now it's mandatory because a worker will try forever triggering the GitHub ban.

vondele:
> > Yes, 4 might be most general. 1 most common. But I think when noob starts his fleet it will be 2 (provided some of the books are in the image, and master is new).
> 
> The 2500 start value (5000 real limit) should handle that?

probably would have that as a positive side effect.  

actually, the book-keeping could be avoided with @ppigazzini solution. Have the worker send upon connection to the server the current value of X-RateLimit-Remaining. If it is smaller than x we serve an old test. That would be robust, no need for the 2500/40 trick, and would work also if the user has a different limit than 5000.

tomtor:
> actually, the book-keeping could be avoided with @ppigazzini solution. Have the worker send upon connection to the server the current value of X-RateLimit-Remaining. If it is smaller than N we serve an old test. That would be robust, no need for the 2500/40 trick, and would work also if the user has a different limit than 5000.

That would be a small worker change and a minimal server change.

noobpwnftw:
I also keep getting deprecation emails about Github API use with tokens. I recommend a satellite cache sever(could be as simple as a docker image) that someone can run and configure workers to download stuff from there. In the future it might also be possible to provide compiled binaries. I don't know but it seems similar to CI.

vondele:
> That would be a small worker change and a minimal server change.

I think that would be useful. If we have both limit and remaining of 
```json
  "rate": {
    "limit": 5000,
    "remaining": 4999,
    "reset": 1372700873
  }
```
we can make a reasonable choice on the server (e.g. old test for remaining < limit / 10)

BTW, looks like the server might have a github api problem right now

ppigazzini:
DEV updated.

tomtor:
I added the worker changes.

When remaining API calls is below 40 it pauses. This should allow auto updating of @noobpwnftw worker fleet.

In addition rate info is passed to the server for `request_task`.

worker/worker.py
```diff
@@ -80,6 +82,15 @@ def on_sigint(signal, frame):
   ALIVE = False
   raise Exception('Terminated by signal')
 
+rate = None
+
+def get_rate():
+  global rate
+  rate = requests.get('https://api.github.com/rate_limit', timeout=HTTP_TIMEOUT).json()['rate']
+  remaining = rate['remaining']
```
vdbergh, line 90: Would it be useful to handle the case that the api request fails?

tomtor: The idea is to catch the exception in the main loop?

vdbergh: Ok. That seems safe...

tomtor:
I changed the limits: worker 30 (stop worker updates) en server 50 (switch to already compiled tests)

vondele:
will these limits work for normal workers (which will have a limit of 60). Answer: I think yes.

ppigazzini:
DEV and workers updated.

ppigazzini:
The 11 workers ran fine with authentication. Now I'm waiting to switch to the unauthenticated use case.

ppigazzini:
DEV updated, started 12 unauthenticated workers.

tomtor:
@ppigazzini I squashed

noobpwnftw:
@ppigazzini Please merge and raise version only for workers, then I'll update by hand. I have other modifications in the worker code that requires a local merge.

ppigazzini:
@noobpwnftw understood. I got a problem with a worker during a DEV server update on the fly, so I'm continuing some tests to be sure of iron out all problems.

tomtor:
@ppigazzini I am playing a bit and I am surprised that without authenticating I have just 26 calls left (2 running workers). 

Perhaps its better to set the limits to limit/10 (=6 for worker) and limit/5 (=12 for server)?

vondele:
or something like sqrt(limit) and 2 * sqrt(limit) which gives quite nice limits for 60 and 5000

ppigazzini:
@tomtor DEV updated.
Yesterday after updating DEV with https://github.com/glinscott/fishtest/pull/662/commits/f9813e4b733ddb2f1a60b0c8a53d08b9fa4d6a77 a couple of workers suddenly went to 16-20 calls left, never able to recover with the passing of time. I want to play asap a bit with those discontinuity. At the moment I'm running authenticated workers to finish some SPSA tests.

ppigazzini:
@tomtor Tests are OK for the whole day. At your convenience please squash and raise the version worker side.

tomtor:
@ppigazzini done, thanks for all testing! @vondele @noobpwnftw thanks for the feedback!

ppigazzini:
@noobpwnftw please update your fleet at you convenience :)
Thank you @tomtor :)

ppigazzini:
@noobpwnftw Let me know if I can start the auto update for other workers.

noobpwnftw:
Yes please @ppigazzini . I have just checked that all my workers had been updated.

noobpwnftw:
@tomtor @ppigazzini It looks like with the regular worker count(~150) the throughput appears to be less than before(My fishtest worker will switch to other tasks and look back in case of 1.failure 2.empty queue). Is that due to API limit?

vondele:
what are current limits / remaining that your workers are seeing ?

noobpwnftw:
```
Exception accessing host:
HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: /rate_limit (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f2849269da0>: Failed to establish a new connection: [Errno 111] Connection refused',))

```
```

[root@localhost ~]# sudo -u fishtest curl -v https://api.github.com/rate_limit
* About to connect() to api.github.com port 443 (#0)
*   Trying 13.250.168.23...
* Connection refused
* Failed connect to api.github.com:443; Connection refused
* Closing connection 0
curl: (7) Failed connect to api.github.com:443; Connection refused
```

Limit seems normal because workers bail early due to weird issue(:
```

{
  "resources": {
    "core": {
      "limit": 60,
      "remaining": 59,
      "reset": 1589180040
    },
    "graphql": {
      "limit": 0,
      "remaining": 0,
      "reset": 1589180735
    },
    "integration_manifest": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1589180735
    },
    "search": {
      "limit": 10,
      "remaining": 10,
      "reset": 1589177195
    }
  },
  "rate": {
    "limit": 60,
    "remaining": 59,
    "reset": 1589180040
  }
}

```

ppigazzini:
@noobpwnftw I'm trying to trigger the problem on my pc, but at the moment this script runs fine with and without authentication:
```
#!/bin/bash
for i in {0..9999}; do
  if ! (( $i % 10 )) ; then
    curl -v -n https://api.github.com/rate_limit
  else
    curl -v -n https://api.github.com/rate_limit &> /dev/null
  fi
  sleep 0.1
done
```

vdbergh:
Perhaps the worker should not generate an exception when the access to https://api.github.com/rate_limit fails ? To me it looks like an error which will often be harmless.

tomtor:
> Perhaps the worker should not generate an exception when the access to https://api.github.com/rate_limit fails ? To me it looks like an error which will often be harmless.

It will now retry (access rate_limit API) after a short while (IMO this is correct).

If we catch the exception (because we want to handle it differently), how should we proceed?

- Just drop the `rate` data from `request_task` Fishtest API?
- Pass fake rate limit data to fishtest `request_task` ?

And do we risk exceeding the API rate limit as a consequence?

vondele:
I think we should indeed retry a couple of times e.g. after 5, 10, 20, and 60s and after that proceed with remaining: 0 and limit: 5000. If we can't get the rate_limit, we have to assume we can not access github from the worker, and the worker should only work on precompiled tests. Should we also increase the number of binaries cached, since we now rely on this more and binary sizes are anyway relatively small ?

```diff
diff --git a/worker/games.py b/worker/games.py
index 3dc00fb..1427f5d 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -493,11 +493,11 @@ def run_games(worker_info, password, remote, run, task_id):
   if not os.path.exists(testing_dir):
     os.makedirs(testing_dir)
 
-  # clean up old engines (keeping the 25 most recent)
+  # clean up old engines (keeping the 50 most recent)
   engines = glob.glob(os.path.join(testing_dir, 'stockfish_*' + EXE_SUFFIX))
-  if len(engines) > 25:
+  if len(engines) > 50:
     engines.sort(key=os.path.getmtime)
-    for old_engine in engines[:len(engines)-25]:
+    for old_engine in engines[:len(engines)-50]:
       try:
          os.remove(old_engine)
       except:
```

tomtor:
> I think we should indeed retry a couple of times e.g. after 5, 10, 20, and 60s

I prefer giving up on first failure, just retry after 200 games (precompiled) batch.

PR #664, opened 2020-05-10
Show username in page titles when viewing a user's tests

linrock:
So if you have a browser tab open while viewing a user's tests, it will show their username in the tab title instead of just "Stockfish Testing Framework"


ppigazzini:

PR #665, opened 2020-05-10
Show link to original test when viewing a rescheduled test

linrock:
Implements vondele's suggestion here:
https://github.com/glinscott/fishtest/issues/625#issuecomment-617950457

So you can easily see which test a re-run came from and then get to it.
 
---


ppigazzini:
Tested OK on DEV, PROD updated. Thank you @linrock :)

PR #667, opened 2020-05-11
Fix some chars showing up as html entities in test commit msgs

linrock:
So apostrophes don't show up as `&#x27;`

Before:
new_tag `tDblpawns (Tune doubled pawns penalty using vondele&#x27;s tuning )`

After:
new_tag `tDblpawns (Tune doubled pawns penalty using vondele's tuning )`


ppigazzini:
Tested working on DEV, PROD updated, thank you @linrock :)

Issue #674, opened 2020-05-18
machine limit for yorkman


vondele:
@ppigazzini can we increase the limit for yorkman (e.g. 16). https://github.com/official-stockfish/Stockfish/issues/2683

BTW, should we increase the default limit to 16 anyway?

ppigazzini:
@vondele "yorkman" limit raised to 40 (the usual second step), I keep open this issue.

PR #676, opened 2020-05-18
Change default machine limit to 16

tomtor:
See https://github.com/glinscott/fishtest/issues/674


ppigazzini:

Issue #677, opened 2020-05-19
No tasks available, but there are?


vondele:
could there be an issue right now, I see the number of workers is low, and there is the following report on fishcooking:

https://groups.google.com/d/msg/fishcooking/YxwtpTY6u5M/Zd6zEnRgBAAJ

noobpwnftw:
My workers have the same issue - no task available.

xoto10:
If we update one of the pending jobs from prio -1 to 0 does it get any workers?

vondele:
yes

xoto10:
I updated my pawnc1 test to prio -1 some time ago (nearly 4 hours ago?) but it is still running and looks like it has received new workers fairly recently - something seems to be wrong with the select / sort of remaining tasks??

xoto10:
I updated my dblfi1 test to prio 1 and it's grabbed all the workers. Something odd happening with priority?
(I've put the test back to 0, so the workers will free up before long)

vondele:
yes, confirmed, putting another task at prio 1, and it gets all workers (i.e. also all those that are otherwise not participating). @tomtor any idea?

xoto10:
Haha, I would have chosen another stc job, but yes. At least the workers are doing something now. As a workaround we could set all active jobs to prio 1 for now.
I assume the first thing is to just stop and start the server when someone is available to do that?

vondele:
well, we should indeed manage stuff by hand till the problem has been fixed. At least it is a workaround for the time being. Let's start by raising LTC jobs to prio 1.

xoto10:
That prio 1 job is taking all the workers from the other tests, maybe set it back to prio 0?

vondele:
I'm not sure that if prio is back to 0, workers will go to other tasks or just go idle. Let's keep at least a few jobs prio 1 till things are fixed.

xoto10:
well true. But we could have looked to see. Now almost all workers are locked on LTC jobs and hardly anything will change for an hour.
I suggest putting some stc jobs to 1 if that's what we're doing, they keep the system flexible, re-allocating workers as they come free ... well, if the reallocation works

vondele:
OK, feel free to change.

xoto10:
Seems we know how to workaround for now, set all live jobs to prio 1.
Or you could set them all to 0 and see what happens.
Whatever.

xoto10:
[...]

PR #678, opened 2020-05-21
Query for unfinished_runs with only the id attribute

tomtor:
Speedup by dropping tasks from unfinished_runs query

See #677

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -271,6 +271,13 @@ class RunDb:
       if task['active'] and task['last_updated'] < old:
         task['active'] = False
 
+  def get_unfinished_runs_id(self):
+    with self.run_cache_write_lock:
+      unfinished_runs = self.runs.find({'finished': False},
+                                       {'_id': 1},
+                                       sort=[('last_updated', DESCENDING)])
+      return unfinished_runs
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor I will test on DEV if this PR breaks something, please suggest how to test if the PR solves the problem.

tomtor:
@ppigazzini I adapted the `~/x/view_unfinished.py` script on prod to test the speedup, and it gives a hugh speedup:

```elapsed time: 0.031681```

So if the PR does not break fishtest (sorting or new test detection broken) it should fix the issue

ppigazzini:
@tomtor DEV updated, I'm start checking if fishtest is broken :)

tomtor:
@ppigazzini  Thanks. Last push should be non functional so no need to restart your test...

ppigazzini:
@tomtor DEV updated. I will start the test in a couple of hours after my commute.

ppigazzini:
@tomtor I run several tests on DEV with a looping mongodump, the workers switched always to the right tests after lowering/raising the priorities.
I wait the squash to merge.

ppigazzini:

tomtor:
Thanks @ppigazzini for testing this!

PR #681, opened 2020-05-23
Fix variable shadowing

noobpwnftw:


tomtor:
Good catch! @noobpwnftw

tomtor:
Will restart after mongodump finishes (around 10:20)

ppigazzini:
@tomtor I use `bash x/sync_repo.sh` to automate the server update, why should you wait the backup finish? (just asking for information).

tomtor:
> why should you wait the backup finish?

A few days ago I noticed that a restart of fishtest also restarted MongoDb in the syslog (That suprised me, I do not see that now in the last syslog). I was afraid that a restart of mongodb might interfere with a running mongodump.

ppigazzini:
> > why should you wait the backup finish?
> 
> A few days ago I noticed that a restart of fishtest also restarted MongoDb in the syslog (That suprised me, I do not see that now in the last syslog). I was afraid that a restart of mongodb might interfere with a running mongodump.

A few day ago I restated mongod by hand to try to debug the workers idling problem :)
Neither `x/sync_repo.sh` nor `x/update_fishtest.sh` restart mongod.

tomtor:
> A few day ago I restated mongod by hand to try to debug the workers idling problem

Ahh, I see, I was surprised to see that in the log.
So, without a restart of mongodb there is indeed no reason to wait for mongodump to finish....

PR #682, opened 2020-05-23
Fix race condition with request_task/stop_run

tomtor:
Dropping 'cores' from run in stop_run causes sort to fail in
request_task

```
May 23 12:30:48 tests pserve[23821]:   File "/home/fishtest/fishtest/fishtest/fishtest/rundb.py", line 524, in <lambda>
May 23 12:30:48 tests pserve[23821]:     r['cores'] / r['args']['itp'] * 100.0,
May 23 12:30:48 tests pserve[23821]: KeyError: 'cores'

```

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -661,7 +661,6 @@ class RunDb:
     if run is None:
       run = self.get_run(run_id)
       save_it = True
-    run.pop('cores', None)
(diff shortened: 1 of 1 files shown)


tomtor:
@ppigazzini Needs no testing IMO...

ppigazzini:
PROD updated without testing, take a look ;)

Issue #685, opened 2020-05-24
locale & pgn


vondele:
https://groups.google.com/forum/?fromgroups=#!topic/fishcooking/SOpgUxPYZa0

looks like depending on the locale the generated PGNs can not be uploaded. I'd like to keep the PGNs easy to parse, so enforcing ascii makes sense. Can we launch cutechess while setting the equivalent of export LC_ALL=C in the worker?

d3vv:
Guess it's all about:

https://doc.qt.io/qt-5/qlocale.html

Sure need something like: `QLocale::setDefault(QLocale::C);`

But in a good way  if fishtest will abilities to parse unicode for date into info-string..

d3vv:
In simple, just cut season and country field..

tomtor:
See https://github.com/glinscott/fishtest/pull/686#issuecomment-633481723

tomtor:
See google thread:

Quote:

Ah thanks, that (8bit codepage) explains the error and why it DOES work for some users.

The next question is why some Window users have Cyrillic and utf-8 and some have Cyrillic with an 8bit codepage.

I guess it depends on window version (7/8/10) and the cutechess binary used.

Anyway, handling 8bit encoded pgns by fishtest is not a real option, so IMO ignoring them (throwing exception) is correct.

tomtor:
We could translate to ascii in the worker before upload when not utf-8. Not sure if that is worth the trouble.

vondele:
I guess this pull should be closed, as it neither solves the problem nor is the preferred way to solve it. Right?

tomtor:
@vondele Yes it can be closed...

tomtor:
Should close the PR, not the issue

tomtor:
command:

`chcp 65001`

in the command-prompt window of a fishtest worker should switch to utf-8 (untested)

kotenok2000:
Do i have to put it in one_click_run.bat?

tomtor:
Yes, please do.

Let me know if it works.

Op za 6 jun. 2020 06:03 schreef kotenok2000 <notifications@github.com>:

kotenok2000:
No it didn’t

tomtor:
Is fixed with #698

PR #686, opened 2020-05-24
Use C locale for cutechess

vondele:
other locales can yield unexpected date formats.

fixes https://github.com/glinscott/fishtest/issues/685

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -421,7 +421,9 @@ def launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play, batch_size
   cmd = cmd[:idx] + ['option.%s=%d'%(x['name'], round(x['value'])) for x in b_params] + cmd[idx+1:]
 
   print(cmd)
-  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
+  my_env = os.environ.copy()
+  my_env['LC_ALL'] = 'C'
+  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS, env=my_env)
 
   try:
     return parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play, batch_size, tc_limit)


vondele:
That's an untested fix. I'm not sure this is actually Linux only, or would work for windows as well..

tomtor:
Or

```
import locale
locale.setlocale(locale.LC_ALL, 'C')
```

Not sure if this effects the invocation of `cutechess` but it might be more portable.
I expect that your code also work on Windows.

vondele:
meanwhile the report in the fishtest thread suggests that modifying the env doesn't work. The user is on windows.

tomtor:
@vondele I wonder if it would not be better to change

https://github.com/glinscott/fishtest/blob/6a9242a10503a5333571178f92db78e8d04e1992/worker/worker.py#L177

to `with open(pgn_file, encoding='utf-8') as f`

because at https://github.com/glinscott/fishtest/blob/6a9242a10503a5333571178f92db78e8d04e1992/worker/worker.py#L179

we expect `utf-8` and why would we be bothered by a utf-8 PGN file?

I do not think this would impact your parsing, how do you parse? 
With Python? With a C program?
Do you interpret dates?

vondele:
I guess that might also be a solution... I don't have strong opinions. Personally, I'm opening pgn files only with python-chess, and open them with: 
```
pgnfilein = open(filename, "r", encoding="utf-8-sig", errors="surrogateescape")
```
I'm assuming that this will be doing the right thing. Usually I don't look at the dates. 
It does mean that we have pgns where the headers can't easily be read by everybody (Фінляндія (літо) being Finland (summer time) is new for me ;-))
Nevertheless, I would be fine with this change as well.

tomtor:
From the google discussion:

https://tests.stockfishchess.org/api/pgn/5eca3862a586eee45aa2abca-61.pgn

DOES have utf-8 content, so normally (sometimes/often?) utf-8 PGN upload succeeds.
I suspect that the failing PGN is not utf-8 but has a different encoding (a cyrillic codepage?)

tomtor:
See google thread...

tomtor:
Close PR but leave #685 open for now

PR #687, opened 2020-05-24
Less logging: no task queue warnings

tomtor:
@ppigazzini prevents filling up the log.


ppigazzini:
fishtest reinstalled on PROD, thank you @tomtor :)

Issue #688, opened 2020-05-26
Resolved New / Base not reliable


xoto10:
After the latest sf merges, patch diffs are not showing on github.

I think the problem is because we now only display and use the first 7 chars of the github commit reference. This is generally enough but in the case of current master it seems that it isn't (I assume there is a clash).
e.g. This diff doesn't display: https://github.com/vondele/Stockfish/compare/d40d04c...3140aa0
but this one does: https://github.com/vondele/Stockfish/compare/d40d04c1...3140aa0

Perhaps we should add a couple of characters to the commit refs to avoid this?

ppigazzini:
Great catch @xoto10 :)

PR #689, opened 2020-05-26
Use 10 digits of the sha

vondele:
avoids rare cases where 7 is not enough.

Fixes https://github.com/glinscott/fishtest/issues/688

### DIFF
--- a/fishtest/fishtest/helpers.py
+++ b/fishtest/fishtest/helpers.py
@@ -3,7 +3,7 @@ def tests_repo(run):
 
 def master_diff_url(run):
   return "https://github.com/official-stockfish/Stockfish/compare/master...{}".format(
-    run['args']['resolved_base'][:7]
+    run['args']['resolved_base'][:10]
   )
 
 def diff_url(run):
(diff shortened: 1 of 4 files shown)


ppigazzini:
@vondele DEV now shows the diff
https://dfts-0.pigazzini.it/tests/view/5ecc351980f2c838b965580b

Please raise also the worker version.
https://github.com/glinscott/fishtest/blob/fea6729d1b52bbb04824ba7d4401a0eb171dfeab/worker/worker.py#L31

vondele:
done.

vondele:
BTW, just ran into this  
https://stackoverflow.com/questions/18134627/how-much-of-a-git-sha-is-generally-considered-necessary-to-uniquely-identify-a 
linux kernel uses 12 digits, but 10 should be enough for most projects seemingly.

ppigazzini:
Test on DEV are all OK.
Let me know if you want merge now with 10 digits or wait tomorrow :)

vondele:
either is fine. 10 digits will be OK till we have 100k commits :-)

ppigazzini:

@noobpwnftw please update your workers fleet at your convenience :)
After that I will force the update for other workers.

PR #691, opened 2020-05-27
Use 10 digits of the sha. Part 2.

vondele:
fix a mised case in the templates.


vondele:
Sorry, overlooked one case in the templates. Not urgent, just consistent.

ppigazzini:

PR #695, opened 2020-05-30
Protect against wrongly formats in tc. Workaround.

vondele:
workaround for current fishtest internal error


ppigazzini:
Should be fixed by #696 
Thank you @vondele :)

Issue #690, opened 2020-05-27
fishtest dead?


Vizvezdenec:
I get 504 error when trying to refresh.

ppigazzini:
Server rebooted. I will check the logs in my first coffee break.

crossbr:
Again?

"Internal Server Error

The server encountered an unexpected internal server error

(generated by waitress)"

vondele:
@ppigazzini did an update a little while ago. I assume he's aware.

ppigazzini:
```
May 30 19:43:47 tests.stockfishchess.org pserve[24316]:   File "/home/fishtest/fishtest/fishtest/fishtest/util.py", line 255, in estimate_game_duration
May 30 19:43:47 tests.stockfishchess.org pserve[24316]:     time_tc = float(chunks[0])
May 30 19:43:47 tests.stockfishchess.org pserve[24316]: ValueError: could not convert string to float: '15Npsq'
```

vondele:
@ppigazzini hmm bogus TC being passed to this function. Possibly a garbage test?

vondele:
https://tests.stockfishchess.org/tests/user/fauzi2 doesn't display. I think it is one of his tests (psq tuning).

vondele:
@ppigazzini https://github.com/glinscott/fishtest/pull/695 might be a workaround.

ppigazzini:
@vondele correct, I deleted that test.
I was able to have fishtest up and running with:
```
  else:
    try:
      time_tc = float(chunks[0])
    except:
      time_tc = 666
```

Previously I restarted fishtest to raise the worker version to 79 server side.

vondele:
Right the tc seems like a pasto : `15Npsq+0.15`

crossbr:
500 Internal Server Error

tomtor:
@ppigazzini 

Mongodb had crashed, did a restart of the mongod service

Vizvezdenec:
bump

ppigazzini:
@tomtor vps restarted 3 hours ago

crossbr:
I'm getting "504 Gateway Time-out" and my worker is timing out.

Vizvezdenec:
main page fails to refresh.

crossbr:
@Vizvezdenec still up for me.

Vizvezdenec:
[...]

PR #696, opened 2020-05-31
Add error check for time control

tomtor:
See #690

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -324,6 +324,9 @@ def validate_form(request):
     'info': request.POST['run-info'],
   }
 
+  if not re.match('^([1-9]\d*/)?\d+(\.\d+)?(\+\d+(\.\d+)?)?$', data['tc']):
+    raise Exception('Bad time control format')
+
   if request.POST.get('rescheduled_from'):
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor DEV updated, it rejects only bad TCs starting with chars, not TCs with chars following a number. Is it the indented behavior?
e.g. bad TC accepted:
- 10xxx+xxx
- 10xxx/xxx
- 10+xxx
- 10/xxx
- 10/

tomtor:
@ppigazzini No that is not intended :-)
Will fix

tomtor:
@ppigazzini Did test it now myself, should be much more robust now.

ppigazzini:
@tomtor it rejects also <n_games_in_floating>/<tc> :)

ppigazzini:

tomtor:
> it rejects also 6.66/6.66 :)

And 0/40

ppigazzini:
> > it rejects also 6.66/6.66 :)
> 
> And 0/40

Great suggestion: 0+0 and 1/0 accepted :)

tomtor:
> and 1/0 accepted

That one is easy to fix...

PR #698, opened 2020-06-06
Strip non utf-8 characters from PGN files

tomtor:
Should fix EDIT:  https://github.com/glinscott/fishtest/issues/685

Also reduce some line lengths to less than 80 chars

@ppigazzini I tested this with python 2 and 3 workers, but not with a PGN containing non-utf8 characters. Perhaps you can create one with a particular Windows setting? If not then it's OK for me to commit and let @kotenok2000 test?

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -28,7 +28,7 @@ from updater import update
 from datetime import datetime
 from os import path
 
-WORKER_VERSION = 79
+WORKER_VERSION = 80
 ALIVE = True
 HTTP_TIMEOUT = 15.0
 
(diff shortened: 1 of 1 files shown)


kotenok2000:
I substituted  file work
er.py  with file from tomtor:non-utf8. Now my pgns upload.
[5edc25adf29b40b0fc95ab65-18.pgn.txt](https://github.com/glinscott/fishtest/files/4741277/5edc25adf29b40b0fc95ab65-18.pgn.txt)

tomtor:
@kotenok2000 Thanks for testing!

@ppigazzini I raised worker version to 80

ppigazzini:
@noobpwnftw please update your fleet at your convenience :)

noobpwnftw:
@ppigazzini Updated.

PR #703, opened 2020-06-16
Add a default test contempt parameter

mstembera:
Add a default parameter for contempt to fishtest.  This allows us the flexibility to use a separate default contempt for fishtest testing from SF own default contempt.  Currently it's set to 24 to keep things the same but I don't know what the best value for testing purposes should be.  Discussion is welcome here.
See comments in https://github.com/official-stockfish/Stockfish/pull/2740

Note: I am unfamiliar with .mak files so if my changes aren't correct please advise.  Also I'm not sure if lines 146 and 150 should also be changed and if so what the exact syntax should be.  Ty.

### DIFF
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -97,19 +97,19 @@
       <label class="field-label leftmost">Test type</label>
       <div class="btn-group choose-test-type">
         <div class="btn" id="fast_test"
-             data-options='{"tc": "10+0.1", "threads": 1, "options": "Hash=16", "bounds": "standard STC"}'>
+             data-options='{"tc": "10+0.1", "threads": 1, "options": "Hash=16 Contempt=24", "bounds": "standard STC"}'>
           short (STC)
         </div>
         <div class="btn" id="slow_test"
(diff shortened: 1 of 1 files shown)


tomtor:
> Also I'm not sure if lines 146 and 150 should also be changed and if so what the exact syntax should be.

Looks good to me, but you'll have to change lines 146 and 150 too. These contain the default data for a new test.

Just change `'Hash=16'` to `'Hash=16 Contempt=24'`

mstembera:
@tomtor Fixed. Thanks.

ppigazzini:
Tested OK on DEV, PROD updated, thank you @mstembera :)

snicolet:
This pull request makes the issue in https://github.com/glinscott/fishtest/issues/704 even more urgent (and delicate).

If user reschedules a STC test which was run with Contempt=10 and uses the "LTC" button to change the time control, then the LTC test will start with Contempt=24, and there is a good probability that the user will not notice.

mstembera:
@snicolet  The rescheduling issue you point out was already there before this PR.  If a user ran STC w/ custom contempt and rescheduled it would have also defaulted back.   I fully agree that it would be nice to not overwrite user set options when rescheduling.

PR #705, opened 2020-06-16
Improve signup message

vondele:
triggered by a message on fishcooking: https://groups.google.com/g/fishcooking/c/41mKxgCwzc0/m/TSN7VxU6BgAJ and after having seen multiple duplicate registrations.


ppigazzini:

PR #709, opened 2020-06-21
List 25 items on view pages

tomtor:
Reduce from 50 for less IO and faster response.

See https://github.com/glinscott/fishtest/issues/690#issuecomment-647110579


ppigazzini:
PROD updated, finger crossed :)

PR #713, opened 2020-06-24
Try to stop test after a cutechess warning

vdbergh:
This is mainly meant for SPSA tunes. It is easy to set up an SPSA tune incorrectly so that in fact no tuning is done.

Note: to make this work I had to redirect stderr of cutechess-cli to stdout. I assume this is harmless but it should be kept in mind.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -311,6 +311,14 @@ def parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play, 
       return { 'task_alive': True }
 
     # Parse line like this:
+    # Warning: New-eb6a21875e doesn't have option ThreatBySafePawn
+    if line.split()[0]=='Warning:' and 'doesn\'t have option' in line:
+      message = r'Cutechess-cli says: "%s"' % line.strip()
+      result['message']=message
+      send_api_post_request(remote + '/api/stop_run', result)
+      raise Exception(message)
+
+    # Parse line like this:
     # Finished game 1 (stockfish vs base): 0-1 {White disconnects}
     if 'disconnects' in line or 'connection stalls' in line:
       result['stats']['crashes'] += 1
@@ -421,7 +429,7 @@ def launch_cutechess(cmd, remote, result, spsa_tuning, games_to_play, batch_size
   cmd = cmd[:idx] + ['option.%s=%d'%(x['name'], round(x['value'])) for x in b_params] + cmd[idx+1:]
 
   print(cmd)
-  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
+  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
 
   try:
     return parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play, batch_size, tc_limit)


tomtor:
You need to comment this line in api.py

https://github.com/glinscott/fishtest/blob/190a05b3d5e7cfaf1c27325f7faf041b84671201/fishtest/fishtest/api.py#L218

It is a safe guard that prevents new user from stopping any job.

worker/games.py
```diff
@@ -310,6 +310,14 @@ def parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play,
       kill_process(p)
       return { 'task_alive': True }
 
+    # Parse line like this:
+    # Warning: New-eb6a21875e doesn't have option ThreatBySafePawn
+    if line.split()[0]=='Warning:':
+      message = r'Cutechess-cli says: "%s"' % line.strip()
+      result['message']=message
+      send_api_post_request(remote + '/api/stop_run', result).json()
```
tomtor, line 318: `.json()` is not needed

worker/games.py
```diff
@@ -310,6 +310,14 @@ def parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play,
...
+      raise Exception(message)
+
```
tomtor, line 320: The signature code also has this fragment:

```
finally:
    if concurrency > 1:
      busy_process.communicate(enc('quit\n'))
      busy_process.stdin.close()
```

Not sure if that is really needed.

worker/games.py
vdbergh: Thanks. This was a left over of a test print. I will delete it of course.

worker/games.py
tomtor: I don't think so, it is only needed for the signature check.

vdbergh: I will think about it... It has to do with cleaning up.

tomtor:
@vdbergh Looks good to me! Should work if https://github.com/glinscott/fishtest/pull/713#issuecomment-648673627 is taken care of.

vdbergh:
I updated the PR. 

I made the test more specific. There may also be benign warnings, e.g. after a cutechess version update.

tomtor:
@vdbergh Is this finished, ie no longer WIP?

vdbergh:
That's right. No longer WIP. I thought I changed the title but must have forgotten to save.

ppigazzini:
@vdbergh it works fine with linux/WSL workers but it doesn't work with a windows worker. I'll debug tonight.

vdbergh:
@ppigazzini Perhaps it has to do with the redirection of stderr to stdout in popen? Maybe this does not work on windows?

ppigazzini:
@vdbergh I started the worker windows just before leaving for the commute, I'll debug when at home.

ppigazzini:
@vdbergh the Warning doesn't show with the STDERR redirect
```python
    sys.stdout.write(line)
    sys.stdout.flush()
    
    print("debug start")
    print(line)
    print("debug end")

    # Have we reached the end of the match?  Then just exit
    if 'Finished match' in line:
```


```
Fetched SPSA parameters successfully in 0.187502s
['D:\\__test_worker\\ori\\worker00\\testing\\cutechess-cli.exe', '-repeat', '-rounds', '3', '-games', ' 2', '-tournament', 'gauntlet', '-site', 'https://tests.stockfishchess.org/tests/view/5ef6505203fa0bac787144e4', '-event', 'Batch 4: spsa_tune vs spsa_tune', '-srand', '2632243431', '-resign', 'movecount=3', 'score=400', '-draw', 'movenumber=34', 'movecount=8', 'score=20', '-concurrency', '3', '-openings', 'file=noob_3moves.epd', 'format=epd', 'order=random', 'plies=16', '-engine', 'name=New-eb6a21875e', 'cmd=stockfish_eb6a21875e433fa54f42b69a62172d3912192ba8', 'option.Hash=16', 'option.Contempt=24', 'option.mThreatBySafePawnNNNNNNN=329', 'option.eThreatBySafePawnNNNNNN=13', '-engine', 'name=Base-eb6a21875e', 'cmd=stockfish_eb6a21875e433fa54f42b69a62172d3912192ba8', 'option.Hash=16', 'option.Contempt=24', 'option.mThreatBySafePawnNNNNNNN=213', 'option.eThreatBySafePawnNNNNNN=76', '-each', 'proto=uci', 'tc=0.870+0.009', 'option.Threads=1']
TC limit 17.396362529322385 End time: 2020-06-26 21:57:01.639401
Indexing opening suite...
debug start
Indexing opening suite...

debug end
Started game 1 of 6 (New-eb6a21875e vs Base-eb6a21875e)
debug start
Started game 1 of 6 (New-eb6a21875e vs Base-eb6a21875e)

```

vdbergh:
Hmm. Unfortunately I do not know how to fix that. There are other ways of redirecting stderr but I don't have a Windows machine, so I cannot test it.

I assume that there are enough Linux workers so that a faulty test will be stopped by some Linux worker...

ppigazzini:
@vdbergh the Windows cutechess-cli writes the STDERR after the whole match output, so "Finished match" triggers the cutechess-cli kill and the STDERR is lost. The STDERR redirection is fine on Windows, the patch works commenting out the "Finished match" block.

* Windows
```
./cutechess-cli.exe -repeat -rounds 4 -tournament gauntlet -resign movecount=3 score=400 -draw movenumber=34 movecount=8 score=20 -concurrency 2 -engine cmd=stockfish.exe option.Hash=8 option.Threads=2 -engine cmd=stockfish.exe option.Hash=8 option.Threads=2 option.XXXXXX=2 -each proto=uci tc=1+0.01 -openings file=noob_3moves.epd format=epd order=random plies=16
Indexing opening suite...
Started game 2 of 4 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT)
Started game 1 of 4 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT)
Finished game 1 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT): 1-0 {White wins by adjudication}
Score of Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT: 1 - 0 - 0  [1.000] 1
Started game 3 of 4 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT)
Finished game 2 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT): 1-0 {White wins by adjudication}
Score of Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT: 1 - 1 - 0  [0.500] 2
Started game 4 of 4 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT)
Finished game 3 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT): 1-0 {White wins by adjudication}
Score of Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT: 2 - 1 - 0  [0.667] 3
Finished game 4 (Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT): 1/2-1/2 {Draw by 3-fold repetition}
Score of Stockfish 181019 64 POPCNT vs Stockfish 181019 64 POPCNT: 2 - 1 - 1  [0.625] 4
Elo difference: 88.7 +/- nan, LOS: 71.8 %, DrawRatio: 25.0 %
Finished match
Warning: 2 opening repetitions vs 1 games per encounter
Warning: Stockfish 181019 64 POPCNT doesn't have option XXXXXX
Warning: Stockfish 181019 64 POPCNT doesn't have option XXXXXX
```

vdbergh:
Thanks! This is rather strange. One would not expect stderr to be buffered in any way... In principle ignoring Finished Match would be fine for an SPSA test since the matches are very short. But it is not very nice.

I can try to redirect both stdout and stderr to PIPE and then to merge the two streams at some later point.

ppigazzini:
@vdbergh I'm building the `cutechess-cli` form master to view if this strange behavior has been fixed.

vdbergh:
I am not sure if it is cutechess-cli's fault. It might be the implementation of popen in Python.

vdbergh:
Ah sorry :( :( Your example is already raw cutechess-cli output and Popen is fine. I had misunderstood.  So it is really cutechess-cli....

ppigazzini:
@vdbergh On Windows python redirect correctly STDOUT and STDERR, tested parsing the output of `stockfish.exe bench` 

```python
import sys
import subprocess
import threading
import time

try:
    from queue import Queue, Empty
except ImportError:
    from Queue import Queue, Empty  # python 2.x

ON_POSIX = 'posix' in sys.builtin_module_names

def enqueue_output(out, queue):
    try:
        for line in iter(out.readline, b''):
            queue.put(line)
        out.close()
    except:
        return

def parse_output(p):
    q = Queue()
    t = threading.Thread(target=enqueue_output, args=(p.stdout, q))
    t.daemon = True # thread dies with the program
    t.start()
    
    # read line without blocking
    while True:
        try:
            line = q.get_nowait() # or q.get(timeout=.1)
        except Empty:
            print('no output yet')
            if p.poll():
                break
            time.sleep(1)
            continue
    
        sys.stdout.write(line)
        sys.stdout.flush()
    
        if "second" in line:
            break

if __name__ == '__main__':
#  cmd=["cutechess-cli.exe", "-repeat", "-rounds", "4", "-tournament", "gauntlet", "-resign", "movecount=3", "score=400", "-draw", "movenumber=34", "movecount=8", "score=20", "-concurrency", "2", "-engine", "cmd=stockfish.exe", "option.Hash=8", "option.Threads=2", "option.XXXXXX=10", "-engine", "cmd=stockfish.exe", "option.Hash=8", "option.Threads=2", "-each", "proto=uci", "tc=1+0.01", "-openings", "file=noob_3moves.epd", "format=epd", "order=random", "plies=16"]

  cmd=["stockfish.exe", "bench"]
  
  print(cmd)
  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True, bufsize=1, close_fds=ON_POSIX)

  parse_output(p)
```

vdbergh:
There does not seem to be much to do about this, except filing a bug report with cutechess-cli. 

Perhaps it is best for now to just add a comment to the worker that things do not work on windows due to a bug in cutechess-cli? The test will anyway be stopped by a Linux worker.

ppigazzini:
@vdbergh the PR is fine. I just verified that `cutechess-cli` STDOUT/STDERR is OK when started in CMD/powershell, and it's wrong when started in a msys2 shell. But the worker is started in a CMD...
I will take a couple of days to try to figure out why/when the STDERR is delayed.

vdbergh:
Maybe something related to this https://stackoverflow.com/questions/50121192/how-can-i-setup-my-msys2-terminal-to-not-buffer-output ? 

Is the cutechess-cli you are using compiled inside the msys2 environment?

ppigazzini:
@vdbergh I updated https://github.com/glinscott/fishtest/pull/713#issuecomment-651193827 with the bat scripts used to start the python script.
`cutechess-cli` has STDERR delayed after STDOUT when called by the python script (python form msys2 and from python.org).
`stockfish bench` has STDERR/STDOUT correctly interleaved when called by the same python script.
Raw `cutechess-cli` has STDERR/STDOUT correctly interleaved when started with a bat.
Raw `cutechess-cli` has STDERR delayed when started from a default msys2 shell (adding `-defterm` according your link solve this problem)

`cutechess-cli` and `stockfish` are built with msys2.

ppigazzini:
@vdbergh it seems that the STDERR STDOUT order cannot be guaranteed, so the STDERR redirection on STDOUT should be fragile also for Linux.

Some references use in my tests:
https://stackoverflow.com/questions/31833897/python-read-from-subprocess-stdout-and-stderr-separately-while-preserving-order
https://stackoverflow.com/questions/375427/non-blocking-read-on-a-subprocess-pipe-in-python
https://stackoverflow.com/questions/18421757/live-output-from-subprocess-command


The Linux/Windows worker seems fine with 2 separated pipes.

```python
  q_stderr = Queue()
  q_stdout = Queue()
  t_stderr = threading.Thread(target=enqueue_output, args=(p.stderr, q_stderr))
  t_stdout = threading.Thread(target=enqueue_output, args=(p.stdout, q_stdout))
  t_stderr.daemon = True
  t_stdout.daemon = True
  t_stderr.start()
  t_stdout.start()

  end_time = datetime.datetime.now() + datetime.timedelta(seconds=tc_limit)
  print('TC limit {} End time: {}'.format(tc_limit, end_time))

  num_games_updated = 0
  while datetime.datetime.now() < end_time:

    try:
      line_stderr = q_stderr.get_nowait()
      
      if line_stderr:
        sys.stderr.write(line_stderr)
        sys.stderr.flush()        
      
        # Parse line like this:
        # Warning: New-eb6a21875e doesn't have option ThreatBySafePawn
        if line_stderr.split()[0]=='Warning:' and 'doesn\'t have option' in line_stderr:
          message = r'Cutechess-cli says: "%s"' % line_stderr.strip()
          result['message']=message
          send_api_post_request(remote + '/api/stop_run', result)
          raise Exception(message)      
    except Empty:
      pass

    try:
      line = q_stdout.get_nowait()
    except Empty:
      if p.poll() is not None:
        break
      time.sleep(1)
      continue

    sys.stdout.write(line)
    sys.stdout.flush()


# other code

  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
```

vondele:
yes two pipes will be needed I think.

vdbergh:
@ppigazzini @vondele Putting stderr=subprocess.STDOUT is a common pattern which is also mentioned in the Python documentation. It should work....

We seem to be witnessing some kind of bug in the interaction of Qt with msys2. It is conceivable that it is solved using two pipes. One would need to two threads to monitor them, but I think they can write to the same queue. So there should be no need for two separate queues for stderr and stdout.

vondele:
while it is a common pattern, I think there is simply no ordering guarantee between what is written to stdout and stderr, even if both are redirected to the same file. https://stackoverflow.com/questions/45316295/the-order-of-stdout-and-stderr-in-bash

vdbergh:
@vondele Yes but the ordering is irrelevant in this case. Except that we do not want stdout and stderr to be buffered. How they are interleaved doesn't matter.

ppigazzini:
@vdbergh It does matter for our code: the process could be killed before the STDERR reading.
Setting `bufsize=0` to have unbuffered pipes doesn't change the behavior, same starting `python -i -u`

Anyway I'm debugging a python performance hit on Windows :(

vdbergh:
> @vdbergh It does matter for our code: the process could be killed before the STDERR reading.

@ppigazzini If STDERR output is delayed so much that this behavior is triggered then I would consider this  a bug (somewhere), even if it is technically not illegal. 

Since this behavior is without consequence and appears only in specific configurations I don't think we should clutter the worker code excessively to accommodate for it. Especially since the proposed solutions may not actually robustly solve the underlying theoretical issue.

I think what you were proposing earlier may be the best solution after all: do not kill cutechess-cli after "Finished match". Cutechess-cli should anyway finish by itself after the match is over.

ppigazzini:
@vdbergh "Finished match" triggers the return, no other lines are processed, we should change the code in this way (it works but need some test and refactoring)

```python
    if 'Finished match' in line:
      # The following assertion will fail if there are games without result.
      # Does this ever happen?
      assert(num_games_updated==games_to_play)
      print('Finished match cleanly')

      # required by Windows worker
      line = q.get_nowait()
      while line:   #### TEST IF THIS CAN HANG
        if line.split()[0]=='Warning:' and 'doesn\'t have option' in line:
          message = r'Cutechess-cli says: "%s"' % line.strip()
          result['message']=message
          send_api_post_request(remote + '/api/stop_run', result)
          raise Exception(message)
        else:
          sys.stdout.write(line)
          sys.stdout.flush()
        line = q.get_nowait()        
      kill_process(p)
      return { 'task_alive': True }
```

`kill_process()` calls `p.stdout.close()` avoiding that python will hog the CPU. I solved the performance hit of the two pipes code adding a `p.stderr.close()` in `kill_process()`

We should check if it's possible to clean up all the `kill_process(p)` calls from `parse_cutechess_output()` with:
```python
  try:
    task_state = parse_cutechess_output(p, remote, result, spsa, spsa_tuning, games_to_play, batch_size, tc_limit)
  except Exception as e:
    print('Exception running games')
    traceback.print_exc(file=sys.stderr)
  else:
    kill_process(p)
    return task_state
```

vdbergh:
@ppigazzini Is it really necessary to do _anything_ after "Finished match"? 

`p.poll()` will return `False` later and the `while` loop will exit through the `break` statement.

Then `cutechess-cli` will be killed anyway (probably it is dead already) and `{'task_alive':True}` will be executed.

ppigazzini:
@vdbergh yes on Windows to catch the STDERR.

`kill_process()` it's really a subprocess management. Without `p.stdout.close()`, for each cutechess-cli call, you will have a zombie that processes STDOUT for the eternity :)

vdbergh:
@ppigazzini I guess I am missing something. This is the current code. 

```
    if 'Finished match' in line:
      # The following assertion will fail if there are games without result.                                                                                                            
      # Does this ever happen?                                                                                                                                                          
      assert(num_games_updated==games_to_play)
      print('Finished match cleanly')
      kill_process(p)
      return { 'task_alive': True }
```

If  we make it 
```
    if 'Finished match' in line:
      # The following assertion will fail if there are games without result.                                                                                                            
      # Does this ever happen?                                                                                                                                                          
      assert(num_games_updated==games_to_play)
      print('Finished match cleanly')
```
and the while loop exits afterwards via the break statement in
```
   try:
      line = q.get_nowait()
    except Empty:
      if p.poll():
        break
      time.sleep(1)
      continue

```
then this code after the loop will be executed
```
  now = datetime.datetime.now()
  if now >= end_time:
    print('{} is past end time {}'.format(now, end_time))
  kill_process(p)
  return { 'task_alive': True }
```
I see no essential difference....

ppigazzini:
@vdbergh I agree that this function needs a clean up.
I already fixed that break with #718
I didn't examinated how works the `end_time` check, it could be a trick to avoid hangs.

vdbergh:
@ppigazzini 

The end time is not really relevant in our case. It is the fact that
``` 
  kill_process(p)
  return { 'task_alive': True }
```
will always be called when cutechess-cli exits (and closes its stdout and stderr), even if we don't do it after "Finished match" (and in fact `kill_process(p)` will even be redundant).

At least this is what I infer from the code.

ppigazzini:
@vdbergh If it's OK for you I would like to merge this PR.
Then I will start to work on a new PR to clean up the code of these functions testing the workers on some OS/Python combinations.
I will try also to build cutechess-cli in a different way to try to solve the STDERR delaying on python.

vdbergh:
@ppigazzini  I think this would be the best strategy!

ppigazzini:
@vdbergh Thank you for the PR and the valuable insights :)

PR #715, opened 2020-06-24
Handle longer (> 7 digits) bench signatures

### LINKED ISSUE #714: Bench signature cutoff
The current script doesn't handle the full number which is Bench: 20101877
while fishtest displays: Bench: 2010187
https://tests.stockfishchess.org/tests/view/5ef32cb2122d6514328d7866
Test where bench signature is manually set are not affected:
https://tests.stockfishchess.org/tests/view/5ef3200c122d6514328d785a

tomtor:
Should fix #714

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -259,7 +259,7 @@ def users_monthly(request):
 
 
 def get_master_bench():
-  bs = re.compile(r'(^|\s)[Bb]ench[ :]+([0-9]{7})', re.MULTILINE)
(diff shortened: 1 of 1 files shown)


ppigazzini:
@tomtor DEV updated, a new test shows correctly the bench 20101877

ppigazzini:
@tomtor the original test still test shows the truncated bench 2010187, let me know if this is the expected behavior.
https://dfts-0.pigazzini.it/tests/view/5ef32cb2122d6514328d7866

tomtor:
> let me know if this is the expected behavior.

Yes, it is... 

Thanks for testing!

PR #717, opened 2020-06-26
Make parsing of spsa parameters resistant to blank lines.

vdbergh:
Test case:

len("".split(","))

returns 1, not 0.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -288,10 +288,10 @@ def parse_spsa_params(raw, spsa):
   params = []
   for line in raw.split('\n'):
     chunks = line.strip().split(',')
-    if len(chunks) == 0:
+    if len(chunks)==1 and chunks[0]=="": # blank line
       continue
     if len(chunks) != 6:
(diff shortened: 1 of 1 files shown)


fishtest/fishtest/views.py
```diff
@@ -288,10 +288,10 @@ def parse_spsa_params(raw, spsa):
   params = []
   for line in raw.split('\n'):
     chunks = line.strip().split(',')
-    if len(chunks) == 0:
+    if len(chunks)==1 and chunks[0]=="": # blank line
```
ppigazzini, line 291: @vdbergh what about to check also for a line with only spaces? 
```python
if len(chunks)==1 and (not chunks[0] or chunks[0].isspace()): # blank line or only spaces
```

vdbergh: I think this is not necessary since `line` was already stripped when defining `chunks`.

ppigazzini:
Tested OK on DEV, PROD updated, thank you @vdbergh :)

PR #718, opened 2020-06-30
Fix subprocess.Popen.poll() check

ppigazzini:
A None value indicates that the process hasn't terminated yet.
Typically, an exit status of 0 indicates that the process ran successfully.


tomtor:
Looks good to me!

PR #719, opened 2020-07-03
Enable a new algorithm for choosing SPSA hyperparameters.

vdbergh:
This PR adds a checkbox to the "New Test" form. Checking this option will rewrite the hyperparameters furnished by the tuning code in Stockfish in such a way that the SPSA tune will finish within 0.5 Elo from the optimum (with 95% probability) assuming the function 'parameters->Elo' is quadratic and varies 2 Elo per parameter over each specified parameter interval. The hyperparameters are relatively conservative and their performance will degrade gracefully if the actual variation is different.  The theoretical basis for choosing the hyperparameters is given in this document
    
https://github.com/vdbergh/spsa_simul/blob/master/doc/theoretical_basis.pdf
    
The formulas can be checked by simulation which is done here
    
https://github.com/vdbergh/spsa_simul
    
It is important for the current version of the algorithm to be able to estimate the draw ratio. The PR uses a heuristic method to do so which makes various assumptions: book='noob_3moves.epd', contempt=24, threads=1 and no nodestime.

### DIFF
--- a/fishtest/fishtest/static/js/spsa_new.js
+++ b/fishtest/fishtest/static/js/spsa_new.js
@@ -1,0 +1,240 @@
+"use strict";
+
+/*
+Quick and dirty way to get the 95% quantile of the
+chi2 distribution.
+*/
+
+function chi2_95_approximation(df){
+    /* Wilson and Hilferty approximation */
+    var z95=1.6448536269514722;
+    var t=2/(9*df);
+    return df*Math.pow(z95*Math.pow(t,0.5)+1-t,3);
+}
+
+function chi2_95(df){
+    /* Table for df=1,..,99 */
+    var chi2_95_=[3.8414588206941236, 5.9914645471079799, 7.8147279032511765, 9.487729036781154, 11.070497693516351, 12.591587243743977,
+		  14.067140449340167, 15.507313055865453, 16.918977604620448, 18.307038053275143, 19.675137572682491, 21.026069817483066,
+		  22.362032494826941, 23.68479130484058, 24.99579013972863, 26.296227604864235, 27.587111638275324, 28.869299430392626,
+		  30.143527205646159, 31.410432844230929, 32.670573340917294, 33.9244384714438, 35.17246162690806, 36.415028501807299,
+		  37.652484133482766, 38.885138659830055, 40.113272069413611, 41.337138151427411, 42.556967804292668, 43.772971825742182,
+		  44.985343280365129, 46.194259520278457, 47.399883919080921, 48.602367367294178, 49.801849568201824, 50.99846016571064,
+		  52.192319730102874, 53.383540622969278, 54.572227758941736, 55.758479278887037, 56.942387146824075, 58.124037680867971,
+		  59.303512026899838, 60.480886582336431, 61.656233376279538, 62.829620411408186, 64.001111972218013, 65.170768903569808,
+		  66.338648862968768, 67.5048065495412, 68.669293912285838, 69.832160339848173, 70.993452833782186, 72.153216167023089,
+		  73.311493029083195, 74.46832415930939, 75.623748469376167, 76.777803156061395, 77.930523805230379, 79.081944487848745,
+		  80.232097848762848, 81.381015188899141, 82.528726541471897, 83.675260742721008, 84.820645497656727, 85.964907441231006,
+		  87.108072195321995, 88.250164421874018, 89.391207872507835, 90.531225434880668, 91.670239176054821, 92.808270383107839,
+		  93.945339601192217, 95.081466669243298, 96.216670753503891, 97.350970379033186, 98.484383459340307, 99.616927324283921,
+		  100.74861874635026, 101.87947396543589, 103.00950871222616, 104.13873823027392, 105.26717729686055, 106.39484024272251,
+		  107.52174097071949, 108.64789297350764, 109.77330935028799, 110.89800282268439, 112.02198574980785, 113.1452701425555,
+		  114.26786767719352, 115.38978970826668, 116.51104728087367, 117.63165114234559, 118.75161175336743, 119.87093929856709,
+		  120.98964369660951, 122.10773460981952, 123.22522145336157];
+    if(df<=99){
+	return chi2_95_[df-1];
+    }else{
+	return chi2_95_approximation(df);
+    }
+}
+
+var spsa_setup_default={
+    "num_params"   : 1,
+    "draw_ratio"   : 0.61,
+    "precision"    : 0.5,
+    "c_ratio"      : 1/6,
+    "lambda_ratio" : 3,
+    "params"       : [
+	{"name": "dummy",
+	 "start": 50,
+	 "min" : 0,
+	 "max" : 100,
+	 "elo" : 2,
+	 "c" : null,
+	 "r" : null}
+    ],
+    "num_games"    : null
+};
+
+
+function deepcopy(o){
+    return JSON.parse(JSON.stringify(o));
+}
+
+function spsa_compute(spsa_setup){
+    const C=347.43558552260146;
+    var s=deepcopy(spsa_setup);
+    var chi2=chi2_95(s.num_params);
+    var r=s.precision/(C*chi2*(1-s.draw_ratio)/8);
+    var lambda=new Array(s.num_params);
+    var i;
+    for(i=0;i<s.num_params;i++){
+	s.params[i].c=s.c_ratio*(s.params[i].max-s.params[i].min);
+	s.params[i].r=r;
+	var H_diag=-2*s.params[i].elo/Math.pow((s.params[i].max-s.params[i].min)/2,2);
+	lambda[i]=-C/(2*r*Math.pow(s.params[i].c,2)*H_diag);
+    }
+    var ng;
+    s.num_games=-1;
+    for(i=0;i<s.num_params;i++){
+	ng=Math.round(s.lambda_ratio*lambda[i]);
+	if(ng>s.num_games){
+	    s.num_games=ng;
+	}
+    }
+    return s;
+}
+
+/*
+Below is some code to estimate the draw ratio from the time
+control. The algorithm is very naive. It uses interpolation for
+a few data points valid for the book "noob_3moves.epd".
+*/
+
+function tc_to_seconds(tc){
+/*
+Convert cutechess-cli like tc time[/moves][+inc] to seconds/move.
+*/
+    var inc=0;
+    var moves=68; /* Fishtest average LTC game duration. */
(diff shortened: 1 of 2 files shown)


vondele:
@vdbergh IMO, this should come with a few lines of description, for example the guidelines on picking the interval, and the underlying assumptions. Basically, to help the user to do the right thing without having the read the long thread, the simulator, etc... 

As one review comment 'New algorithm' is almost always a poor name. Do we have something more descriptive, and better in the longer term? (robust, quadratic, guaranteed, ... !? )

vondele:
Minor nit 'the tuning branch' is now outdated, since the code is in master. What about 'the tuning routines in Stockfish'? 

Otherwise, this looks good, and presumably ready for some more in-depth testing... @ppigazzini should we have a look at this on the dev server?

ppigazzini:
DEV updated.

ppigazzini:
@vdbergh with 12 parameters it triggers `Number of games must be <= 800000`

vdbergh:
> @vdbergh with 12 parameters it triggers `Number of games must be <= 800000`

There are some restrictions on the form values in the current js code which I did not touch since I do not know where they come from. @tomtor ?

vondele:
I think this is a check that was put in place to avoid mistakes. However, it might make sense to relax things?

I think the form has a good 'feel'. 'SPSA A' parameter isn't changing, I don't know enough what it does to judge if that's right or wrong.

vdbergh:
@vondele The value of A does not matter if alpha is zero but I guess psychologically it would be better to set it to zero also.  Also alpha and gamma should actually be mathematically zero but I did not test yet if this does not cause some server issue (I don't think it will).

ppigazzini:
@vondele with a normal SPSA the "A" value (suggested 5-10% of total iterations) avoids big values for "ak" at the start of the optimization `ak = a / (k + 1 + A)^alpha`

@vdbergh I will try to set by hand 0 values to alpha and gamma

ppigazzini:
Found a SPSA with only 4 parameters to make some tests :)

https://dfts-0.pigazzini.it/tests/view/5f02ff19919f102ab14f68b3

tomtor:
> > @vdbergh with 12 parameters it triggers `Number of games must be <= 800000`
> 
> There are some restrictions on the form values in the current js code which I did not touch since I do not know where they come from. @tomtor ?

Yes, the 800000 limit is in place to guard against very large chunk collections. 

It was introduced because a user increased the number of games by just adding '00' to the end of the default and it crashed fishtest. Could probably be raised somewhat further, but we do need to restrict it. Do we really want/need (tuning) tests with > 800000 games? In the past there were somewhat successful runs with 100 parameters, but allocating the number of games as computed by this PR (10000000?) for 100 parameters is not possible anyway.

vondele:
@tomtor, indeed makes sense still to keep an upper limit, and collect some experience with tuning a smaller number of parameters first. Maybe it just needs an additional line of info in the info-box that right now, the maximal number of parameters that can be tuned is approx. N since we limit the number of games per test to 800k. If it turns out we're highly successful with the new approach we can start thinking about how to expand it to more parameters.

vdbergh:
@ppigazzini It is currently not possible to set alpha=gamma=0 manually since things are reset on submit. In any case this is not terribly important as `580000**0.001=1.013` (so very close to 1). 

The second and the third test are doing exactly the same thing. This is ok if we want to check that things are repeatable (which is always a very good idea), but I am not sure if there are enough resources are available to run 1.7 million games in a reasonable time frame, even if they are 1+0.01. I can currently only contribute 3 cores as the server I used to use can no longer run the latest version of cutechess-cli.

ppigazzini:
@vdbergh I restarted the server on DEV after changing by hand `tests_run.mak` here:
```
504                 $("input[name='spsa_alpha']").val(0.00);
505                 $("input[name='spsa_gamma']").val(0.00);
```

This code already allows 0 values
```
218       <label class="field-label rightmost">SPSA Alpha</label>
219       <input type="number" min="0" step="0.001" name="spsa_alpha"
220              class="third-size no-arrows"
221              value="${args.get('spsa', {'alpha': '0.602'})['alpha']}" />
222
223       <label class="field-label" style="margin-left: 7px">SPSA Gamma</label>
224       <input type="number" min="0" step="0.001" name="spsa_gamma"
225              class="third-size no-arrows"
226              value="${args.get('spsa', {'gamma': '0.101'})['gamma']}" />
```

ppigazzini:
> The second and the third test are doing exactly the same thing. This is ok if we want to check that things are repeatable (which is always a very good idea), but I am not sure if there are enough resources available to run 1.7 million games in a reasonable time frame, even if they are 1+0.01.

My idea is to check the PR code on DEV submitting some SPSA tests.
We could check the convergence on PROD pasting the parameters value by hand and selecting a SPSA test with good parameters, eg

https://dfts-0.pigazzini.it/tests/view/5f0331d41289d603ccb2ebd4


>  I can currently only contribute 3 cores as the server I used to use can no longer run the latest version of cutechess-cli.

Let me know the distro.

vdbergh:
CentOS 6. Yes very old... I need to upgrade it but it takes some planning but as the server is headless and in a room I do not have a key for it takes too much planning...

vondele:
BTW, I think we should use the testing on DEV to establish the implementation is behaving as expected. Testing if this new implementation is really helpful we should do on PROD, just because the DEV setup doesn't have the resources. Doing a few tests on PROD with the parameters as obtained on DEV might be useful however.

vondele:
So, started one test on PROD https://tests.stockfishchess.org/tests/view/5f03768359f6f035328944db

I started them from the current values... probably an alternative would be to start somewhat away from the optimum and see if one goes back.

ppigazzini:
DEV updated, started some tests.

vondele:
spsa_new_params
Just an update with intermediate results & comparison of the two running SPSA runs mentioned before. The runs starting with displaced values have roughly found master values, but those starting from master have moved a bit away.  I might do an intermediate SPRT run on the latter values to see if they are better.

vondele:
spsa_new_params_final
attached image is the full SPSA runs. I would say all looks consistent. We find more or less the values of master back.

I tried the near final results in a SPRT test, but like the first one this was not showing a passing result, suggesting master is already optimal. https://tests.stockfishchess.org/tests/view/5f09ad1559f6f0353289488c

So, I think this more or less shows that the new scheme is doing fine. I think we can consider merging this.

ppigazzini:
@vdbergh please open a new PR for the stochastic rounding. I would like:
 - some tests of stochastic rounding with an integer parameter in a little range (e.g. depth in 11-15) and other parameters. If the only good depth is 13, all the SPSA iterations with depth=12 or depth=14 IMO are wasted and in that case a bias could be a good thing to have.
 - a clean up of the unnecessary code from server/worker
 - a good commit message to make clear that SPSA works with real numbers, and that the stochastic rounding has:
   - primary goal to deal with Stockfish integer parameters (here a Stockfish parameter scaling is useless)
   - secondary goal is to round all parameters values for the cutechess-cli parameter input constrain, (here a good Stockfish parameter scaling avoids a precision loss). This could be dropped with a future cutechess-cli release able to deal with floating point parameters values.

vdbergh:
@ppigazzini  Cutechess-cli accepting real numbers does not make any difference since then the rounding needs to happen inside SF (as SF uses fixed precision). 

You need to do the rounding somewhere and in my opinion it is best done outside SF since then it is generic and the user does not have think about it.

ppigazzini:
@vdbergh so it's more complicated that I thought, please open a new PR to collect the information.

vdbergh:
@ppigazzini I don't really see the point of opening a new PR for what is just one line of code representing common sense (as I said stochastic rounding is common in machine learning). 

I could add to the info window that the SPSA algorithm should not be used for parameters that can just take a few values and that there is also no point in artificially introducing scaling to try to achieve this (this actually has nothing to do with the new algorithm, it is even more true for the algorithm in master). 

For the new algorithm an interval of length 6 represents the bare minimum (this could be enforced). Then c is 1 so that the gradient descent actually has some room to play.

It is an interesting question what would be a good tuning algorithm for a parameter that can take only a few values.  A simple round robin tournament seems tempting but then one does not use the fact that nearby parameter values are expected to be correlated. Some form of Bayesian tuning seems to be the most natural method to exploit this.

ppigazzini:
@vdbergh the previous PR seems to be tested OK by @vondele, so if this is fine with you I would like to merge it asap to have some feedback from the framework.

Stochastic rounding was here for years and nobody used it because bad documentation, please mind also that some fishtest contributors could lack Stockfish knowledge (for example me).

So I would like to  have a new PR for stochastic rounding:
 - to document why we need a rounding function and why we are choosing the stochastic rounding
 - to clean up the server/worker code. Merging now https://github.com/glinscott/fishtest/pull/719/commits/ede07e37025f5b7990d50355ce095f6575143c63 will leave some dead code (the "deterministic" rounding) and some unnecessary code (the double level rounding in worker code)

vdbergh:
> @vdbergh the previous PR seems to be tested OK by @vondele, so if this is fine with you I would like to merge it asap to have some feedback from the framework.
> 
> Stochastic rounding was here for years and nobody used it because bad documentation, please mind also that some fishtest contributors could lack Stockfish knowledge (for example me).
> 
> So I would like to have a new PR for stochastic rounding:

Ok I guess I can do that. 

As I said it is hard to say that we "need" stochastic rounding. It is just that it avoids that one has to think about rounding. It is similar to repeated openings. It may be hard to prove that this is truly harmful. But if there are no repeated openings (or as few as possible) one simply does not have think about it.

> 
> * to document why we need a rounding function and why we are choosing the stochastic rounding
> * to clean up the server/worker code. Merging now [ede07e3](https://github.com/glinscott/fishtest/commit/ede07e37025f5b7990d50355ce095f6575143c63) will leave some dead code (the "deterministic" rounding) and some unnecessary code (the double level rounding in worker code)

vdbergh:
@ppigazzini I reverted the last commit and squashed the PR.

vdbergh:
Thanks! Now let's see....

ppigazzini:

tomtor:
@vdbergh Thanks, this is a very nice update to the fishtest SPSA tuning feature!

ppigazzini:
@vdbergh in a corner case (e.g. integer not scaled parameter with narrow range, old SPSA algorithm) stochastic rounding could change the ordering of `round(p+c)` and `round(p-c)` (e.g when `N<p-c<p+c<N+1`). I done some math and the expected difference should be still `2c`, so the occasionally strange reversed derivative shouldn't be a problem (at least for only one parameter). Please double check this case at your convenient time :)

PR #720, opened 2020-07-03
Fix stderr empty string

ppigazzini:


ppigazzini:
@vdbergh STDERR sometimes has an empty string. This fix the exception.

PR #722, opened 2020-07-05
Clean up parse_cutechess_output()

ppigazzini:
Clean up the code removing the calls to `kill_process()`.
In this way also the Windows worker, where at the moment the cutechess-cli
STDERR is queued after the STDOUT, is able to process the STDERR.
Thanks to @vdbergh for the suggestion.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -154,11 +154,11 @@ def kill_process(p):
       subprocess.call(['taskkill', '/F', '/T', '/PID', str(p.pid)])
     else:
       p.kill()
-    p.wait()
-    p.stdout.close()
   except:
     print('Note: ' + str(sys.exc_info()[0]) + ' killing the process pid: ' + str(p.pid) + ', possibly already terminated')
-    pass
+  finally:
+    p.wait()
+    p.stdout.close()
 
 def adjust_tc(tc, base_nps, concurrency):
   factor = 1600000.0 / base_nps # 1.6Mnps is the reference core, also used in fishtest views.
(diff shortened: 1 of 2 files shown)


ppigazzini:
At the moment the PR is running successfully with several OS and python versions: Linux, WSL, portable worker (python 2.x), python 3.x form msys2 and python.org.
Anyway this PR needs a review e some other tests.

With a SPSA with a bad parameter:
- a Linux workers stops immediately the run that is moved in the "Failed" state
- a Windows worker stops the run after the first 2*concurrency games, the run isn't moved in a "Failed" state

tomtor:
>     * a Windows worker stops the run after the first 2*concurrency games, the run isn't moved in a "Failed" state

I will have a look at that.

tomtor:
> > ```
> > * a Windows worker stops the run after the first 2*concurrency games, the run isn't moved in a "Failed" state
> > ```
> 
> I will have a look at that.

The cause is this line: https://github.com/glinscott/fishtest/blob/261d0874fb41b19cbc4ba7b7c51094b87b6030c2/fishtest/fishtest/views.py#L834

A run is only considered failed when no games have been played.

I see several options:

- leave it as is
- remove the results when api stop_run() is called (will have a minor impact on the worker attributed game count)
- Introduce a 'failed' flag in the run state and use that instead of "are any games played?"

@ppigazzini I prefer the last option, I will file a separate PR for this...

ppigazzini:
@tomtor PR tested without problems for nearly 2 weeks on DEV, merged raising the version worker side.
@noobpwnftw please update the workers fleet at your convenience :)

tomtor:
Thanks @ppigazzini

noobpwnftw:
@ppigazzini Workers updated.

PR #723, opened 2020-07-08
Update tests_view.mak

unaiic:
Separate % from the actual text and put it in parentheses, just as happens above (_Adjust priority (higher is more urgent)_).

No functional change.


ppigazzini:
@tomtor Are you OK with this change?

tomtor:
Yes it is ok for me.

Op di 14 jul. 2020 17:21 schreef ppigazzini <notifications@github.com>:

ppigazzini:
PROD updated, AUTHORS updated #724, thank you @unaiic :)

PR #726, opened 2020-07-17
Add failed flag for runs

tomtor:
Supports https://github.com/glinscott/fishtest/pull/722


ppigazzini:
Tested OK with all workers, PROD updated. Thank you @tomtor :)

Issue #732, opened 2020-07-25
NNUE integration work.


vondele:
some work will be needed to allow for NNUE testing, as mentioned https://github.com/official-stockfish/Stockfish/issues/2823 

I'm of course open to suggestions and welcome all help here. 

Enabling or disabling NNUE can be done with a 'Use NNUE' option. If this is true we must download (either from the official-stockfish/networks or from the testing branch if this doesn't exist) the net and pass this with a cutechess option.

There are broadly three tasks I see right now.
- [x] communicate to users
- [x] building the binary
  - [x] requires C++17 (done in the makefile)
  - [x] picks the right target for building (https://github.com/glinscott/fishtest/pull/738)
  - [x] only participate in NNUE enabled test if hardware has avx2 or better?
- [x] getting the network
  - [x] get from the binary the name (UCI output provides EvalFile, the default value nn-sha256[12 digits].bin)
  - [x] download if needed, using new infrastructure (merge: https://github.com/glinscott/fishtest/pull/736)
- [x] instructing cutechess to do the right thing (merge: https://github.com/glinscott/fishtest/pull/738)
- [x] change `Contempt=24` in `Use NNUE=true` for `Test options` (to be done after merge of NNUE in master)

vondele:
One item that came to my mind, while we could compile for x86-64-avx2 on machines that support it, for regression tests vs. SF11 we would need to compile that branch still with x86-64-modern, as it is the only thing it supports. This transition needs some solution.

nodchip:
I have several concerns.

**AVX2 on some AMD CPUs**
AMD CPUs from Excavator micro architecture to Zen+ micro architecture executes AVX2 instructions using 128-bit registers internally, and it causes speed down.  In those CPUs, the results of comparisons between Stockfish NNUE and the original Stockfish will be unfair.
An idea to solve this problem is to create a black list of CPUIDs.  I'm not sure if this can be implemented on fishtest.

**BMI2 on AMD CPUs**
BMI2 instructions are slow on AMD CPUs.  If we disable BMI2 on AMD CPUs, nps will be increased.  On the other hand, BMI2 instructions are fast on Intel CPUs.  We could use x86-64-avx2 for AMD CPUs, and x86-64-bmi2 for Intel CPUs.

**Three-way struggle**
[...]

PR #733, opened 2020-07-25
[WIP] Download networks for NNUE

vondele:
this is a WIP patch to download networks for NNUE, using the same setup as we do for books.

after compilation binaries are queried for the EvalFile name, which is a unique name for the net.
This net is downloaded from the official repo if not found locally.
This should allow for testing the NNUE branch with 'official' books.

WIP is how to test new nets, i.e. that are not yet part of the official repo.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -573,6 +573,17 @@ def run_games(worker_info, password, remote, run, task_id):
     os.remove(zipball)
     os.chmod(cutechess, os.stat(cutechess).st_mode | stat.S_IEXEC)
 
+  # clean up old networks (keeping the 20 most recent)
+  networks = glob.glob(os.path.join(testing_dir, 'nn-*.nnue'))
+  if len(networks) > 20:
+    networks.sort(key=os.path.getmtime)
+    for old_net in networks[:-20]:
+      try:
+         os.remove(old_net)
+      except:
+         print('Note: failed to remove an old network ' + str(old_net))
+         pass
+
   # Add EvalFile with full path to cutechess options, and download networks if not already existing
   net_base = required_net(base_engine)
   if net_base:
--- a/worker/games.py
+++ b/worker/games.py
@@ -71,7 +71,7 @@ def required_net(engine):
       net = line.split(' ')[6].strip()
 
   p.wait()
-  p.stderr.close()
+  p.stdout.close()
 
   if p.returncode != 0:
      raise Exception('uci exited with non-zero code %d' % (p.returncode))
--- a/worker/games.py
+++ b/worker/games.py
@@ -573,9 +573,13 @@ def run_games(worker_info, password, remote, run, task_id):
     os.remove(zipball)
     os.chmod(cutechess, os.stat(cutechess).st_mode | stat.S_IEXEC)
 
-  # Download networks if not already existing
+  # Add EvalFile with full path to cutechess options, and download networks if not already existing
   net_base = required_net(base_engine)
+  if net_base:
+     base_options = base_options + ["option.EvalFile=%s"%(os.path.join(testing_dir, net_base))]
   net_new = required_net(new_engine)
+  if net_new:
+     new_options = new_options + ["option.EvalFile=%s"%(os.path.join(testing_dir, net_new))]
 
   for net in [net_base, net_new]:
     if net:
--- a/worker/games.py
+++ b/worker/games.py
@@ -5,6 +5,7 @@ import datetime
 import json
 import os
 import glob
+import hashlib
 import stat
 import subprocess
 import shutil
--- a/worker/games.py
+++ b/worker/games.py
@@ -594,14 +594,10 @@ def run_games(worker_info, password, remote, run, task_id):
 
   for net in [net_base, net_new]:
     if net:
-      try:
-        if not os.path.exists(os.path.join(testing_dir, net)) or not validate_net(testing_dir, net):
-           download_net(testing_dir, net)
-           if not validate_net(testing_dir, net):
-              raise Exception('Failed to validate the network: %s ' % (net))
-      except Exception as e:
-        sys.stderr.write('Exception from downloading network: %s \n' % (net))
-        print(e)
+      if not os.path.exists(os.path.join(testing_dir, net)) or not validate_net(testing_dir, net):
+         download_net(testing_dir, net)
+         if not validate_net(testing_dir, net):
+            raise Exception('Failed to validate the network: %s ' % (net))
 
   # pgn output setup
   pgn_name = 'results-' + worker_info['unique_key'] + '.pgn'
--- a/worker/games.py
+++ b/worker/games.py
@@ -61,6 +61,34 @@ def github_api(repo):
 def enc(s):
   return s.encode('utf-8')
 
+def required_net(engine):
+  net = None
+  print('Obtaining EvalFile of %s ...' % (os.path.basename(engine)))
+  p = subprocess.Popen([engine, 'uci'], stdout=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
+
+  for line in iter(p.stdout.readline,''):
+    if 'EvalFile' in line:
+      net = line.split(' ')[6].strip()
+
+  p.wait()
+  p.stderr.close()
+
+  if p.returncode != 0:
+     raise Exception('uci exited with non-zero code %d' % (p.returncode))
+
+  return net
+
+def download_net(testing_dir, net):
+  NETWORKS_REPO_URL = 'https://dfts-0.pigazzini.it/api/nn/'
+  url = NETWORKS_REPO_URL + net
+  r = requests.get(url, allow_redirects=True)
+  open(os.path.join(testing_dir, net), 'wb').write(r.content)
+
+def validate_net(testing_dir, net):
+  content = open(os.path.join(testing_dir, net), "rb").read()
+  hash = hashlib.sha256(content).hexdigest()
+  return hash[:12] == net[3:15]
+
 def verify_signature(engine, signature, remote, payload, concurrency):
   if concurrency > 1:
     busy_process = subprocess.Popen([engine], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
@@ -544,6 +572,22 @@ def run_games(worker_info, password, remote, run, task_id):
     os.remove(zipball)
     os.chmod(cutechess, os.stat(cutechess).st_mode | stat.S_IEXEC)
 
+  # Download networks if not already existing
+  net_base = required_net(base_engine)
+  net_new = required_net(new_engine)
+
+  for net in [net_base, net_test]:
+    if net:
+      try:
+        if not os.path.exists(os.path.join(testing_dir, net)) or not validate_net(testing_dir, net):
+           download_net(testing_dir, net)
+           if not validate_net(testing_dir, net):
+              raise Exception('Failed to validate the network: %s ' % (net))
+      except Exception as e:
+        sys.stderr.write('Exception from downloading network: %s \n' % (net))
+        print(e)
+
+  # pgn output setup
   pgn_name = 'results-' + worker_info['unique_key'] + '.pgn'
   if os.path.exists(pgn_name):
     os.remove(pgn_name)
--- a/worker/games.py
+++ b/worker/games.py
@@ -576,7 +576,7 @@ def run_games(worker_info, password, remote, run, task_id):
   net_base = required_net(base_engine)
   net_new = required_net(new_engine)
 
-  for net in [net_base, net_test]:
+  for net in [net_base, net_new]:
     if net:
       try:
         if not os.path.exists(os.path.join(testing_dir, net)) or not validate_net(testing_dir, net):


vondele:
This addresses part of https://github.com/glinscott/fishtest/issues/732

The TODO needs resolving, maybe @tomtor has an idea how to do this best ? 

If this could be tested on DEV, in principle one could start a first NNUE test.

gonzalezjo:
Here's the experience from the Leela side of things. tl;dr at the bottom.


In the earlier stages for Leela Chess, we had issues serving networks to all clients. It's a pretty pathological situation: as soon as a new network is released, if you're unlucky--say, for example, that nothing else is happening on Fishtest--you will have hundreds of clients all simultaneously request to download a large file from your server. At the very least, this raises bandwidth costs considerably. At worst, it creates challenges for testers with a poor connection to the server, who might receive have issues with the downloaded file (corruption, incomplete downloads, etc.,) that lead to misleading results and other technical issues.


While this isn't as big of an issue now as it was for Leela, as network sizes continue to increase (work on increasingly large networks is already under way) that might quickly change.


At the time, I reached out to GitHub to see if we could use GitHub to share the files. This was their response:

> While we aren't able to provide specific guidelines with regard to bandwidth used to deliver releases, we do monitor bandwidth usage with the goal of maintaining high performance for all of our users. Per our Terms of Service, section C.7:
> 
>> If we determine your bandwidth usage to be significantly excessive in relation to other GitHub customers, we reserve the right to suspend your Account or throttle your file hosting until you can reduce your bandwidth consumption.
> 
> Given the volume of releases necessary for your project, this would most certainly affect GitHub's performance for other users.
> 
> We do love to hear about and encourage our users' new projects, but in this case, I'm afraid we strongly discourage you from proceeding.
> 
> I'm sorry for any disappointment. Please let us know if you have any other questions.


We ended using Cloudflare as a caching CDN, upgrading the server, and sending compressed (just gzip, which is surely far from optimal) neural networks.


tl;dr:

It is a bad idea to use GitHub to host the networks. Using the Stockfish site to host the networks may be problematic in certain cases, which _could_ result in unexpected client behavior. There are mitigations that can work, like compressing the neural networks and using the free Cloudflare plan, and upgrading the server.

noobpwnftw:
I own some CDN services, you'll get those for free.

vondele:
thanks @gonzalezjo and @noobpwnftw for the info... 

might thus require more effort to get this implemented. Probably good to brainstorm a bit for ideas, as I'm quite new to this. We need both a way for a fishtest worker to find a specific net, but also for testers to upload their nets prior to the tests.

noobpwnftw:
My take: AWS S3 or OSS counterparts(for example, https://github.com/chrislusf/seaweedfs). Then put a CDN in front of it to serve downloads with cache enabled for a lower long term operational cost.

gonzalezjo:
If noob is willing to provide access to CDN services, then that seems to solve the problem. Networks can still be included in repositories as long as [a CDN](https://github.com/schme16/gitcdn.xyz) of some sort prevents us from raising traffic alarms at GitHub. 

@noobpwnftw would you be able to mirror `https://api.github.com/repos/username/reponame/zipball/branchname`?

My concern with S3 is that someone could run up the costs by making a bunch of requests with cache-invalidation headers attached.

noobpwnftw:
I can also host a server and attach CDN to it.

vondele:
having a dedicated server or API for uploading & downloading networks would work for me. From the workflow point of view, a dev could upload a network (with the unique name based on sha), prior to starting a test on fishtest, change the EvalFile UCI option on a branch and run the test. Fishtest workers (or users of the engine) would just get the file from the server using the API. Making the network 'official' after a passed SPRT test would just update the UCI option with a PR on Stockfish. The server would store and serve networks indefinitely (or at least those that are successful).

Does that make sense? Do we need to take care about malicious uploads/downloads?

If it does, since I have no experience with this, who could do the implementation?

I think this is one of the most pressing items for integration, so let's try to sort this out.

tomtor:
> Does that make sense? Do we need to take care about malicious uploads/downloads?

Yes, the current fishtest could offer a new upload network GUI which authenticates with a very basic network storage/download API impemented.

This API could be as simple as an FTP server with an account/password configured in the fishtest server configuration. We must first decide on this API, an FTP server hosted by @noobpwnftw would allow a quick implementation.

Regarding malicious uploads, I have no knowledge about the network format, I assume it specifies a collection of fused-multiply-add vectors? It would be trivial for the fishtest GUI to check the header/format of this file.

vondele:
For the API, if we would have something simple for downloads that would help, for example
```
curl -sL https://networks.stockfishchess.org/download/nn-c157e0a5755b.bin
```
would give us the `nn-c157e0a5755b.bin` if it was previously uploaded (or anything similar from server name / path etc).

I don't think we can do much format verification, as I expect the networks to change as the structure changes. We should verify on upload that the sha in the name matches the first digits of the actual sha.

For malicious, I was more thinking about some user uploading 100GB of networks, but we might be able to limit this. However, if we limit the uploads to users registered on fishtest, we at least have basic authentication.

gonzalezjo:
The upload API key / FTP login / whatever for the server that hosts the networks should definitely be stored as a secret environment variable or something for sure. As long as we do that, tie uploads to Fishtest submissions, tie Fishtest submissions to accounts, and put reasonable file size limits in place, I don't see any issues.

tomtor:
> put reasonable file size limits in place

How big are these networks?

dorzechowski:
20-30 MB currently (uncompressed). We start with 20 MB networks.

vondele:
right now 20Mb, but 30Mb is on the corner. Others get 150Mb.

tomtor:
Compression (gzip/bzip2) would make it 4 or 5 times smaller?

vondele:
about 2x for gzip, but probably depends on the content. It is already binay.

gonzalezjo:
30 MB nets aren't just around the corner: they exist and they're in use. There are specialized integer/float compression schemes (@oscardssmith did a lot of work experimenting with this for Leela, so maybe he can provide insights) that can outperform traditional techniques alone, but I don't remember if the savings were significant.


On a random network I downloaded (Sergio Vieri's `experimental-20200725-2237.bin`) DEFLATE gets a ratio of 53%, LZMA2 gets a ratio of 45%, lz4 gets a ratio of 89.05%. Zopfli:  All using default parameters.


One potentially easy source of savings is just quantizing the network files. i.e., if things are stored as 32-bit floats, then storing things as 16-bit floats could cut the file size in half.

vondele:
Don't think we should use any fancy compression. 20-30Mb is not a huge download these days (except if we'd have millions of them of course).

vondele:
@gonzalezjo do you have any numbers for the volume downloaded in the Leela project?

tomtor:
> Don't think we should use any fancy compression. 20-30Mb is not a huge download these days (except if we'd have millions of them of course).

I agree, gzip is OK.

oscardssmith:
My experiments showed that bz2 was pretty good for generic compression methods. I don't think anything too custom will help. The advantage lc0 had was that since nets were always based off of an old net, delta encoding was a reasonable possibility.

gonzalezjo:
> @gonzalezjo do you have any numbers for the volume downloaded in the Leela project?

In June 2018, it was 6 TB of bandwidth per month. Adding Clousflare dropped egress volume to 2TB/month.

noobpwnftw:
Compression ratio isn't much of an issue when bandwidth is free and never heard of any CDN incapable of doing at least GZIP for you unless expressively disabled.

noobpwnftw:
> > @gonzalezjo do you have any numbers for the volume downloaded in the Leela project?
> 
> In June 2018, it was 6 TB of bandwidth per month. Adding Cloudflare dropped egress volume to 2TB/month. Noob’s CDN could probably achieve a far better hit rate than the free Cloudflare plan.
> 
> > Don't think we should use any fancy compression. 20-30Mb is not a huge download these days (except if we'd have millions of them of course).
> 
> I don’t think Brotli is particularly fancy. On most web servers, you enable it in the same place that you enable gzip support. Any web library that sends Accept-Encoding: br (curl should support this) will automatically receive the Brotli compressed version. I don’t think it’s even possible to disable Brotli without disabling gzip on Cloudflare.
> 
> Of course, things are different if you have users download .gz files rather than allowing the web server/CDN to compress files automatically. I think doing that would be a bad idea, though: it’s just duplicating work that the server and/or CDN should be doing anyway.

In fact, I fail to see how our use case is anything exotic, including Leela. As far as I understand, if your origin server is on any of those public clouds(AWS, GCP, Azure, etc), and use any of those CDN services(Cloudflare, Akamai, etc), your origin server egress to the CDN edge is completely free, and since there is a free Cloudflare plan too, all you need to pay is just the rent of a server.

https://blog.cloudflare.com/bandwidth-alliance/

Nothing I provide is exceptional, it only saves you the server rent.

gonzalezjo:
Cloudflare is one of the only big few CDNs that doesn't let you push files to their servers for caching. With Leela and with Stockfish, we're fairly exotic in that almost all of the bandwidth consumed by a resource comes from the first few seconds of it being uploaded. 


With Leela, consistently, every single machine--no matter where it is in the world--would pull a new, very large neural network at the exact same time. Because we couldn't push anything to Cloudflare's datacenters, because they have so many, because Leela (and Stockfish) are so international, and because it would take a few hits from a region for something to be cached, Leela ended up getting both (1) pretty bad cache hit rates, and (2) pathologically bad server network speeds for a few minutes after every network release. 


tl;dr: Extremely non-uniform traffic arrival times, very uniform geographical distribution of downloaders, very large blobs to download, and lack of push zone support make for a poor cache hit rate.

noobpwnftw:
CDNs aren't flat, there are multiple cache tiers and finally edges, even with such an use case, to my understanding, it is still highly unlikely that requests can hit origin in a burst, especially with the more anti-DDOS oriented services like Cloudflare. Basically we tend to limit concurrent origin fetch on the same resource to a very small number by sacrificing TTFB, and redistribute the content from cache tiers. Also there are ranged caching, that splits big files into smaller chunks in order to reduce cache miss and TTFB, might want to see if your free plan has that.

gonzalezjo:
I found that very informative, thanks @noobpwnftw. I won't pretend to be intimately familiar with how Cloudflare's infrastructure works, especially since the data is from July 2018. I would be surprised if they hadn't improved since then. So I will not comment on most of what you said, though I appreciate your insights.


That said, I am _absolutely confident_ that origin fetches are done from multiple regions for the free version of Cloudflare. I know this because Cloudflare has a blog post about how this feature enhances their paid "Argo" service. You can scroll to the ["Argo Tiered Cache" section here](https://blog.cloudflare.com/argo/) for a brief explanation.

tl;dr: 

> Cloudflare caches content very close to end users, but if a piece of content is not in cache, the Cloudflare edge PoP must contact the origin server to receive the cacheable content. 

> Argo Tiered Cache lowers origin load, increases cache hit ratios, and improves end user experience by first asking other Cloudflare PoPs if they have the requested content when a cache miss occurs

It seems like Argo is a bit smarter than just going to the region closest to origin, but nonetheless it is something that users must pay for.

noobpwnftw:
Yeah well I'm sure they do have those today, probably just implemented differently as opposed to having those as de-facto.
No comment on their business model, for example they'll charge for additional cache rules and so on, or Argo.

My takeaway is that with a free plan you still get free edge egress, plus with bandwidth alliance with public clouds you get free origin egress, so those details doesn't really matter anyways. Just saying that infrastructure cost will not going to be a big problem even without me providing it.

tomtor:
> What needs to be developed is basically just a file saver, listens to connections from current fishtest server for uploads(since fishtest server already authenticated the user and got uploaded net), sanitize it and save it somewhere on disk.

That is what a simple FTP server could do.

> Then a simple nginx configuration which serves those saved files via HTTP.

The NGINX could serve the ftp upload directory.

Anyway, I do not particularly favor FTP for transport.

An alternative would be a minimal Python HTTP server which accepts a POST (with basic authentication when your server has HTTPS or Digest-authentication otherwise) which contains the network data and the POST path specifies the filename. This program could also serve the files, there is no need for an additional NGINX when traffic is low.

@noobpwnftw: What do you consider when you mention "sanitize it"?

I expect a program as described above is written many times, without the sanitize part.

Nginx, can also be used:

https://www.nginx.com/resources/wiki/modules/upload/

@noobpwnftw What is not clear to me is why the current fishtest server could not be the origin for the CDN.

ppigazzini:
> If this could be tested on DEV, in principle one could start a first NNUE test.

@vondele of course we'll test the PR when ready, simply let me know. I will suggest some improvements for the `subprocess` code, I need to re find the references...

vondele:
Thanks! If we move away from github for storing the networks (as we probably should) this PR is in part obsolete. So, I think I need to have some patience. 

Nevertheless, the NNUE branch is nearing completion, so getting the net download and testing infrastructure done is increasingly going up in priority.  Unfortunately, I have no experience in the webserver tech, so I'm afraid I won't be able to contribute much to the discussion/coding (I'll also be away from computers Wed/Thu).

vondele:
I think it is better to take some time to do it right initially, after all, the initial testing can be done locally. The fishtest integration is important for the final merge, and the future development of master, including the nets.

One nice thing with the upload is that the nets will remain available to everybody and test will be repeatable in the future. BTW, the upload page should presumably mention that the nets uploaded are made available under a CC0 or CC BY 4.0 license, and keep minimal 'metadata' such as the data submitted and the fishtest user ID.

tomtor:
I currently have not much time to write code for this but fishtest should get as minimal first implementation:
- An upload `sha.nnue` network page which shows CC-license info.
- Store this network file in MongoDb with id `sha` and username if it does not already exist (code is similar to PGN upload)
- Serve this network if it is requested by the worker (code similar to PGN download)

This minimal code can be expanded later if a CDN feature is added.

vondele:
@tomtor that sounds right from my point of view. Maybe @linrock has a chance to contribute to the implementation?

noobpwnftw:
> @noobpwnftw: What do you consider when you mention "sanitize it"?
> 
> I expect a program as described above is written many times, without the sanitize part.
> 
> Nginx, can also be used:
> 
> https://www.nginx.com/resources/wiki/modules/upload/
> 
> @noobpwnftw What is not clear to me is why the current fishtest server could not be the origin for the CDN.

By sanitizing I mean first you don't want to host public downloads for arbitrary files, like porn or some other malicious files, second you might need a consistent versioning/naming scheme instead of "my very first shiny net.bin" or so.

If your current fishtest server is on one of those clouds that provides egress bandwidth waiver for CDN interconnect, then it is fine. Otherwise it may get costly for both.

vondele:
@noobpwnftw for the naming I proposed the simple nn-SHA[12 digits].nnue 

I'm not sure how to sanitize the files practically. So far they are just binary blobs, one could embed anything. I'm not sure they have a particular structure. Would you suggest some webservice that can do that ? 

We will require a fishtest login to upload, which at least is a bit of a barrier.

noobpwnftw:
@vondele Implementing a simple form of https://github.com/official-stockfish/Stockfish/blob/5c616bc46bfb6975494aae11f46a166adf47e0c9/src/nnue/evaluate_nnue.cpp#L145 would be sufficient, the current net does have headers with hash checksums, will need that to guard against sending nets that won't load anyways.

tomtor:
I will make a minimal upload network PR the next few days unless someone else wants to work on this.

vondele:
thanks!

BTW, I'm not sure we should sanitize by parsing the headers and checksums. I assume that code will evolve as networks evolve, and it will be tedious to keep fishtest updated, and it is better to keep it simple.

Initially the fact that we require fishtest login for upload should be enough IMO. Longer term, I think we could consider our approval mechanism, e.g. we could allow public downloads only after a fishtest approver has approved a corresponding test, or unblocked the download in the pending list (like we do with new users).

vondele:
new version of this patch, using the API of #736. Still hard-coded to the DEV server. 

I've added some of the subprocess code @ppigazzini posted in a different PR.

worker/games.py
```diff
@@ -61,6 +61,34 @@ def github_api(repo):
...
+  net = None
+  print('Obtaining EvalFile of %s ...' % (os.path.basename(engine)))
+  p = subprocess.Popen([engine, 'uci'], stdout=subprocess.PIPE, universal_newlines=True, bufsize=1, close_fds=not IS_WINDOWS)
+
+  for line in iter(p.stdout.readline,''):
+    if 'EvalFile' in line:
+      net = line.split(' ')[6].strip()
+
+  p.wait()
+  p.stderr.close()
```
ppigazzini, line 74: `p.stderr` was not piped, so here you need `p.stdout.close()`

vondele: thanks!

vondele:
@ppigazzini if you have time, it would be useful to test this worker change with the current dev server (while it has the upload/download patch there as well). In principle, it should be enough to do a first test of the current nnue-player-wip branch vs. master (say a small amount of fixed number of games, proper results need some more work), or, more likely, find some bugs.

ppigazzini:
@vondele workers updates with this PR, feel free to add some tests.

vondele:
@ppigazzini started a test...

ppigazzini:
@vondele worker output:
```
Exception running games:
  File "worker.py", line 154, in worker
  File "/home/usr00/_git/__test_folder00/fishtest/worker/games.py", line 576, in run_games
    for net in [net_base, net_test]:
NameError: name 'net_test' is not defined
```

vondele:
OK, fixed.

vondele:
OK, network download seems to be OK now. However, this runs into the problem that cutechess-cli needs to send the EvalFile option with the full path to the file before it can work. I'll need to fix that as well.

vondele:
now playing seems to work as well. From the first few games, even with the 'modern' compile (and not avx2 as it should be) performance is roughly equal.

vondele:
This patch still needs some additional code to remove locally old downloaded networks (keeping the most recent as we do with binaries).

ppigazzini:
@vondele workers updates, I start some testing.

vondele:
Thanks! I connected my worker with the changes and it seemed OK. 

Note that to test the NNUE functionality the test options should include `Use NNUE=true` for the nnue-player-wip branch.

I'm meanwhile working on the architecture selection code.

vondele:
so, still needs some work. On failure to download, the worker should skip participation in the test... that's seemingly not working:
```
Obtaining EvalFile of stockfish_0d2965f1754eb2026dfef1b998aac52fb0f3f42d ...
Exception from downloading network: nn-c157e0a5755b.nnue 
HTTPSConnectionPool(host='dfts-0.pigazzini.it', port=443): Max retries exceeded with url: /api/nn/nn-c157e0a5755b.nnue (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7fd70d1835f8>: Failed to establish a new connection: [Errno 110] Connection timed out',))
Verifying signature of stockfish_0d2965f1754eb2026dfef1b998aac52fb0f3f42d ...
Verifying signature of stockfish_d89730d5c8dcf10eb9e1d91a81f903d9fc3c949a ...
CPU factor : 0.640937 - tc adjusted to 6.409+0.064
[...]
```
Any hints on how to achieve this?

worker/games.py
```diff
@@ -544,6 +573,37 @@ def parse_options(s):
...
+  for net in [net_base, net_new]:
+    if net:
+      try:
+        if not os.path.exists(os.path.join(testing_dir, net)) or not validate_net(testing_dir, net):
+           download_net(testing_dir, net)
+           if not validate_net(testing_dir, net):
+              raise Exception('Failed to validate the network: %s ' % (net))
+      except Exception as e:
+        sys.stderr.write('Exception from downloading network: %s \n' % (net))
+        print(e)
```
tomtor: You could rethrow the Exception here (`raise`), ore just remove the try/except code completely. EDIT: I see you just did :-)

I expect that the error will be handled correctly in the caller. Just like the `raise` in line 601

vondele: Thanks! Yes, seems to work.

vondele:
I'll close this one in favor of the more complete https://github.com/glinscott/fishtest/pull/738

vondele:
Thanks for the contributions by everyone BTW.

Issue #731, opened 2020-07-25
SPSA tests have timeouts


tomtor:
```
Finished game 18 (Base-50fdcc9c94 vs New-50fdcc9c94): 1-0 {White wins by adjudication}
Score of New-50fdcc9c94 vs Base-50fdcc9c94: 6 - 6 - 15  [0.500] 27
Finished game 26 (Base-50fdcc9c94 vs New-50fdcc9c94): 1/2-1/2 {Draw by adjudication}
Score of New-50fdcc9c94 vs Base-50fdcc9c94: 6 - 6 - 16  [0.500] 28
  Task updated successfully in 1.44457s
Elo difference: 0.0 +/- 85.8, LOS: 50.0 %, DrawRatio: 57.1 %
Finished match
Finished match cleanly
2020-07-25 09:11:03.450461 is past end time 2020-07-25 09:11:03.450451
Fetched SPSA parameters successfully in 0.045672s
```

I notice this since a few days. Cutechess does not exit occasionally.

@ppigazzini Any ideas?

tomtor:
While it hangs python is 150% busy

I suspect it is introduced with 

https://github.com/glinscott/fishtest/commit/98e846949a39aeb7c741f463b0307edaca5372db

tomtor:
When it loops it has a line with length 0 here:

https://github.com/glinscott/fishtest/blob/c16bc060f6c049fde2a1205f2ca8bfba44bfded2/worker/games.py#L301

and sometimes it exits after some looping due to:

https://github.com/glinscott/fishtest/blob/c16bc060f6c049fde2a1205f2ca8bfba44bfded2/worker/games.py#L297

tomtor:
When I add:

```
    if len(line) == 0 and p.poll() is not None:
       break
```
before

https://github.com/glinscott/fishtest/blob/c16bc060f6c049fde2a1205f2ca8bfba44bfded2/worker/games.py#L301

it works OK but it is not clear to me why this effect only SPSA.

tomtor:
> it works OK but it is not clear to me why this effects only SPSA.

It really is some kind of tricky race condition:

```
if len(line) == 0 and p.poll() is not None:
       print("break")
       break
```

works fine, but removing the `print("break")` makes it hang on occasion...

tomtor:
What also seems to work is adding:
```
      return { 'task_alive': True }
```

after
[...]

PR #734, opened 2020-07-27
Fix "tc_limit" for SPSA

ppigazzini:
For SPSA raise "tc_limit" only once and not at any iteration.
~~Raise also the "scaled_tc" decimal places to 5 for ultra low time control.~~

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -184,7 +184,7 @@ def adjust_tc(tc, base_nps, concurrency):
   else:
     time_tc = float(chunks[0])
 
-  # Rebuild scaled_tc now
+  # Rebuild scaled_tc now: cutechess-cli and stockfish parse 3 decimal places
   scaled_tc = '%.3f' % (time_tc * factor)
   tc_limit = time_tc * factor * 3
   if increment > 0.0:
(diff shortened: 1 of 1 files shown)


vondele:
cutechess will only parse 3 digits of the tc, as far as I know. Similarly, in Stockfish time has millisecond resolution only.

ppigazzini:
Thanks @vondele, added a comment about cutechess-cli and stockfish time control resolution.

tomtor:
@ppigazzini Nice you found this bug.

I think this explains why I saw https://github.com/glinscott/fishtest/issues/731 only on SPSA tests. It did probably also happen on SPRT tests, but the timeout was much faster, so I did not notice it hanging.

ppigazzini:
> @ppigazzini Nice you found this bug.
> 
> I think this explains why I saw #731 only on SPSA tests. It did probably also happen on SPRT tests, but the timeout was much faster, so I did not notice it hanging.

SPRT and NumGames return here skipping the timeout
https://github.com/glinscott/fishtest/blob/c16bc060f6c049fde2a1205f2ca8bfba44bfded2/worker/games.py#L373-L375

PR #736, opened 2020-07-29
NN upload/download support

tomtor:
See #733 

@ppigazzini Note that I added a new index for `nns` in `create_indices.py`

### DIFF
--- a/fishtest/fishtest/templates/nn_upload.mak
+++ b/fishtest/fishtest/templates/nn_upload.mak
@@ -1,0 +1,41 @@
+<%inherit file="base.mak"/>
+
+<header style="text-align: center; padding-top: 7px">
+  <legend>Neural Network Upload</legend>
+
+  <section class="instructions" style="margin-bottom: 35px">
+    <p>
+    Please read the
+    <a href="https://github.com/glinscott/fishtest/wiki/Creating-my-first-test">Testing Guidelines</a>
+    before uploading your network.
+    </p>
+    <p>
+    By uploading you license your network under a
+    <a href="https://creativecommons.org/share-your-work/public-domain/cc0/">CC0</a> license.
+    </p>
+    <p>
+    You are the author of this network or have obtained the network with a CC0 license.
+    </p>
+    <p>
+    Your uploaded network will be available for public download and listed on the
+    <a href="/nns">NN stats page</a>.
+    </p>
+  </section>
+</header>
+
+<div>
+  <form id="upload-nn" action="${request.url}" method="POST" enctype="multipart/form-data">
+    <div class="control-group">
+    <label class="control-label">Select your Network file (nn-[SHA256 first 12 digits].nnue):</label>
+      <div class="controls">
+        <input id="network" name="network" type="file" value="" />
+      </div>
+    </div>
+
+    <div class="control-group">
+      <div class="controls">
+        <button type="submit" class="btn btn-primary">Agree to License & Upload</button>
+      </div>
+    </div>
+  </form>
+</div>
--- a/fishtest/fishtest/__init__.py
+++ b/fishtest/fishtest/__init__.py
@@ -51,6 +51,7 @@ def main(global_config, **settings):
 
   config.add_route('home', '/')
   config.add_route('login', '/login')
+  config.add_route('nn_upload', '/upload')
   config.add_route('logout', '/logout')
   config.add_route('signup', '/signup')
   config.add_route('user', '/user/{username}')
@@ -59,6 +60,7 @@ def main(global_config, **settings):
   config.add_route('users', '/users')
   config.add_route('users_monthly', '/users/monthly')
   config.add_route('actions', '/actions')
+  config.add_route('nns', '/nns')
 
   config.add_route('tests', '/tests')
   config.add_route('tests_machines', '/tests/machines')
@@ -88,6 +90,7 @@ def main(global_config, **settings):
   config.add_route('api_upload_pgn', '/api/upload_pgn')
   config.add_route('api_download_pgn', '/api/pgn/{id}')
   config.add_route('api_download_pgn_100', '/api/pgn_100/{skip}')
+  config.add_route('api_download_nn', '/api/nn/{id}')
   config.add_route('api_get_elo', '/api/get_elo/{id}')
 
   config.scan()
--- a/fishtest/fishtest/actiondb.py
+++ b/fishtest/fishtest/actiondb.py
@@ -23,6 +23,9 @@ class ActionDb:
   def new_run(self, username, run):
     self._new_action(username, 'new_run', run)
 
+  def upload_nn(self, username, network):
+    self._new_action(username, 'upload_nn', network)
+
   def modify_run(self, username, before, after):
     self._new_action(username, 'modify_run',
                      {'before': before, 'after': after})
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -1,9 +1,11 @@
 import copy
 import base64
+import zlib
 from datetime import datetime
 
 import requests
 from pyramid.httpexceptions import HTTPUnauthorized, exception_response
+from pyramid.httpexceptions import HTTPFound
 from pyramid.view import view_config, view_defaults, exception_view_config
 from pyramid.response import Response
 
@@ -209,6 +211,18 @@ class ApiView(object):
     if urls is None:
       raise exception_response(404)
     return urls
+
+
+  @view_config(route_name='api_download_nn')
+  def download_nn(self):
+    nn = self.request.rundb.get_nn(self.request.matchdict['id'])
+    if nn is None:
+      raise exception_response(404)
+    #self.request.response.content_type = 'application/x-chess-nnue'
+    #self.request.response.body = zlib.decompress(nn['nn'])
+    #return self.request.response
+    return HTTPFound("https://data.stockfishchess.org/nn/"
+                    + self.request.matchdict['id'])
 
 
   @view_config(route_name='api_stop_run')
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -38,6 +38,7 @@ class RunDb:
     self.userdb = UserDb(self.db)
     self.actiondb = ActionDb(self.db)
     self.pgndb = self.db['pgns']
+    self.nndb = self.db['nns']
     self.runs = self.db['runs']
     self.deltas = self.db['deltas']
 
@@ -164,6 +165,25 @@ class RunDb:
   def get_pgn_100(self, skip):
     return [p['run_id'] for p in
             self.pgndb.find(skip=skip, limit=100, sort=[('_id', DESCENDING)])]
+
+  def upload_nn(self, userid, name, nn):
+    self.nndb.insert_one({'user': userid, 'name': name, 'downloads': 0})
+                          # 'nn': Binary(zlib.compress(nn))})
+    return {}
+
+  def get_nn(self, name):
+    # nn = self.nndb.find_one({'name': name})
+    nn = self.nndb.find_one({'name': name}, {'nn': 0})
+    if nn:
+      self.nndb.update_one({'name': name}, {'$inc': { 'downloads': 1 }})
+      return nn
+    return None
+
+  def get_nns(self, limit):
+    return [dict(n, time=n['_id'].generation_time) for n in
+            self.nndb.find({}, {'nn': 0},
+                           limit=limit, sort=[('_id', DESCENDING)])]
+
 
   # Cache runs
   run_cache = {}
--- a/fishtest/fishtest/templates/actions.mak
+++ b/fishtest/fishtest/templates/actions.mak
@@ -13,6 +13,7 @@ Show only:
   <option value="purge_run">Purge Run</option>
   <option value="block_user">Block/Unblock User</option>
   <option value="update_stats">System Events</option>
+  <option value="upload_nn">Upload NN file</option>
 </select>
 &nbsp;From user:
 <input id="user" type="text" name="user" class="submit_on_enter">
--- a/fishtest/fishtest/templates/base.mak
+++ b/fishtest/fishtest/templates/base.mak
@@ -37,11 +37,15 @@
         <li><a href="/tests/user/${request.authenticated_userid}">My tests</a></li>
       %endif
       <li><a href="/tests/run">New</a></li>
+      %if request.authenticated_userid:
+        <li><a href="/upload">Upload NN</a></li>
+      %endif
 
       <li class="nav-header">Misc</li>
       <li><a href="/users">Users</a></li>
       <li><a href="/users/monthly">Top Month</a></li>
       <li><a href="/actions">Actions</a></li>
+      <li><a href="/nns">NN stats</a></li>
       <li><a href="/html/SPRTcalculator.html?elo-0=-0.5&elo-1=1.5&draw-ratio=0.61&rms-bias=0" target="_blank">SPRT Calc</a></li>
 
       <li class="nav-header">Github</li>
--- a/fishtest/fishtest/templates/nns.mak
+++ b/fishtest/fishtest/templates/nns.mak
@@ -1,0 +1,28 @@
+<%inherit file="base.mak"/>
+<h3>Neural Net download and statistics</h3>
+
+<p>
+These networks are available for download under
+<a href="https://creativecommons.org/share-your-work/public-domain/cc0/">CC0</a> license.
+</p>
+
+<table class="table table-striped table-condensed" style="max-width: 700px;">
+ <thead>
+  <tr>
+   <th>Time</th>
+   <th>Network</th>
+   <th>Username</th>
+   <th>Downloads</th>
+  </tr>
+ </thead>
+ <tbody>
+ %for nn in nns:
+  <tr>
+   <td>${nn['time'].strftime("%y-%m-%d %H:%M:%S")}</td>
+   <td><a href="api/nn/${nn['name']}">${nn['name']}</a></td>
+   <td>${nn['user']}</td>
+   <td>${nn.get('downloads', 0)}</td>
+  </tr>
+ %endfor
+ </tbody>
+</table>
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -5,6 +5,7 @@ import time
 import threading
 import re
 import html
+import hashlib
 
 import requests
 from pyramid.security import remember, forget, authenticated_userid, has_permission
@@ -62,6 +63,56 @@ def login(request):
     request.session.flash(token['error'], 'error')  # 'Incorrect password'
   return {}
 
+
+@view_config(route_name='nn_upload', renderer='nn_upload.mak',
+             require_csrf=True)
+def upload(request):
+  userid = authenticated_userid(request)
+  if not userid:
+    request.session.flash('Please login')
+    return HTTPFound(location=request.route_url('login'))
+  if request.method != 'POST':
+    return {}
+  try:
+    filename = request.POST['network'].filename
+    input_file = request.POST['network'].file
+    network = input_file.read()
+    errors = []
+    if len(network) >= 100000000:
+      errors.append('Network must be < 100MB')
+    if not re.match(r"^nn-[0-9a-f]{12}\.nnue$", filename):
+      errors.append('Name must match "nn-[SHA256 first 12 digits].nnue"')
+    hash = hashlib.sha256(network).hexdigest()
+    if hash[:12] != filename[3:15]:
+      errors.append('Wrong SHA256 hash: ' + hash[:12]
+                    + ' Filename: ' + filename[3:15])
+    if request.rundb.get_nn(filename):
+      errors.append('Network already exists')
+    if errors:
+      for error in errors:
+        request.session.flash(error, 'error')
+      return {}
+  except:
+    request.session.flash('You must specify a network filename', 'error')
+    return {}
+
+  try:
+    with open(os.path.expanduser('~/fishtest.upload'), 'r') as f:
+      upload_server = f.read().replace('\n', '')
+      upload_server = upload_server + '/' + filename
+      response = requests.post(upload_server, data=network)
+      if response.status_code != 200:
+        print('Network upload failed: ' + str(response.status_code))
+        request.session.flash('Network upload failed: '
+                              + str(response.status_code), 'error')
+        return {}
+  except Exception as e:
+    print("NN Upload fails or not configured: " + str(e))
+
+  request.actiondb.upload_nn(authenticated_userid(request), filename)
+  request.rundb.upload_nn(userid, filename, network)
+
+  return HTTPFound(location=request.route_url('nns'))
 
 @view_config(route_name='logout', require_csrf=True, request_method='POST')
 def logout(request):
@@ -122,6 +173,16 @@ def signup(request):
   return {}
 
 
+@view_config(route_name='nns', renderer='nns.mak')
+def nns(request):
+
+  nns_list = []
+
+  for nn in request.rundb.get_nns(100):
+    nns_list.append(nn)
+  return {'nns': nns_list}
+
+
 @view_config(route_name='actions', renderer='actions.mak')
 def actions(request):
   search_action = request.params.get('action', '')
@@ -137,6 +198,9 @@ def actions(request):
     if action['action'] == 'update_stats':
       item['user'] = ''
       item['description'] = 'Update user statistics'
+    elif action['action'] == 'upload_nn':
+      item['user'] = ''
+      item['description'] = 'Upload ' + action['data']
     elif action['action'] == 'block_user':
       item['description'] = (
           'blocked' if action['data']['blocked'] else 'unblocked')
@@ -259,7 +323,7 @@ def users_monthly(request):
 
 
 def get_master_bench():
-  bs = re.compile(r'(^|\s)[Bb]ench[ :]+([0-9]+)', re.MULTILINE)
+  bs = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
   for c in requests.get(
       'https://api.github.com/repos/official-stockfish/Stockfish/commits').json():
     if not 'commit' in c:
@@ -324,7 +388,7 @@ def validate_form(request):
     'info': request.POST['run-info'],
   }
 
-  if not re.match('^([1-9]\d*/)?\d+(\.\d+)?(\+\d+(\.\d+)?)?$', data['tc']):
+  if not re.match(r"^([1-9]\d*/)?\d+(\.\d+)?(\+\d+(\.\d+)?)?$", data['tc']):
     raise Exception('Bad time control format')
 
   if request.POST.get('rescheduled_from'):
@@ -348,21 +412,21 @@ def validate_form(request):
     if 'commit' not in c:
       raise Exception('Cannot find branch in developer repository')
     if len(data['new_signature']) == 0:
-      bs = re.compile(r'(^|\s)[Bb]ench[ :]+([0-9]+)', re.MULTILINE)
+      bs = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
       m = bs.search(c['commit']['message'])
       if m:
         data['new_signature'] = m.group(2)
       else:
         raise Exception("This commit has no signature: please supply it manually.")
     if len(data['info']) == 0:
-        data['info'] = ('' if re.match('^[012]?[0-9][^0-9].*', data['tc'])
+        data['info'] = ('' if re.match(r"^[012]?[0-9][^0-9].*", data['tc'])
                         else 'LTC: ') + strip_message(c['commit']['message'])
 
   # Check that the book exists in the official books repo
   if len(data['book']) > 0:
     api_url = 'https://api.github.com/repos/official-stockfish/books/contents'
     c = requests.get(api_url).json()
-    matcher = re.compile(r'\.(epd|pgn)\.zip$')
+    matcher = re.compile(r"\.(epd|pgn)\.zip$")
     valid_book_filenames = [file['name'] for file in c if matcher.search(file['name'])]
     if data['book'] + '.zip' not in valid_book_filenames:
       raise Exception('Invalid book - ' + data['book'])
@@ -402,7 +466,7 @@ def validate_form(request):
     api_url = data['tests_repo'].replace('https://github.com',
                                          'https://api.github.com/repos')
     api_url += '/commits'
-    bs = re.compile(r'(^|\s)[Bb]ench[ :]+([0-9]+)', re.MULTILINE)
+    bs = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
     for c in requests.get(api_url).json():
       m = bs.search(c['commit']['message'])
       if m:
--- a/fishtest/utils/create_indexes.py
+++ b/fishtest/utils/create_indexes.py
@@ -53,6 +53,10 @@ def create_pgns_indexes():
   print('Creating indexes on pgns collection')
   db['pgns'].create_index([('run_id', DESCENDING)])
 
+def create_nns_indexes():
+  print('Creating indexes on nns collection')
+  db['nns'].create_index([('name', DESCENDING)])
+
 def create_users_indexes():
   db['users'].create_index('username', unique=True)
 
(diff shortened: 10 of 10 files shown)


tomtor:
@ppigazzini @vondele I added the NN upload code. I believe it's complete and I did test it locally.

tomtor:
@vondele Is the SHA calculation SHA1 or SHA256? I implemented SHA1.

tomtor:
Changed to sha256

ppigazzini:
@tomtor DEV updated

fishtest/fishtest/templates/nn_upload.mak
```diff
@@ -0,0 +1,35 @@
...
+
+  <section class="instructions" style="margin-bottom: 35px">
+    <p>
+    Please read the
+    <a href="https://github.com/glinscott/fishtest/wiki/Creating-my-first-test">Testing Guidelines</a>
+    before uploading your network.
+    </p>
+    <p>
+    By uploading you license your network under a
+    <a href="https://creativecommons.org/share-your-work/public-domain/cc0/">CC0</a> or
```
ppigazzini, line 14: @tomtor @vondele I'm wondering how can work an "or" between two licenses.

tomtor: Yes, I wondered a bit about that too. 

Dual licensing is however possible in general. The party using the nn has to comply with one of the offered licenses.

@vondele might have suggestions for other wording.

ppigazzini: With a "xor" we should add a couple of checkboxes and keep trace of the user choice.

tomtor:
@ppigazzini On dev I get with a 20MB file (standard net size):

```
413 Request Entity Too Large
nginx/1.14.0 (Ubuntu)
```

ppigazzini:
@tomtor nginx should work now with file up to 100MB.
Here the error uploading a 20MB net.
```
Jul 30 18:49:09 vps1572612706 pserve[11719]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pymongo/message.py", line 1073, in _raise_document_too_large
Jul 30 18:49:09 vps1572612706 pserve[11719]:     " bytes." % (doc_size, max_size))
Jul 30 18:49:09 vps1572612706 pserve[11719]: pymongo.errors.DocumentTooLarge: BSON document too large (20000222 bytes) - the connected server supports BSON document sizes up to 16793598 bytes.
```

I'm worried about saving the net files in mongodb: IMO saving the net files in the file system and the info in mongodb is a safer move. No problem experimenting on DEV, though.

fishtest/fishtest/templates/nn_upload.mak
```diff
@@ -0,0 +1,35 @@
...
+    <a href="https://creativecommons.org/licenses/by/4.0/">CC BY 4.0</a> license.
+    </p>
+  </section>
+</header>
+
+<div>
+  <form id="upload-nn" action="${request.url}" method="POST" enctype="multipart/form-data">
+    <div class="control-group">
+    <label class="control-label">Select your Network file (nn-SHA[12 digits].nnue):</label>
```
ppigazzini, line 23: IMO better `nn-[SHA first 12 digits].nnue`

tomtor:
> saving the net files in the file system

If we compress first then we could store source files of about 30MB. 
We could start with that. The filesystem needs more code and we have to arrange a backup or just ignore backupping them

fishtest/fishtest/templates/nn_upload.mak
tomtor: Fixed

ppigazzini:
DEV updated.

fishtest/fishtest/templates/nn_upload.mak
vondele: We should pick one, it can't be an or, we wouldn't know which one. The main difference between CC0 and CC BY is that we have the attribution clause. CC0 is 'easiest' in that we have no requirements on sharing the data. If we go for CC BY we would need to be able to list the networks with their author (i.e. capture that info on upload and at least list networks with their author somewhere). Possibly best to go CC0, since if we would violate the terms of the license (i.e. not able to ack), the license would be invalid

vondele:
Thanks! Let me have a look at it.

vondele:
so, the api seems to work as advertised, and is from my point of view good. I can't judge the technical part (like putting stuff in mongdb, bandwidth, CDN etc), so trust your judgement there.

Some questions/wishes:
* can we retrieve the date of the upload from the db, or should we add a field to it?
* would be nice if a NN upload would appear on the 'Actions' tab, if possible (will allow us to track).
* should we limit the number of uploads a user can have per day (e.g. 4 a day per user, or 10 per day total), or limit automatic uploads (e.g. the captcha). People training networks might otherwise do 'an hourly dump'. We could also just mention this on the docs page, however, and just contact the author if we see a problem (but implementing nothing could cause the server to fail quickly).

tomtor:
@vondele yes, the date can be retrieved from the MongoDb id.

I will add an Action for uploads.

tomtor:
> * should we limit the number of uploads a user can have per day (e.g. 4 a day per user, or 10 per day total), or limit automatic uploads (e.g. the captcha). People training networks might otherwise do 'an hourly dump'. We could also just mention this on the docs page, however, and just contact the author if we see a problem (but implementing nothing could cause the server to fail quickly).

The fishtest server cannot store huge amounts of networks. Can old networks be removed? Could we store eg just the last 10 of each user? I think we should start with some guidance rules on the upload page and implement measures later.

The CDN implementation would also solve the storage requirement. The fishtest server would just have to transfer the network and does not need to store it.

@ppigazzini or @noobpwnftw could implement the CDN  upload server with an NGINX configuration.

fishtest/fishtest/templates/nn_upload.mak
tomtor: CC0 picked

vondele:
Ideally we can keep networks forever, would be really needed for any network that we reference to in the official stockfish branch.

I'm definitely OK with writing guidance rules before we do anything more involved. My worry would be somebody who floods the server before we have time to block that user, so even a global counter that limits uploads to 10, and is reset once a day would work for me.

fishtest/fishtest/templates/nn_upload.mak
```diff
@@ -11,8 +11,7 @@
     </p>
     <p>
     By uploading you license your network under a
-    <a href="https://creativecommons.org/share-your-work/public-domain/cc0/">CC0</a> or
-    <a href="https://creativecommons.org/licenses/by/4.0/">CC BY 4.0</a> license.
+    <a href="https://creativecommons.org/share-your-work/public-domain/cc0/">CC0</a>.
```
vondele, line 14: Maybe we should add one more sentence.
`You are the author of this network or have obtained the network with a CC0 license.`

tomtor: Done

noobpwnftw:
CDN will not cache things forever, eviction rules are complicated, the origin server must to be able to serve the content as long as it is needed. for NGINX on the origin, you need 1) sufficiently large "client_max_body_size", 2) appropriate "expires" header to indicate how long should the content be cached, 3) don't redirect HTTP->HTTPS as CDNs will do SSL termination and have the option to redirect there.

For service, you can use a Cloudflare free plan, which will need you to use their DNS for the domain name server, then proxy/passthrough on a per host(subdomain) basis, set up SSL and so on. For mine, need your subdomain SSL cert, then I set up the rest and will give you a CNAME which you should point the DNS of the subdomain to. Non-free Cloudflare plan also has CNAME option so that you don't need to park the entire domain DNS.

tomtor:
> CDN will not cache things forever

The CDN does not need to cache forever, but the server to which fishtest uploads the networks should preserve the (important?) networks.

> For mine, need your subdomain SSL cert,

Well, we do not really need a fishtest related domain. Any domain will do because the domain name is only used by the worker script to fetch networks.

So we need someone to supply an upload network server with sufficient storage and who is willing to configure the upload server, DNS and (Cloudflare) CDN setup.

vondele:
@tomtor the largest number of downloads will probably come from users after an official release, so having a domain name with something 'stockfish' is nice. However, if stockfishchess.org just redirects the download, the name of the server doesn't really much. Anyway, to the extend I can judge (and that's just  only a little bit) I'm fine with the current setup to start.

noobpwnftw:
So you are going to separate fishtest server and the file server after all? I thought people want to host it with the current server.

Right, any domain will do, but will still need those to set up any CDN(although some will just do ACME for you).

tomtor:
> So you are going to separate fishtest server and the file server after all? I thought people want to host it with the current server.

The current setup is that fishtest does the bookkeeping, authorization, upload-form AND the actual storage AND download api.

But fishtest server has limited resources so I prefer offloading the network downloads (and permanent storage) to another provider.

We can start with the current setup, but I would like to transfer uploaded networks (with a simple HTTP-POST to the network-storage server) as soon as something is arranged. I will add this POST code soon, so we just have to configure it when details are clear.

ppigazzini:
@tomtor DEV updated

tomtor:
> so even a global counter that limits uploads to 10, and is reset once a day would work for me.

@vondele We could limit the interval between uploads to eg 1 hour or 3 hours or ...?

vondele:
I think the global counter is better in the non-malicious case (i.e. allows two NN authors to start a test at the same time), a malicious uploader will be able to do a couple of uploads in any case, preventing non-malicious uploads, but without impact on server operation.

vondele:
another thing that might be interesting in the long run would be a download counter for each network. Could also be useful if ever we need to purge stuff.

tomtor:
> We could limit the interval between uploads to eg 1 hour or 3 hours or ...?

I was not clear but this limit would be per user.

>a download counter for each network

If we offload the downloads to a CDN or other server we will not be able to count on the fishtest server. Or we should allways use the fishtest api and redirect requests to the bulk download URL.

vondele:
OK, with limit per user would work better, but e.g. a user that makes an upload mistake (i.e. uploads wrong net) would be punished by a 3h wait. A counter that resets once per day still sounds better. BTW, we won't have compute resources to test more than 10 nets per day, I would say, so it is not bad to have (in fact we could throttle through this net testing if it would become excessive, but I don't think it will be a problem). 

I would always use the fishtest API and redirect requests. It is also not necessary to be absolutely accurate (i.e. if somebody uses the bulk download URL and it is not counted it is fine.). Note that in the WIP README.md I already refer to the fishtest download: https://github.com/official-stockfish/Stockfish/blob/nnue-player-wip/README.md#files (let me know if that works for you).

tomtor:
> let me know if that works for you

@vondele Yes, that's fine.

noobpwnftw:
I've made a file server and sent logins to @vondele .
You can work on setting it up and probably register a domain so that CDN setup can continue.

vondele:
@noobpwnftw thanks! 

You're OK I share this with @tomtor and @ppigazzini right?

noobpwnftw:
@vondele Yes, of course.
You might want to also change the credentials since it's the defaults.

tomtor:
> @vondele Yes, of course.
> You might want to also change the credentials since it's the defaults.

@noobpwnftw Thanks, that's very nice!

Best to disallow password login and only allow SSH public key login.

I have added my key, we can disable password login as soon as @noobpwnftw and @vondele and @ppigazzini added theirs.

I'll arrange with @ppigazzini to discuss the best upload implementation. Either NGINX of a dedicated Python script.

I prefer an NGINX configuration (see https://www.nginx.com/resources/wiki/modules/upload/ ) because we will have the smallest number of dependencies and @ppigazzini has experience with the current fishtest NGINX configuration.

Meanwhile I added a simple NN-download stats page to this PR.

tomtor:
@ppigazzini https://dfts-0.pigazzini.it/nns gives an error but works fine for me locally?

Do you have a hint from the logs?

vondele:
@tomtor key added.

ppigazzini:
@tomtor key added

ppigazzini:
> @ppigazzini https://dfts-0.pigazzini.it/nns gives an error but works fine for me locally?
> 
> Do you have a hint from the logs?

```
Aug 01 13:45:27 vps1572612706 pserve[14738]:     _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/mako/runtime.py", line 947, in _exec_template
Aug 01 13:45:27 vps1572612706 pserve[14738]:     callable_(context, *args, **kwargs)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "base_mak", line 76, in render_body
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "nns_mak", line 43, in render_body
Aug 01 13:45:27 vps1572612706 pserve[14738]: pyramid_mako.MakoRenderingException:
Aug 01 13:45:27 vps1572612706 pserve[14738]: Traceback (most recent call last):
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/pyramid_mako/__init__.py", line 148, in __call__
Aug 01 13:45:27 vps1572612706 pserve[14738]:     result = template.render_unicode(**system)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/mako/template.py", line 482, in render_unicode
Aug 01 13:45:27 vps1572612706 pserve[14738]:     self, self.callable_, args, data, as_unicode=True
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/mako/runtime.py", line 883, in _render
Aug 01 13:45:27 vps1572612706 pserve[14738]:     **_kwargs_for_callable(callable_, data)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/mako/runtime.py", line 920, in _render_context
Aug 01 13:45:27 vps1572612706 pserve[14738]:     _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/env/lib/python3.6/site-packages/mako/runtime.py", line 947, in _exec_template
Aug 01 13:45:27 vps1572612706 pserve[14738]:     callable_(context, *args, **kwargs)
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/fishtest/templates/base.mak", line 106, in render_body
Aug 01 13:45:27 vps1572612706 pserve[14738]:     <main>${self.body()}</main>
Aug 01 13:45:27 vps1572612706 pserve[14738]:   File "/home/usr00/fishtest/fishtest/fishtest/templates/nns.mak", line 19, in render_body
Aug 01 13:45:27 vps1572612706 pserve[14738]:     <td>${nn['downloads']}</td>
Aug 01 13:45:27 vps1572612706 pserve[14738]: KeyError: 'downloads'
```

noobpwnftw:
@tomtor key added.

ppigazzini:
@noobpwnftw @tomtor @vondele password authentication disabled.

tomtor:
@ppigazzini I added handling the missing key, but old nets will not have download statistics. Best to drop 'nns':

```
mongo
> use fishtest_new
> db.nns.drop()
```

tomtor:
@ppigazzini 
> If the field does not exist, $inc creates the field and sets the field to the specified value.

So downloading should add the download counter if it does not exist. So no need to drop.

ppigazzini:
@tomtor DEV updated, db cleaned. I will return online in the night, feel free to work on DEV.

tomtor:
@ppigazzini All ok now: https://dfts-0.pigazzini.it/nns

Have a nice day/evening

vondele:
some minor nits. 

* The stats page should presumably mention that these networks are available for download under CC0 license. 
* Some title for the stats page ? `Neural Net download and statistics` ? 
* It would be nice for the upload page to link to the stat spage, something like `Your uploaded network will be available for public download and listed on the [NN stats page](https://dfts-0.pigazzini.it/nns)`

ppigazzini:
DEV updated

tomtor:
@ppigazzini Did install and start NGINX on the file server. Will look at POST uploading now.

tomtor:
@ppigazzini I gave up on quickly installing the NGINX opload module (would require rebuilding NGINX from source).

So I did a minimal Python setup:

* Copied https://gist.github.com/mdonkers/63e115cc0c79b4f6b8b3a6b797e485c7 to `server.py` dir in `/root/postserver`
* Made a virtual env `env` but this is not needed, only standard Python3
* The `server.py` listens on port 8080 and stores POSTS in `/root/uploads`
* Test with `curl --data-binary @date.txt localhost:8080/file.txt` after your started `python server.py`

To proceed the NGINX configuration would have to map POSTS on eg `/secret_url/filename` to `localhost:8080/filename`.
In addition it would have to serve files requested as `nn/filename` from `/root/uploads/filename`

If HTTPS is configured that should be safe, `secret_url` should be configured on the fishtest server in `~/fishtest.upload`

tomtor:
For reference the current server.py (could be stripped further if needed):

```
#!/usr/bin/env python3
"""
Very simple HTTP server in python for logging requests
Usage::
    ./server.py [<port>]
"""

# From https://gist.github.com/mdonkers/63e115cc0c79b4f6b8b3a6b797e485c7

from http.server import BaseHTTPRequestHandler, HTTPServer
import logging

class S(BaseHTTPRequestHandler):
    def _set_response(self):
        self.send_response(200)
        self.send_header('Content-type', 'text/html')
        self.end_headers()

    def do_GET(self):
        logging.info("GET request,\nPath: %s\nHeaders:\n%s\n", str(self.path), str(self.headers))
        self._set_response()
        self.wfile.write("GET request for {}".format(self.path).encode('utf-8'))

    def do_POST(self):
        content_length = int(self.headers['Content-Length']) # <--- Gets the size of data
        post_data = self.rfile.read(content_length) # <--- Gets the data itself
        logging.info("POST request,\nPath: %s\nHeaders:\n%s\n\nBody length:\n%s\n",
                #str(self.path), str(self.headers), post_data.decode('utf-8'))
                str(self.path), str(self.headers), len(post_data))

        self._set_response()
        self.wfile.write("POST request for {}".format(self.path).encode('utf-8'))

        with open("/root/uploads/" + self.path, "wb") as f:
            f.write(post_data)

def run(server_class=HTTPServer, handler_class=S, port=8080):
    logging.basicConfig(level=logging.INFO)
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    logging.info('Starting httpd...\n')
    try:
        httpd.serve_forever()
    except KeyboardInterrupt:
        pass
    httpd.server_close()
    logging.info('Stopping httpd...\n')

if __name__ == '__main__':
    from sys import argv

    if len(argv) == 2:
        run(port=int(argv[1]))
    else:
        run()
```

tomtor:
@ppigazzini Regarding starting `server.py`, I assume you would like to do that with `systemd`?

NGINX/systemd is your area, but please let me know if you would prefer that I have a look at implementing this.

ppigazzini:
@tomtor done
```
systemctl enable nginx
systemctl restart nginx
```

tomtor:
@ppigazzini @vondele @noobpwnftw 

Next step is to arrange the DNS/CDN configuration.

It's still not clear for me what the (minimal) required steps are.

We will redirect downloads from `https://tests.stockfishchess.org/api/nn/...` to the CDN by answering with a `HTTP 301` and a new URL which points to the CDN download URL.

Is creating a DNS entry for eg `fishnets.stockfishchess.org` needed? Should this be a CNAME which points to the CDN?

We also need a DNS entry for the upload server IP as supplied by @noobpwnftw in order to configure HTTPS and a Let's Encrypt certificate.

@daylen could create these DNS entries?

For my own convenience I created http://fishnets.v7f.eu which points to the upload server

vondele:
@tomtor , minor nit, should we use `data` instead of `fishnets` just to keep the name more generic? Maybe there will be future use that is different from nn.

ppigazzini:
@tomtor please test the server
```
less /etc/systemd/system/simple_http.service


[Unit]
Description=simple HTTP server in python
After=network.target

[Service]
Type=simple
ExecStart=/root/post-server/env/bin/python3 /root/post-server/server.py
Restart=on-failure
RestartSec=3
User=root
WorkingDirectory=/root/post-server

[Install]
WantedBy=multi-user.target
```

ppigazzini:
@tomtor I updated the env with `env/bin/python3 -m pip install --upgrade pip setuptools wheel`
Dose the `simple_http` must run with root privileges?

noobpwnftw:
You'll need to decide what domain/subdomain you wish to use, it does not need to be .stockfishchess.org unless you have access to DNS of the domain. Once you decided on the domain, then it is pretty straightforward, you can get a free SSL cert by TXT validation. The cert has to be renewed every once a while on the CDN due to expiry, this is going to be manual.

tomtor:
@ppigazzini I am currently struggling to have NGINX serve files from eg `/usr/share/nginx/html/nn` or any other directory.

I get `forbidden` . I restored the original `nginx.conf` and modified `server.py` to upload to `/var/nn` for now.

tomtor:
> Does the `simple_http` run with root privileges?

Yes, but it only needs to write in the upload directory so it's fine for me to change the UID

tomtor:
> please test the server

@ppigazzini It runs OK! Thanks!!!

tomtor:
>  I am currently struggling to have NGINX serve files from eg `/usr/share/nginx/html/nn` or any other directory.

@ppigazzini Solved:

`chcon -Rt httpd_sys_content_t /var/nn`

tomtor:
For reference the current version:

```
#!/usr/bin/env python3

# From https://gist.github.com/mdonkers/63e115cc0c79b4f6b8b3a6b797e485c7

from http.server import BaseHTTPRequestHandler, HTTPServer
import logging

class S(BaseHTTPRequestHandler):
    def _set_response(self):
        self.send_response(200)
        self.send_header('Content-type', 'text/html')
        self.end_headers()

    def do_GET(self):
        logging.info("GET request,\nPath: %s\nHeaders:\n%s\n", str(self.path), str(self.headers))
        self._set_response()
        self.wfile.write("GET request for {}".format(self.path).encode('utf-8'))

    def do_POST(self):
        content_length = int(self.headers['Content-Length']) # <--- Gets the size of data
        post_data = self.rfile.read(content_length) # <--- Gets the data itself
        logging.info("POST request,\nPath: %s\nHeaders:\n%s\n\nBody length:\n%s\n",
                str(self.path), str(self.headers), len(post_data))
        self._set_response()
        self.wfile.write("POST request for {}".format(self.path).encode('utf-8'))

        try:
          with open("/var/nn/" + self.path[self.path.rfind('/')+1:], "wb") as f:
            f.write(post_data)
        except Exception as e:
          self.wfile.write(str(e))

def run(server_class=HTTPServer, handler_class=S, port=8080):
    logging.basicConfig(level=logging.INFO)
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    logging.info('Starting httpd...\n')
    try:
        httpd.serve_forever()
    except KeyboardInterrupt:
        pass
    httpd.server_close()
    logging.info('Stopping httpd...\n')

if __name__ == '__main__':
    from sys import argv

    if len(argv) == 2:
        run(port=int(argv[1]))
    else:
        run()
```
```

tomtor:
I have enabled https for the upload server with the temporary domain name: https://fishnets.v7f.eu

Nets are served from `/nn/`

I will reconfigure it as soon as the DNS name is choosen and configured by @vondele and @daylen

@ppigazzini I will contact you by private mail with the contents for `~/fishtest.upload`

vondele:
I would be happy with data.stockfishchess.org if that's possible. 

It would be good if @daylen could configure it, or just indicates that somebody else should do it.

tomtor:
> I would be happy with data.stockfishchess.org if that's possible.

Sure, that's fine!

ppigazzini:
@tomtor DEV updated, `~/fishtest.upload` written :)

daylen:
Happy to configure data.stockfishchess.org. Just to confirm, we'd like that to point to https://fishnets.v7f.eu/?

noobpwnftw:
@daylen Please point "data.stockfishchess.org" to CNAME: "data.stockfishchess.org.cnamegslb.com".

noobpwnftw:
@ppigazzini I've pointed the CDN to the file server, so far only HTTP, please create a cert for "data.stockfishchess.org" on the file server so I can deploy it. Force redirection not needed.

daylen:
@noobpwnftw Done. Let me know if you'd like me to enable CloudFlare's proxy for that subdomain. If we are just serving static resources it could potentially speed up download times/reduce origin server load.

noobpwnftw:
@daylen Oh, is it already on Cloudflare? If so, please just set it to the same IP address as https://fishnets.v7f.eu/ and enable proxy. No need to use my CDN when you already got one.

daylen:
@noobpwnftw Done! A record created and Cloudflare proxy enabled.

noobpwnftw:
@daylen And it now works without problem. Thanks!

ppigazzini:
@tomtor DEV updated

tomtor:
@ppigazzini On DEV I now get a 404 error on upload.

Probably something wrong with `~/fishtest.upload` which still should point to `fishnets.v7f.eu` (the upload server) and not to `data.stockfishchess.org` (the CDN download site)

tomtor:
@daylen @vondele @noobpwnftw @ppigazzini  The upload server has now as only DNS entry and certificate `fishnets.v7f.eu` and (`173-248-224-124.static.imsbiz.com`)

It could be reconfigured to another name, eg. `uploads.stockfishchess.org` (@daylen would have to add a DNS entry, and we would have to reconfigure NGINX and certbot on the upload server) but that is not urgent.

I have enable crontab `certbot` renewal on the server.

noobpwnftw:
You can just configure the IP address on the fishtest server(since only need access from there and nobody else), can even add IP whitelist on the file server to accept only from fishtest server. Not really need https anyways.

ppigazzini:
> @ppigazzini On DEV I now get a 404 error on upload.
> 
> Probably something wrong with `~/fishtest.upload` which still should point to `fishnets.v7f.eu` (the upload server) and not to `data.stockfishchess.org` (the CDN download site)

Sorry, a pasto, fixed now :(

tomtor:
> You can just configure the IP address on the fishtest server(since only need access from there and nobody else), can even add IP whitelist on the file server to accept only from fishtest server. Not really need https anyways.
> 
> I thought users upload to fishtest server, then it writes to file server, which means the actual communication is just between fishtest and file, and this endpoint can be kept private for good.

Yes, but that (white-listing fishtest server IP) would make development/testing for me and @ppigazzini difficult.
A longer white-list introduces maintanance issues.

I am happy with the current setup, so I propose squashing this PR and giving it a try on prod.

noobpwnftw:
Well, you can just use a "super secret" URL path for the upload location, which only exists in the configuration files on the servers. :)

ppigazzini:
> I have enabled crontab `certbot` renewal on the server.

The official certbot instruction suggests a different way to write the crontab

https://certbot.eff.org/lets-encrypt/centosrhel8-other

ppigazzini:
> I am happy with the current setup, so I propose squashing this PR and giving it a try on prod.

@tomtor if you are ok I will load the PR on PROD without a merge.

tomtor:
> The official certbot instruction suggests a different way to write the crontab

Good catch, my entry would have failed with a running NGINX.

Should be OK now

tomtor:
> if you are ok I will load the PR on PROD without a merge.

Sure!

ppigazzini:
> > if you are ok I will load the PR on PROD without a merge.
> 
> Sure!

Done :)

tomtor:
> Done :)

@ppigazzini 

Forgot nginx change:

`413 Request Entity Too Large`

ppigazzini:
> > Done :)
> 
> @ppigazzini
> 
> Forgot nginx change:
> 
> `413 Request Entity Too Large`
> 
> EDIT: changed to 50M

Changed to 100M according to the new setup script.

vondele:
Great progress... I'll only be able to work on this in the evenings this week..

Minor nit, for the uploads to PROD, we should ask the authors of the nets to upload them, in particular, @gekkehenker for `nn-c157e0a5755b.nnue` Just so the user name matches the creator. Maybe @gekkehenker is fine with this, or he can create a fishtest login and we try to correct this.

tomtor:
> Maybe @gekkehenker is fine with this, or he can create a fishtest login and we try to correct this.

Yes, I dropped the uploads. Note to myself that I have to run `create_indexes.py nns` after the first official network is uploaded

gekkehenker:
Thank you for notifying me.

I see I had to agree to the CC0 license. File is uploaded.

vondele:
Thanks!

@gekkehenker testing doesn't work yet... needs the pending worker changes, I'll stop the test.

gekkehenker:
> 
> 
> Thanks!
> 
> @gekkehenker testing doesn't work yet... needs the pending worker changes, I'll stop the test.
> 
> Edit: two reasons for asking you.. 1) you are listed as the author 2) license makes clear the network can be used by others.

I let my enthousiasm get the better of me...

> two reasons for asking you.. 1) you are listed as the author 2) license makes clear the network can be used by others.

Yes, this prevents possible misunderstandings / trouble.

vondele:
It seems the functionality is working (many thanks, BTW!). I have updated the worker scripts to use 
`NETWORKS_REPO_URL = 'https://tests.stockfishchess.org/api/nn/'` and have a similar download functionality in the Makefile (https://github.com/official-stockfish/Stockfish/pull/2896)

What's the plan to proceed from here? 

Do we need some changes to the upload server, further testing, or do we merge like this ?

In particular, do we do the 'stress test' (i.e. NNUE test with @noobpwnftw worker fleet on the dev server or on the prod server)?

tomtor:
All network downloads are off loaded so I see no reason for increased load on the prod server with NNUE tests, so testing on prod is fine.

vondele:
Do you think we could run in `HTTP_TIMEOUT = 15.0` (games.py) on the workers if the download takes long? For me a download takes about 10s now.... or does that just is the time till the initial server response, not the completion of the request?

dorzechowski:
I believe this timeout expires if server is not reponding to a request for a defined number of seconds, nothing to do with how long it takes to complete.

tomtor:
> I believe this timeout expires if server is not reponding to a request for a defined number of seconds, nothing to do with how long it takes to complete.

Exactly.

vondele:
OK, so I'm assuming this PR is set for merging. I'll update the todo list, this will replace the networks repo.

ppigazzini:

daylen:
Catching up here. @noobpwnftw just want to double check that your server can withstand the bandwidth onslaught of tons of people downloading 20 MB network files? According to @glinscott with lczero, CloudFlare still let ~20% of requests through.

noobpwnftw:
@daylen Yes, it will handle such bursts.

ppigazzini:
@tomtor I changed the systemd unit name from "simple_http" to "post-server", so use `systemctl start post-server`

ppigazzini:
> @ppigazzini On DEV I now get a 404 error on upload.
> 
> Probably something wrong with `~/fishtest.upload` which still should point to `fishnets.v7f.eu` (the upload server) and not to `data.stockfishchess.org` (the CDN download site)

@tomtor I'm wondering if we need to update `~/fishtest.upload` on PROD

tomtor:
> I'm wondering if we need to update `~/fishtest.upload` on PROD

@ppigazzini Why/how would you like to update?

SFisGOD:
@tomtor Thank you for your work in improving fishtest. I wonder if it's possible to add an option to remove non-default nets?

vondele:
remove as in 'do not show' or 'delete data' ?

We don't really want to delete data if we want to keep tests reproducible in the future.

SFisGOD:
@vondele 'Do not show' in NN stats maybe

tomtor:
I wonder how to implement this feature request in a user/GUI friendly way.
- An extra menu entry on the left?
- Show only green nets when not logged in? That would not help the usecase of @SFisGOD because he is probably logged in.
- Hide non-green nets when they are not among the first 20/50/100 entries?
- A toggle on the page (Hide non-default nets).

I am neutral if this is worth the trouble of implementing.
It is not clear to me why the display of the non-default nets are counter productive for the expert fishtest user, because most users visiting this page are experts?

tomtor:
@SFisGOD I will add the toggle on the page...

vondele:
yes, I think a toggle is best.

PR #737, opened 2020-07-29
Improve the subprocess usage

ppigazzini:
- Open the standard streams in text mode and retire the encode function
- Redirect the stdout to devnull
- Close the stderr
- Keep the python 2.x compatibility


tomtor:
@ppigazzini nice cleanups!

vondele:
thanks!

PR #739, opened 2020-08-03
Enable proxy for net download api

vondele:


vondele:
@noobpwnftw does this look good?

noobpwnftw:
Yes, this is perfect.

ppigazzini:
Thank you @vondele :)

Issue #741, opened 2020-08-04
Changing test type with test options


vondele:
When rescheduling a test with non-default options, these options are copied. However, when changing the test type (e.g. STC -> LTC), these options are reset to default. This is something which will bite us after the NNUE merge, since `Use NNUE=true` or `Use NNUE=false` will be more commonly changed. Maybe we can make the buttons that change the test type a little smarter (i.e. only change those options that match need to change)

tomtor:
I will add that to my NN fixups PR

tomtor:
See also #704 (duplicate)

tomtor:
Fixed with #742

PR #742, opened 2020-08-04
Minor NN Updates and fix test buttons

### LINKED ISSUE #704: Lost options when changing TC in the test creation page
Steps to reproduce:

1) open the following passed STC test, which correctly includes "Contempt=0" in the options for base and tested patch: https://tests.stockfishchess.org/tests/view/5ee7f27daae8aec816ab75db

2) click on "reschedule" : the options correctly include "Contempt=0" in the test creation page

3) click on "Long (LTC)": the page changes correctly the hash table settings, but clear all the other options. If the user doesn't notice and launch the test, then it is a surpise that the STC and the LTC tests do not use the same settings... https://tests.stockfishchess.org/tests/view/5ee84d41563bc7aa755ffb6d

Expected behavior: changing the time control should change the hash table size for the test, but leave all other options the same.

tomtor:
Minor updates to NN code:

- Align NN downloads right
- Monospace font for NN
- Smarter new test buttons

Should close #704 and #741

### DIFF
--- a/fishtest/fishtest/templates/nns.mak
+++ b/fishtest/fishtest/templates/nns.mak
@@ -12,16 +12,16 @@ These networks are available for download under
    <th>Time</th>
    <th>Network</th>
    <th>Username</th>
-   <th>Downloads</th>
+   <th style="text-align:right">Downloads</th>
   </tr>
  </thead>
  <tbody>
  %for nn in nns:
   <tr>
    <td>${nn['time'].strftime("%y-%m-%d %H:%M:%S")}</td>
-   <td><a href="api/nn/${nn['name']}">${nn['name']}</a></td>
+   <td><a href="api/nn/${nn['name']}" style="font-family:monospace">${nn['name']}</a></td>
    <td>${nn['user']}</td>
-   <td>${nn.get('downloads', 0)}</td>
+   <td style="text-align:right">${nn.get('downloads', 0)}</td>
   </tr>
  %endfor
  </tbody>
--- a/fishtest/fishtest/templates/tests_run.mak
+++ b/fishtest/fishtest/templates/tests_run.mak
@@ -97,19 +97,19 @@
       <label class="field-label leftmost">Test type</label>
       <div class="btn-group choose-test-type">
         <div class="btn" id="fast_test"
-             data-options='{"tc": "10+0.1", "threads": 1, "options": "Hash=16 Contempt=24", "bounds": "standard STC"}'>
+             data-options='{"tc": "10+0.1", "threads": 1, "options": "Hash=16 Use NNUE=true", "bounds": "standard STC"}'>
           short (STC)
         </div>
         <div class="btn" id="slow_test"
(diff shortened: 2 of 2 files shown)


vondele:
if easy... can we use a fixed font (monospaced) for the name of the net (or even table) ? I.e. so that all names are equally long.

vondele:
BTW, after the merge, I plan to replace 'contempt=24' by 'Use NNUE=true'. If it is easier to remove 'contempt=24' now, this can be done as well.

ppigazzini:
@tomtor DEV updated, IMO there is a bug:
 - `Use NNUE=true` is always added, also for normal test :(
Anyway:
 - `Hash` field is changed according the TC
 - other custom option are maintained

vondele:
@ppigazzini this patch (including the options) would only be OK after a NNUE merge, where I would assume the normal testing would become 'Use NNUE=true' and only eval tests would use 'Use NNUE=false' (while regression tests would be mixed options).

ppigazzini:
Ok, I added the "wip" label.

ppigazzini:
@vondele I scheduled a test with:
- test `Hash=16 Use NNUE=true Contempt=666`
- base `Hash=16 Use NNUE=false Contempt=999`

Rescheduling and changing the TC I got:
 - `Hash=16 Use NNUE=true Use NNUE=false Contempt=999`

vondele:
that would be wrong... options should be separate for base and test.

tomtor:
> options should be separate for base and test.

They are...

I remove the default options which leaves the user options and prepend the new defaults options.

Only when the user entered "Use NNUE=false" then "Use NNUE" is twice in the resulting options, the user specified is last (valid).

I can remove "Use NNUE=[^ ]+" but this will always reset Use NNUE to true.

vondele:
OK, thanks, there was a small edit to the comment that clarified they are separate. 

I don't know exactly how cutechess would deal with duplicated options, but normally it sends the options in alphabetical order, not the user specified order (see also https://github.com/cutechess/cutechess/issues/557). It is likely it would send `Use NNUE=true` last.

tomtor:
I could also leave "Use NNUE=*" in the original and remove it from the prepended defaults, so only Hash is replaced.

vondele:
Yes, only replacing Hash is right, I think.

ppigazzini:
DEV updated, I confirm that change only the "Hash" value.
With old test is not added "Use NNUE="

I wait the squash to merge.

vondele:
Also, merging only after NNUE gets merged in master please.

ppigazzini:
I'm wondering if `Hash=16 Use NNUE=true` as default for a new SPSA test is the right option.

vondele:
This will depend if the SPSA is tuning search or eval. So, I think it is fine like that.

vondele:
@ppigazzini this can be merged when you think it is ready.

ppigazzini:

Issue #743, opened 2020-08-06
show the make ARCH=.... the worker uses


vondele:
It might be interesting to display on fishtest the ARCH the worker uses to build the code (probably enough to just show the one of test). Needs the worker to transmit that info to the server, and some layout changes to the test page.

PR #745, opened 2020-08-07
Send ARCH to server

tomtor:
Should implement #743

### DIFF
--- a/fishtest/fishtest/api.py
+++ b/fishtest/fishtest/api.py
@@ -174,6 +174,7 @@ class ApiView(object):
       task_id=self.task_id(),
       stats=self.request.json_body['stats'],
       nps=self.request.json_body.get('nps', 0),
+      ARCH=self.request.json_body.get('ARCH', '?'),
       spsa=self.request.json_body.get('spsa', {}),
       username=self.get_username()
     )
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -575,12 +575,12 @@ class RunDb:
         self.active_runs[id] = {'time': time.time(), 'lock': active_lock}
       return active_lock
 
-  def update_task(self, run_id, task_id, stats, nps, spsa, username):
+  def update_task(self, run_id, task_id, stats, nps, ARCH, spsa, username):
     lock = self.active_run_lock(str(run_id))
     with lock:
-      return self.sync_update_task(run_id, task_id, stats, nps, spsa, username)
+      return self.sync_update_task(run_id, task_id, stats, nps, ARCH, spsa, username)
 
-  def sync_update_task(self, run_id, task_id, stats, nps, spsa, username):
+  def sync_update_task(self, run_id, task_id, stats, nps, ARCH, spsa, username):
     run = self.get_run(run_id)
     if task_id >= len(run['tasks']):
       return {'task_alive': False}
@@ -613,6 +613,7 @@ class RunDb:
 
     task['stats'] = stats
     task['nps'] = nps
+    task['ARCH'] = ARCH
     if num_games >= task['num_games']:
       # This task is now finished
       if 'cores' in run:
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -276,6 +276,7 @@
    <td>
    %if 'worker_info' in task:
      ${task['worker_info']['uname']}
+     ARCH=${task.get('ARCH', '?')}
    %else:
      Unknown worker
    %endif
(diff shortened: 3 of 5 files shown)


tomtor:
@vondele @ppigazzini Not sure how to get the ARCH from a stockfish which has already been compiled...

tomtor:
> Not sure how to get the ARCH from a stockfish which has already been compiled...

Never mind, I figured it out....

tomtor:
@ppigazzini @vondele I have a problem.

The compile settings are written on 'stdout', the signature message on 'stderr'.

Reading both FDs does not work in the current PR, at least not on Linux, not sure why.

I expect merging both FDs might (also) give trouble on Windows systems.

vondele:
@tomtor somewhat surprising, and maybe ppigazzini has another solution. However, you could write a small function like `required_net` but sending `compiler` instead of `uci` that does it independently from verify bench. It is maybe anyway a good idea to separate things.

tomtor:
> somewhat surprising, and maybe ppigazzini has another solution. However, you could write a small function like `required_net(engine)` but sending `compiler` instead of `uci` that does it independently from verify bench. It is maybe anyway a good idea to separate things.

My code was broken in other ways, I am rewriting with separate processes now.

vondele:
quick question, will the code work with older version that reply:
```
Stockfish 7 64 POPCNT by T. Romstad, M. Costalba, J. Kiiski, G. Linscott
Unknown command: compiler 
``` 
Just returning `unknown` is fine.

tomtor:
It works:

```
Linux 4.15.0-112-generic ARCH=64bit AVX2 SSE42 SSE41 SSSE3 SSE3 POPCNT
```

is shown as info in the test view for my 14 thread Ryzen 1700x

vondele:
OK, nice!

tomtor:
> quick question, will the code work with older version that reply:

That should return `None` now. I can replace it with `?`  (server default) or `Unknown`

vondele:
`ARCH=?` would be nice

ppigazzini:
@tomtor DEV and workers updated.

vondele:
OK, my worker displays properly: `vdv-12cores-1 | Linux 5.4.0-42-generic      ARCH=64bit AVX2 SSE42 SSE41 SSSE3 SSE3 POPCNT`

ppigazzini:
All my workers (installed from zero) are stuck with this message, no time now to debug on the fly:
```
Obtaining EvalFile of stockfish_857e045ced9e20f202e15d825e47b3ab8241dcef ...
Obtaining EvalFile of stockfish_857e045ced9e20f202e15d825e47b3ab8241dcef ...

Exception running games:
  File "worker.py", line 154, in worker
  File "/home/usr00/_git/__test_folder00/fishtest/worker/games.py", line 681, in run_games
    raise Exception('Failed to validate the network: %s ' % (net))
Exception: Failed to validate the network: nn-9931db908a9b.nnue
Task exited
```

vondele:
I copied it over, because the network doesn't exist on the dev server (was my guess?)

ppigazzini:
@vondele but I started also a `Use NNUE=false` test
https://dfts-0.pigazzini.it/tests/view/5f2d78a59ff42b2cd04b75fd

vondele:
The network is downloaded irrespective of the options, if the binary has an EvalFile option. This will be needed in the future anyway to bench / PGO (even if the code for it has to change).

ppigazzini:
@vondele ok, thanks. In the week end I will investigate how to keep DEV up to date.

ppigazzini:
@vondele my workers are running fine after a sync with the PROD db.

tomtor:
@vondele Not related to this PR but I get on aarch64:

```
tom@pine64:~/src/fishtest/worker$ uname -a
Linux pine64 4.19.63-sunxi64 #5.92 SMP Fri Aug 2 00:18:27 CEST 2019 aarch64 aarch64 aarch64 GNU/Linux
```

```
Exception running games:
  File "/tom/src/fishtest/worker/games.py", line 239, in setup_engine
    subprocess.check_call(MAKE_CMD + ARCH + ' -j %s' % (concurrency) + ' profile-build', shell=True)
  File "/usr/lib/python3.6/subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'make COMP=gcc ARCH=x86-64 -j 3 profile-build' returned non-zero exit status 2.
```

Should I create a fishtest or stockfish issue? I thought aarch64 was supported?

vondele:
it is supported for stockfish but not for fishtest, we don't have cutechess available (AFAIK), so the fishtest worker only needs to care about x86. I might be wrong however...

ppigazzini:

PR #747, opened 2020-08-08
Remove x86-64-sse42 target (corresponds to Stockfish pull request)

sf-x:
I'm not sure what exactly I should do here, but in the Stockfish pull request (https://github.com/official-stockfish/Stockfish/pull/2922 ) vondele said that fishtest worker needs to be changed also.
I don't know how to test this change but this seems simple enough?
I understand that versions need to be incremented. This is NOT done here.

Corresponds to the above-mentioned change for the Stockfish Makefile, however, should work now as is (as SSE4.2 is not used by Stockfish now, other than to display that it was compiled with SSE4.2 support).

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -193,18 +193,12 @@ def find_arch_string():
         res='x86-64-bmi2'
      elif '-mavx2' in props['flags'] and 'x86-64-avx2' in targets:
         res='x86-64-avx2'
-     elif '-msse4.2' in props['flags'] and 'x86-64-sse42' in targets:
-        res='x86-64-sse42'
      elif '-mpopcnt' in props['flags'] and '-msse4.1' in props['flags'] and 'x86-64-modern' in targets:
         res='x86-64-modern'
-     elif '-msse4.1' in props['flags'] and 'x86-64-sse41' in targets:
-        res='x86-64-sse41'
      elif '-mssse3' in props['flags'] and 'x86-64-ssse3' in targets:
         res='x86-64-ssse3'
      elif '-mpopcnt' in props['flags'] and '-msse3' in props['flags'] and 'x86-64-sse3-popcnt' in targets:
         res='x86-64-sse3-popcnt'
-     elif '-msse3' in props['flags'] and 'x86-64-sse3' in targets:
-        res='x86-64-sse3'
      else:
         res='x86-64'
   else:


ppigazzini:
@sf-x Tested fine on DEV and my workers.
@vondele if I'm not wrong this PR has a low priority and can be queued to be merged with other worker PRs.

vondele:
yes, lower prio, needs to be synced with upstream changes as well. Let's wait for that PR to merge.

sf-x:
Actually this doesn't need to wait for that stockfish PR, as this simply removes a useless target.

vondele:
I'll come to the stockfish PR, there are some aspects still. 

Worker changes are best done in bunches, as it needs syncing with noob to update his fleet, so few worker changes is desirable. I still need to implement stuff to enable PGO builds in the worker before we should do a worker update.

vondele:
also note that https://github.com/official-stockfish/Stockfish/pull/2844 is relevant, as we would need to check more flags for avx512 in the worker.

ppigazzini:
@sf-x please let me know if you want your real name in fishtest [AUTHORS](https://github.com/glinscott/fishtest/blob/master/AUTHORS) :)

A worker update requires:
 - update by hand for noob fleet (an auto update with so many downloads could trigger a IP fleet ban by github), then
 - auto update for other workers

So usually I prefer to wait that all the open worker PRs are approved and merge them in one step.

sf-x:
> @sf-x please let me know if you want your real name in fishtest [AUTHORS](https://github.com/glinscott/fishtest/blob/master/AUTHORS) :)

No, thanks.

> So usually I prefer to wait that all the open worker PRs are approved and merge them in one step.

OK, just letting you know this need not wait for Stockfish.

vondele:
@sf-x note that the suggested changes at https://github.com/official-stockfish/Stockfish/pull/2922 will remove more targets.

vondele:
@sf-x please update this PR with the targets available in the current form of https://github.com/official-stockfish/Stockfish/pull/2922 and check add the checks for the required -m... flags

@ppigazzini once this is updated (even before the corresponding stockfish PR is merged), we can update the worker with the 3 pending PRs. This should enable me to move forward with the stockfish CI / PGO patch, which I'm eager to test and commit (could be another +- 5 Elo ;-) ), and more importantly will lead to meaningful bench changes upon NNUE-only commits.

ppigazzini:
> @ppigazzini once this is updated (even before the corresponding stockfish PR is merged), we can update the worker with the 3 pending PRs. This should enable me to move forward with the stockfish CI / PGO patch, which I'm eager to test and commit (could be another +- 5 Elo ;-) ), and more importantly will lead to meaningful bench changes upon NNUE-only commits.

@vondele of course :)

sf-x:
> @sf-x please update this PR with the targets available in the current form of [official-stockfish/Stockfish#2922](https://github.com/official-stockfish/Stockfish/pull/2922) and check add the checks for the required -m... flags

@vondele  I'm not sure what to add, this seems to need a pure removal (three targets gone, one which is renamed in Stockfish wasn't used as it was the same as `x86-64-modern`.

vondele:
OK, seems good to me.

ppigazzini:
Tested fine on DEV, thank you @sf-x :)

PR #749, opened 2020-08-09
Prepare for PGO builds with hybrid bench

vondele:
Download net before building, copy from cache if possible

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -16,6 +16,7 @@ import time
 import traceback
 import platform
 import struct
+import re
 import requests
 import copy
 from base64 import b64decode
@@ -70,6 +71,20 @@ def required_net(engine):
      raise Exception('uci exited with non-zero code %d' % (p.returncode))
 
   return net
+
+def required_net_from_source():
+  """ Parse ucioption.cpp to find default net"""
+  net = None
+  with open('ucioption.cpp','r') as srcfile:
+    for line in srcfile.readlines():
+       if 'EvalFile' in line and 'Option' in line:
+          p = re.compile('nn-[a-z0-9]{12}.nnue')
+          m = p.search(line)
+          if m:
+             net = m.group(0)
+
+  return net
+
 
 def download_net(remote, testing_dir, net):
   url = remote + '/api/nn/' + net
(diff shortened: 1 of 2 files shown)


vondele:
seems to work OK with my worker on https://dfts-0.pigazzini.it/tests/view/5f2fcdc5deeb34d9b11e4d2b

ppigazzini:
@vondele please let me know how test the effect of this PR

vondele:
First, workers should just behave, no crash when they run with this patch, on all tests.

Second, this needs to test on the above test which enables the hybrid bench & PGO. In that case the first step of the build (make net) should reply with
```
Default net: nn-[...].nnue
Already available.
```
What the patch does is trying to make sure it is not the `make net` part of the build that downloads the net, but the python code. The latter is smarter in that it uses the cache of nets in the testing directory, and uses the proxy of the server if the worker is configured to do so.

vondele:
@ppigazzini if your testing looks OK, let's go for it :-)

ppigazzini:
@noobpwnftw I merged several worker PRs and raised the version number, please update the fleet at your convenience :)

ppigazzini:
Thank you @vondele :)

vondele:
OK, I'll merge the PGO patch in master once @noobpwnftw has updated the fleet. After that, next RT.

noobpwnftw:
@ppigazzini @vondele I started rolling updates, there are a few workers that are currently down and unable to update, will send people there. Looks like the fleet will be fully updated by tomorrow.

vondele:
OK, I might start the test before, once a good fraction has updated. The only side effect is that there will be a call to the net download api without using the proxy for every new binary built. If that's no problem, all is fine.

noobpwnftw:
@vondele That is fine as long as failed downloads will be handled. Direct connections to API is not very smooth on my workers.

vondele:
@ppigazzini the worker fleet has been updated, time to enforce the worker level change via the server... this should take some of the strain of the dataserver, and reduce the number of downloads. Current net has been downloaded 54231 times...

ppigazzini:
@vondele done

ppigazzini:
@vondele however a big percentage of the workers is already updated to version 84 so we should check if the caching system is working fine.

vondele:
yes, I agree, it is surprisingly large.

vondele:
the number of downloads of the new default net seems very reasonable now. I think the caching system is working after the enforced worker update.

PR #757, opened 2020-08-13
Handle upload timeouts/retries

### LINKED ISSUE #756: double net uploads
image

seems like the same net can be uploaded twice, or at least is displayed twice

tomtor:
Should fix #756

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -64,9 +64,21 @@ def login(request):
   return {}
 
 
+# Guard against upload timeouts/retries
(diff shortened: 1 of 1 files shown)


vondele:
Thanks maybe `An other upload is in progress, please try again later.`

tomtor:
@ppigazzini I tested this locally and it works OK.

ppigazzini:
I was able to replicate on DEV the problem starting 2 parallel uploads.
This PR fix the problem, PROD updated, thank you @tomtor :)

Issue #758, opened 2020-08-13
feature suggestion for net stats page


vondele:
I realized that it would be possible to implement a feature for the net stats page that is often requested, and that would help us with the illegal uploads. Basically, every uploaded net gets a 'status' field in the database as well which can be 'untested', 'tested', 'recommended' (or 'default'). The stats page shows that status next to the net. 

Benefits: 
* Together with the date, the status would be an indication what the strongest net is ('recommended' of latest date). 
* If 'untested' nets would not be shown to users not logged in, the incentive to upload illegally (and without testing) would be small, since nobody would see the upload. Logged in users would still be able to see the needed info (and act if e.g. downloads are misbehaving). If an illegal upload would be tested, the approver could step in, and furthermore we have the github handle of the illegal uploader (in addition to the email), which would help in case we need to take action.

What I realized is that the implementation should be quite doable. Basically, for every test submitted, we check the resolved repo and download the ucioptions.cpp of that repo, extracting the net as we do in the worker. If that resolved repo is equivalent to official-stockfish/master, it can be marked as 'default', otherwise we can switch to 'tested' from 'untested'

I won't be able to implement this myself easily (since I don't know the server code), but am happy to help.

tomtor:
We could persist as `last_test` attribute the id of each new test in the nn database with the corresponding net. Or would you (also?) like to preserve the first test?

vondele:
interesting, yes, this would make it possible to link to the test. first_test and last_test could be interesting. I think first is maybe more interesting than last, since first will be the actual test of the net against the previous one, but both have a use. The additional columns could just display the date, where a click on the date jumps to the test page.

vondele:
this feature has been merged, thanks @tomtor.

PR #759, opened 2020-08-14
Update NN database with usage

tomtor:
Implements #758

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -360,6 +360,24 @@ def get_sha(branch, repo_url):
     return '', ''
 
 
+def get_net(branch, repo_url):
+  """ Get the net from ucioption.cpp in the repo """
+  api_url = repo_url.replace('https://github.com',
+                             'https://raw.githubusercontent.com')
+  try:
+    api_url = api_url + '/' + branch +  '/src/ucioption.cpp'
+    options = requests.get(api_url).content.decode('utf-8')
+    net = None
+    for line in options.splitlines():
+      if 'EvalFile' in line and 'Option' in line:
+        p = re.compile('nn-[a-z0-9]{12}.nnue')
+        m = p.search(line)
+        if m:
+          net = m.group(0)
+    return net
+  except:
+    raise Exception("Unable to access developer repository: " + api_url)
+
 def parse_spsa_params(raw, spsa):
   params = []
   for line in raw.split('\n'):
@@ -498,6 +516,12 @@ def validate_form(request):
   })
   data['base_same_as_master'] = master_diff.text is ''
 
+  # Test existence of net
+  new_net = get_net(data['new_tag'], data['tests_repo'])
+  if new_net:
+    if not request.rundb.get_nn(new_net):
+      raise Exception("Net not in repository: " + new_net)
+
   # Integer parameters
 
   if stop_rule == 'sprt':
@@ -557,6 +581,31 @@ def del_tasks(run):
     run = copy.deepcopy(run)
     del run['tasks']
   return run
+
+
+def update_nets(request, run):
+  run_id = str(run['_id'])
+  data = run['args']
+  if run['base_same_as_master']:
+    base_net = get_net(data['base_tag'], data['tests_repo'])
+    if base_net:
+      net = request.rundb.get_nn(base_net)
+      if not net:
+        # Should never happen:
+        raise Exception("Net not in repository: " + base_net)
+      if 'is_master' not in net:
+        net['is_master'] = True
+        request.rundb.update_nn(net)
+  new_net = get_net(data['new_tag'], data['tests_repo'])
+  if new_net:
+    net = request.rundb.get_nn(new_net)
+    if not net:
+      return
+    if 'first_test' not in net:
+      net['first_test'] = { 'id': run_id, 'date': datetime.datetime.utcnow() }
+    net['last_test'] = { 'id': run_id, 'date': datetime.datetime.utcnow() }
+    request.rundb.update_nn(net)
+
 
 @view_config(route_name='tests_run', renderer='tests_run.mak', require_csrf=True)
 def tests_run(request):
@@ -685,6 +734,7 @@ def tests_approve(request):
   if request.rundb.approve_run(run_id, username):
     run = request.rundb.get_run(run_id)
     run = del_tasks(run)
+    update_nets(request, run)
     request.actiondb.approve_run(username, run)
     cached_flash(request, 'Approved run')
   else:
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -171,6 +171,9 @@ class RunDb:
                           # 'nn': Binary(zlib.compress(nn))})
     return {}
 
+  def update_nn(self, net):
+    self.nndb.replace_one({'_id': ObjectId(str(net['_id']))}, net)
+
   def get_nn(self, name):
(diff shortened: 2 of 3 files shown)


tomtor:
@vondele WIP, still have to show the data on the NN stats page....

fishtest/fishtest/views.py
```diff
@@ -360,6 +362,24 @@ def get_sha(branch, repo_url):
     return '', ''
 
 
+def get_net(branch, repo_url):
+  """ Get the net from ucioption.cc in the repo """
```
vondele, line 366: ucioption.cc -> ucioption.cpp

fishtest/fishtest/views.py
```diff
@@ -558,6 +578,22 @@ def del_tasks(run):
     del run['tasks']
   return run
 
+
+def update_nets(request, run_id, data):
+  if data['base_tag'] == 'master':
+    global MASTER_NET
+    MASTER_NET = get_net('master', data['tests_repo'])
```
vondele, line 585: Does this take care that this 'master' branch / commit is also part of the official stockfish, i.e. I could push a net to my master branch on a private repo and start a test, but that wouldn't make it official. The test approval page shows if base is in master or not, so much have some logic there.

fishtest/fishtest/views.py
tomtor: fixed

fishtest/fishtest/views.py
tomtor: `validate_form()` checks for equal signature, but also has ` data['base_same_as_master']` so I will use that instead.
Thanks for this feedback!

vondele: btw, I now see you use MASTER_NET as a global, i was rather thinking that any net that ever has been in the master repository would be marked in the database (i.e. db['is_master'] = db['is_master'] or data['base_same_as_master']). In that way we could mark all nets that have been 'default net' in the history of development differently in the list. The current default in master will probably always be the one with the most recent submission date.

tomtor: Ok, changed that...

tomtor:
@ppigazzini Should be ready for some testing

ppigazzini:
@tomtor DEV updated

vondele:
looks good in my testing:

image

the two default nets get green, a tested  but non-default net remains white.

One question, do we want to add the feature that users that are not logged in do not see the 'untested' nets ? That would be the trick to make the illegal uploads less interesting.

ppigazzini:
@vondele I started only two tests (SPRT, NumGames) with high priority, both with "master" vs "master", I'm wondering why are marked several net with the time stamp:
https://dfts-0.pigazzini.it/tests/view/5f36b1d01822588a3156711e
https://dfts-0.pigazzini.it/tests/view/5f36b1e91822588a31567120

tomtor:
> One question, do we want to add the feature that users that are not logged in do not see the 'untested' nets ? That would be the trick to make the illegal uploads less interesting.

It would still be downloadable from the API URL.

I am not sure why these nets are uploaded (what does the uploader gain?)

Any way, it's easy to add if it makes sense.

vondele:
I also submitted two tests (to get the older master and the tested non-master net)

Actually that raises another question. Do we updated the database when the test is submitted or when it is approved. I don't have too strong an opinion.

tomtor:
I don't have to change any code when we update when submitted ;-)

vondele:
> > One question, do we want to add the feature that users that are not logged in do not see the 'untested' nets ? That would be the trick to make the illegal uploads less interesting.
> 
> It would still be downloadable from the API URL.
> 
> I am not sure why these nets are uploaded (what does the uploader gain?)
> 
> Any way, it's easy to add if it makes sense.

yes, they could be downloaded if their sha is known, however, we don't 'advertise' or offer them for download. We're somewhat more on the safer side with content that is not clearly licensed. So, if easy, I would do it.

I'm also not sure what the uploader gains, but I think he enjoys the public visibility of the net, possibly uses it do promote distribution. .. No idea really.

vondele:
> I don't have to change any code when we update when submitted ;-)

good enough a reason. At that point we anyway have the github handle of the person. It would become listed on the page before an approver can have a look, however.

tomtor:
> It would become listed on the page before an approver can have a look, however.

I will change it...

ppigazzini:
DEV updated.

vondele:
and it works :-)

tomtor:
Should update now on approval....

tomtor:
Hmm. The exceptions (no net) don't make sense on approval.

Could move one to the new test code.

vondele:
actually net == None can happen for tests running on older versions. However if net is not none, the net should be in the database. If net is not none, and the net could not be found in the database, it should not be possible to submit the test (I think).

vondele:
One question I would have is how to add some meaningful data to the prod server once we apply this. I can, once merged on master, easily run a few tests on prod with old master so all former default nets get some 'green' label, and a few tests on a branch for the other net so they get the 'tested' status and become visible. Or did you have a different plan?

tomtor:
> One question I would have is how to add some meaningful data to the prod server once we apply this. I can, once merged on prod, easily run a few tests on prod with old stockfish master so all former default nets get some 'green' label, and a few tests on a branch for the other net so they get the 'tested' status and become visible. Or did you have a different plan?

We could also write a small script to update the current NN database on prod, but I prefer approving some real tests (the script is a bit more work)

vondele:
yes, that's what I was thinking as well.

ppigazzini:
DEV updated, clean MongoDB synced with PROD.

tomtor:
@ppigazzini @vondele For me it's OK to merge. Did some testing and updating stats on Approval looks OK.

ppigazzini:
@vondele please run the tests to populate the NN db on PROD :)

vondele:
OK, will do now.

tomtor:
No green labels on prod?

vondele:
yes, to my surprise. Something seems not quite OK with the 'master' detection

vondele:
the commits sha are from master, but of course are not HEAD on master. That's something we need to be able to deal with, if I do multiple commits in one push, a net might become default but never be 'publically' HEAD.

tomtor:
@vondele You would have to help me with the github query. I could manually update NN DB to add the green flag to at least one net.

vondele:
I'll try to help figure out, what do you want to do exactly?

vondele:
(right now no need to manually make things green, let's see if we can fix it, and I can run the tests again).

tomtor:
The reschedules of old tests on dev to mark nets green DID work.

vondele:
Maybe related to how I created the tests on master? I didn't reschedule but used the sha (so it is not a branch name master).

vondele:
actually, that's probably it, I have always used SF11 as base... stupid me.

vondele:
I'll submit new tests, reverted.

tomtor:
@vondele Looks good!

vondele:
yes, now next step, I'll try to submit tests to get the tested nets visible, and not green :-)
this will take a little more time, need to do some searching in history.

vondele:
so, seems to work. As far as I know the database is now complete. 
I think sergio didn't test nn-5a3983417e4f.nnue 
and nn-f6005244adb0.nnue Amplaytro is not tested either yet.

vondele:
seemingly the test based on the wip-nnue-player branch didn't compile/work added a final test for nn-c157e0a5755b.nnue

Thanks for the work, guys!

Issue #727, opened 2020-07-17
Residual calculation


tomtor:
A long running test like:

https://tests.stockfishchess.org/tests/view/5f01a78059f6f0353289439b

has a lot of workers with bad Residual (yellow or red).

@vdbergh Is it possible that the Residual calculation is impacted by the smaller batch size (originally 1000 games now 200) or that tests with more worker slots are more likely to have a larger residual and that some compensation is possible/needed?

ppigazzini:
The chi2 code aggregates the results by worker, so theoretically changing the batch size could change the number of unique workers in the test and change a bit the residuals. I don't know how works the workers allocation algorithm, though.

In that very long running test the worker "tvijlbrief-7cores-3" ran 18 batches for 3600 games, "tvijlbrief-32cores-24" ran 45 batches for 9000 games, so IMO the games batch size should be ineffective for the residuals.

A red worker is purged only when:
https://github.com/glinscott/fishtest/blob/261d0874fb41b19cbc4ba7b7c51094b87b6030c2/fishtest/fishtest/util.py#L141

and the 0.001 of the tests will be auto purged when ran with perfectly identical workers.

Vizvezdenec:
https://tests.stockfishchess.org/tests/view/5f2df92061e3b6af64882012
Update on issue - with NNUE it seems that there are some workers that are "cut off" even if they are not bad ones because of higher draw rates.
I suggest to up residual purging cutoff to smth like 4.
I see a lot of this type of workers in my tests.
https://tests.stockfishchess.org/tests/view/5f2e45e761e3b6af6488204d - look at all this purged workers and they don't look THAT bad for me honestly...

Vizvezdenec:
@vondele there seem to be smth broken badly
https://tests.stockfishchess.org/tests/view/5f2e45e761e3b6af6488204d
this test can't converge because it gets to some +5 / +3.5 LLRs and then workers that don't even look bad get massively purged.

vondele:
@vdbergh this is probably an issue now. The performance for workers will differ much more between workers, because hardware differences can't be (easily) corrected. The residual calculation will always differ significantly between workers having e.g. avx2 and those that do not. So, we could increase the threshold, or see if we can refine the scheme. One option for that would be https://github.com/glinscott/fishtest/pull/745 which will record the 'arch' string of the worker, and one could think about having the residual computed per 'arch' ?

vondele:
[...]

PR #760, opened 2020-08-14
Change default auto-purge to false.

vondele:
currently, with NNUE we expect the different hardware to have slightly different result.
Auto-purge would need to take the ARCH=.. tags into account to work correctly.
As a zero-th order fix, switch default to off.

partially addresses #727

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -76,7 +76,7 @@ class RunDb:
               spsa=None,
               username=None,
               tests_repo=None,
-              auto_purge=True,
+              auto_purge=False,
               throughput=100,
               priority=0):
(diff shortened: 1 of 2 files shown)


vondele:
needs a careful look, I actually don't know the syntax of .mak templates.

ianfab:
For consistency I would suggest to also change the default in the python code at https://github.com/glinscott/fishtest/blob/2d37c397097f0baf1204d67304ced494ead82c6d/fishtest/fishtest/rundb.py#L79 although I think it does not have any effect since the argument is anyway always provided and the default therefore is rather meaningless.

ppigazzini:
DEV updated, it's working fine.

ppigazzini:

PR #761, opened 2020-08-14
Reduce somewhat the minimum NPS

vondele:
with NNUE some machines get excluded (e.g. opteron cores), which we could keep

Non-urgent worker change.


JojoM-sf:
I have been testing this figure (500000) for 3-4 days now. It seems to work fine, residuals are ok.

PR #762, opened 2020-08-14
Detect vnni targets

vondele:
adjust to master makefile with avx512 vnni

Non-urgent worker change


vondele:
updated to match the most recent Makefile.

ppigazzini:
Companion Stockfish PR https://github.com/official-stockfish/Stockfish/pull/3038

PR #765, opened 2020-08-17
Clarify the color coding, and use of nets

vondele:


tomtor:
Thanks @vondele

tomtor:
@vondele Adding `<br>` at the end of lines may increase the readability.

vondele:
yes, right now this is not very nice on a wide screen. Ideally it was about as wide as the table.. I don't know how to do this, I you can improve with whatever trick, feel free.

PR #766, opened 2020-08-18
Fix last active

tomtor:
Last active is not updated for users not active last 30 days

### DIFF
--- a/fishtest/utils/delta_update_users.py
+++ b/fishtest/utils/delta_update_users.py
@@ -50,6 +50,7 @@ def process_run(run, info, deltas=None):
 
     try:
       info[username]['last_updated'] = max(task['last_updated'], info[username]['last_updated'])
+      info[username]['task_last_updated'] = max(task['last_updated'], info[username]['last_updated'])
     except:
       info[username]['last_updated'] = task['last_updated']
 
@@ -67,7 +68,10 @@ def build_users(machines, info):
   for u in info.keys():
     user = info[u]
     try:
-      user['last_updated'] = delta_date(user['last_updated'])
+      if isinstance(user['last_updated'], str):
+        user['last_updated'] = delta_date(user['task_last_updated'])
+      else:
+        user['last_updated'] = delta_date(user['last_updated'])
     except:
       pass
     users.append(user)


ppigazzini:
DEV updated

tomtor:
@ppigazzini I have a copy running in prod to see if https://tests.stockfishchess.org/users "Last Active" updates for non active users (ie nr of inactive days should increase).

The current version of the update script is really a bit of a hack (with `try ... except` instead of proper logic). I also noticed that numbers changed quite a lot after running a full scan on prod.

We could also consider running a full scan eg each hour at 31 minutes past the hour and optional an incremental run at 01/16/46 (or not all, it does not create much value IMO. @vondele What is your opinion on these stats). Then this PR is not needed.

tomtor:
```
bking_US | 1288 days ago
```
just increased to 1289 👍

vondele:
users seem to be noticing...
https://groups.google.com/g/fishcooking/c/irzQeN_yMcc/m/MTVrFpFzBgAJ

Updating the number of non-active days seems OK, I wonder how this affects the CPU hours?

vondele:
CPU hours could actually change for various reasons, after all, the only accurate count is the number of played games, the CPU hours is actually estimated (IIRC).

tomtor:
@vondele @ppigazzini Was a full scan run after: 

https://github.com/glinscott/fishtest/commit/66c8fd676e65f76530770d0a4d813acc536b127b#diff-9441661765713fd204b9fc7b8a946258

That would explain why there was a significant change after recalculation of CPU hours

vondele:
it would definitely explain things (and suggest a reason to a daily/hourly full scan)

JojoM-sf:
It used to be done so, a while ago, if I remember correctly. CPU hours were incremented every half hour.

tomtor:
> It used to be done so, a while ago, if I remember correctly. CPU hours were incremented every half hour.

It is done every 15 minutes, this is not changed the last few years. The point is that it is done incrementally and if the games to CPU-hours calculation is changed then a full rebuild of the stats is needed.

JojoM-sf:
But the 15 minute update is not transparent on the public user page, is it? Figures appear to be incremented on some other schedule or condition, it would seem – this used to be different in the past. Same goes for „last active“ on the user stats page as opposed to the main page: On the latter an active tester will get something like „minutes ago“ or „seconds ago“, while the former (or the „Top Month“ stats, for that matter) might show hours or even days.

tomtor:
> But the 15 minute update is not transparent on the public user page, is it? Figures appear to be incremented on some other schedule or condition, it would seem – this used to be different in the past. Same goes for „last active“ on the user stats page as opposed to the main page: On the latter an active tester will get something like „minutes ago“ or „seconds ago“, while the former (or the „Top Month“ stats, for that matter) might show hours or even days.

Yes, on the user page stats are only incremented after runs are finished. This used to be different, but was changed a year ago to speed up the periodic database scan: https://github.com/glinscott/fishtest/commit/b25530301aebebdbc4b526403408ac37c90fbeb9

PR #769, opened 2020-08-18
Detect x86-32 targets

syzygy1:
This is a follow up to the following PR, which would add support for the targets x86-32-sse41-popcnt and x86-32-sse2.
https://github.com/official-stockfish/Stockfish/pull/3022

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -228,7 +228,12 @@ def find_arch_string():
      else:
         res='x86-64'
   else:
-     res='x86-32'
+     if '-mpopcnt' in props['flags'] and '-msse4.1' in props['flags'] and 'x86-32-sse41-popcnt' in targets:
+        res='x86-32-sse41-popcnt'
+     elif '-msse2' in props['flags'] and 'x86-32-sse2' in targets:
+        res='x86-32-sse2'
+     else:
+        res='x86-32'
 
   print("Available Makefile architecture targets: ", targets)
   print("Available g++/cpu properties : ", props)


vondele:
LGTM

ppigazzini:
LGTM

Is hard to continue to support 32 bit OS in fishtest:
- Windows: msys2 recently dropped the support for the Windows 32 bit installer, the latest installer released (May 2020) is already not able to install new packages
- Linux: I'm expecting to find only old distribution at 32 bit, so we need to build Qt5 static on Ubuntu 14.04 or CentOS 7 to use old libstdc++.so libraries.
  - CentOS 7: isn't able to build Qt5 32 bit because Devtoolset is released only at 64 bit
  - Ubuntu 14.04: is able to build Qt5 5.10 static and cutechess-cli (32/64 bit) only dropping the cutechess Qt5 version check (cutechess v1.2.0 raised the requirement to Qt5 >= 5.11)

vondele:
if it is too hard to continue support for these reasons, I'm fine with not doing it. I haven't seen a single 32 bit worker on fishtest last time I checked.

ppigazzini (approved):

mstembera:
I would not object to no longer supporting any 32 bit machines on fishtest at all. 
- it's more to maintain
- we get very little to no contribution from them
- they have significantly different performance trade offs between classical and nnue than 64 bit which can skew testing results

syzygy1:
@mstembera 32-bit cpus are probably relatively faster at nnue compared to 64-bit cpus of the same age, as long as the proper vector instruction sets are enabled.

The problem seems to be more that old cpus are relatively worse at nnue than newer cpus. That would speak for not supporting older cpus in general, but I don't know if that's what we want.

PR #777, opened 2020-08-27
Finding the default net from sources.

vondele:
the changes needed to embed the net in the binary:
https://github.com/official-stockfish/Stockfish/pull/3070

need a change in the magic for finding the default net from sources on both the server and worker. This should be merged before we can merge the above PR

For backwards compatibility we now need to look for the name of the default net in two places.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -361,13 +361,26 @@ def get_sha(branch, repo_url):
 
 
 def get_net(branch, repo_url):
-  """ Get the net from ucioption.cpp in the repo """
+  """ Get the net from evaluate.h or ucioption.cpp in the repo """
   api_url = repo_url.replace('https://github.com',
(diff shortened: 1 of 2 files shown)


ppigazzini:
@vondele DEV/workers updated with this PR, feel free to test with the Stockfish PR

vondele:
@ppigazzini thanks, this looks like it tests OK. Note that the PR contains both a server and a worker part. They can be committed independently, if that would be needed.

vondele:
so, if we're fine with this patch, I propose we do a worker + server update with it. Once the fleet has updated, we enforce a worker version change, and after that I can do a test of the PR on fishtest, and decide to merge.

ppigazzini:

noobpwnftw:
@ppigazzini Updated.

ppigazzini:
PROD raised to version 85

Issue #764, opened 2020-08-17
Discussion about SPRT bounds.


Vizvezdenec:
After first days of huge elo gains with NNUE everything more or less calmed down so I think it's about time to start discussion of possible change in SPRT bounds.
With NNUE integration our drawrate at fishtest really skyrocketed (no contempt for NNUE also doesn't help ofc). So it makes even harder to catch any improvements, especially at LTC, since for this strong of a play even 1 elo is kinda huge.
On the other hand with this draw rate tests started to converge much faster - prior to NNUE we had 75%~ drawrate at LTC, currently it's 90% which makes LTC converge 2.5 times faster.
So it will be logical to lower SPRT for LTC even further to catch more improvements, smth like [0,25 ; 1,25] (or 0,2 ; 1,2 ) will make LTC SPRT converge in the same ~ number of games as previously but will allow to catch more patches and (as a side result) will also increase confidence in passing test being non-negative.
I'm not saying this should be done now but, imho, this is eventually what we will need to do, current hitrate of green STC getting a green LTC is kinda abysmal in latest days :) 
Well, it's not an immediate measure, but rather a discussion of smth that eventually should probably be done.
What do you think? @vondele 
All calculations were made using fishtest-integrated SPRT calculator.

vondele:
yes, I've been thinking about this. Need to experiment a bit. Right now a patch like https://tests.stockfishchess.org/tests/view/5f3a58f6b38d442594aabe0c has a W/L ratio that is roughly 1.05, which seems like worth having, but it is at 1Elo, and might thus be rejected. So, reducing a little the upper bound might make sense. We might also have to update the simplification bounds, certainly if we adjust the passing Elo threshold to 0.75. 

Maybe @vdbergh has an opinion.

vondele:
Some, some experimenting from my side. First draw rates:

|         | classical | NNUE |
| ---    | --- | --- |
| STC | 61% | 78% |
| LTC  | 75% | 89% |

I wonder if a new net could gain a few Elo, but completely change these stats. A 'contemptuous net' for example. A single merge could, in that case, require a re-adjustment of bounds discussed below.

Based on these draw rates, the current bounds have roughly 100k an expected maximum number of games, while with NNUE it is roughly half of that. If we want to get back to ~100k games max, while making some adjustments so that LTC tests pass earlier, 0 Elo patches have a harder time, and simplifications remain as easy for 0Elo, but harder for -0.5 Elo, we could use
[...]

PR #783, opened 2020-09-03
Rescale TC with new NPS

vondele:
as discussed in https://github.com/glinscott/fishtest/issues/764

use 1.2MNPS instead of 1.6MNPS to account for the bench slowdown after merging NNUE.
This effectively reduces TC by 25%.

No functional change.

### DIFF
--- a/fishtest/fishtest/views.py
+++ b/fishtest/fishtest/views.py
@@ -1137,7 +1137,7 @@ def homepage_results(request):
         machine["last_updated"] = delta_date(machine["last_updated"])
         if machine["nps"] != 0:
             games_per_minute += (
-                (machine["nps"] / 1600000.0)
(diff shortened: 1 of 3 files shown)


vondele:
That's a worker + server change. 

I think this should not influence the hours contributed, but possibly a db update needs to be done.

ppigazzini:
DEV updated

vondele:
thanks, this is missing a server label, but I assume you did that right? 

I have another worker patch coming, if that one looks good as well, I'd like to merge this before the next regression test.

ppigazzini:
@vondele I have updated DEV and workers.
"worker" label : "issue requiring (also) a worker updated"
"server" label : "server side changes, not requiring a worker update"

ppigazzini (approved):

Issue #478, opened 2019-12-24
repeated games from books


vondele:
as is being discussed in https://github.com/official-stockfish/Stockfish/issues/2283 even with a large book, openings can be repeated in fishtest. This is due to cutechess being instructed to pick random openings, in independent cutechess runs with different seeds, which can lead to repetitions.

One way to solve this would be to invoke cutechess with the `order=sequential` option, and give each batch a different starting point with the `start=N` option. E.g. N=batchId * 250/2
```
-openings file=short.pgn format=pgn order=sequential plies=16 start=N
```
However, this doesn't work out of the box:

* The pgn/epd needs to be randomized.. right now openings are in a logical sequence (e.g. noob_3moves), so would lead to highly similar games.
* N can not exceed the number of openings in the book, seems like cutechess doesn't wrap (with the start options), just starts from startpos and ignores the book if N > length.
* Better than batchId * 250/2 would be some formula that uses a quasi-random batch (but obviously without allowing for random repetitions.
* Finally, the start=N options seems not to be working for epd files. At least for the version of cutechess we have. This might be bug fixed upstream.

If we anyway have a random opening from the set, it occurs to me we could get the needed functionality in cutechess. If `order=random`would allow for a `start=N` to specify the location in the random stream, we could easily just use the right starting point in the stream (using the same seed for all workers)

ppigazzini:
@vondele the new cutechess-cli is only 2 months old. Have you tested that version?
ps:  I'm away from my keyboard

noobpwnftw:
Another workaround is to generate opening set per task by picking and shuffling the lines in the chosen opening book from the worker side, then launch cutechess with each of those intermediate books.

vondele:
@ppigazzini compiling qt5 right now :-) let's see if I manage to follow your instructions and build cutechess-cli from sources.

@noobpwnftw yes that would work, and be rather easy for the epd format. For pgn we would need some kind of a basic parser, or convert those to epd as well.

ppigazzini:
@vondele expect to wait some time with a desktop CPU :)
I uploaded on the "books" repo the cutechess-cli built 2 months ago for variant SF (I preferred to use a well tested build in this first switch). I tested only the start of cutechess-cli on recent version of several linux distro, view the list on the wiki page. Some distro (Debian and Gentoo if I recall correctly) required to install a missing package.

vondele:
3h for qt5 on my (old) desktop :) but I do have a working version of cutechess-cli now. 

[...]

PR #785, opened 2020-09-04
Use unique openings

vondele:
each worker uses a unique subset of the opening book, which is randomized with a consistent seed on all workers.

fixes https://github.com/glinscott/fishtest/issues/478

Needs a recent cutechess (1.2).

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -89,6 +89,35 @@ def required_net(engine):
 
     return net
 
+def verify_required_cutechess(cutechess):
+    print("Obtaining version info for %s ..." % (os.path.basename(cutechess)))
+
+    p = subprocess.Popen(
+        [cutechess, "--version"],
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+        bufsize=1,
+        close_fds=not IS_WINDOWS,
+    )
+
+    pattern = re.compile("cutechess-cli ([0-9]*).([0-9]*).([0-9]*)")
+    for line in iter(p.stdout.readline, ""):
+        m = pattern.search(line)
+        if m:
+           print("Found: ",line)
+           major = int(m.group(1))
+           minor = int(m.group(2))
+           patch = int(m.group(3))
+
+    p.wait()
+    p.stdout.close()
+
+    if p.returncode != 0:
+        raise Exception("Failed to find cutechess version info" % (p.returncode))
+
+    if major < 1 or (major == 1 and minor < 2):
+       raise Exception("Requires cutechess 1.2 or higher, found version doesn't match")
+
 
 def required_net_from_source():
     """ Parse evaluate.h and ucioption.cpp to find default net"""
@@ -761,10 +790,11 @@ def run_games(worker_info, password, remote, run, task_id):
     if not "pentanomial" in input_stats:
         input_stats["pentanomial"] = 5 * [0]
 
-    assert (
-        2 * sum(input_stats["pentanomial"])
-        == input_stats["wins"] + input_stats["losses"] + input_stats["draws"]
+    input_total_games = (
+        input_stats["wins"] + input_stats["losses"] + input_stats["draws"]
     )
+
+    assert 2 * sum(input_stats["pentanomial"]) == input_total_games
 
     input_stats["crashes"] = input_stats.get("crashes", 0)
     input_stats["time_losses"] = input_stats.get("time_losses", 0)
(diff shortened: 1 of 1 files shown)


vondele:
I have turned the case of not .epd and .pgn opening books into an assert. I think this shouldn't be possible right now.

vondele:
first testing with just my worker connected seems to work well (all game pairs had unique openings). Would need some updated workers (+updated cutechess) connected to see if this also works well if cutechess runs on different OS.... 

If the same srand, doesn't lead to the same random ordering on openings, we might not fully eliminate duplicates (but improve it).

ppigazzini:
@vondele DEV and workers updated, using cute120

vondele:
All unique fens for 5f523f2bdeb31bb53d26c911

for i in `seq 0 299`; do wget https://dfts-0.pigazzini.it/api/pgn/5f523f2bdeb31bb53d26c911-$i.pgn ; done ; 
grep FEN *pgn | cut -d: -f2 | sort | uniq -c | awk '{print $1}' | sort | uniq -c

I'll run it a second time with the larger pool of workers (this one had already 2).

ppigazzini:
@vondele I use this scripts to manage a test workers fleet on my 32 thread VPS

```
#!/bin/bash
# start_test_fleet.sh
for i in {0..5}; do
  sleep $((${i}*15)) && echo ${i} && nohup bash start_test_worker.sh ${i} &>/dev/null &
done
```

```
#!/bin/bash
# stop_test_fleet.sh
touch ${HOME}/_git/__test_folder{0..5}/fishtest/worker/fish.exit
```

```
#!/bin/bash
# start_test_worker.sh
dir_num=${1:-'00'}
usr_pwd=${2:-'user00'}
test_folder=${HOME}/_git/__test_folder${dir_num}
virtual_env=${test_folder}/fishtest/worker/env

rm -rf ${test_folder}
mkdir -p ${test_folder}
cd ${test_folder}
git clone --single-branch --branch master https://github.com/glinscott/fishtest.git
cd ${test_folder}/fishtest
git config user.email "you@example.com"
git config user.name "your_name"

# add here the upstream branch to be tested
git remote add new-upstream https://github.com/ppigazzini/fishtest
git pull --no-edit new-upstream TEST_ppigazzini_books

# add here the PRs to be tested
git pull --no-edit origin pull/783/head
git pull --no-edit origin pull/785/head

cd "${test_folder}/"fishtest/worker
python3 -m venv ${virtual_env}
${virtual_env}/bin/python3 -m pip install --upgrade pip setuptools wheel
${virtual_env}/bin/python3 -m pip install requests
rm -rf requests

${virtual_env}/bin/python3 worker.py --host dfts-0.pigazzini.it ${usr_pwd} ${usr_pwd} --concurrency 5
```

vondele:
again all game pairs from a unique fen. So, from my point of view this PR ready. Might benefit from a second pair of eyes.

Does does need a merge of the new cutechess in the books repo. @ppigazzini what do you think? It presumably is bug free for our purpose (i.e. non-variants), so I would go ahead if you think that's the thing to do.

vondele:
@ppigazzini is your test worker fleet including the windows worker, or is that standalone.

ppigazzini:
@vondele just started 2 msys2 workers and 2 wsl workers.

According @ianfab that bug should hit only a couple of variants, see https://github.com/ianfab/FishCooking/pull/11#issuecomment-687138341

vondele:
also works fine with your msys2 and wsl workers. 

So I propose I'll update the books repo with the new cutechess binaries. I recall correctly that if we do a worker version update, new cutechess binaries are downloaded by the workers as well, right ?

ppigazzini:
I'm wondering if it's acceptable to have some workers with an old version cutechess custom build.
Otherwise we should add a cutechess version check to be sure that everybody is using the correct version.

vondele:
it is certainly not ideal to have old cutechess versions around. If multiple workers would do this, this would lead to game duplication.

note that we don't build the release version, so that would be hard to get. Just looking at the version number would be sufficient for now.

vondele:
I have added the version check. Works with my worker, and I verified the same check should work with the windows version of cutechess-cli.

worker/games.py
```diff
@@ -89,6 +89,35 @@ def required_net(engine):
...
+           minor = int(m.group(2))
+           patch = int(m.group(3))
+
+    p.wait()
+    p.stdout.close()
+
+    if p.returncode != 0:
+        raise Exception("Failed to find cutechess version info" % (p.returncode))
+
+    if not (major > 1 or (major == 1 and minor >= 2)):
```
ppigazzini, line 118: `if major < 1 or (major == 1 and minor < 2): `

ppigazzini:
Tested OK with cutechess-cli 1.1.0 (raise) and 1.2.0 (works fine) on linux, msys2, wsl.

vondele:
OK, squashed with your review suggestion.

ppigazzini (approved):

ppigazzini:

ppigazzini:
@noobpwnftw please remember to download the updated cutechess-cli v 1.2.0 from https://github.com/official-stockfish/books

noobpwnftw:
Done @ppigazzini

PR #788, opened 2020-09-06
Link username to user tests page from tests_view pages

linrock:
Fixes https://github.com/glinscott/fishtest/issues/625#issuecomment-629713265 as suggested by @silversolver1

> I mean on the page of an actual test where a user can access the stop/purge, reschedule or modify buttons, next to where the value for itp is listed, there is the word "username". here, the user's particular name does not currently link to that user's /tests/user/ page, which is the feature I am suggesting

I agree this is more practical since I've also commonly wanted to go see a list of the user's tests while on an individual test page.

---
Approvers will see a link to the user admin page in addition to the link to the user tests page.

image

### DIFF
--- a/fishtest/fishtest/templates/tests_view.mak
+++ b/fishtest/fishtest/templates/tests_view.mak
@@ -34,8 +34,13 @@
       %if len(arg[2]) == 0:
         <tr>
           <td>${arg[0]}</td>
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:

Issue #792, opened 2020-09-16
Scheduling issue


noobpwnftw:
There are exactly 500x fleet workers, however, a lot of them got "no task" response.
This problem only appears whenever there is SPSA and also worker have a chance failing API requests(either to fishtest or github), after they fail, their task "slot" stays and will not get rescheduled until probably an hour later as the script clears them.

noobpwnftw:
This however does not appear to be a problem when there is no SPSA running.

vondele:
let me do a test and set all SPSA to -1 priority. See if this fixes things.

vondele:
@noobpwnftw no SPSA jobs should be running right now.

noobpwnftw:
As you can see more workers appear(although old ones are still stale, and total NPS gets larger indicating workers picked up their task).

noobpwnftw:
I'll double fleet workers and see if the observation holds. If so, then you can restart SPSA and see if it get stuck again.

vondele:
one interesting point, even though I stopped and set priority to -1, some of the SPSA runs are in active state 'spontaneously'.

vondele:
I have now switch 4 SPSA runs to prio 0, see what happens. I can also put a test to prio 1.

noobpwnftw:
If you put SPSA on prio 1, I think all other tests will gradually lose workers and only one SPSA will remain active, with only some hundred cores active :)

vondele:
There is some special handling of SPSA runs:
https://github.com/glinscott/fishtest/blob/41723fc7b6a7f9f5b13caadbdf60c6374a54e90f/fishtest/fishtest/rundb.py#L588-L596
maybe the logic is buggy, blocking all tasks as soon as one SPSA run has reached its limit?

vondele:
should that break be a continue instead? @tomtor ?

ppigazzini:
@vondele that "break" is suspect... I replaced it with a "continue" on PROD by hand, feel free to make some tests.

vondele:
that's maybe a little adventurous, I don't know the code well enough. Now that it has been done, let me again put prio back to 0 for these tests.

ppigazzini:
@vondele already reverted, after a second reading the "break" seems fine.

vondele:
Also, it looks like this limit should not be reached easily. The largest tune has 28 params (if I looked correctly), which would limit the cores for that test to ~7500, and we're not near.

FauziAkram:
[...]

PR #794, opened 2020-09-28
Restore waitress default for connections/threads

ppigazzini:
Default values and higher values showed the same performance in some tests with
2000 machines 17000 cores.

Restoring the threads default values allows to save some RAM useful
to raise the MongoDB cache.


tomtor (approved):

ppigazzini:
PR set as WIP.
PROD is already running with the default values set by hand.

IMO better to wait for a SPSA batch resize PR and some tests with the big fleet, see https://github.com/glinscott/fishtest/issues/792#issuecomment-700243014

PR #801, opened 2020-10-03
Use the string format method

ppigazzini:
Make the worker code more consistent using only the format method to formatting
the strings.


ppigazzini:
DEV and workers updated, tests seem fine.

PR #803, opened 2020-10-05
NN upload accept .nnue

vondele:
should presumably simplify the upload in that only .nnue files should be listed in the 'browse' file selector.


vondele:
untested.

vondele:
untested.

ppigazzini:

vondele:
thanks!

Issue #599, opened 2020-04-16
Night Mode


jhellis3:
First, just want to thank everyone who works to keep fishtest improving; the progress has been truly remarkable.  

Second, if it is not too much trouble, I was wondering if someone could do a night/dark theme to help save our eyes after hours.  I could probably try to do something, but it really isn't my area of expertise.  It would probably take me a while, and the end result would likely be worse than someone who knows what they are doing.

SFisGOD:
@linrock Thank you so much for your work in improving fishtest. If you have the time, we would be really grateful if you could implement something like in openbench. They have a nice dark theme there http://chess.grantnet.us/

linrock:
i'll work on a night mode theme today and aim to get a PR ready for review by the end of the day!

Issue #252, opened 2018-04-24
Proposal for an auto-approve feature


Stefano80:
I noticed that since we introduced the last changes https://github.com/glinscott/fishtest/pull/239 concerning autofill of the benches, all devs started actually writing the bench of the branch in the commit message since they saw an advantage with it.

In short, by giving a reward to use good patterns, they immediately use the better pattern.

I was thinking, maybe we can allow auto-approve, using the following design. Are eligible to autoapproval those patches which satisfy following criteria:
- there are only changes in `.h` or `.cpp` files
- the branch passes a Travis CI build on the committer repository
- it has [0,5] bounds or [0,4] with only changes in numerics

Every 15 min we then automatically scan the pending tests and check whether there is a corresponding passing Travis CI build using their API and approve those eligible.

This would give a reward for connecting the repository to Travis, give a reward to submit non [-3,1] patches, and get some workload off the approvers.

Any thoughts?

mstembera:
I thought the purpose of the approval process was to prevent any malicious code from being executed on the worker machines.  Auto approval would be a huge security hole.

tomtor:
> it has [0,5] bounds or [0,4] with only changes in numerics

In practice only [0,4] tests will only have changes to numeric constants. For those changes I do not see the reason/advantage of Travis builds. Auto-approving these tests makes sense to me, because introducing real malicious code with those changes is not possible. DOS attacks, eg allocating enormous amounts of memory, etc., are possible, but I would trust such an automated test more then the current human source code inspection. @vondele What are your thoughts?

vondele:
As an example of a [0,4] test that would fail a travis build see the recent: 
http://tests.stockfishchess.org/tests/view/5adf2bfc0ebc590331056b48
(look at commit comments).

In addition to the security aspect, I like the policy of having all patches 'approved'. At least one other human has seen the idea, can comment on it, can correct, etc. There have been countless times where I was inspired by or learned something from a patch I approved.

I also don't see the need to differentiate [-3,1] from [0,N] patches... both classes are needed to keep the code in good shape.

That the framework is sometimes going idle with some patches pending is, a pity but not terrible.

Stefano80:
Ok, I try to summarize:

1) from the security point of view, there is no risk in auto-approving parameter tunes
2) there are some corner cases in which even parameter tunes could fail on CI 
3) from the security point of view, there could be some risk in auto-approving functional patches, but automatic test should be more efficient in catching malicious code
[...]

PR #806, opened 2020-10-13
Enable toggling between light and dark mode

linrock:
This PR adds a sun and a moon icon to the bottom of the sidebar. Clicking on the icons will switch the site between light and dark modes. Dark mode is implemented as a CSS file that's injected into the page to override the default CSS.

Example of how the homepage looks in dark mode:

image

All suggestions welcome! Some pages that I was unable to view in dev may have unreadable text in dark mode and will need to be fixed. Some standalone pages don't use bootstrap 2 so they haven't been updated to respect dark mode yet. 

Fixes https://github.com/glinscott/fishtest/issues/599

### DIFF
--- a/fishtest/fishtest/static/css/application.css
+++ b/fishtest/fishtest/static/css/application.css
@@ -6,6 +6,10 @@
   }
 }
 
+body {
+  transition: background 0.2s ease-in-out;
+}
+
 .flash-message {
   margin-top: 20px;
 }
@@ -34,4 +38,8 @@ th {
 
 .spsa-param-row td:first-child {
   text-align: left;
+}
+
+#change-color-theme:hover {
+  cursor: pointer;
 }
--- a/fishtest/fishtest/static/css/theme.dark.css
+++ b/fishtest/fishtest/static/css/theme.dark.css
@@ -1,0 +1,127 @@
+/* A custom dark color theme for overriding bootstrap 2.3.2 css rules */
+
+body {
+  background: #111;
+  color: #fafafa;
+}
+
+a {
+  color: #63daff;
+}
+
+a:hover, a:focus {
+  color: #e1ffff;
+}
+
+.btn-link {
+  color: #63daff;
+}
+
+.btn {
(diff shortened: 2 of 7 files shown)


SFisGOD:
@linrock Thank you so much for your work! I'm sure others are as grateful as I am.

SFisGOD:
@linrock I don't know if this is hard to implement but maybe if you have some free time https://github.com/glinscott/fishtest/issues/252#issuecomment-386876894
The auto-LTC feature is really convenient for devs and was requested for a long time now.

ppigazzini:
@linrock DEV updated

ppigazzini:
@linrock the night/day "prev-next-pages" menu have slightly different lengths.

linrock:
@ppigazzini thanks, i fixed the issues you mentioned:

- test stats + live elo pages now respect dark mode if it's enabled
- "prev-next-pages" menu sizes now stay the same when switching between night/day

ppigazzini:
DEV updated, it looks good to me :)

ppigazzini:

Sopel97:
Very nice. Though I won't switch unless other white elements are darkened too, currently there's too much contrast.

jhellis3:
This is great, thank you for doing this.

VoyagerOne:
@linrock 
Awesome work!  Does this work on mobile?
It works on my desk PC but not on Galaxy S9 phone.

ppigazzini:
@VoyagerOne on my Huawei Mate10pro it works fine with both chrome and native browser.

ppigazzini:
@linrock just nit picking after viewing the sun/moon icons on TCEC :)
In night/day mode the status is obvious, so the icon should show the action, see 
https://ux.stackexchange.com/questions/39418/night-day-mode-toggle 

When the status is not obvious the on/off toggle usage is way more complicated :)
https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan?rq=1

Issue #809, opened 2020-10-16
Stuck SPSA runs


tomtor:
As reported on the forum: https://groups.google.com/d/msg/fishcooking/no-PkdKafpA/4TZhEY-uAwAJ

https://tests.stockfishchess.org/tests/view/5f8744f4dcdad978fe8c51fe
https://tests.stockfishchess.org/tests/view/5f875213dcdad978fe8c5211
https://tests.stockfishchess.org/tests/view/5f8744fadcdad978fe8c5200
 
These seem to be somehow "stuck", indicating a number of active cores in
the main list, while there seem to be no active machines on the respective
test pages. Also, the number of games/iteratations remains absolutely
constant.

EDIT_000: added the right link to the forum post

tomtor:
@ppigazzini  I will add some debugging code to prod.

tomtor:
@ppigazzini Could it be that live changes to the python code in `~/fishtest/fishtest` are no longer picked up (after a restart) in the current deployment with `pip install`? We used to `pip install` in development mode.

Any hints because I don't see my debugging output in the logs...

ppigazzini:
@tomtor try filtering with ` sudo journalctl -u fishtest@6543 --since 10:00`

ppigazzini:
@tomtor on DEV I changed `views.py`, restarted with `sudo systemctl restart fishtest@{6543..6544}` and got "AAAAAAAA Your account has been created."

Also on PROD you can setup fishtest merging a github branch, check `~/x/update_fishtest.sh`

ppigazzini:
@tomtor it works after a reboot ...
```
Oct 16 10:55:55 tests.stockfishchess.org pserve[348]: Starting server in PID 348.
Oct 16 10:55:55 tests.stockfishchess.org pserve[348]: Serving on http://localhost:6543
Oct 16 10:55:55 tests.stockfishchess.org pserve[348]: Return task: 120 {'uname': 'Linux 3.10.0-1127.19.1.el7.x86_64', 'architecture': ['64bit', 'ELF'], 'concurrency': 8, 'max_memory': 1834, 'min_threads': 1,
Oct 16 10:55:55 tests.stockfishchess.org pserve[348]: Return task: 773 {'uname': 'Linux 3.10.0-1127.19.1.el7.x86_64', 'architecture': ['64bit', 'ELF'], 'concurrency': 8, 'max_memory': 1834, 'min_threads': 1,
[...]

PR #810, opened 2020-10-16
Increase active timeout settings

tomtor:
Increase timeout settings because we now batch SPSA results.

In addition workers with 7 or more cores will never show as active because they post only results after 200 played games.
This is caused by a batch size of 32*concurrency.

@vondele I remember you suggested to increase this setting? What would you suggest?

We could adjust with concurrency of workers and NPS.

EDIT: See also https://github.com/glinscott/fishtest/issues/809

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -315,7 +315,7 @@ class RunDb:
         self.run_cache_lock.release()
 
     def scavenge(self, run):
-        old = datetime.utcnow() - timedelta(minutes=30)
+        old = datetime.utcnow() - timedelta(minutes=110)
         for task in run["tasks"]:
             if task["active"] and task["last_updated"] < old:
                 task["active"] = False
--- a/fishtest/utils/scavenge.py
+++ b/fishtest/utils/scavenge.py
@@ -13,7 +13,7 @@ from fishtest.rundb import RunDb
 rundb = RunDb()
 
 
-def scavenge_tasks(scavenge=True, minutes=60):
+def scavenge_tasks(scavenge=True, minutes=120):
     """Check for tasks that have not been updated recently"""
     for run in rundb.runs.find({"tasks": {"$elemMatch": {"active": True}}}):
         changed = False
@@ -25,7 +25,7 @@ def scavenge_tasks(scavenge=True, minutes=60):
                 task["active"] = False
                 changed = True
         if changed and scavenge:
-            rundb.runs.save(run)
+            rundb.runs.replace_one({"_id": run["_id"]}, run)
 
 
 def main():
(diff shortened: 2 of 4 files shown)


ppigazzini:
@tomtor `scavenge.py` tested working on DEV loading a DB backup (several workers passed the timeout). 

While here please fix this warning:
```
scavenge.py:31: DeprecationWarning: save is deprecated. Use insert_one or replace_one instead
  rundb.runs.save(run)
```

vondele:
@tomtor, not sure what to suggest. 

Probably it needs to be actually something depending on batch size, concurrency and TC? We can expect games to run for 
sf * estimate_game_duration(tc) * batch_size * threads / concurrency
where sf is some kind of scale factor probably ~ 4.

ppigazzini:
We compute/use a tc_limit in `games.py`

https://github.com/glinscott/fishtest/blob/b5cf10759ab5f6b7329e6cfadbd25600d8e87178/worker/games.py#L451

https://github.com/glinscott/fishtest/blob/b5cf10759ab5f6b7329e6cfadbd25600d8e87178/worker/games.py#L1003-L1004

https://github.com/glinscott/fishtest/blob/b5cf10759ab5f6b7329e6cfadbd25600d8e87178/worker/games.py#L1102

https://github.com/glinscott/fishtest/blob/b5cf10759ab5f6b7329e6cfadbd25600d8e87178/worker/games.py#L586-L590

tomtor:
> While here please fix this warning:

Done

tomtor:
@vondele @ppigazzini I don't want to complicate the code too much, because the timeout is just a safety net for crashed workers. Normally on CTRL-C a worker will send a sign-off message.

So the timeout should just be not too small, but indeed ideally it takes TC/threads/etc into account.

I can live with just using `concurrency` to adjust a large timeout.

vondele:
OK, let's keep it simple, I don't know the code will enough to suggest otherwise :)

tomtor:
@ppigazzini I will change the PR tomorrow to skip scaling for SPSA tests...

tomtor:
@ppigazzini I think this can be commited.

First step is to update the workers, after that we can update the hand-patched code on the prod server with this PR.

It is not critical, we can also update/restart the server earlier, timeout values are better than the old master and the issue is mostly cosmetic.

ppigazzini:
@tomtor great, I can do that this afternoon.

ppigazzini:
@tomtor updated DEV and some workers.

ppigazzini:
Thank you @tomtor :)
PROD updated (I want to merge some night mode PRs). Some git objects had wrong ownership (?!?) and the usual pull failed, I reinstalled fishtest with `x/update_fishtest.h`, now git works fine.

noobpwnftw:
@ppigazzini Updated.

PR #816, opened 2020-10-18
List 500 NN entries

tomtor:
Fix #815 for now.

Maybe add paging/search later.

PR #817, opened 2020-10-18
Add NN paging and show/hide non default nets

tomtor:
@vondele As a followup to #816 some quick paging. Not sure if we really need this, `/users` is more than 1000 entries and works fine too, but better safe than sorry.

@ppigazzini I did some creative testing of this. You could lower page_size by hand on dev to 100 for testing....

### DIFF
--- a/fishtest/fishtest/templates/nns.mak
+++ b/fishtest/fishtest/templates/nns.mak
@@ -9,6 +9,10 @@ and achieved the status of <i>default net</i> during the development of Stockfis
 The recommended net for a given Stockfish executable can be found as the default value of the EvalFile UCI option.
 </p>
 
+<button id="non_default-button" class="btn">
+      ${'Hide non default nets' if non_default_shown else 'Show non default nets'}
+</button>
+
 <table class="table table-striped table-condensed" style="max-width: 900px;">
  <thead>
   <tr>
@@ -23,6 +27,7 @@ The recommended net for a given Stockfish executable can be found as the default
  <tbody>
  %for nn in nns:
   %if request.authenticated_userid or 'first_test' in nn:
+  %if non_default_shown or 'is_master' in nn:
   <tr>
    <td>${nn['time'].strftime("%y-%m-%d %H:%M:%S")}</td>
    %if 'is_master' in nn:
@@ -45,6 +50,18 @@ The recommended net for a given Stockfish executable can be found as the default
    <td style="text-align:right">${nn.get('downloads', 0)}</td>
   </tr>
   %endif
+  %endif
  %endfor
  </tbody>
 </table>
+<p>
+ %if prev_page:
+  <a href="nns?page=${prev_page}">&lt; Later networks</a>
+ %endif
+ %if prev_page and next_page:
+  <span>&#8209;</span>
+ %endif
+ %if next_page:
+  <a href="nns?page=${next_page}">Earlier networks &gt;</a>
+ %endif
+</p>
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -192,11 +192,11 @@ class RunDb:
             return nn
         return None
 
-    def get_nns(self, limit):
+    def get_nns(self, limit, skip=0):
         return [
             dict(n, time=n["_id"].generation_time)
             for n in self.nndb.find(
-                {}, {"nn": 0}, limit=limit, sort=[("_id", DESCENDING)]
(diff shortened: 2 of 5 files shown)


tomtor:
@vondele We could lower the `500` pagesize somewhat now that we have paging?

ppigazzini:
@tomtor DEV updated with `page_size = 5` for hallucinated testing, some pages have lesser than 5 entries (???).

tomtor:
> some pages have lesser than 5 entries

Some entries are hidden when not logged in

tomtor:
@ppigazzini Added show/hide of non default nets...

tomtor:
@ppigazzini Forgot to bump the javascript version....

ppigazzini:
DEV updated with 50 pages. LGTM.

tomtor:
With 50 i get only one green network one the first page when hiding non-default nets.

ppigazzini:
> With 50 i get only one green network one the first page when hiding non-default nets.

I thought that intended action was to hide the non default nets of the page.

tomtor:
> I thought that intended action was to hide the non default nets of the page.

Sure, I meant that lowering from 250 to 50 is not a good idea.

I will squash the PR, I think it's OK to merge.

ppigazzini:
50 is creative testing :)

vondele:
I'm maybe a bit late with the comment, what about using 'earlier networks' instead of 'more networks' and I'm wondering if we should have a way to go back 'later networks'. 
I would have a slight preference for showing all nets by default.

tomtor:
@ppigazzini Added Joost's suggestions and changed page size to 250

ppigazzini:
DEV updated, page=50 for creative debug, I will swap the "earlier" and "later" positions to have the same order of pages from pagination().

BTW I applied unsuccessfully my js coding skillset (copy&paste&hope_it_works) with the pagination code :)

https://github.com/glinscott/fishtest/blob/c7acb76feed19f3cf96bed5f27e7e1c15d02799c/fishtest/fishtest/templates/run_table.mak#L39-L59

tomtor:
> I applied unsuccessfully my js coding skillset

@ppigazzini This would need (a lot of) additonal code in `views.py` to work.

I swapped Later/Earlier as suggested by you

fishtest/fishtest/templates/nns.mak
```diff
@@ -45,6 +50,15 @@ The recommended net for a given Stockfish executable can be found as the default
    <td style="text-align:right">${nn.get('downloads', 0)}</td>
   </tr>
   %endif
+  %endif
  %endfor
  </tbody>
 </table>
+<p>
```
ppigazzini, line 57: This seems to work on DEV:
```
<p>
 %if prev_page:
  <a href="nns?page=${prev_page}">&lt Later networks</a>
 %endif
 %if prev_page and next_page:
  &#8209;
 %endif
 %if next_page:
  <a href="nns?page=${next_page}">Earlier networks &gt</a>
 %endif
</p>
```

tomtor: @ppigazzini Looks much better! Thanks, will add that.

ppigazzini: Please double check my syntax :)

tomtor: I spotted the missing `;` but for some reason `&#8209;` shows also red in my FF page source view?

ppigazzini: Perhaps `<span>&#8209;</span>` ?
https://www.toptal.com/designers/htmlarrows/punctuation/non-breaking-hyphen/

ppigazzini: DEV updated with `<span>` tag.

tomtor: `<span>`looks good

vondele: maybe 'newer' and 'older' are better than 'earlier' and 'later' ...

ppigazzini:

tomtor:
Thank you @ppigazzini :)

SFisGOD:
@tomtor Thanks! I think it's better that the default view only shows the green nets?

tomtor:
> @tomtor Thanks! I think it's better that the default view only shows the green nets?

@vondele prefered it this way, I think I do too. It remembers your choice, so it's not a real issue.

fishtest/fishtest/templates/nns.mak
ppigazzini: @vondele done.

image

PR #818, opened 2020-10-18
Mark reported core count on tests if suspicious (main test page).

xoto10:
[ For discussion, not necessarily this exact code change. May not be appropriate for SPSA, I am not sure. ]

Nice work guys on #792 and #809 !

In a similar matter, but not for spsa jobs, I think jobs often appear active but actually have no active cores running if many workers are withdrawn from fishtest at one time. If fishtest is getting more jobs at the time, the number of cores to allocate to individual jobs decreases and the job does not even get a few extra cores to add to the missing cores which fishtest thinks are still there.
Depending on timeouts / the normal fluctuation of worker allocation, fishtest eventually corrects itself, but it can be frustrating if the user realises their job is not progressing.

Could we do a patch something like this PR to add a subtle indication that the cores number may be unreliable? This would give us a bit more feedback on the main test page - we could see how this goes and then maybe gradually use this to correct some of the causes and/or use a similar technique to try to improve the reported core count and hours remaining at the top of the main test page.

The code here says ` hours = 1 ` but I would hope to change this to something shorter (10 mins or so?) if it works ok. (How often are LTC results sent back if running on 1 core?)

### DIFF
--- a/fishtest/fishtest/templates/run_table.mak
+++ b/fishtest/fishtest/templates/run_table.mak
@@ -123,7 +123,8 @@ ${pagination()}
           %endif
           @ ${run['args']['tc']} th ${str(run['args'].get('threads',1))}
           <br>
-          ${('cores: '+str(run['cores'])) if not run['finished'] and 'cores' in run else ''}
+          ${('cores: '+("~" if (datetime.utcnow() - run["last_updated"] > timedelta(hours=1)) else "")
+                      +str(run['cores'])) if not run['finished'] and 'cores' in run else ''}
         </td>
 
         <td style="word-break: break-word;">


xoto10:
e.g. my current test https://tests.stockfishchess.org/tests/view/5f8af3d505363b249ba45919 which I have set to prio -1, but it is currently still showing active as fishtest thinks the grandphish machine is still running this job. The machine stopped running this job hours ago after 56 games out of 200 :

In Task list:
597 | grandphish2-16cores-11 | Windows 10      ARCH=64bit BMI2 AVX2 SSE41 SSSE3 SSE2 POPCNT
 | 2020-10-18 **17:19:49** | 056 / 200 | 5 | 2 | 49 | 0 | 0 | 0.484

On test detail page:
start time | 2020-10-17 13:38:29
last updated | 2020-10-18 **17:55:09**

On main test page (Active list):
20-10-17 | xot | wev1 | diff | LLR: 0.49 (-2.94,2.94) {-0.25,1.25}
Total: 138656 W: 14493 L: 14271 D: 109892
Ptnml(0-2): 630, 11140, 45610, 11274, 674 | sprt           @ 10+0.1 th 1                      **cores: 16**


The time now is **18:50** UTC.

If it works as expected my code change would display ` cores: ~16 ` against this job, and perhaps in time related changes could help us display jobs like this more accurately.

tomtor:
This is somewhat of an issue when the @noobpwnftw fleet is restarted for a worker update.

The old workers do not sign off, probably because they are not killed with a SIGINT or SIGTERM.
After 110 minutes we clear the old workers. Lowering the 110 minutes raises the risk of dropping the results of slow workers, eg for (future) VLTC tests. I agree that currently 110 minutes is very much on the safe side. Changing it to 60 would have the same effect as your PR.

xoto10:
@tomtor 
Yes, I realised after my last comment that the 110 minute scavenge fixed the test I referred to above. Originally I thought it was a slightly different problem - but it's good that we understand what's happening.

Perhaps we can separate the issues slightly? For updating individual jobs, we want to be relatively confident that workers have disconnected, so a longer timeout seems sensible. For displaying the summary stats at the top of the /tests page we could look for the "probable" position rather than requiring more certainty.

At the moment, this patch doesn't change fishtest operation, it just alters the display slightly to highlight tests that might have an issue. This could be a problem if it encourages users to update tests when we would prefer them to just wait and let fishtest sort itself out, or it could be a useful indication to users that fishtest is monitoring what is happening; "~" added to core count after 60 mins ... disconnected cores removed after 110 mins. Perhaps it would highlight jobs we don't currently notice?

I'm interested in making the core count at the top of the main page more accurate if we can, e.g. at the moment it says ~2000 cores, but is that just because noob's fleet has reconnected? Perhaps we could calculate `cores` via allocated tasks that have updated recently rather than via machines that we think are (still) connected, maybe a change to get_machines() ?
Also I might look again at my patch to count the games played / minute instead of estimating it, I think it would be good if we can make these stats more accurate.

xoto10:
e.g. in get_machines() we could do something like :
```
machine['active_cores'] = 0 if (datetime.utcnow() - run["last_updated"] > timedelta(minutes=15))
                     else machine['concurrency']
```
and then use ` m['active_cores'] ` instead of ` m['concurrency'] ` when calculating ` cores ` ?

vondele:
could active workers be asked to ping the server every 1min (5min) and we base decisions on that, so we don't have to guess timeouts etc.

xoto10:
I've created https://github.com/glinscott/fishtest/compare/master...xoto10:cores2 along the lines of my idea above.

@vondele 
Yes, if the workers did a regular ping, perhaps this active_cores idea could be adpated to match.

ppigazzini:
> could active workers be asked to ping the server every 1min (5min) and we base decisions on that, so we don't have to guess timeouts etc.

With a worker ping the STC/LTC/VLTC tests will be threated in the same way but, If I'm not wrong, we should need anyway some timeouts to deal with a server downtime (e.g. mongodb compact collections) or a network interruption.

xoto10:
In practice, the workers are generally updating the server within a few minutes already, so using the current last_updated value should be good enough to get a much better total core count than we currently have. e.g. right now most workers have last_updated < 10 minutes ago, with a handful between 10 and 20 minutes. I assume those with last_updated > 20 minutes ago have disconnected from those jobs.

vondele:
my suggestion was ping the server independently of running games (i.e. their batching, their TC, etc.) so the question which workers are active should be easier to answer. (right now, the (rare) single longest games could be taking ~2400s for 60s+0.6s tests, as in 2*(60+800*0.6)*16/7 (800 ply game, we have seen longer, 2 sides, tc scaling for a 700000 nps worker).

tomtor:
> could active workers be asked to ping the server every 1min (5min) and we base decisions on that, so we don't have to guess timeouts etc.

That would be the only robust solution. We could scavenge simply on 3 our more missed pings.

xoto10:
Sure, individual games can be long, but usually there will other cores in the same worker running games and causing updates to the server. Looking at the machine list it looks to me like a 10 minute cut-off on the existing last_updated field would get a pretty good count 10 minutes after any large-scale dropout of machines. Maybe one or two machines will be excluded when they should be included, but I don't think this is a big deal, the main thing is to exclude the many / high core count machines reasonably soon after they drop out. Better to sometimes be 40 cores below the "correct" core count than sometimes 400 cores above the "correct" count! My cores2 branch is only a few lines of code and should be a big improvement on waiting for the 110 minute scavenge, but if we are happy to code an accurate count using a heartbeat then that would be good too. Whatever you guys decide :)

PR #829, opened 2020-10-22
Raise worker version to 90 (server side)

ppigazzini:


ppigazzini:
Todo after merging:
- [ ] set minutes=30 by hand and restart the server
https://github.com/glinscott/fishtest/blob/8889d4c2126bba29e257a0ece279e948239f32ef/fishtest/fishtest/rundb.py#L329
- [ ] drop `utils/scavenge.py`
- [ ] delete `scavenge.py` in crontab (commented out now)
- [ ] update the server setup scripts
- [ ] restore the default when all workers are updated to v89
https://github.com/glinscott/fishtest/blob/8889d4c2126bba29e257a0ece279e948239f32ef/fishtest/fishtest/rundb.py#L329

ppigazzini:
Raised version to 90 for #831

PR #832, opened 2020-10-24
Retire utils/scavenge.py

ppigazzini:
The server uses now a better scavenging process based on workers heartbeat, see
https://github.com/glinscott/fishtest/commit/08d525c4af2016adbbfbe844a325c5ce9a67b847


tomtor:
@ppigazzini I noticed that some of the workers from the @noobpwnftw fleet on version 90 still show update times > 1 minute. That should not happen unless network connectivity is not stable.

ppigazzini:
I'm expecting this, if I'm not wrong the solution is to set a proper timeout here (now 30 by hand), 15 should works:
https://github.com/glinscott/fishtest/blob/e521fcc06200ff060c4086aeb2e502718ad0a828/fishtest/fishtest/rundb.py#L332

tomtor:
> I'm expecting this

I was not, I wonder if the current worker which is 8 minutes idle is really still working.

If the heartbeat is not send in 8 attempts, it probably will also fail sending the worker results.

The whole idea of the original PR is lowering the scavenge timeout and detecting bad workers, raising from 3 to 15 minutes is not very attractive as a solution.

ppigazzini:
IMO the PR replace the "non-uniform ping" (wrt the workers - upload of games batch results) with a "uniform ping" (heartbeat), but we still need a proper timeout to deal with a good worker on a bad connection and PROD maintenance tasks, see my https://github.com/glinscott/fishtest/pull/818#issuecomment-712064502

Experimenting on DEV (nginx, systemd unit file, PR etc.) I view the workers keep crunching tests and uploading the result as soon as DEV returns reachable, with a too low timeout I fear that that work could be wasted.

ppigazzini:
@tomtor all workers on v90, timeout lowered to 15 to monitoring the workers.

tomtor:
Maintenance should be handled by not scavenging before being up for eg 2 minutes.
I will create a PR for that later.

Issue #822, opened 2020-10-19
workers with many cores


vondele:
High core workers (64, 88, etc) can't meaningfully participate in fishtest because cutechess can't deal with that many concurrent processes. The user can fix this manually by running multiple workers with reduced concurrency. The alternative would be we deal with this in the worker code, spawning multiple cutechess runs with reduced concurrency (e.g. 16). This happened with at least 3 users so far.

vondele:
BTW, a simple workaround would be to assign these workers to LTC jobs. At high TC this problem is nearly absent.

tomtor:
> BTW, a simple workaround would be to assign these workers to LTC jobs. At high TC this problem is nearly absent.

@vondele I do not understand why it is no problem at LTC, the number of processes is the same?

Is the cutechess issue too much IO between the master and child processes?

Anyway, is the only file that need to be unique when we start multiple cutechess instances the `.pgn` file?

vondele:
I don't know what the exact cutechess issue is, but I've observed it before locally myself, and now it is clear on fishtest as well. Clearly following 60 games at LTC is easier than at STC (quick IO, but also at STC there will be more often moves that are close to the time limit).

Yes, probably the only file that needs to be unique for starting multiple cutechess instances is the .pgn file, assuming they work on the same test of course.

vondele:
I have also filed this as a cutechess issue, but we should try to fix this in the short term 

https://github.com/cutechess/cutechess/issues/630

ppigazzini:
> I have also filed this as a cutechess issue, but we should try to fix this in the short term
> 
> [cutechess/cutechess#630](https://github.com/cutechess/cutechess/issues/630)

The issue reports cutechess-cli 1.1.0, in fishtest we are using 1.2.0. Typo?

vondele:
happens with both, I happen to pick an older local version in a testing dir.

alwey:
@vondele : Could you please make a local test with the old cutechess-0.6.0 for comparison?
I guess the high rate of PV conversion from LAN to SAN might trigger the problem. The old version does not convert PVs into SAN.

vondele:
@alwey I can test but it would be easiest if there is a linux binary for cutechess-0.6.0 somewhere available. 
I didn't see any on github, do you know if they can be found somewhere ?

alwey:
@vondele: Indeed, there is no binary for version 0.6.0 on the cutechess site.

[...]

Issue #825, opened 2020-10-20
Configure number of exec files in .fishtest/worker/testing/


JojoM-sf:
(See https://groups.google.com/g/fishcooking/c/ZLU5Ry07HKM/m/BrSxUaJIBAAJ).

Would it be a great effort to add a switch to the worker script enabling it to configure the number or the amount of files being cached in the testing folder?

I have to manually reduce the respective value (50) in line 856 of games.py every time we are getting an update, as I am regularly hitting my assigned hard drive quota since the advent of NNUE.

JojoM-sf:
Can, after update #91, this issue be closed and would I be authorised to do this?

ppigazzini:
@JojoM-sf yes :)

JojoM-sf:
Ok, then ... ;)

PR #833, opened 2020-10-24
Do not assign workers with large concurrency to STC jobs

vondele:
STC jobs suffer from time losses on machines with large concurrency,
probably because cutechess can't keep up with the IO. Workaround this
issue by assigning workers with a concurrency >= 32 to LTC and multi-threaded
jobs

fixes https://github.com/glinscott/fishtest/issues/822

### DIFF
--- a/fishtest/fishtest/rundb.py
+++ b/fishtest/fishtest/rundb.py
@@ -601,6 +601,15 @@ class RunDb:
                 and run["args"]["threads"] <= max_threads
                 and run["args"]["threads"] >= min_threads
                 and need_tt <= max_memory
+                # To avoid time losses in the case of large concurrency and short TC,
+                # probably due to cutechess-cli as discussed in issue #822,
+                # assign those workers to LTC or multi-threaded jobs.
+                and (
+                    max_threads < 32
+                    or estimate_game_duration(run["args"]["tc"])
+                    * run["args"]["threads"]
+                    > estimate_game_duration("30+0.3")
+                )
             ):
                 task_id = -1
                 cores = 0


vondele:
as usual needs a careful look, untested.

I need to remove/modify https://github.com/glinscott/fishtest/wiki/Running-the-worker%3A-overview#worker-limitations-for-systems-with-a-large-number-of-cores if this is merged.

tomtor:
But these machines will be idle without LTC jobs so it is not a real sollution? 

A (python) wrapper which starts several classic workers with reduced cores would be better?

vondele:
I  agree it is rather pragmatic as an approach. Practically, it is very rare for fishtest not to have LTC jobs running, usually just before going idle. 

The real solution is a fix to the game playing engine cutechess (I did try c-chess-cli, and it seems to be less sensitive, but still timeouts). 

More advanced workarounds are more work, and more fragile. For example a wrapper script will need maintaining, probably needs to merge pgns, and also will have like more idle time on the machine (needs to wait till all subprocesses finish, without work-stealing). However, would be an alternative.

tomtor:
I would suggest spawning multiple processes with the current worker code each with its own work directory path as parameter. That gains also better efficiency because while compiling other workers continue.

The subworkers should not update `fishtest.cfg`. I expect that the additional required code is not much, eg 15 lines in the current worker code (add working dir argument) and 25 lines for the wrapper.

vondele:
Ah, at a higher level thus. I'm not opposed to it, it is probably more general, and i can't judge effort/benefit ratio. If you want to give it a try, we just wait merging this one.

ppigazzini:
> I would suggest spawning multiple processes with the current worker code each with its own work directory path as parameter.

Some notes:
 - with nnue the `testing` directory is growing fat (see https://github.com/glinscott/fishtest/issues/825, on my test workers I pull a branch to keep only 4 binaries), we should keep only one copy
 - the CPU factor could be wrongly estimated according to the random load from other workers, I try to avoid this at least at the start with a timeout 
```
#!/bin/bash
for i in {0..5}; do
  sleep $((${i}*30)) && echo ${i} && nohup bash start_test_worker.sh ${i} &>/dev/null &
done
```

@vondele if I'm not wrong the SF binary embeds the nnue now, so I'm wondering if we can now retire the nnue download

vondele:
@ppigazzini the directory growing bigger.... we can in principle switch off embedding of the binary in fishtest, all should still work. The cache of nets could probably also be reduced a bit.

I agree with your comments about possible problems with spawning multiple workers..  IMO, it is probably safer to go with this patch for now, after all these workers are still very rare. We could revisit this when these workers become more common.

tomtor:
> Some notes:
> 
>     * with nnue the `testing` directory is growing fat (see #825, on my test workers I pull a branch to keep only 4 binaries), we should keep only one copy of `testing` directory
> 
>     * the CPU factor could be wrongly estimated according to the random load from other workers, I try to avoid this at least at the start with a timeout. BTW this should be automatically solved w/ only one master thread making builds:

I don't think these points are real issues for the big machines on which we will spawn multiple workers.

Anyway, it will make the code more complex (so if I implement my proposed PR I will not handle that in the first versions), so probably best to start with the current @vondele PR which looks good to me.

ppigazzini:
DEV and workers updated.

ppigazzini (approved):
All tests are good on DEV. The comment IMO could be improved, eg
```
# cutechess-cli causes time losses in STC jobs for workers with very large concurrency (issue #822),
# assign those workers to LTC or multi-threaded jobs.

vondele:
I have adjusted the comment, but don't want blame cutechess-cli for it.. the real cause will be clear once it is fixed.

ppigazzini:

PR #835, opened 2020-10-24
Save some space in testing dir

vondele:
build binaries without embedded net, keep a smaller cache of nets

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -411,7 +411,7 @@ def setup_engine(
 
         subprocess.check_call(
             MAKE_CMD + ARCH + " -j {}".format(concurrency) + " profile-build",
-            shell=True,
+            shell=True, env=dict(os.environ, CXXFLAGS="-DNNUE_EMBEDDING_OFF")
         )
         try:  # try/pass needed for backwards compatibility with older stockfish, where 'make strip' fails under mingw.
             subprocess.check_call(
@@ -924,11 +924,11 @@ def run_games(worker_info, password, remote, run, task_id):
     # verify that an available cutechess matches the required minimum version
     verify_required_cutechess(cutechess)
 
-    # clean up old networks (keeping the 20 most recent)
+    # clean up old networks (keeping the 10 most recent)
     networks = glob.glob(os.path.join(testing_dir, "nn-*.nnue"))
-    if len(networks) > 20:
+    if len(networks) > 10:
         networks.sort(key=os.path.getmtime)
-        for old_net in networks[:-20]:
+        for old_net in networks[:-10]:
             try:
                 os.remove(old_net)
             except:


vondele:
probably also fixes https://github.com/glinscott/fishtest/issues/825

ppigazzini:
DEV and workers updated.

JojoM-sf:
@vondele Could, supplementing the change in the number of NNUE files from 20 to 10, also the number of SF files be reduced a bit? Let's say from 50 to 30? I am running 6 or 7 workers in a distributed environment, each in its own directory, and my overall HD quota is somewhere close to 8 GB.

vondele:
@JojoM-sf the new binaries should be now back to the old sizes from 20Mb to 300Kb ... would that still be too large?

JojoM-sf:
Each sf file 300K instead of 20M, is that correct?

vondele:
yes

JojoM-sf:
So something like 50*300KB = 15MB for each worker? No problem, in that case.

JojoM-sf:
Didn't know that the size of sf files is reduced so much? Why is that?

ppigazzini (approved):
Workers on linux, windows subsystem for linux, windows msys2, tested successfully:
 - build small stockfish binary
 - keep only 10 nets

vondele:
50*300K + 10 * 20M + book << 1G

vondele:
> Didn't know that the size of sf files is reduced so much? Why is that?
Disabled embedding on fishtest.

JojoM-sf:
My quota is 7813M ... So for seven workers maximum is somewhat less than 1G ... if its really somewhat smaller than 1G, no problem.

ppigazzini:
Stockfish binary size w/o embedded net:
 - 300K - linux
 - 1.3M - windows

JojoM-sf:
In my case, it's Linux. So << will indeed mean "considerably less than", something like 250M per worker. Thanks very much. By the way, I have been starting to try and learn some Python and at least understand the worker code. (See @Pasquale's "wishful thinking" remark in #809). Very difficult for me, but maybe some day I will be able to help you.

ppigazzini:
@JojoM-sf check my script to setup a python development environment on Ubuntu/Windows Subsystem for Linux: `pyven`, `pipx` and `jupyterlab`. 
https://gist.github.com/ppigazzini/73c2dd665116f0efddbf0bc85d6ef309
Inspired by https://jacobian.org/2019/nov/11/python-environment-2020/

Please mind that at the start I setup `vim` as default editor :)
[Learn vim For the Last Time: A Tutorial and Primer](https://danielmiessler.com/study/vim/)

JojoM-sf:
@ppigazzini Thanks very much indeed. As my usual working machine is an Apple laptop, could you also recommend a beginner's set-up for OSX?

tomtor:
> @ppigazzini Thanks very much indeed. As my usual working machine is an Apple laptop, could you also recommend a beginner's set-up for OSX?

@JojoM-sf vscode is very nice: https://code.visualstudio.com/

JojoM-sf:
@tomtor Thx ...

ppigazzini:
@JojoM-sf simply follow the instruction for OSX :)
- pyenv (with Homebrew and Xcode command line tools)
  https://github.com/pyenv/pyenv#homebrew-on-macos
  ```bash
  brew update
  brew install pyenv
  ```
  https://github.com/pyenv/pyenv/wiki#suggested-build-environment
  ```bash
  xcode-select --install
  # optional, but recommended:
  brew install openssl readline sqlite3 xz zlib
  ```
  Then install python 3.8.6 and set it as global (in this way you will be sure to keep safe the system python):
  ```
  pyenv install 3.8.6
  pyenv global 3.8.6
  ```

- pipx
  https://github.com/pipxproject/pipx
  ```bash
  brew install pipx
  pipx ensurepath
  ```
  Then install some useful python cli tools with pipx:
  ```
  # format python code, use this for fishtest
  # black --exclude=worker/requests .
  pipx install black
  # sort python imports, use this for fishtest
  # isort --profile black --skip worker/requests .
  pipx install isort
  # lint python code
  pipx install flake8
  ```

JojoM-sf:
So, xcode as an alternative. Thanks, Pasquale.

JojoM-sf:
Another question about the folders of the form `_testing_<10digits>.<6digits>`, also to be found in increasing numbers in `/fishtest/worker`: Do they get automatically cleared, or would I need to keep track of this?

ppigazzini:
@JojoM-sf we keep 3 backups, the user could have a custom `cutechess-cli`
https://github.com/glinscott/fishtest/blob/8d14a5b835da33aa2eddb31543ffbb4c44ebadbf/worker/updater.py#L72-L85

JojoM-sf:
Thx very much, with snippets like this I will be able to understand the code step by step.

ppigazzini:
Thank you @vondele :)
Also merged https://github.com/glinscott/fishtest/pull/839 to save disk space.

PR #834, opened 2020-10-24
Rename fishtest server code dir

tomtor:
@ppigazzini This fixes a ton of problems for Vscode IDE and some mentals problems for me with the old nested `fishtest/fishtest` directory naming.

PR #836, opened 2020-10-24
Delay scavenge on boot

tomtor:
See https://github.com/glinscott/fishtest/pull/832#issuecomment-715900620

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -25,6 +25,8 @@ from fishtest.util import (
     remaining_hours,
 )
 
+boot_time = datetime.utcnow()
+
 last_rundb = None
 
 
@@ -327,6 +329,8 @@ class RunDb:
         self.run_cache_lock.release()
 
     def scavenge(self, run):
+        if datetime.utcnow() < boot_time + timedelta(seconds=150):
+            return False
         # print("scavenge ", run["_id"])
         dead_task = False
         old = datetime.utcnow() - timedelta(minutes=3)


ppigazzini:
PR applied by hand to PROD.
I raised the scavenge timeout to 37 minutes, took PROD offline for 7 minutes to apply  #834, several workers were still lost in space after 18 minutes since the server restart.

image

Now I lowered the scavenge timeout to 3 minutes and there are only 54 workers.

JojoM-sf:
Error message looked like this:

`Exception from calling update_task:
Expecting value: line 1 column 1 (char 0)
Exception from calling heartbeat:
Expecting value: line 1 column 1 (char 0)`

I had to ctrl-C and manually log back on.

ppigazzini:
Scavenge timeout raised to 15 minutes, only 57 workers.

ppigazzini (approved):

JojoM-sf:
Noob's workers are gone, it would seem :(

ppigazzini:
Raising scavenge timeout to 15 minutes not helped.

@noobpwnftw please report the message error from your workers.

ppigazzini:
@tomtor Merged, reinstalled the server with `x/update_fishtest.sh`, raised the scavenge timeout to 15 minutes by hand.

ppigazzini:
@tomtor scavenge timeout lowered to default 3 minutes, we have 178 workers now.

vondele:
BTW, there was some comment by @noobpwnftw  in discord yesterday: 
image
not quite sure if that made it to one of the issues here.

tomtor:
@vondele @ppigazzini @noobpwnftw Ah, `sys.exit()` must be preceeded by `ALIVE = False`.

Will make a PR for that

tomtor:
@ppigazzini There are 0 "dead tasks" removed by the currently running pserv with 3 minutes timeout, which is good.

Workers which are shown with longer timeouts are probably just waiting for a task, the heartbeat code on the server ignores these beats by design.

tomtor:
> Workers which are shown with longer timeouts are probably just waiting for a task

Hmm, after inspecting the code I do not think this is the cause. Anyway, currently no machines >1 minute idle, so that's ok.

tomtor:
@ppigazzini Currently an issue with idle times but a Python MongoDb error in the log...

tomtor:
@ppigazzini after a restart of mongo and pserv correct times.

tomtor:
> Currently an issue with idle times but a Python MongoDb error in the log...

@ppigazzini Should be fixed with #844

PR #838, opened 2020-10-25
Replace sys.exit() by os._exit() to stop heartbeat

tomtor:
See https://github.com/glinscott/fishtest/pull/836#issuecomment-716122229

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -90,6 +90,9 @@ def on_sigint(signal, frame):
     ALIVE = False
     raise Exception("Terminated by signal")
 
+def worker_exit():
+    os._exit(1)
+
 
 rate = None
 
@@ -162,7 +165,7 @@ def worker(worker_info, password, remote):
         if "version" not in req:
             print("Incorrect username/password")
             time.sleep(5)
-            sys.exit(1)
+            worker_exit()
 
         if req["version"] > WORKER_VERSION:
             print("Updating worker version to {}".format(req["version"]))
@@ -321,7 +324,7 @@ def main():
             args.extend([username, password])
         else:
             sys.stderr.write("{} [username] [password]\n".format(sys.argv[0]))
-            sys.exit(1)
+            worker_exit()
 
     # Write command line parameters to the config file
     config.set("login", "username", args[0])
@@ -338,7 +341,7 @@ def main():
     protocol = options.protocol.lower()
     if protocol not in ["http", "https"]:
         sys.stderr.write("Wrong protocol, use https or http\n")
-        sys.exit(1)
+        worker_exit()
     elif protocol == "http" and options.port == "443":
         # Rewrite old port 443 to 80
         options.port = "80"
@@ -355,7 +358,7 @@ def main():
 
     if cpu_count <= 0:
         sys.stderr.write("Not enough CPUs to run fishtest (it requires at least two)\n")
-        sys.exit(1)
+        worker_exit()
 
     uname = platform.uname()
     worker_info = {
--- a/worker/games.py
+++ b/worker/games.py
@@ -456,7 +456,8 @@ def adjust_tc(tc, base_nps, concurrency):
         sys.stderr.write(
             "This machine is too slow to run fishtest effectively - sorry!\n"
         )
-        sys.exit(1)
+        from worker import worker_exit
+        worker_exit()
 
     # Parse the time control in cutechess format
     chunks = tc.split("+")


tomtor:
@ppigazzini Could be combined with #835

ppigazzini:
@tomtor "combined" in the means of "merged with the same version raise" I suppose.
I'm testing locally another worker patch to be merged in the same version raise.

tomtor:
> "combined" in the means of "merged with the same version raise" I suppose.

Yes, that was what I meant :-)

ppigazzini:
@tomtor You have just deleted `ALIVE = False` from `worker_exit()`, is it unnecessary?

tomtor:
> @tomtor You have just deleted `ALIVE = False` from `worker_exit()`, is it unnecessary?

Yes, I believe it is, `os_exit()` will exit immediately

worker/worker.py
```diff
@@ -162,7 +165,7 @@ def worker(worker_info, password, remote):
         if "version" not in req:
             print("Incorrect username/password")
             time.sleep(5)
-            sys.exit(1)
+            worker_exit()
```
ppigazzini, line 168: Change while here :)
- `print("Incorrect username/password")` replaced by `sys.stderr.write("Incorrect username/password\n")`

ppigazzini (approved):
@tomtor tested working fine with linux/WSL/msys2 workers

worker/worker.py
ppigazzini: @tomtor Ignore my suggestion, after a worker code review we should make a dedicated PR to have more consistency in the output.

ppigazzini:
Thank you @tomtor :)
@noobpwnftw please updated the fleet at convenient time :)

noobpwnftw:
I have a feature request, so that I don't have to local edit every time and have conflicts.

My fishtest workers are invoked by an external scheduler, which looks for error code to decide what to do.
Whenever the machine is idle, it launches fishtest worker and expects it to finish one job.
If the process exits with code 0, it relaunches the same task, if the process exits with code 1, it schedules some other task.
Also removes all delay sleeps as the machine can do something else and try again later, helpful when there is some network issue with github or empty queue or server down.

Can this be implemented by a flag in the fishtest worker code?

vondele:
@noobpwnftw I think so, do you have a diff of the current implementation to see what it involves?

noobpwnftw:
`diff --git a/worker/worker.py b/worker/worker.py
index e40568f..7abc6b2 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -189,8 +189,7 @@ def worker(worker_info, password, remote):
         sys.stderr.write("Exception accessing host:\n")
         print(e, file=sys.stderr)
         #    traceback.print_exc()
-        time.sleep(random.randint(10, 60))
-        return
+        worker_exit()

     print("Task requested in {}s".format((datetime.utcnow() - t0).total_seconds()))
     if "error" in req:
@@ -200,8 +199,7 @@ def worker(worker_info, password, remote):
     if "task_waiting" in req:
         print("No tasks available at this time, waiting...\n")
         # Note that after this sleep we have another ALIVE HTTP_TIMEOUT...
-        time.sleep(random.randint(1, 10))
-        return
+        worker_exit()

     success = True
     global RUN, TASK_ID
@@ -352,7 +350,7 @@ def main():
     print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))

     try:
-        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count() - 1)
+        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count())
     except:
         cpu_count = int(options.concurrency)

@@ -383,7 +381,7 @@ def main():
             print("Stopped by 'fish.exit' file")
             break
         if not success:
-            time.sleep(HTTP_TIMEOUT)
+            worker_exit()
         success = worker(worker_info, args[1], remote)
`

noobpwnftw:
```
index e40568f..7abc6b2 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -189,8 +189,7 @@ def worker(worker_info, password, remote):
         sys.stderr.write("Exception accessing host:\n")
         print(e, file=sys.stderr)
         #    traceback.print_exc()
-        time.sleep(random.randint(10, 60))
-        return
+        worker_exit()

     print("Task requested in {}s".format((datetime.utcnow() - t0).total_seconds()))
     if "error" in req:
@@ -200,8 +199,7 @@ def worker(worker_info, password, remote):
     if "task_waiting" in req:
         print("No tasks available at this time, waiting...\n")
         # Note that after this sleep we have another ALIVE HTTP_TIMEOUT...
-        time.sleep(random.randint(1, 10))
-        return
+        worker_exit()

     success = True
     global RUN, TASK_ID
@@ -352,7 +350,7 @@ def main():
     print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))

     try:
-        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count() - 1)
+        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count())
     except:
         cpu_count = int(options.concurrency)

@@ -383,7 +381,7 @@ def main():
             print("Stopped by 'fish.exit' file")
             break
         if not success:
-            time.sleep(HTTP_TIMEOUT)
+            worker_exit()
         success = worker(worker_info, args[1], remote)

```

tomtor:
@noobpwnftw I will create a PR for this.

Does anyone have a suggestion how to name this flag?

noobpwnftw:
Well, the idea is that the worker becomes a "slice" runner for a certain task, does not hog around when there is no job, using all available resources and report if the job is successfully run which helps the parent to decide whether it should run more often.

So, I think something like "--slice"?

noobpwnftw:
Another thing, as for currently we cannot specify different NNUE net other than what is written in code, and Makefile downloads the net, there is no need for the worker to download the net and cache them, they are never used.

vondele:
"--slice" or "--fleet" or just "--noob" ;-) ? 

not sure I understand the net comment, the cache is being used, i.e. the Makefile only downloads if the net is not present in the src directory, and the logic in the worker script should copy the right net from the cache in the build directory. So, the net is downloaded only once. Also, since we disabled embedding in fishtest recently, that's where the binaries look for the net. Maybe something is not right in the worker code if you see different things happening?

tomtor:
> does not hog around when there is no job, using all available resources

@noobpwnftw While in `sleep()` it will not use CPU resources, but it will consume some memory resources.

I wonder in what cases another worker task will be able to get some work when original worker code decided to sleep().
The "no task" and "host not accessible" will apply to another worker invocation as well?

noobpwnftw:
@vondele if that is already fixed then you can ignore the net part, I had the impression that the worker will download the net but the makefile will download it again, and cache the compiles with embedded net.

@tomtor the comma separates the case entirely, I mean 1) does not hog when no task, 2) when there is task, use all cores(see the `concurrency - 1`  part).
By other tasks I mean something other than fishtest, for example, in an HPC environment, you have a supervising task manager which will launch your desired task on a fixed set of VM instances whenever they are free, currently I have to spawn thousands of new VM instances and touch the exit file to have them decommissioned, then remove the exit file and launch the worker process again. This is not actually trivial, I don't know if the job has finished or not, or whether it started successfully, there can be more granular control over this.

tomtor:
@noobpwnftw Thanks for your answer, it's clear to me now.

PR #842, opened 2020-10-25
Fix test_server.sh to use new server directory

linrock:
Updating this script since the server directory was renamed from `fishtest` to `server`


ppigazzini (approved):

ppigazzini:
I missed badly this when inspecting the files :(
Thank you @linrock :)

PR #844, opened 2020-10-26
Make flush buffers more robust

tomtor:


server/fishtest/rundb.py
```diff
@@ -288,45 +288,45 @@ def flush_all(self):
...
+                self.run_cache[oldest]["dirty"] = False
+                self.run_cache[oldest]["ftime"] = time.time()
+                # print("SYNC")
+                with self.run_cache_write_lock:
+                    self.runs.replace_one(
+                        {"_id": ObjectId(oldest)}, self.run_cache[oldest]["run"]
+                    )
+        except:
+            print("Flush exception")
+        finally:
```
ppigazzini, line 326: DEV updated.
If I'm not wrong the PR, with this `finally:` section, simply releases the lock and starts the timer also after an exception. This looks good to me.

tomtor: Yes, that's one part of the PR. In addition I removed the writing without lock which I added some months ago to gain a potential speedup under heavy load. That is the root cause for the rare exception which I have seen twice (yesterday and a few weeks ago).

ppigazzini: I missed the `self.run_cache_lock.release()  # Release the lock while writing`.

tomtor: It might be that we need to grab the lock also on an heartbeat update, but I do not think that is needed. Let's wait to see if the log gets any "flush exception" messages in the future.

PR #845, opened 2020-10-26
Add --fleet/--use_all_cores worker option

tomtor:
See discussion at https://github.com/glinscott/fishtest/pull/838

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -38,6 +38,8 @@ WORKER_VERSION = 91
 ALIVE = True
 HTTP_TIMEOUT = 15.0
 
+FLEET = False
+
 
 def setup_config_file(config_file):
     """ Config file setup, adds defaults if not existing """
@@ -72,6 +74,8 @@ def setup_config_file(config_file):
         ("parameters", "concurrency", "3"),
         ("parameters", "max_memory", str(int(mem / 2 / 1024 / 1024))),
         ("parameters", "min_threads", "1"),
+        ("parameters", "fleet", "False"),
+        ("parameters", "use_all_cores", "False"),
     ]
 
     for v in defaults:
@@ -90,8 +94,9 @@ def on_sigint(signal, frame):
     ALIVE = False
     raise Exception("Terminated by signal")
 
-def worker_exit():
-    os._exit(1)
+
+def worker_exit(val=1):
+    os._exit(val)
 
 
 rate = None
@@ -116,6 +121,7 @@ def get_rate():
 RUN = None
 TASK_ID = None
 
+
 def heartbeat(worker_info, password, remote):
     global ALIVE, RUN, TASK_ID
     print("Start heartbeat")
@@ -126,7 +132,7 @@ def heartbeat(worker_info, password, remote):
         count += 1
         if count == 60:
             count = 0
-            print("Send heartbeat... ", end='')
+            print("Send heartbeat... ", end="")
             payload["run_id"] = str(RUN["_id"]) if RUN else None
             payload["task_id"] = TASK_ID
             req = requests.post(
@@ -144,7 +150,7 @@ def heartbeat(worker_info, password, remote):
 
 
 def worker(worker_info, password, remote):
-    global ALIVE
+    global ALIVE, FLEET
 
     payload = {"worker_info": worker_info, "password": password}
 
@@ -189,6 +195,8 @@ def worker(worker_info, password, remote):
         sys.stderr.write("Exception accessing host:\n")
         print(e, file=sys.stderr)
         #    traceback.print_exc()
+        if FLEET:
+            worker_exit()
         time.sleep(random.randint(10, 60))
         return
 
@@ -199,6 +207,8 @@ def worker(worker_info, password, remote):
     # No tasks ready for us yet, just wait...
     if "task_waiting" in req:
         print("No tasks available at this time, waiting...\n")
+        if FLEET:
+            worker_exit()
         # Note that after this sleep we have another ALIVE HTTP_TIMEOUT...
         time.sleep(random.randint(1, 10))
         return
@@ -314,6 +324,24 @@ def main():
         dest="min_threads",
         default=config.get("parameters", "min_threads"),
     )
+    parser.add_option(
+        "-f",
+        "--fleet",
+        dest="fleet",
+        default=config.get("parameters", "fleet"),
+    )
+    parser.add_option(
+        "-a",
+        "--use_all_cores",
+        dest="use_all_cores",
+        default=config.get("parameters", "use_all_cores"),
+    )
+    parser.add_option(
+        "-w",
+        "--only_config",
+        dest="only_config",
+        default=False,
+    )
     (options, args) = parser.parse_args()
 
     if len(args) != 2:
@@ -335,8 +363,11 @@ def main():
     config.set("parameters", "concurrency", options.concurrency)
     config.set("parameters", "max_memory", options.max_memory)
     config.set("parameters", "min_threads", options.min_threads)
-    with open(config_file, "w") as f:
-        config.write(f)
+    config.set("parameters", "fleet", options.fleet)
+    config.set("parameters", "use_all_cores", options.use_all_cores)
+
+    global FLEET
+    FLEET = options.fleet == "True"
 
     protocol = options.protocol.lower()
     if protocol not in ["http", "https"]:
@@ -349,16 +380,40 @@ def main():
         # Rewrite old port 80 to 443
         options.port = "443"
     remote = "{}://{}:{}".format(protocol, options.host, options.port)
-    print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))
 
     try:
-        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count() - 1)
-    except:
+        if options.use_all_cores == "True":
+            cpu_count = multiprocessing.cpu_count()
+        else:
+            cpu_count = int(options.concurrency)
+            if cpu_count > multiprocessing.cpu_count() - 1:
+                print(
+                    (
+                        "\nYou cannot have concurrency {} but at most:\n"
+                        "{} with option --concurrency\n"
+                        "{} with option --use_all_cores\n"
+                    ).format(
+                        options.concurrency,
+                        multiprocessing.cpu_count() - 1,
+                        multiprocessing.cpu_count(),
+                    ),
+                    file=sys.stderr,
+                )
+                worker_exit()
+    except Exception as e:
+        print(e, file=sys.stderr)
         cpu_count = int(options.concurrency)
 
     if cpu_count <= 0:
         sys.stderr.write("Not enough CPUs to run fishtest (it requires at least two)\n")
         worker_exit()
+
+    with open(config_file, "w") as f:
+        config.write(f)
+    if options.only_config == "True":
+        worker_exit(0)
+
+    print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))
 
     uname = platform.uname()
     worker_info = {
@@ -383,6 +438,8 @@ def main():
             print("Stopped by 'fish.exit' file")
             break
         if not success:
+            if FLEET:
+                worker_exit()
             time.sleep(HTTP_TIMEOUT)
         success = worker(worker_info, args[1], remote)
 


tomtor:
@vondele @ppigazzini I wonder if we would have to separate the `cpu_count` check in a separate option because many users with high core CPUs are likely to currently patch the fishtest worker to remove that check.

vondele:
I don't think many users patch the code, but we could have an `--use-all-cores` flag. Do we document options somewhere, since this should mention that only dedicated machines should use this flag, IMO.

tomtor:
`--use-all-cores` could be limited to machines >= 8 cores, to limit wrong usage somewhat.

ppigazzini:
> I don't think many users patch the code, but we could have an `--use-all-cores` flag. Do we document options somewhere, since this should mention that only dedicated machines should use this flag, IMO.

https://github.com/glinscott/fishtest/wiki/Running-the-worker%3A-overview#worker-parameters

vondele:
so, we do have a check in the code that 'silently' overrides the --concurrency option? So probably, the right thing to do is to stop if concurrency is larger than what we allow for, and have this flag --use-all-cores disable this check?

ppigazzini:
> `--use-all-cores` could be limited to machines >= 8 cores, to limit wrong usage somewhat.

`multiprocessing.cpu_count()` is 8 for a quad core + HT.

If we trust the user we should simply document `--use-all-cores`

vondele:
I agree that just documenting the flag is good enough. Wrong usage will be limited. Worst case, this user will see a lot of timeouts for his worker.

ppigazzini:
> @vondele @ppigazzini I wonder if we would have to separate the `cpu_count` check in a separate option because many users with high core CPUs are likely to currently patch the fishtest worker to remove that check.

@tomtor Do you patch your workers? :)

tomtor:
> > @vondele @ppigazzini I wonder if we would have to separate the `cpu_count` check in a separate option because many users with high core CPUs are likely to currently patch the fishtest worker to remove that check.
> 
> @tomtor Do you patch your workers? :)

No, I have switched from 30 threads for one worker to 2 workers with each 16 threads ;-)

ppigazzini:
> Worst case, this user will see a lot of timeouts for his worker.

I'm missing the root of the timeouts. If I'm not wrong we are talking about a `--use-all-cores` independents from `--fleet`.
In the the contrary case the worker at worst will exit.

tomtor:
> In the the contrary case the worker at worst will exit

The worker will exit when no task is available, that is not the desired behavior for users who have a dedicated machine for running fishtest workers.

vondele:
> > Worst case, this user will see a lot of timeouts for his worker.
> 
> I'm missing the root of the timeouts. If I'm not wrong we are talking about a `--use-all-cores` independents from `--fleet`.
> In the the contrary case the worker at worst will exit.

I'm talking about timeouts in games played, assuming we want 1 thread idle on a typical machine to give some space to OS and normal user activity. If the user forces 'full occupation' by --use_all_cores and he's not running on an idle machine, I expect we will see some losses on time.

ppigazzini:
@tomtor @vondele In the workers setup scripts I'm currently abusing the `fish.exit` to simply write/update the `fishtest.cfg`.
What about adding a `--write-config`?

ppigazzini:
> If the user forces 'full occupation' by --use_all_cores and he's not running on an idle machine, I expect we will see some losses on time.

Fast test on DEV with a worker@8cores (quad+HT):
- tc=1+0.01 - 2 losses in 1200 games
- tc=0.5+0.005 - 0 losses in 1200 games

noobpwnftw:
I have never come across any timeouts due to occupying all cores for years, at high concurrency count, cutechess spend too much time parsing UCI output causing timeouts, this is another matter.

OTOH, we should give user full control over how many threads he wish to run, capping at available cores, not available cores -1.

tomtor:
> so, we do have a check in the code that 'silently' overrides the --concurrency option? So probably, the right thing to do is to stop if concurrency is larger than what we allow for, and have this flag --use-all-cores disable this check?

@vondele I did implement this in the current PR, the old behavior is commented out for now.

worker/worker.py
```diff
@@ -352,7 +389,17 @@ def main():
     print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))
 
     try:
-        cpu_count = min(int(options.concurrency), multiprocessing.cpu_count() - 1)
+        if options.use_all_cores == "True":
+            cpu_count = multiprocessing.cpu_count()
+        else:
+            cpu_count = int(options.concurrency)
+            if cpu_count > multiprocessing.cpu_count() - 1:
+                sys.stderr.write("You cannot have concurrency {} but at most {}\n", options.concurrency, multiprocessing.cpu_count() - 1)
```
vondele, line 397: maybe refer to the other option ?
"You cannot have concurrency {} but at most {}, override this check with the option --use_all_cores\n"

tomtor:
@ppigazzini `--write_config` value is written in ``fishtest.cfg`, but that that should not happen.

It's not clear how to prevent that, I would expect that the current code would not write that option.

worker/worker.py
```diff
@@ -76,6 +76,7 @@ def setup_config_file(config_file):
         ("parameters", "min_threads", "1"),
         ("parameters", "fleet", "False"),
         ("parameters", "use_all_cores", "False"),
+        ("parameters", "write_config", "False"),
```
ppigazzini, line 79: It' seems to work commenting out these two lines
`# ("parameters", "write_config", "False"),`

worker/worker.py
```diff
@@ -336,6 +337,12 @@ def main():
         dest="use_all_cores",
         default=config.get("parameters", "use_all_cores"),
     )
+    parser.add_option(
+        "-w",
+        "--write_config",
+        dest="write_config",
+        default=config.get("parameters", "write_config"),
```
ppigazzini, line 344: `# default=config.get("parameters", "write_config"),`

ppigazzini: Tonight I will have more time to try/test.

ppigazzini: @tomtor `optparse` is deprecated, suggested `argparse` (but this could break python2), see https://docs.python.org/3/library/optparse.html

This example use `argparse` and `configparse`
http://www.jtrive.com/developing-command-line-applications-with-argparse-and-configparser.html

worker/worker.py
tomtor: @ppigazzini Thanks, I overlooked step 2. Should have figured that out myself :-/

worker/worker.py
tomtor: Fixed

worker/worker.py
tomtor: Done, thanks.

worker/worker.py
tomtor: @ppigazzini We still have some Python 2 workers....

ppigazzini: @tomtor you sometimes uses a python 2 worker :)

tomtor: Yes, by accident...

ppigazzini: Python 2 has `argparse` but we should learn the hard way the usage differences wrt Python 3: too much pain :)
https://docs.python.org/2/howto/argparse.html

ppigazzini:
New options work fine.

Some notes:
- `--write_config` is a bit confusing and should be renamed `--only_config`
- `--only_config=True` should exit with 0 (useful in bash scripting)
- `worker_exit()` now only call `os._exit(1)`, so could be replaced by `os._exit(0|1)`
- the concurrency message is a bit cryptic about the "override", IMO better something like this:
```
                    (
                        "\nYou cannot have concurrency {} but at most:\n"
                        "{} with option --concurrency\n"
                        "{} with option --use_all_cores\n"
                    ).format(
                    options.concurrency,
                    multiprocessing.cpu_count() - 1,
                    multiprocessing.cpu_count()),
                    file=sys.stderr
                )
```
- we could reorder the code to check the user arguments before to write `fishtest.cfg`. Useful for the worker setup script:
```
    if cpu_count <= 0:
        sys.stderr.write("Not enough CPUs to run fishtest (it requires at least two)\n")
        os._exit(1)

    with open(config_file, "w") as f:
        config.write(f)
    if options.write_config == "True":
        os._exit(0)

    print("Worker version {} connecting to {}".format(WORKER_VERSION, remote))

    uname = platform.uname()
    worker_info = {
        "uname": uname[0] + " " + uname[2],
        "architecture": platform.architecture(),
```

tomtor:
> * `worker_exit()` now only call `os._exit(1)`, so could be replaced by `os._exit(0|1)`

main will exit with `0` so we should use `1` to see the difference?

worker/worker.py
```diff
@@ -314,6 +324,24 @@ def main():
...
+    )
+    parser.add_option(
+        "-a",
+        "--use_all_cores",
+        dest="use_all_cores",
+        default=config.get("parameters", "use_all_cores"),
+    )
+    parser.add_option(
+        "-w",
+        "--only_onfig",
```
ppigazzini, line 341: Typo, should be ` "--only_config",`

tomtor: Oops, fixed

ppigazzini:
> main will exit with `0` so we should use `1` to see the difference?

Many Unix programs exit with `0` for correct execution, `n!=0` for error in execution.
https://docstore.mik.ua/orelly/unix3/upt/ch35_12.htm

Useful to use in sh scripting:
`bash test.sh && (echo $?; echo "^^^ exit value. OK" ) || (echo $?; echo "^^^ exit value. KO")`

I could write something like these for the worker setup scripts:
- asking the user to restart the setup script:
```
python3 worker.py ${usr_name} ${usr_pwd} --concurrency=${usr_concurrency} --only_config=True || echo "Restart the script using a proper concurrency value."
```
- using a default value:
```
python3 worker.py ${usr_name} ${usr_pwd} --concurrency=${usr_concurrency} --only_config=True || (echo "Reverting to default --concurrency=2. Restart the script to set a new proper concurrency value."; python3 worker.py ${usr_name} ${usr_pwd} --concurrency=2 --only_config=True)
```

tomtor:
@ppigazzini I did not understand your remark about `_exit(0|1)` and I expect you didn't understand my reply? :-)

Anyway, is the code in the current PR ok?

ppigazzini:
@tomtor IMO `--only_config` should followed by `os._exit(0)` because in that case the execution is correct. The other exit cases are errors, so it's correct to use `os._exit(1)`

ppigazzini:
> Anyway, is the code in the current PR ok?

I can thoroughly test the PR with the 3 different OS this night :)

tomtor:
@ppigazzini Changed exit value for `--only_config`

tomtor:
> could you test this script on your workstation?

Sure, Ubuntu 18.04.5 LTS:

```
CPU:  AMD Ryzen 7 1700X Eight-Core Processor
Processors: 1
Physical cores:  8
Sibling cores:  16
Total physical cores: 8
Total sibling cores: 16
On line cores: 16
```

and Ubuntu 20.04.1 LTS

```
CPU:  AMD Ryzen 9 3950X 16-Core Processor
Processors: 1
Physical cores:  16
Sibling cores:  32
Total physical cores: 16
Total sibling cores: 32
On line cores: 32
```

ppigazzini:
Thanks, works fine also for AMD CPUs.

ppigazzini:
@tomtor LGTM.
Opened #848 to fix a bug in the old username/password reading code.

ppigazzini (approved):

ppigazzini:
@noobpwnftw update the fleet at your convenience :)

noobpwnftw:
OK, so I should add `--fleet --use_all_cores` to the command line?

vondele:
yes

noobpwnftw:
Do I need `--only_config` and what does it do?

vondele:
I don't think you need that option, IIRC this would use write the config file. @ppigazzini ?

noobpwnftw:
Updated.

noobpwnftw:
Updated.

noobpwnftw:
Updated.

tomtor:
> OK, so I should add `--fleet --use_all_cores` to the command line?

Yes, `--fleet=True --use_all_cores=True`

The config option just saves the parameters in the `fishtest.cfg` file and exits.

noobpwnftw:
Updated. I have no more local edits, if this works, from now on you can push worker updates without my manual intervention.

ppigazzini:
PROD updated, triggered workers update.
Updated wiki with: documentation of the new worker options; linux worker setup script printing the whole CPU information and using `--only_config` option (instead of the `fish.exit` trick)
Updated the windows setup script printing the whole CPU information and using `--only_config` option

In both setup scripts I trust the user for the `concurrency` value, after giving the whole CPU information and the suggested max value.

```
# print CPU information
cpu_model=$(grep "^model name" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
n_cpus=$( grep "^physical id" /proc/cpuinfo | sort | uniq | wc -l)
online_cores=$(grep "^bogo" /proc/cpuinfo | wc -l)
n_siblings=$(grep "^siblings" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
n_cpu_cores=$(grep "^cpu cores" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
total_siblings=$((${n_cpus} * ${n_siblings}))
total_cpu_cores=$((${n_cpus} * ${n_cpu_cores}))
printf "CPU model : ${cpu_model}\n"
printf "CPU       : %3d  -  Online cores    : %3d\n" ${n_cpus} ${online_cores}
printf "Siblings  : %3d  -  Total siblings  : %3d\n" ${n_siblings} ${total_siblings}
printf "CPU cores : %3d  -  Total CPU cores : %3d\n" ${n_cpu_cores} ${total_cpu_cores}

# read the fishtest credentials and the number of cores to be contributed
echo
echo "Write your fishtest username:"
read usr_name
echo "Write your fishtest password:"
read usr_pwd
echo "Write the number of cores to be contributed to fishtest:"
echo "(max suggested 'Total CPU cores - 1')"
read n_cores
```

Thank you @tomtor :)

PR #848, opened 2020-10-28
Fix logic for cli positional arguments

ppigazzini:
Use a proper logic for the positional arguments ("username", "password"):
- two positional arguments, use the cli data
- zero posizional arguments, use non zero data from config file
- others, write error message and exit

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -34,7 +34,7 @@ from os import path
 from games import run_games
 from updater import update
 
-WORKER_VERSION = 91
+WORKER_VERSION = 92
 ALIVE = True
 HTTP_TIMEOUT = 15.0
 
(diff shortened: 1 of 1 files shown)


worker/worker.py
```diff
@@ -316,19 +316,20 @@ def main():
...
-        else:
-            sys.stderr.write("{} [username] [password]\n".format(sys.argv[0]))
-            worker_exit()
+    username = config.get("login", "username")
+    password = config.get("login", "password", raw=True)
+
+    if len(args) == 2:
+        username = args[0]
+        password = args[1]
+    elif len(username) == 0 or len(password) == 0:
```
ppigazzini, line 325: @tomtor I replicated the old logic to use the config information for `len(args) != 2`, otherwise we can use it only for `len(args) == 0` with `elif len(args) != 0 or len(username) == 0 or len(password) == 0:`

ppigazzini: @tomtor updated to use the latter unambiguous logic.

Issue #851, opened 2020-11-01
'active workers' not doing any test


vondele:
This test has 2 workers which appear active, but don't run any test:

https://tests.stockfishchess.org/tests/view/5f9e7c786a2c112b60691e39
![image](https://user-images.githubusercontent.com/4202567/97808229-03260f80-1c66-11eb-9861-2e2336429e98.png)

their 'active' time is updated and up-to-date, but the test is now over, except for those two workes. 
So probably hung somewhere, but not the heartbeat?

vondele:
I have now stopped and restarted the test to see if that fixes the issue with those two batches.

vondele:
A simple stop/modify didn't help so far. I've stopped it again, and will wait for a bit more before restarting again, to be sure it gets different workers.

JojoM-sf:
Also, Noob's machines will have to be assigned to some other test in the meantime :)

JojoM-sf:
The two respective machines are back, it would seem ... active, but doing any testing.

vondele:
yes, now after a longer wait, these batches are being computed.

JojoM-sf:
Done, and with a good result.
Issue resolved, I think.

vondele:
no issue not quite resolved. We need to understand how these workers ended up in that state.

JojoM-sf:
True. But then maybe @noobpwnftw would have to tell us about the output he got ...

snicolet:
@vondele Could it be that the workers were hit by the bug discussed in https://github.com/official-stockfish/Stockfish/issues/3216 , so that Stockfish would crash on bench for them?

vondele:
no, I don't think so, first, that bug is newer than this issue, and a crashing compile should not lead to this behavior. (Also these are noob workers, so not windows, and stable). Hard to guess what it would be, but possibly something network that doesn't timeout? idk.

tomtor:
I noticed that at least one of the `dew` machines constantly report failures almost directly after receiving/completing their task?

```
Nov  6 15:01:05 tests pserve[1708]: Allocate run: 5fa40e365d72639a7acef79e task_id: 310 to 6c3861c6-370e-4f83-81dd-8705a5ea5224
[...]

PR #853, opened 2020-11-04
Do not scavenge finished runs

### LINKED ISSUE #852: Fishtest logs shows dead tasks on completion of test
Fishtest server logs show regular (every few hours) errors with dead tasks for workers which are running just fine.

tomtor:
Should fix #852

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -329,7 +329,7 @@ class RunDb:
             self.start_timer()
 
     def scavenge(self, run):
-        if datetime.utcnow() < boot_time + timedelta(seconds=150):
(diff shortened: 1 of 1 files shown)


tomtor:
@ppigazzini Applied patch to prod by hand

vondele:
could this also fix https://github.com/glinscott/fishtest/issues/851 ?

tomtor:
> could this also fix #851 ?

No, I do not think that is related. This patch only applies to finished tests.

tomtor:
@ppigazzini Hmm, at 15:07 in the log the same issue, but this time not at the end of test.

tomtor:
Test for finished run was allready in code

PR #858, opened 2020-11-07
Add unique_key for failed_task API call

tomtor:
See https://github.com/glinscott/fishtest/issues/851#issuecomment-723424511


ppigazzini:
@tomtor test workers updated

ppigazzini:
@tomtor PROD updated.

PR #861, opened 2020-11-07
Start using avx512 again.

vondele:
Recent testing shows that avx512 can be up to 10% faster than avx2 on certain chips.

Additionally, this makes sure the avx512 arch will be tested on fishtest, improving coverage of the code.


ppigazzini (approved):

ppigazzini:
Updated wiki script:
https://github.com/glinscott/fishtest/wiki/Building-stockfish-on-Windows#building-stockfish

Issue #859, opened 2020-11-07
do gcc version check in the worker


vondele:
avoid things like https://github.com/glinscott/fishtest/issues/851#issuecomment-723428479 with e.g. worker exiting cleanly with error message in this case.

PR #862, opened 2020-11-07
gcc version check

vondele:
check that gcc is >= 7.3

fixes https://github.com/glinscott/fishtest/issues/859

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -14,8 +14,12 @@ import threading
 import time
 import traceback
 import uuid
+import subprocess
+import platform
 
 import requests
+
+IS_WINDOWS = "windows" in platform.system().lower()
 
 try:
     from ConfigParser import SafeConfigParser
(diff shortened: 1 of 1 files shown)


vondele:
might need a second pair of eyes to decide if this is the right place to put the check, etc.

ppigazzini:
Looks good w/ workers ubuntu, WSL, msys2.

Output for ubuntu 16.04:
```
Worker started in /root/_git/__test_folder00/fishtest/worker ...

Memory: 1949937664
The g++ version must be greater or equal to 7.3, but g++ is version 5.4. A g++ update is required.
```

vondele:
thanks... and the worker quits after that error message, right ?

ppigazzini:
Yes, quit :)

worker/worker.py
```diff
@@ -285,6 +289,48 @@ def worker(worker_info, password, remote):
...
+
+    try:
+        major = int(major)
+        minor = int(minor)
+    except:
+        raise Exception("Failed to parse g++ version.")
+
+    if not (major > 7 or (major == 7 and minor >= 3)):
+        raise Exception(
+            "The g++ version must be greater or equal to 7.3, but g++ is version {}.{}. A g++ update is required.".format(
```
ppigazzini, line 326: I do like something shorter:, e.g.
```
"Found g++ version {}.{}: please update to g++ version 7.3 or greater."
```

worker/worker.py
```diff
@@ -285,6 +289,48 @@ def worker(worker_info, password, remote):
...
+        stdin=subprocess.PIPE,
+        universal_newlines=True,
+        bufsize=1,
+        close_fds=not IS_WINDOWS,
+    )
+
+    p.stdin.close()
+    for line in iter(p.stdout.readline, ""):
+        if "__GNUC__" in line:
+            major = line.split(" ")[2].strip()
```
ppigazzini, line 306: IMO could be simplified in `major = line.split()[2]`
whitespace is the default separator for split() and split() delete all whitespaces
```
>>> a = "   aaa   bbb   ccc   ddd   "
>>> a
'   aaa   bbb   ccc   ddd   '
>>> a.split()
['aaa', 'bbb', 'ccc', 'ddd']
```
split and strip are used when splitting for something different from the space
https://stackoverflow.com/questions/4071396/split-by-comma-and-strip-whitespace-in-python

ppigazzini: The previous split syntax is dangerous :)
```
>>> a = "   aaa   bbb   ccc   ddd   "
>>> a.split(" ")
['', '', '', 'aaa', '', '', 'bbb', '', '', 'ccc', '', '', 'ddd', '', '', '']
```

ppigazzini (approved):
Looks good to me, workers are fine with an up to date gcc (ubuntu, WSL, msys2), the worker exit when:

* no gcc installed:
```
Worker started in /root/_git/__test_folder00/fishtest/worker ...

Memory: 1882824704
[Errno 2] No such file or directory: 'g++'
root@u16:~#
```
* old gcc installed:
```
Worker started in /root/_git/__test_folder00/fishtest/worker ...

Memory: 1834590208
Found g++ version 5.4: please update to g++ version 7.3 or later.
root@u16:~#
```

ppigazzini:
PROD updated, worker update triggered, thank you @vondele :)

PR #866, opened 2020-11-12
Allow finer control on num games tests

vondele:
start from 1000 in steps of 1000. Useful for some very quick tests.


ppigazzini (approved):

ppigazzini:

Issue #872, opened 2020-12-11
Workflow for nets?


vdbergh:
On my local fishtest I tried to schedule a test of master vs master. I got...

Net not in repository: nn-62ef826d1a6d.nnue

Is this supposed to be correct? I thought nets were not part of the repository but stored in a separate location.

vondele:
the local fishtest doesn't provide the api or doesn't have the local database needed for downloading the nets. Probably needs to be synced from fishtest master. The nets themselves are in a separate location.

vdbergh:
The local fishtest was installed with Pasquale's script. It is up to date with master....

ppigazzini:
@vdbergh I always take the lazy way to upload in the local fishtest server:
http://172.23.3.112/upload

the required net downloaded from:
https://tests.stockfishchess.org/nns

If I'm not wrong this action simply write the mandatory info in the db without side effects, @tomtor can help here :)

vdbergh:
The net nn-62ef826d1a6d.nnue is present https://tests.stockfishchess.org/nns .

So the error message I get ("Net not in repository: nn-62ef826d1a6d.nnue") is not correct ...

Am I right? If so I can debug it.

vondele:
your local fishtest needs a database of nets so that your local server can provide the redirect for https://my.local.server/api/nn/nn-018430ac30dd.nnue easiest is to upload that net to your local server and populate that database.

vdbergh:
Ok I get it but this is an annoying extra step if one wants to do a quick test.

Is it necessary? The worker will download the net anyway right?

vondele:
Don't recall where this is coming from exactly, but I think it is the server checking that the net required to build the binary is available. If not, it will fail to build and all workers that start that test will fail. It is a reasonable thing for the server to check.

For a quick check, maybe start a test of SF_11, that should circumvent the check.

vdbergh:
@vondele Thanks. At least I now understand the issue. For now I will upload the net to my local Fishtest. The default net doesn't change all that often so it is not too much work.

@Pasquale Maybe this should be documented a bit better? Then the message could also be more illuminating (i.e. if the server detects it is not the official one, it could include a reference to the FishtestWiki).

vdbergh:
[...]

PR #873, opened 2020-12-11
Hopefully clearer feedback in case of a missing net.

vdbergh:
The message is now like this:

```The net <net name>, used by <branch>, is not known to Fishtest. Please upload it to: <fishtest_url>/upload.```

where <fishtest_url> is determined via request.host_url.

Addresses: #872

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -21,7 +21,6 @@ from fishtest.util import (
     format_results,
 )
 
-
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:
Tested fine on DEV, PROD updated, thank you @vdbergh :)

PR #875, opened 2020-12-20
List SPRT bounds also in "normalized Elo" on the raw stats page.

vdbergh:
Normalized Elo is inversely proportional to the square root of the number of games it takes on average to detect a given strength difference with a given level of significance. In this PR it is given by
```
logistic_elo/(2*standard_deviation_per_game)
```
which in case of a perfectly balanced book and small Elo differences is equal to
```
 logistic_elo/sqrt(1-draw_ratio)
```
In other words if the draw ratio is zero, the book is perfectly balanced, and Elo differences are small then normalized Elo and logistic Elo coincide. We say that zero is the "reference draw ratio". A reference draw ratio different from zero would also work. This seems less intuitive however.

The advantage of expressing SPRT bounds in normalized Elo (which is currently not done) would be that the expected test length no longer depends on the draw ratio or the book (cfr the issue with non NNUE tests which take very long on average).

Normalized Elo is discussed here

http://hardy.uhasselt.be/Fishtest/normalized_elo.pdf

In this PR we use a different normalization factor which is perhaps more intuitive.

Here is a table with conversions for various draw ratios (assuming a perfectly balanced book)

| draw ratio | 0.0 | 0.3 | 0.5 | 0.6 | 0.7 | 0.8 | 0.9 |
|- |------|-----|-----|----|-|-|-|
| Normalized Elo| 5.00 | 5.00 | 5.00| 5.00 | 5.00 |5.00 |5.00|
| Logistic Elo| 5.00| 4.18 | 3.54 | 3.16 | 2.74 | 2.24 | 1.58|

**Note** A `SPRT(0,5)`  with bounds expressed in normalized Elo takes about `42k` games to complete (expected worst case) regardless of the book or the draw ratio. One can verify this with the SPRT calculator by just putting in the above values in logistic Elo for the corresponding draw ratio (and `RMS bias=0`).

### DIFF
--- a/server/fishtest/templates/tests_stats.mak
+++ b/server/fishtest/templates/tests_stats.mak
@@ -1,5 +1,5 @@
 <%
-   import copy
+   import math,copy
    import fishtest.stats.stat_util
    import fishtest.stats.LLRcalc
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
@vdbergh DEV updated, started some tests.

ppigazzini:
@vdbergh logs clean, let me know if I can merge.

vdbergh:
@ppigazzini I checked the pages for old tests and there appear to be no exceptions.
So for me it is ok to merge. I have squashed the PR.

ppigazzini:

PR #877, opened 2020-12-30
Add Actions workflow status badge

ppigazzini:
See the [official documentation](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge)


tomtor:
@ppigazzini Thanks, missed that.

PR #885, opened 2021-03-03
Reformat code with black and isort

ppigazzini:
`black --exclude=worker/requests .`
`isort --profile black --skip worker/requests .`

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -1,6 +1,7 @@
 import os
 import sys
 
+from fishtest.rundb import RunDb
 from pyramid.authentication import AuthTktAuthenticationPolicy
 from pyramid.authorization import ACLAuthorizationPolicy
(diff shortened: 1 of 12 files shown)


ppigazzini:
Draft, two files are not reformatted, error log

```
error: cannot format /home/usr00/_git/fishtest/fishtest/fishtest/builder.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_l593qxap.log

error: cannot format /home/usr00/_git/fishtest/fishtest/fishtest/stat_util.py: INTERNAL ERROR: Black produced code that is not equivalent to the source.  Please report a bug on https://github.com/psf/black/issues.  This diff might be helpful: /tmp/blk_wetpseqt.log

```

PR #886, opened 2021-03-03
Set pyramid to version 1.10

ppigazzini:
With the new pyramid 2.0 some server tests are failing with the message:
```
    from pyramid.session import UnencryptedCookieSessionFactoryConfig
ImportError: cannot import name 'UnencryptedCookieSessionFactoryConfig' from 'pyramid.session' (unknown location)
```

See:
https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -6,7 +6,7 @@ README = ""
 CHANGES = ""
 
 requires = [
-    "pyramid",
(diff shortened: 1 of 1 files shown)


ppigazzini:
PROD updated.

PR #890, opened 2021-03-06
Show complete python version

ppigazzini:
Also raise worker version and trigger workers update


ppigazzini:

PR #892, opened 2021-03-06
Fix "src/" folder search

### LINKED ISSUE #889: Stuck tests
I submitted these two tuning sessions:
https://tests.stockfishchess.org/tests/view/6042f6a3ddcba5f0627bb717
https://tests.stockfishchess.org/tests/view/6042f8eaddcba5f0627bb71a
but even after seeing workers get assigned to them and waiting a long time no tests ever ran.
(Other tuning sessions were running fine at the time.)  It may be that due to the Eigen/SVD library it takes SF a bit longer to start up causing some kind of time out? but that's pure speculation on my part.  I don't know what is preventing these tests from running?  Any help would be appreciated.
Thanks Fisherman

ppigazzini:
Search only in the first subfolder level.
Also raise the workers version and trigger the workers update.

Fix #889

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -10,7 +10,7 @@ from pyramid.httpexceptions import HTTPFound, HTTPUnauthorized, exception_respon
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 
-WORKER_VERSION = 95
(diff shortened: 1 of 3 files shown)


ppigazzini:
Tested fine on DEV, PROD updated.

PR #895, opened 2021-03-22
Use always os.path.join()

ppigazzini:


ppigazzini:
Runs fine with DEV and some workers.

PR #897, opened 2021-03-22
Use short try block

ppigazzini:


ppigazzini:
Runs fine with all workers and DEV.

Issue #865, opened 2020-11-12
SPRT bounds V2


Vizvezdenec:
So, I did new draw rates measurements.
https://tests.stockfishchess.org/tests/view/5fabbba367cbf42301d6a831
https://tests.stockfishchess.org/tests/view/5fabbbb067cbf42301d6a833
https://tests.stockfishchess.org/tests/view/5fabbbb967cbf42301d6a835
https://tests.stockfishchess.org/tests/view/5fabbbd067cbf42301d6a837
New draw rates 
Single core - 80,8% / 91,7%
Multicore - 86,5% / 93,7%.
When we did current bounds we had (implying that we lowered baseline speed from 1,6 mnps to 1,2 mnps so from this https://github.com/glinscott/fishtest/issues/764 topic we should actually take drawrates for 7,5 and 45 seconds respectively) - 
Single core - 76,4% / 88,1%
So we have LTC % of decisive games decreased by 30% effectively - I think it's about time to do somewhat of a revamp of SPRT bounds.
What stats did we have : 
STC:
https://tests.stockfishchess.org/html/SPRTcalculator.html?elo-0=-0.25&elo-1=1.25&draw-ratio=0.764&rms-bias=30
106,3k games peak, 0,5 elo 50% pass rate, 12,3% regression rate
LTC:
116,8k games peak, 0,75 elo 50% pass rate, 1,2% regression rate
With new bounds we should probably have peak and regression rate to be approximately the same with lowering 50% pass rate elo as much as we can.
So my suggestion will be : 
STC : 
{-0,2 ; 1,15} 
https://tests.stockfishchess.org/html/SPRTcalculator.html?elo-0=-0.2&elo-1=1.15&draw-ratio=0.808&rms-bias=30
105,9k games peak, 0,475 elo 50% pass rate, 11,2% regression rate
LTC : 
{0,2 ; 1,0}
https://tests.stockfishchess.org/html/SPRTcalculator.html?elo-0=0.2&elo-1=1.0&draw-ratio=0.917&rms-bias=30
123,6k games peak, 0,6 elo 50% pass rate, 1,2% regression rate
So overall this bounds will take like 5% more time for full LTC to pass but will allow more frequent patches and actually will lower regression passing probability by 10%.

Vizvezdenec:
[...]

PR #901, opened 2021-03-28
New SPRT bounds for tests

snicolet:
New bounds for STC and LTC tests in fishtest.

The main change is that the new bounds will make LTC tests quite easier to pass Elo-wise.
We have to shrink the intervals to keep good precision, so the total cost will be about three
times longer runs in the worst case.

See discussions in https://github.com/glinscott/fishtest/issues/865 and https://discord.com/channels/435943710472011776/813919248455827515/825466226994053170

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -186,10 +186,10 @@
     <div class="flex-row stop_rule sprt">
       <label class="field-label leftmost stop_rule sprt">SPRT bounds</label>
       <select name="bounds" class="stop_rule sprt" style="width: 246px">
-        <option value="standard STC">Standard STC {-0.25, 1.25}</option>
-        <option value="standard LTC">Standard LTC {0.25, 1.25}</option>
-        <option value="regression STC">Non-regression STC {-1.25, 0.25}</option>
-        <option value="regression LTC">Non-regression LTC {-0.75, 0.25}</option>
+        <option value="standard STC">Standard STC {-0.2, 1.1}</option>
+        <option value="standard LTC">Standard LTC {0.2, 0.9}</option>
+        <option value="regression STC">Non-regression STC {-1, 0.2}</option>
+        <option value="regression LTC">Non-regression LTC {-0.7, 0.2}</option>
         <option value="custom" ${is_rerun and 'selected'}>Custom bounds...</option>
       </select>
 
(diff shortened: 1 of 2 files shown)


server/fishtest/templates/tests_run.mak
```diff
@@ -335,10 +335,10 @@
   });
```
vondele, line 336: Maybe add a comment here:
```
## In the future, this should be changed to normalized Elo,
## and converted within fishtest to logistic Elo whenever needed.
## See also https://github.com/glinscott/fishtest/issues/865#issuecomment-808808220
```

vondele:
Thanks, looks good to me!

server/fishtest/templates/tests_run.mak
snicolet: Not so sure how to enter a comment in a .mak file, in fact :-)

Are they html files? javascript files? python files?

snicolet: @ppigazzini @tomtor What is the best way to put comment in a .mak file?

ppigazzini: Mako template inline comment : `##`
https://stackoverflow.com/questions/20906838/markdown-conflict-in-mako

ppigazzini: @vondele wrote the comment the right way :)

snicolet: thanks! :-)

vondele: yes, I actually had a patch prepared that matched exactly yours except for the comments :-)

snicolet:
Pull request updated with the comment

ppigazzini:
@snicolet @vondele [DEV](https://dfts-0.pigazzini.it/) updated, LGTM, please have a double check.

ppigazzini (approved):

vondele:
looks good.

ppigazzini:
PROD updated, thank you @snicolet @vondele :)

vondele:
Thank you for the quick turnaround @ppigazzini  !

vdbergh:
I apologize for letting this issue lapse a little bit. I have the code to make Fishtest use normalized Elo internally (it would just be another Elo model besides logisistic and BayesElo).  

Still, making such changes is a bit stressful for various reasons. Hence the procrastination.

The next two months I will be extremely busy but after that I will implement the change.

vondele:
@vdbergh excellent! Sure, implementing that is a lot of work, and thanks for offering. Please, take the time it needs.

PR #902, opened 2021-03-28
Clean up and update to pymongo 4 "utils" scripts

ppigazzini:


tomtor:
@ppigazzini  `index_pending.py` could be removed, `create_indexes.py` does all that's needed?

ppigazzini:
Yes, it works fine in the server setup script
```bash
# add mongodb indexes
sudo -i -u ${user_name} << 'EOF'
${VENV}/bin/python3 ${HOME}/fishtest/server/utils/create_indexes.py actions flag_cache pgns runs users
EOF
```

server/utils/upgrade.py
```diff
@@ -82,6 +82,89 @@ def show(p):
 
 worker_dict = {}
 
+<<<<<<< HEAD
+=======
```
dav1312, line 86: Merge conflict?

server/utils/upgrade.py
```diff
@@ -82,6 +82,89 @@ def show(p):
...
+
+        # Two other things that changed
+        if "ARCH" not in worker_info:
+            if "ARCH" in task:
+                worker_info["ARCH"] = task["ARCH"]
+                del task["ARCH"]
+            if "nps" in task:
+                worker_info["nps"] = task["nps"]
+                del task["nps"]
+>>>>>>> 1f00e05... Clean up utils scripts and update to pymongo 4.x
```
dav1312, line 167: Same?

PR #906, opened 2021-04-07
Add SPRT bounds for classical tests.

chuckstables:
Implementing the proposed bounds for classical tests:

classical LTC: Elo{ 0.4;1.7} nElo{0.75,3.25} LTC regression rate 0.9%, 160k games peak, 1.05 elo 50% pass, draw rate 0.73416
classical STC: Elo{-0.3,1.6} nElo{-0.5,2.5} STC regression rate 11.8%, 113.5k games peak, 0.65 elo 50% pass, draw rate 0.60113

See https://discord.com/channels/435943710472011776/813919248455827515/829456512841547787 for the relevant discussion. This is being done to bring the classical STC/LTC bounds in line with the current normalized elo bounds. As an aside, this is my first pull request, my apologies if I've done something wrong here.

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -188,6 +188,8 @@
       <select name="bounds" class="stop_rule sprt" style="width: 246px">
         <option value="standard STC">Standard STC {-0.2, 1.1}</option>
         <option value="standard LTC">Standard LTC {0.2, 0.9}</option>
+	<option value="classical STC">Classical STC {-0.3, 1.6}</option>
+	<option value="classical LTC">Classical LTC {0.4, 1.7}</option>
         <option value="regression STC">Non-regression STC {-1, 0.2}</option>
         <option value="regression LTC">Non-regression LTC {-0.7, 0.2}</option>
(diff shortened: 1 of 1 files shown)


chuckstables:
Viz expressed a belief over on discord that we should aim to keep the expected number of games at midpoint at or below 100k given that classical tests shouldn't be taking up too much of the servers computing power. 

An argument in favor of this being that NNUE is around 200 ELO stronger than classical, so the bulk of our resources should be focusing on tests with the NNUE enabled.

An argument against this is that we are planning on implementing normalized elo in fishtest to have bounds automatically adjusted based on draw rates, and that it's too messy to have two different sets of normalized ELO gaining bounds. 

See discord discussion on this topic: https://discord.com/channels/435943710472011776/813919248455827515/829521603339485245

vondele:
this looks good to me. Indeed, I see this as a temporary measure until we switch to nElo internally. Right now this is needed to avoid very long runs with classical Eval.

ppigazzini (approved):
DEV updated, LGTM.

ppigazzini:
PROD updated, thnak you @chuckstables 
By the way I must update the AUTHORS file, let me know if you want also your real name :)

PR #914, opened 2021-04-17
Use normalized Elo input for Fishtest

vdbergh:
Specify SPRT bounds in normalized Elo.
    
The normalized bounds are indicated using `<>` brackets. Currently they are
```    
    standard STC  : <-0.5, 2.5>
    standard LTC  : < 0.5, 3.5>
    regression STC: <-2.5, 0.5>
    regression LTC: <-2.5, 0.5>
```    
leading to a peak expected test duration of `116K` games.
    
Specific changes:
    
Update LLRcalc.py, sprt.py and stat_util.py to understand normalized Elo bounds.
    
Introduce a new Elo model 'normalized' for the sprt object.
    
Clean up the raw stats page. In particular give variables referring to 'sensitivity' a 'nt' (normalized t) prefix. 'Exact LLR' is renamed to 'Multinomial LLR'.
    
Add appropriate bound conversions to the raw stats page.
    
Show normalized LLR's on the raw stats page.
    
Generic bounds formatting ([],{} or <>).

Put normalized bounds in tests_run.mak. Send the Elo model via a hidden field but validate on the server.
    
Change defaults for the SPRT calculator.
    
Fix some cosmetic bugs in the latest SPRT calculator.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -18,6 +18,7 @@ from fishtest.userdb import UserDb
 from fishtest.util import (
     calculate_residuals,
     estimate_game_duration,
+    format_bounds,
     format_results,
     get_worker_key,
     post_in_fishcooking_results,
@@ -420,14 +421,9 @@ class RunDb:
                 if "sprt" in run["args"]:
                     sprt = run["args"]["sprt"]
                     elo_model = sprt.get("elo_model", "BayesElo")
-                    if elo_model == "BayesElo":
-                        run["results_info"]["info"].append(
-                            ("[%.2f,%.2f]") % (sprt["elo0"], sprt["elo1"])
-                        )
-                    else:
-                        run["results_info"]["info"].append(
-                            ("{%.2f,%.2f}") % (sprt["elo0"], sprt["elo1"])
-                        )
+                    run["results_info"]["info"].append(
+                        format_bounds(elo_model,sprt["elo0"],sprt["elo1"])
+                    )
         return (runs, pending_hours, cores, nps)
 
     def get_finished_runs(
--- a/server/fishtest/static/js/calc.js
+++ b/server/fishtest/static/js/calc.js
@@ -81,21 +81,22 @@ function draw_charts(){
     // pseudo globals
     pass_chart.elo_start=elo_start;
     pass_chart.elo_end=elo_end;
-    pass_chart.N=N;
-    const specials=[0.05,0.95];
+    const specials=[elo0,elo1];
+    var anchors=[];
+    pass_chart.anchors=anchors;
     for (var i=elo_start*N; i<=elo_end*N; i+=1) {
-        var elo=i/N;
+	var elo=i/N;
+	anchors.push(elo);
 	var elo_next=(i+1)/N;
 	var c=sprt.characteristics(elo);
-	var c_next=sprt.characteristics(elo_next);
 	data_pass.push([elo,null,{v:c[0],f:(c[0]*100).toFixed(1)+'%'}]);
 	data_expected.push([elo,null,{v:c[1],f:(c[1]/1000).toFixed(1)+'K'}]);
-	for(const s of specials){
-	    if((c[0]<s && c_next[0]>=s) || (c[0]<=s && c_next[0]>s)){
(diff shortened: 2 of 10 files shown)


ppigazzini:
DEV updated, loaded the latest DB backup.

ppigazzini:
DEV updated, loaded the latest DB backup.

vdbergh:
@ppigazzini Could you remind me what the URL is of DEV?

ppigazzini:
https://dfts-0.pigazzini.it/tests

ppigazzini:
@vdbergh DEV updated after your latest force-push.
I don't view the "SPRT Classical bounds" added with https://github.com/glinscott/fishtest/pull/906

vdbergh:
The point is that normalized Elo obsoletes the classical bounds. The classical bounds were needed because classical tests have a lower draw ratio and hence they take much longer to complete with bounds expressed in logistic Elo. Normalized Elo automagically compensates for this.

Master even has a comment in tests_run.mak about this.

vondele:
Thanks, that looks cool!

If I can suggest bounds that are a bit more rounded (and based of the experience in the past few days), I would suggest
```
    standard STC: <-0.5, 2.5>
    standard LTC: <0.5, 3.5>
    regression STC: <-2.5, 0.5>
    regression LTC: <-2.5, 0.5>
```
so, all have the same width, and LTC is a little easier to pass.

Are there any particular ways we can check this patch?

vdbergh:
@vondele I will change the bounds as you suggest.

Concerning checking. Of course I simulated the formulas using essentially the same files as the ones that are in the stats directory.

The most crucial thing is the LLR computation. That's why on the raw stats page it is computed in several different ways. There are the three old methods which use the logistic bounds obtained from the normalized bounds. Now there is a fourth method which computes the LLR directly from the normalized bounds (this is the method that is now used for the tests, as it is rather elegant). All 4 methods should give approximately the same results (_the_ LLR is ill defined because of the unknown nuisance parameters).  

In the pentanomial case it is easy to make a mistake with sqrt(2) factors. That's why everything is also computed using the trinomial model. Of course these results are a bit different but they still provide a sanity check on the pentanomial values.

The Elo computation simply uses converted bounds and is less critical anyway.

vondele:
OK, I'm having a look at the raw stats page. 
```
Normalized Elo | -19.00
```
Is it possible to have the error / confidence interval here as well ?

An other thing, for a fixed number of games test, it would be cool to have both the Elo and the nElo result in the output
```
 ELO: 22.62 +-40.5 (95%) LOS: 86.5%
Total: 200 W: 66 L: 53 D: 81
Ptnml(0-2): 9, 21, 34, 20, 16 
```
Similarly is https://dfts-0.pigazzini.it/html/live_elo.html?607c43efeae418362d7afb9c now using nElo or Elo ? Might be useful to switch to nElo?

vdbergh:
Currently all outputs are still logistic (e.g. the SPRTcalculator). I would prefer to keep the scope of this PR to the inputs. This PR tries to solve the technical problem of keeping resource usage constant.

Afterwards we can discuss the outputs. I guess there would be no issues giving the normalized Elo numbers beside the logistic ones.

You are right that the normalized Elo for SPRT tests does not have a confidence interval. This is the naive Elo estimate which is biased for SPRT tests (but it is of course a valid statistic). For fixed length tests it comes with an honest confidence interval which also takes the variance of the variance into account. I did not know how to compute this for SPRT tests so that's why it is open. I have forgotten why I list the number itself. 

If we want normalized outputs then this should be rethought.

vondele:
OK, I fine with limiting the scope to just inputs.

It is nice to have the nElo output in the statistics, especially as it is getting a more important status (as the basis for the current SPRT, and it might become a criterion for the releases).

ppigazzini:
DEV updated.

ppigazzini:
DEV updated.

ppigazzini:
DEV updated :)

vdbergh:
So it seems the test yielded no surprises....

vdbergh:
This PS has been squashed to a single commit.

ppigazzini:

vdbergh:
:)

PR #916, opened 2021-04-20
Update views.py to allow for slightly larger nns

NightlyKing:
This is needed for a test I want to run. It will use the HalfkAE5-v2 arch (256x2->32->32->1) from Sopel. The net is 115,525,612 Bytes in size.


vondele:
this looks OK to me, at least to enable testing.

ppigazzini (approved):

ppigazzini:
PROD updated, `nginx.conf` updated on the file server.
Thank you @NightlyKing :)

Issue #919, opened 2021-05-07
Booking keeping issue with nets?


vondele:
something is strange... 
both 
https://tests.stockfishchess.org/tests/view/608d83b895e7f1852abd2c6d 
and 
https://tests.stockfishchess.org/tests/view/608d2c7a95e7f1852abd2c1f 
mention nn-1a9420c9c1d3.nnue as the new_net.
The git diff on the same page shows only one of the two tests with this net, the first link seemingly being wrong.

https://tests.stockfishchess.org/nns shows I uploaded 6 nets, but only 5 show they have been tested (but hte download numbers are such that it is likely all 6 have actually been tested) https://tests.stockfishchess.org/tests/user/vdv shows I tested 3 standard nets (Elo around -30) and 3 Lc0 data nets (Elo around -60). With one canceled test in between (date: 21-05-01 )

As if the git diff or our book-keeping gets it wrong.

TopoIogist:
There seems to be a race condition in the views.py that fetches the net
https://github.com/glinscott/fishtest/blob/8308edf4e1b245d3f922e4439f221094dae90a3e/server/fishtest/views.py#L416

It uses the git branch name to fetch the net

```
def get_net(branch, repo_url):
    """ Get the net from evaluate.h or ucioption.cpp in the repo """
    api_url = repo_url.replace(
        "https://github.com", "https://raw.githubusercontent.com"
    )
    try:
        net = None

        url1 = api_url + "/" + branch + "/src/evaluate.h"
```

Correct would be to use the sha commit id, as the branch could have changed in the meantime.

PR #920, opened 2021-05-08
Use commit sha to retrieve net.

TopoIogist:
When rescheduling a test, fishtest previously used the cached commit sha to retrieve
the code to execute, but fetched the net using only the branch name.

This leads to race conditions if the default net of the corresponding branch was changed in
the meantime as observed in issue https://github.com/glinscott/fishtest/issues/919
To fix the problem use the commit sha also to fetch the net.

Note: I do not have a local fishtest installation to test this fully, so please test with care before merging.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -413,7 +413,7 @@ def get_sha(branch, repo_url):
         return "", ""
 
 
-def get_net(branch, repo_url):
+def get_net(commit_sha, repo_url):
     """ Get the net from evaluate.h or ucioption.cpp in the repo """
     api_url = repo_url.replace(
         "https://github.com", "https://raw.githubusercontent.com"
@@ -421,7 +421,7 @@ def get_net(branch, repo_url):
     try:
         net = None
 
-        url1 = api_url + "/" + branch + "/src/evaluate.h"
+        url1 = api_url + "/" + commit_sha + "/src/evaluate.h"
         options = requests.get(url1).content.decode("utf-8")
         for line in options.splitlines():
             if "EvalFileDefaultName" in line and "define" in line:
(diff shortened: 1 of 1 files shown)


ppigazzini:
@TopoIogist @vondele I updated [DEV](https://dfts-0.pigazzini.it/) with the PR but not sure how to test it. Feel free to make some test on DEV (user00:user00 approver, user01:user01 normal user)

vondele:
I have created a test using a new net (nn-f7d87b7a1789.nnue) and that seems to be correctly detected by dev (i.e. appears on the test page). However, a worker fails to build that test, probably because somehow that net is not correctly found. That might be unrelated to the patch and rather related to setup of dev?

ppigazzini:
DEV hasn't a nn server, so uploading a new net creates only an entry in the DB.

vondele:
OK, so that makes sense. Let me try to do a test with an existing net, in that case.

ppigazzini:
@vondele by the way after updating the new nn server I will try to setup a nn server on DEV :)

vondele:
with an existing net the test runs fine. I think the change looks good (but I'm not the expert of the code).

ppigazzini:
@vondele if I'm not wrong this is the other change required (with `fishtest.upload`) to use the new nn server: https://github.com/glinscott/fishtest/blob/bb78b0b39ad3ad6f640d31005b97aab316ad0562/server/fishtest/api.py#L223-L225
to:
```python
return HTTPFound( 
    "https://fishnet-new.v7f.eu/nn/" + self.request.matchdict["id"]
) 
```

DEV should be ready to use only the new nn server.

At your convenience please try to upload a new valid net and run a test.

vondele:
saw this only now... yes works fine.

vondele:
OK, that will be for at daylen.. 

personally, I think it would be good to update data.stockfishchess.org to point to the new server.

vondele:
@ppigazzini this one can probably be merged as well.

ppigazzini:
PROD updated, thank you @TopoIogist :)

PR #924, opened 2021-05-14
Use format <master>..<test> for diff URL

Lolligerhans:
```
This shows the full diff when master has diverged from the test
branch.

This is helpful no to disguise an outdated master when chekcing the test
diff.
```

The "new test" form will complain when master in the base repo is not up to date. But it does _not_ warn when the old master is _used_ instead. With this change, the approver obtains a chance to notice this error.

### DIFF
--- a/server/fishtest/helpers.py
+++ b/server/fishtest/helpers.py
@@ -4,18 +4,30 @@ def tests_repo(run):
     )
 
 
+def tests_repo_api(run):
+    """Return URL for GitHub API of the repository"""
+    return run["args"].get(
+        "tests_repo", "https://github.com/official-stockfish/Stockfish"
+        ).replace(
+            "//github.com/", "//api.github.com/repos/", 1)
+
+
 def master_diff_url(run):
     return "https://github.com/official-stockfish/Stockfish/compare/master...{}".format(
         run["args"]["resolved_base"][:10]
     )
 
 
-def diff_url(run):
+def diff_url(run, *, api_url=False):
+    """Produces GitHub diff URL
+
+    Uses the two-dotted format for non-API, the three-dotted format for API."""
     if run["args"].get("spsa"):
         return master_diff_url(run)
     else:
-        return "{}/compare/{}...{}".format(
-            tests_repo(run),
+        return "{}/compare/{}{}{}".format(
+            tests_repo(run) if not api_url else tests_repo_api(run),
             run["args"]["resolved_base"][:10],
+            ".." if not api_url else "...",
             run["args"]["resolved_new"][:10],
         )
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -329,7 +329,7 @@
     }
 
     // Fetch the diff and decide whether to show it on the page
-    const diffApiUrl = "${h.diff_url(run)}".replace("//github.com/", "//api.github.com/repos/");
+    const diffApiUrl = "${h.diff_url(run, api_url=True)}";
     $.ajax({
       url: diffApiUrl,
       headers: {


Lolligerhans:
*Now changed in the correct I believe

Lolligerhans:
One disadvantage is that the individual commits can no longer be listed.

ppigazzini:
@Lolligerhans DEV updated, the PR seems to have broken the diff, see
https://dfts-0.pigazzini.it/tests/view/609d84615085663412d08e2f
https://tests.stockfishchess.org/tests/view/609d84615085663412d08e2f

Lolligerhans:
@ppigazzini It appears to be working as intended. What seems broken?
Two dots `..` show the "raw" diff of the code that is being tested against each other (like one might expect).
Three dots `...` show the commits between the `test` commit and the _common ancestor_ of the `test` and `base` branch.

I believe the second to be safer because it shows when `base` has grown after `test` has diverged. Maybe I am the only one who ever made this mistake. :)
```
           C --- D  <--master
         /
A ----- B  *
         \
           E --- F  <--test

(*) B is the common ancestor of test and base.
```
In this case, three dots `master...test` show the diff of `E + F`, but the test evaluates `- D - C + E + F`.
Two dots `master..test` reflect the tested code. I am not sure if GitHub can be made to also list the involved commits in this case  (if possible).

Lolligerhans:
It appears that, at the moment, GitHub even provides a link to a three-dotted diff automatically. In my experience that would only occasionally be helpful or needed.

ppigazzini:
@Lolligerhans open the two links provided, the first one (PR) lacks the Diff section

Lolligerhans:
Yes, that seems to disappear. I have no idea why.

I have never used this section and would personally sacrifice it, but I suppose others might use it.

G-Lorenz:
I made few attempts:
deleting a dot in this [url](https://api.github.com/repos/glinscott/fishtest/compare/master...Lolligerhans:diff-all) led to an error, while deleting it [here](https://github.com/glinscott/fishtest/compare/master...Lolligerhans:diff-all.diff) led to the desired output. I think that the mistake came in this [line](https://github.com/glinscott/fishtest/blob/master/server/fishtest/templates/tests_view.mak#L332), so probably restoring the three dots format in the api calling in this line is needed. Despite the strange call the api would behave as `git log base..head` (two dots), according to the [guide](https://docs.github.com/en/rest/reference/repos#compare-two-commits).

Lolligerhans:
Simply using two dots `..` for the API part and three dots `...` for web diff, as suggested by G-Lorenz. I opted to expand the diff_url() function so nobody needs to remember why some dots are replaced.

I don't have a testing setup but am somewhat unfamiliar with Fishtest and Python, so please verify the implementation.

G-Lorenz:
There is an error:
image

Writing `.replace("//github.com/", "//api.github.com/repos/", 1)` (instead of max=1) should fix it. I tested this fix on my local DEV and it seems to work:
image

ppigazzini:
@Lolligerhans DEV updated.

ppigazzini:
@vondele I need your review from a SF developer/approver point of view before merge this PR :)
At example the "Copy apply-diff command" returns:
- `curl -s https://github.com/BM123499/Stockfish/compare/594e2ac999...4daab5ca11.diff | git apply` - master
- `curl -s https://github.com/BM123499/Stockfish/compare/594e2ac999..4daab5ca11.diff | git apply` - PR

Lolligerhans (commented):
The commit title is wrong here, I meant of course "Use **three-dotted** diff URL with API"

ppigazzini:
@Lolligerhans you should squash the commits before the merge, so you can change the commit title. You should be able to modify the PR title, too.

vondele:
Do we have an example test where this makes a difference?

I notice that the comments are now not visible in the github diff
https://dfts-0.pigazzini.it/tests/view/609d6c345085663412d087e6
this needs an additional click, that's somewhat of a disadvantage, I usually look at the inline diff for approving a test, and at the github link if it has non-zero comments.

Lolligerhans:
@vondele This is meant not to make a difference (in presented diff) unless base is selected erroneously, like [here](https://tests.stockfishchess.org/tests/view/609cef177746e3dc74ffaed6). When viewing the diff on GitHub, manually replace `...` in the URL with `..` to see the difference.
I might be the only one to ever (have been) make(ing) this mistake.

Unfortunately, the commit list disappears (since we generally no longer compare two ends of a commit chain). On the GiuHub web interface they insert a link to the current scheme (which then shows separate commits, too). This will be 1 additional click to see the conversation/comments on GitHub.

The diff embedded on the test page should remain as is - we noticed it disappears when trying to use two dots.

G-Lorenz:
@Lolligerhans after this PR the embedded diff will show the complete ("two dots") diff, since [this line](https://github.com/glinscott/fishtest/blob/716682d55a33c8393a502227542432c1f5686cff/server/fishtest/templates/tests_view.mak#L315) is left unchanged.
On the other hand `const DiffApiUrl` is needed in order to find information about the diff, e.g. number of comments or number of lines. If I understand correctly (both the code and the guide) using three dots in DiffApiUrl would behave as `git log base..head` (two dots).

@vondele I noticed that the `diff` in the upper part of the page and the `view on GitHub` in the diff section link to the same url: maybe we could embed the second one in "(N comments)" using the old url (see picture). In this case we will have two complete diff (external and embedded) and a direct link to comments (as before).

immagine

Lolligerhans:
Can we make the test page show the result of this instead?

`git branch <test_branch> --contains <base_commit>`

Sorry, I think I was too hasty with the diff thing.

snicolet:
One solution is to propose both diff as link options in each test page:

a) on the main fishtest page, let the diff after each page open diff...  (as in current server)

b) in the individual page of a test:

b1) at the top, propose both diff... and diff.. links (they will usually give the same, but power users can click on diff.. if they want)
b2) in the middle of each page, continue using diff... as in current server, but propose the two options for "Copy apply-diff command"

   -  Copy apply-diff... command
   -  Copy apply-diff.. command

snicolet:
This way we can still follow the comments in the individual commits if we want, but we have access to the real hard diff when the branches have diverged

dubslow:
> Do we have an example test where this makes a difference?

Well now we've found one lol, https://tests.stockfishchess.org/tests/view/63f31bc2e74a12625bccc7e7

has just been PRd to stockfish but was run with a flawed, and thus useless, test which would likely have been caught with a `..` diff rather than the `...` diff

PR #925, opened 2021-05-14
Replace freegeoip.app country code queries with a local database.

Sopel97:
This PR replaces the code that queries a remote server for the country code (for flag display) with a query to a local database (fastip) provided by @noobpwnftw. The database is a single .bin file (stored compressed with .bz2, <3MiB in size). Currently the path to the default database file is hardcoded and the file is placed along the sources (Any better alternatives? Don't store in git and use a config of some sort?). Both country code and the country full name is retrieved, but more information is stored for some ip ranges if needed in the future. Also, the `geoip_checked_at` value is stored in the database, how needed is it and could it be renamed/removed?


tomtor:
`geoip_checked_at` is used to expire the cached entries. If we use a static database expiring makes no sense, the `.bin` file should be updated at regular intervals

noobpwnftw:
the `.bin` file is updated daily, but not really necessary for this purpose. If someone whines about their country mis-identified, we can update it. Meanwhile this drops external API queries for performance reasons.

tomtor:
@ppigazzini I am OK with this change...

ppigazzini:
DEV updated.
To simplify the maintenance I propose these steps:
- `.bin` external to git repo
- instructions to download/create an updated `.bin` file

PROD and DEV are running with python 3.9.5, so we could start using some useful python standard libs to simplify the code (e.g. `pathlib`)

Sopel97:
So I guess we want to keep this database file outside of version control and provide it as a dependency. How to go about this in the best way? Hardcode the location and required the file to be there? Or do we have some way of configuration?

ppigazzini:
If the online IP geo localization hurts the fleet performance I'm to drop the function rather than use a local one that make harder the setup/maintenance.

I'm wondering if we can have the user setting (an optional) nation ID worker side.

PR #928, opened 2021-05-17
Stop the heartbeat when stopping the worker

noobpwnftw:
Trigger heartbeat thread exit on exceptions.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -99,6 +99,8 @@ def on_sigint(signal, frame):
 
 
 def worker_exit(val=1):
+    global ALIVE
+    ALIVE = False
     os._exit(val)
 
 


vondele:
I assume that might be the fix for the workers that are seemingly stuck indefinitely?

noobpwnftw:
Yes.

ppigazzini (approved):

ppigazzini:
Merged, raised the worker version to trigger the update.
Thank you @noobpwnftw :)

vdbergh:
An obvious question however is why a worker sometimes get stuck in the first place...

vondele:
if the fleet launches, certain things could timeout (like net download from server), which could cause an exception building the code. It would be interesting to know what it is, but it most likely is some contention/timeout issue somewhere.

PR #940, opened 2021-06-19
Fix deprecation warning for collection_names

zungur:
Fixes a deprecation warning during the initial setup of the server.
`collection_names()` is deprecated since pymongo 3.7.
See: [https://pymongo.readthedocs.io/en/stable/changelog.html#changes-in-version-3-7-0](https://pymongo.readthedocs.io/en/stable/changelog.html#changes-in-version-3-7-0)

### DIFF
--- a/server/utils/create_indexes.py
+++ b/server/utils/create_indexes.py
@@ -76,7 +76,7 @@ def create_flag_cache_indexes():
 
 
 def print_current_indexes():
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:
Tested OK on DEV; PROD updated, thank you @zungur :)

Issue #947, opened 2021-06-21
Do not use AVX512 until hardware improves


noobpwnftw:
Current CPUs are still faster with AVX2 than AVX512.

PR #948, opened 2021-06-21
Disable AVX512 for now.

vondele:
noob reports 20% slowdown compared to avx2

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -336,6 +336,7 @@ def find_arch_string():
             and "x86-64-avx512" in targets
         ):
             res = "x86-64-avx512"
+            res = "x86-64-bmi2" #use bmi2 until avx512 performance becomes actually better
         elif (
             "-mbmi2" in props["flags"]
             and "x86-64-bmi2" in targets


vondele:
fixes https://github.com/glinscott/fishtest/issues/947

ppigazzini (approved):

ppigazzini:
Version raised to trigger the worker update :)

NightlyKing:
I think a better way is to have some automatic bench once to test which arch to use (e.g. after installing the worker). Alternatively we could ask the user what to use. On some CPUs (latest i9s) it gains as much as noob is losing on his xeons (which struggle quite a bit with the down clocking).

ppigazzini:
@vondele something related to bench. I recently compared two builds with the same ARCH but different gcc version (10.2 cygwin vs. 10.3 msys2) and the bench result seems to depend strongly on depth/NNUE `bench 16 1 ${depth} default depth [NNUE]`

| depth | NNUE | 10.2 vs 10.3 speedup |
|-|-|-|
| 13 | n | 0.006041 |
| 14 | n | 0.018216 |
| 15 | n | 0.050885 |
| 16 | n | 0.033239 |
| 13 | y |  0.005921 |
| 14 | y | -0.040629 |
| 15 | y | -0.033918 |
| 16 | y | -0.020917 |

This is the result with gcc 10.3 and two different PGOs: `bench 16 1 13 default depth` vs `bench 16 1 13 default depth NNUE`

| depth | NNUE | PGO normal vs PGO NNUE speedup |
|-|-|-|
| 13 | n | -0.002464 |
| 14 | n | -0.004735 |
| 15 | n | -0.003849 |
| 13 | y | -0.000016 |
| 14 | y | 0.001413 |
| 15 | y | 0.001973 |

PR #950, opened 2021-06-22
Add Discord link

SFisGOD:
The invite link used is from here https://github.com/official-stockfish/Stockfish/issues/2345


ppigazzini:
PROD updated, thank you @SFisGOD :)

PR #954, opened 2021-06-23
Add password strength requirement

### LINKED ISSUE #953: Internal server error during password change
If a user attempts to change their password on their profile page and types a non-matching verification password, the user encounters the following error:
```
Internal Server Error
The server encountered an unexpected internal server error
(generated by waitress)
```
The error is recoverable from the user's perspective with a page refresh. No changes are submitted to the userdb.

zungur:
- Fixes a bug where a non-matching verification password in user profile causes internal server error.
- Introduces an offline, password strength validation for signup and password change operations. Existing passwords in the userdb are not subject to a minimum strength evaluation.
- Adds notifications upon email and/or password updates on user profile page. Relevant errors are displayed as well, accordingly.
- Skips email update if the email field matches the db entry. 
- The mako template for user profile page specifies the same minimum 8 character requirement for the password as the signup

Fixes #953

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -20,7 +20,7 @@
   <div class="control-group">
     <label class="control-label">New password:</label>
     <div class="controls">
-      <input name="password" type="password"/>
+      <input name="password" type="password" pattern=".{8,}" title="Eight or more characters: a password too simple or trivial to guess will be rejected"/>
     </div>
   </div>
   <div class="control-group">
--- a/server/fishtest/templates/signup.mak
+++ b/server/fishtest/templates/signup.mak
@@ -25,11 +25,11 @@
       <label class="control-label">Password:</label>
       <div class="controls">
         <input name="password" type="password" pattern=".{8,}"
-               title="Eight or more characters" required="required"/>
+               title="Eight or more characters: a password too simple or trivial to guess will be rejected" required="required"/>
       </div>
     </div>
     <div class="control-group">
-      <label class="control-label">Repeat password:</label>
+      <label class="control-label">Verify password:</label>
       <div class="controls">
         <input name="password2" type="password" required="required"/>
       </div>
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -7,6 +7,7 @@ from email.mime.text import MIMEText
 import fishtest.stats.stat_util
 import numpy
 import scipy.stats
+from zxcvbn import zxcvbn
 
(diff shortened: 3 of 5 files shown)


ppigazzini:
[DEV](https://dfts-0.pigazzini.it/tests) updated. I will test the PR tomorrow, attention diverted by @snicolet step down :(

server/fishtest/templates/user.mak
```diff
@@ -20,7 +20,7 @@
   <div class="control-group">
     <label class="control-label">New password:</label>
     <div class="controls">
-      <input name="password" type="password"/>
+      <input name="password" type="password" pattern=".{8,}" title="Eight or more characters"/>
```
ppigazzini, line 23: "Eight or more non trivial characters: password too simple will be rejected"

zungur: I agree. Even though we display a warning notification which explains why the password is rejected, it's better to give a heads up to the user. I think the signup page should also incorporate the same type of warning here:
https://github.com/glinscott/fishtest/blob/master/server/fishtest/templates/signup.mak#L28

ppigazzini: I agree with the signup page improvement.

zungur: Pushed changes for both pages, thank you

ppigazzini:
@zungur I played on DEV and looks good. If this is finished, please squash the commits :)

ppigazzini:
How to squash several commits:
- `git checkout master && git fetch -p upstream && git reset --hard upstream/master && git checkout password_strenght`
- `git rebase -i master`
- replace "pick" with "squash" for all the commits but the first one
- write the final commit message (eg copying moving deleting the several commits messages)
- `git push -f origin password_strenght`

zungur:
Thank you, that was very helpful for a first timer :)

ppigazzini:
Looks good :)

I learned git working on this project :)

In this case master was updated a couple of times but there are no changes conflicting with yours, the GitHub status was "This branch has no conflicts with the base branch when rebasing", so no changes was required.

In case of conflict then is required a rebase against the new master.
- `git checkout master && git fetch -p upstream && git reset --hard upstream/master && git push -f origin master && git checkout password_strenght`
- `git rebase -i master`
- the editor opens with the right file and it highlights the 2 rows in conflict. Write the right code and save.


`git rebase -i master` is the git swiss army knife, better to play a bit with it.

ppigazzini:
PROD updated, thank you @zungur :)

PR #959, opened 2021-06-27
Use .. instead of ... for github compare.

Sopel97:
...: https://github.com/MichaelB7/Stockfish/compare/0470bcef0e...be08a5bf65
..: https://github.com/MichaelB7/Stockfish/compare/0470bcef0e..be08a5bf65

untested, not sure if it will work in all cases

### DIFF
--- a/server/fishtest/helpers.py
+++ b/server/fishtest/helpers.py
@@ -5,7 +5,7 @@ def tests_repo(run):
 
 
 def master_diff_url(run):
-    return "https://github.com/official-stockfish/Stockfish/compare/master...{}".format(
+    return "https://github.com/official-stockfish/Stockfish/compare/master..{}".format(
         run["args"]["resolved_base"][:10]
     )
 
@@ -14,7 +14,7 @@ def diff_url(run):
     if run["args"].get("spsa"):
         return master_diff_url(run)
     else:
-        return "{}/compare/{}...{}".format(
+        return "{}/compare/{}..{}".format(
             tests_repo(run),
             run["args"]["resolved_base"][:10],
             run["args"]["resolved_new"][:10],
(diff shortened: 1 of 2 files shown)


G-Lorenz:
Is this similar to #924 ?

G-Lorenz:
I tested this PR in my local DEV: inline diff is broken. This is due to GitHub API, see https://github.com/glinscott/fishtest/pull/924#issuecomment-841476534 or directly Lolligerhans' code.

I suggest you to read also the comments since vondele's review: he poses an issue that I think it should be solved. There are some proposals but the discussion doesn't come to an end.

ppigazzini:
[DEV](https://dfts-0.pigazzini.it/) updated with the PR. It breaks the git diff like #924, see

https://dfts-0.pigazzini.it/tests/view/609d84615085663412d08e2f
https://tests.stockfishchess.org/tests/view/609d84615085663412d08e2f

Sopel97:
> 
> 
> [DEV](https://dfts-0.pigazzini.it/) updated with the PR. It breaks the git diff like #924, see
> 
> https://dfts-0.pigazzini.it/tests/view/609d84615085663412d08e2f
> https://tests.stockfishchess.org/tests/view/609d84615085663412d08e2f

what exactly is broken here?
obraz
obraz

ppigazzini:
The git diff disappears

Untitled

Sopel97:
> 
> 
> The git diff windows disappears:
> 
> Untitled

this is the first time I see this feature, and I'm unable to find it on https://tests.stockfishchess.org/tests/view/609d84615085663412d08e2f

Sopel97:
so if it's not possible to get the inline diff correct it should be dropped entirely?

ppigazzini:
> So if it's not possible to get the inline diff correct it should maybe be dropped entirely? It is useless if it's not correct.

I agree with you that a feature with wrong output should be or fixed or dropped.
I'm not a Stockfish developer, never used that feature, so like in #924 I ask to the Stockfish maintainer (cc @vondele) what to do.

vondele:
I use the inline diff all the time... but of course it would be good if it were correct.

PR #966, opened 2021-07-07
Fix message about --concurrency set to 0 or less in worker.py

Cubox:


ppigazzini:
https://github.com/glinscott/fishtest/blob/302ad0d56421b132a2cee4cb80611b61c33a6a65/worker/worker.py#L440

https://github.com/glinscott/fishtest/blob/302ad0d56421b132a2cee4cb80611b61c33a6a65/worker/worker.py#L459-L461

`cpu_count` is the `concurrency` set by the user, that line simply drop a wrong non positive `concurrency`.

With your proposed change it will be impossible to run a worker with a single core.

The message could be improved `"Not enough CPUs to run fishtest (it requires at least one)\n"`

Cubox:
I assumed, from the message, that fishtest required at least two cores.
Either the if or the message can be changed, depending on what the actual requirement is.

ppigazzini:
The requirements are:
 - "max worker cores = CPU cores - 1" (or "CPU cores" with `--use_all_cores`)
 - "min worker cores = 1"

The original message is talking about the min CPU cores (it was written before the introduction of the with `--use_all_cores`) but that check fails only if the user set a non positive `concurrency`.

"Not enough CPUs to run fishtest: set --concurrency at least one\n"

Cubox:
Here, this should be good

ppigazzini (approved):

ppigazzini:
Some requests:
- improve a bit my bad English and quoting the parameter: `"Not enough CPUs to run fishtest: set '--concurrency' to at least one\n"`
- confirm your real name for the AUTHORS file :)

I will merge this PR at the next worker version raise.

ppigazzini:
Merged, thank you @Cubox :)

Cubox:
Oh I missed that message, sorry.
I'll make another PR with the typo fix if that's still needed
My real name is Andy Pilate.

ppigazzini:
https://github.com/glinscott/fishtest/blob/eab590b433cb1d7a044faf1e5c82c6c793c2e6ff/AUTHORS#L1-L6

PR #971, opened 2021-07-17
Refine memory check a bit

vondele:
in addition to the TT, we allocate quite some memory for the net.
Add a base estimate of memory usage per process.

This might fix some timeout observed on workers, that might be paging.


ppigazzini (approved):
Tested fine on DEV and some workers.
Mem usage from top 5135, mem estimated by fishtest 5332 (need_tt =  992 need_base =  4340)
PROD updated.
Thank you @vondele :)

PR #974, opened 2021-07-27
Stop NN upload process after exception

ppigazzini:
Update the NN db only after a successful upload.


vondele:
I think this looks OK. Tom might be better in reviewing this.

tomtor:
Looks good to me!

ppigazzini:
Tests on DEV are ok.
Thank you @vondele @tomtor for reviewing :)

PR #980, opened 2021-08-02
Show the worker UUID in machines/tasks tables

ppigazzini:
Show only the first UUID group of hex digits

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -33,6 +33,8 @@ def get_worker_key(task):
     suffix = UUID_MAP[username][uuid]
     if suffix != 0:
         worker_key += "-" + str(suffix)
+    if len(uuid) != 0:
+        worker_key += "-" + uuid.split("-")[0]
 
     return worker_key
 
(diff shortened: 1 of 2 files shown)


ppigazzini:
@tomtor please review this PR at your convenience, I'm not very confident with my knowledge of the server logic :)
DEV updated, it seems working fine also for very old tests.

server/fishtest/util.py
```diff
@@ -33,6 +33,8 @@ def get_worker_key(task):
     suffix = UUID_MAP[username][uuid]
     if suffix != 0:
         worker_key += "-" + str(suffix)
+    if len(uuid) != 0:
```
tomtor, line 36: Is len(uuid) ever 0? Can this condition be removed?

tomtor (commented):
Looks good to me, but see comment

server/fishtest/util.py
ppigazzini: `uuid=""` for a very old test that lacks `"unique_key"`:
https://github.com/glinscott/fishtest/blob/55f4f9f66d618c6f627aec4222acb116cecb3b27/server/fishtest/util.py#L26

See this test on DEV:
https://dfts-0.pigazzini.it/tests/view/5134df478f0c3e3c12c7465c

tomtor: Ah, clear.

tomtor (approved):
Nice!

Issue #978, opened 2021-08-01
I recommend to increase batch sizes for powerful hardware (Dual Xeon)


Technologov:
Hi,

I own several Dual Xeon Platinum 8176 servers, those are 56-core/112-thread machines.
The current batch size in Fishtest is only 200 games, according to the logs.

That means that every couple of minutes my servers re-download and re-compile Stockfish, a CPU effort that gets wasted, while it could be put to good use (...into more games being played).

I suggest that the batch size should depend on the hardware, with a small batch for Raspberry Pi, but a much bigger batches for Dual Xeon servers.

-Alexey "Technologov ", 01.08.2021

Sopel97:
hmm, is there anything particularily hard preventing us from assigning, say, `min(cores*20, game_left_to_limit)` game batches?

Technologov:
Let's try even bigger batches... 1 hour of computer time so (cores *60 ) for LTC games and (cores*60*6) for STC games

Sopel97:
https://github.com/glinscott/fishtest/blob/55f4f9f66d618c6f627aec4222acb116cecb3b27/server/fishtest/rundb.py#L58

Vizvezdenec:
I honestly don't remember why we even have such a small batches like 200.
With current fishtest mostly being run on 8+ core machines and also now some massive 50+ cores servers it would make a lot of sense to make batches bigger, like 1000-2000 games easily.

Vizvezdenec:
@vondele

ppigazzini:
@Technologov 
> That means that every couple of minutes my servers re-download and re-compile Stockfish, a CPU effort that gets wasted, while it could be put to good use (...into more games being played).

Wrong, the worker keeps the latest 50 binaries to optimize the download + build process, and the latest 10 nest to optimize the download, see
https://github.com/glinscott/fishtest/blob/9f4c421aee89e2eb92e3eddf270a994adec1958c/worker/games.py#L865-L873

https://github.com/glinscott/fishtest/blob/9f4c421aee89e2eb92e3eddf270a994adec1958c/worker/games.py#L937-L945

By the way the little VPS running the fishtest server scales very well, serving without problem 4803 machines at 29678 cores, see #553 where a SF developer whined "It's not that easy to beat speed of tests that complete if 20k cores are running - it's hard to produce any meaningful ideas at this tempo." :)

@Vizvezdenec 
> I honestly don't remember why we even have such a small batches like 200.

[...]

PR #988, opened 2021-08-05
Set task size to 1000 games

Vizvezdenec:
PR following 
https://github.com/glinscott/fishtest/issues/978
With current fishtest machine distribution reasons for having such a small batch size are probably non-existent and with recent influx of cores such a small batch size can hurt fishtest performance.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -46,7 +46,7 @@ class RunDb:
         self.runs = self.db["runs"]
         self.deltas = self.db["deltas"]
 
-        self.chunk_size = 200
+        self.chunk_size = 1000
 
         global last_rundb
         last_rundb = self
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -666,7 +666,7 @@ def validate_form(request):
         # Limit on number of games played.
         # Shouldn't be hit in practice as long as it is larger than > ~200000
         # must scale with chunk_size to avoid overloading the server.
-        data["num_games"] = 2000 * request.rundb.chunk_size
+        data["num_games"] = 400 * request.rundb.chunk_size
     elif stop_rule == "spsa":
         data["num_games"] = int(request.POST["num-games"])
         if data["num_games"] <= 0:


ppigazzini:
[DEV](https://dfts-0.pigazzini.it) updated, started 2 workers at 1 core and at 23 cores (same server CPU bmi2 at 16 cores/32 threads) to measure the time required to finish a 1000 games task.

Some notes:
- the `estimate_game_duration` pushes a worker with concurrency > 23 cores to LTC task, see:
https://github.com/glinscott/fishtest/blob/82ce909ef592cb0ff91652379b9b586eec400196/server/fishtest/rundb.py#L622-L631
- the worker at 56 cores in #978 (Dual Xeon Platinum 8176 56-cores/112-threads) can finish a 200 games LTC task in at least 12.5 minutes ([see discord](https://discord.com/channels/435943710472011776/813919248455827515/872561289833087007))

ppigazzini:
@Vizvezdenec @vondele some data.

MaxKlaxxMiner 4 cores has now the highest MNps at 2.17, from a STC pgn:
- [TimeControl "4.972+0.05"] (0.4972 TC)
- [GameStartTime "2021-08-07T11:52:02.554 Mitteleuropäische Sommerzeit"] - first game start
- [GameEndTime "2021-08-07T12:03:33.124 Mitteleuropäische Sommerzeit"] - last game end
- task time duration: 00:10:10:57
- average game time: 12.2114

armo9494 24 cores has a Ryzen 5950x (one of most powerful CPU), from a LTC pgn:
- [TimeControl "37.658+0.377"] (0.6276 TC)
- [GameStartTime "2021-08-04T15:00:50.137 Västeuropa, sommartid"]
- [GameEndTime "2021-08-04T15:14:21.460 Västeuropa, sommartid"]
- task time duration: 00:13:31:323
- average game time: 90.147

Estimation of armo9494 23 cores at STC (worst case for a STC test) 
- [TimeControl "6.276+0.006"] (0.6276 TC)
- average game time: 12.2114 * 0.6276 / 0.4972 = 15.4141
- task time duration: 00:02:18.727

vdbergh:
I think we could deal with problem of the "1 core worker nibbling away forever at the last batch of 1K LTC games" by overcommitting. E.g. pretend that a test of 60000 games is a test of 400000 games and stop it when 60000 games  have been received. To make this work the batch size (the frequency that the worker calls home) should be set to a low value (e.g. 8 like for SPRT tests). The chunk size can be enlarged arbitrarily however.

ppigazzini:
IIRC the code create the whole tasks list at the start, so we should overcommit only a bit to avoid a memory problem.
I collected some info/link about the batch size, see https://github.com/glinscott/fishtest/issues/978#issuecomment-892791683
If I read correctly the code the worker calls home every 8 games for SPRT/NumGames and 4*concurrency for SPSA.

Anyway @vondele suggested to assign low cores workers to STC (I failed to find a fast way to implement this)

Raising the task size should raise a bit the percentage of task switch to another worker after a worker stop, not a big problem but I have in my todo list the check of that code.

vdbergh:
Well for SPRT the default length is 400000 games. Since the vast majority of tests are SPRT, it seems the memory problem is not very big.

I think calling home is the same for SPSA and Numgames, only SPRT is special (also IIRC).

MJZ1977:
As said before, I think that increasing the number of default batch to 1000 can lead to a very long batchs with 1 core for SPSA and fixed number tests.
First we need to limit 1 or 2 core workers to SPRT, and after we can increase batch sizes to 1000. Or, change default batch size to 1000 only for SPRT...

Vizvezdenec:
I don't see much problem with either of this, honestly.  
For example, tuning runs can be stopped w/o last 200 games being completed since they wouldn't produce much variation. 
Same can be done for fixed core tests if they are to check smth quick and for regression tests it doesn't really matter when they actually finish, honestly.

jhellis3:
Can it not be made core count dependent? Say 200 * x for example.

ppigazzini:
> Can it not be made core count dependent? Say 200 * x for example.

I fear that nobody of the active fishtest developers know the server code so well to made that change in reasonable time. At example now an unfinished batch is assigned to the first free worker.

ppigazzini:
Solved with dynamic task #1021 thanks to @vdbergh :)

PR #992, opened 2021-08-13
Stop the test after failing to setup an engine

ppigazzini:
In case of a bug thas is OS dependent we can have some workers that waste CPU
failing to setup the engine while the test seems to run fine in the dashboard.
The fishtest server already stops a test only for user with at least 1000 hours
of CPU contributed, so a beginner mistake in setup the worker toolchain
cannot stop a test.


ppigazzini:
This should fix problems like this https://github.com/official-stockfish/Stockfish/pull/3651

Thanks to @SFisGOD to provide the faulty branch to write/test the PR :)

PR #996, opened 2021-08-22
Update to a new default book

vondele:
Switches to a new book by Stefan Pohl: UHO_XXL_+0.90_+1.19.epd

See
https://github.com/official-stockfish/books/commit/63b0f2095923d98d4dcb6d7b0ee79404f83c4270

See discussion and tests in
https://github.com/official-stockfish/Stockfish/issues/3323

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -307,7 +307,7 @@ fb=lambda e0,e1:format_bounds(elo_model,e0,e1)
       <label class="field-label leftmost">Book</label>
       <input type="text" name="book" id="book"
              style="width: 229px"
-             value="${args.get('book', 'noob_3moves.epd')}" />
+             value="${args.get('book', 'UHO_XXL_+0.90_+1.19.epd')}" />
 
       <label class="field-label book-depth"
              style="width: 87px; display: none">Book depth</label>


vondele:
There are a few more references to noob_3moves.epd in the comments, they appear to relate to the SPSA, and I don't think they need simultaneous updating (refers to the auto SPSA parameters, IIUC).

ppigazzini:
DEV updated; created SPRT, NumGames, SPSA tests with the new book; joined Linux, WSL2, msys2 workers. I'm starting to review the references in the comment.

ppigazzini:
> There are a few more references to noob_3moves.epd in the comments, they appear to relate to the SPSA, and I don't think they need simultaneous updating (refers to the auto SPSA parameters, IIUC).

That is the info box for the auto SPSA parameters, so the reference must stay. There is also a reference to `contempt = 24`, so I'm wondering if that code makes still sense with the NNUE Stockfish.

By the way the SPSA seems to tune very well the NNUE parameters.

vondele:
yeah, I agree that reference must stay, and probably one could refit that code for the new book / new engine. However, as you say, SPSA is pretty successful these days, so it wouldn't be urgent to change.

ppigazzini:
DEV and workers look good to me, let me know when I can merge and update the server.

vondele:
please go ahead.

ppigazzini:

PR #997, opened 2021-08-23
Make the defaults for the SPRTcalculator compatible with the new book.

vdbergh:
Make the defaults for the SPRTcalculator compatible with the new book.


ppigazzini (approved):

ppigazzini:
DEV looked good, PROD updated, thank you @vdbergh :)

PR #999, opened 2021-08-25
Correct the chi2 p-value computation for an unbalanced book.

vdbergh:
Currently the chi2 test assumes that the games in a task are independently drawn from a trinomial distribution. This assumption is no longer valid for an unbalanced book. This is very visible from the p-values which are often > 99%. Under the null hypothesis p-values should be uniformly distributed.

Since currently the ww and ll frequencies will be too small for the chi2 test to be valid, we coalesce them with the wd and ld frequencies.

The main test page still shows the trinomial wdl frequencies. This will be addressed in a separate PR.

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -45,6 +45,7 @@ def get_chi2(tasks, bad_users):
 
     # Aggregate results by worker
     users = {}
+    has_pentanomial = None
     for task in tasks:
         task["worker_key"] = get_worker_key(task)
         if "worker_info" not in task:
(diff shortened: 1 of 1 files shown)


vdbergh:
For a chi2 test the expected frequencies should not be too small. See e.g. the last page of

https://www.open.ac.uk/socialsciences/spsstutorial/files/tutorials/chi-square.pdf

I agree that the comment is not very precise (the issue is not p[0] and p[5] but the aggregated quantities obtained by summing over the tasks). I will improve the comment and reformat the code.

ppigazzini:
I think to have found why that minimum requirement for the expected frequencies.
The Pearson's chi squared statistic computed by the data must have a limiting chi square distribution, so also the `residual = (observed - expected) / sqrt(expected)` must be very similar to a standard normal distribution.
The data in a contingency table follow usually a binomial/multinomial distribution and the binomial distribution is approximated by a normal distribution only for value >= 5
https://en.wikipedia.org/wiki/Binomial_distribution#Normal_approximation

PR #1004, opened 2021-08-27
Enable asymmetric TC on the server.

vdbergh:
This contains the server changes to make asymmetric TC work.


vdbergh:
I have squashed the PR.

ppigazzini:
I found those problems on DEV:
- "NumGames" test now has a result box with "Pending..." when active or finished (yesterday was fine, showing the progress of the games)
- "SPRT" and "NumGames" result boxes are not clickable when active (not sure if yesterday they were fine)

vdbergh:
Thanks. The "pending..." is purely cosmetic. It is because there were 0 lost games. 

I'll have a look at the clickability of the SPRT and NumGames boxes.

vdbergh:
Actually I do not understand what the issue is with the SPRT and NumGames boxes. 

I just looked at the 'new test' page and everything seemed to be fine. Can you describe the steps to reproduce?

ppigazzini:
The boxes with:
```
89712/100000 iterations
179424/200000 games played
```
There is not the link to the live_elo page.

ppigazzini:
> Actually I do not understand what the issue is with the SPRT and NumGames boxes.
> 
> I just looked at the 'new test' page and everything seemed to be fine. Can you describe the steps to reproduce?

My fault, live_elo link works for SPRT test :(

ppigazzini:
DEV updated.

If I'm not wrong the result of a NumGames test with extreme time odds is reported only in stats page:
https://dfts-0.pigazzini.it/tests/view/612b4ecaf8259937c4db9094
https://dfts-0.pigazzini.it/tests/stats/612b4ecaf8259937c4db9094

What about these change (also not conditional)?
- "tc" -> "base_tc" in https://dfts-0.pigazzini.it/tests/view/612b4888f8259937c4db9080
- "TC" -> "Base TC" in https://dfts-0.pigazzini.it/tests/stats/612b4888f8259937c4db9080

vdbergh:
About the NumGames test. This is a known issue with fishtest. Tests with 0% or 100% score are not quite handled correctly. This is a left over from old days when the formulas did not work in those cases. Now there is no technical reason for this anymore and I can fix it, but I would prefer to do that in a separate PR. 

Changing tc to base_tc would entail a lot of changes in the code (run['args']['tc'] would have to be changed to run['args']['base_tc']). It don't think it would be worth it, and in any case it would be better to do it in a separate PR.

I can change TC to Base TC in the raw stats page. 

I looked on DEV and it seems the border is not very visible in case the result widget is red. I will try a different color (unless you think the border is a bad idea @ppigazzini @vondele).

ppigazzini:
Ok, thank you for the explanation about "tc", DEV updated. I will be off line for a couple of hours.

vdbergh:
Dev now has some result widgets with **black** borders (but not all of them)… Is it this PR?

ppigazzini:
Yes, I reloaded the server after your squash, but I view still red borders.

ppigazzini:
Check the test comment, I started also several "Normal" TC test.
By the way, if you haven't added commit to be tested I will merge these PR (raising the worker version).

vdbergh:
It happens with safari on an iPad (but not with chrome). Some borders are black (but others are pink, there seems to be no rule).

I will investigate tomorrow (I cannot do view page source on an iPad, to see if the black borders have any relation to the html code).

vdbergh:
Firefox is also fine. I suspect it is a bug in safari.

ppigazzini:
Microsoft Edge is ok.

vdbergh:
Thanks.

I checked that the html code for the black and pink borders is the same.

However an html validator indicates an error in the inline css syntax (also on the official Fishtest). I will fix this but I doubt it will help.

vdbergh:
This is how it looks on Safari on the iPad:

7EF40E1C-ED49-47ED-88EE-53612588A529

vdbergh:
Anyway the border is only an indication that the test is non standard (which will occur very rarely) and its colour is not really important. So if this is really a Safari bug then it is only a very small cosmetic problem.

server/fishtest/templates/elo_results.mak
```diff
@@ -2,9 +2,14 @@
...
   def get_run_style(run):
+    ret=''
     if 'style' in run['results_info']:
-      return 'background-color:' + run['results_info']['style']
-    return ''
+      ret+='background-color:' + run['results_info']['style']+';'
+    tc=run['args']['tc']
+    new_tc=run['args'].get('new_tc',tc)
+    if tc!=new_tc:
+       ret+='border-style: solid; border-color: Pink;  border-width: medium; border-radius: 0;'
```
ppigazzini, line 11: - double space after "Pink;"
- for red border, the double semicolon ";;" is right? (in chrome using only one don't change the result)
```
<pre style="background-color:;border-style: solid; border-color: Pink;  border-width: medium; border-radius: 0;;white-space:nowrap;" class="elo-results">
```
- for black border:
  - PROD:
  ```
  <pre style="background-color:;white-space:nowrap;" class="elo-results">
   ```
  - DEV (double semicolon ";;"):
   ```
  <pre style="background-color:;;white-space:nowrap;" class="elo-results">
   ```

vdbergh:
@ppigazzini Thanks for investigating, Yes there are various possible issues with the inline style. Tomorrow I will make it clean.

ppigazzini:
DEV updated.
On Windows looks good with Chrome, Firefox, Opera and Edge (I cannot test with Safari), the inline styling looks good:
- black border
```
<pre style="white-space:nowrap;background-color:yellow;" class="elo-results">```
```
- pink border
```
<pre style="white-space:nowrap;background-color:#44EB44;border-style:solid;border-color:Pink;border-width:medium;border-radius:0px;" class="elo-results">
```

vdbergh:
Safari on ipad still has the problem. But it really seems to be a bug with Safari. Since it is only a small cosmetic issue I think we can ignore it,

ppigazzini:
> Safari on ipad still has the problem. But it really seems to be a bug with Safari. Since it is only a small cosmetic issue I think we can ignore it,

You are using the iPad with the wrong hand!

As soon as you will squash, I will merge the PRs.

ppigazzini (approved):

ppigazzini:
PROD updated, triggered the workers update, thank you @vdbergh :)

SFisGOD:
@vdbergh Hash should be 64 https://tests.stockfishchess.org/tests/view/612d3a36bb4956d8b78eb5e8 https://tests.stockfishchess.org/tests/view/612d33d4bb4956d8b78eb5e2

vdbergh:
@SFisGOD Fixed. Thanks.

PR #1005, opened 2021-08-28
Enable fake net upload in a nice way.

vdbergh:
This is to make the installation of a local fishtest (for testing) a bit easier by avoiding the need of having a local cdn as well.

If the fishtest user has a file 'fake_net_upload' in its home directory then an "uploaded" net will be recorded in the local database but it will not be uploaded to any cdn.

It is the assumption that in that case the net already exists in the official stockfish cdn.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -120,21 +120,22 @@ def sync_upload(request):
         request.session.flash("You must specify a network filename", "error")
         return {}
 
-    try:
-        with open(os.path.expanduser("~/fishtest.upload"), "r") as f:
-            upload_server = f.read().replace("\n", "")
-            upload_server = upload_server + "/" + filename
-            response = requests.post(upload_server, data=network)
-            if response.status_code != 200:
-                print("Network upload failed: " + str(response.status_code))
-                request.session.flash(
-                    "Network upload failed: " + str(response.status_code), "error"
-                )
-                return {}
-    except Exception as e:
-        print("Network upload fails or not configured: " + str(e))
-        request.session.flash("Network upload fails or not configured", "error")
-        return {}
+    if not os.path.exists(os.path.expanduser("~/fake_net_upload")):
+        try:
+            with open(os.path.expanduser("~/fishtest.upload"), "r") as f:
+                upload_server = f.read().replace("\n", "")
+                upload_server = upload_server + "/" + filename
+                response = requests.post(upload_server, data=network)
+                if response.status_code != 200:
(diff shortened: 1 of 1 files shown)


vdbergh:
Perhaps this can be made more discoverable by making fishtest.upload an ini file. 

upload_server=....
skip_upload=true/false

ppigazzini:
IMO both ways are a hack to be used only with an unofficial DEV fishtest setup (no post-server).

I'm wondering if you could simply hack your DEV bulling a branch from your repo, see https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup#fishtest-server-script-to-test-a-prbranch

```bash
# add here the upstream branch to be tested
#git remote add <your_upstream> https://github.com/<your_username>/fishtest
#git pull --no-edit <your_upstream> <your_branch>
```

I use that bash script for all tests on DEV, I use also the branch pulling trick in a bash script to lower some defaults (e.g. cache binaries and nets) in my test workers.

vdbergh:
I agree it is a bit of hack, but I believe that for fishtest to attract developers it should be somewhat easy to install. Personally I would hesitate to run a webserver on my laptop just to make some occasional additions to fishtest. 

The solution you propose works fine but it does not solve the problem for other people.

ppigazzini:
I ran the DEV fishtest server only in a virtual machine or a linux container (lxd/lxc), so that is not a problem for me. I must routinely test that the server setup script is still working fine.

ppigazzini:
I'm using lxc/lxd on Ubuntu (and also on Windows with Hyper-V) to run linux containers with different distros/versions (it's possible also to run VMs in lxd/lxc, never tried).
```
# first host configuration, I accept the defaults
sudo lxd init
# container creation and starting, shortcut:
# lxc launch ubuntu:20.04 u2004
lxc init ubuntu:20.04 u2004
lxc start u2004
# exec a command on the container, here open a shell
lxc exec u2004 -- bash
# copy file to/from container
lxc file push <filename> u2004/root/<filename>
lxc file pull u2004/root/<filename> <filename> 
# stop container
lxc stop u2004
# list containers and delete container
lxc ls
lxc delete u2004
# list images and delete image
lxc image ls
lxc image delete <image_fingerprint>
# add/remove proxy to forward host port to container port
lxc config device add u2004 myport80 proxy listen=tcp:0.0.0.0:80 connect=tcp:127.0.0.1:80
lxc config device remove u2004 myport80
```

vdbergh:
Using a container seems like a good option. I need to make room on my laptop though for another virtual machine.

PR #1008, opened 2021-08-30
Dark mode refinements

SFisGOD:
Closes https://github.com/glinscott/fishtest/pull/1008

### DIFF
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -26,12 +26,16 @@ a:hover, a:focus {
   outline: none;
 }
 
+.form-control {
+  background-color: #b1ada7;
+}
+
 .hljs {
   filter: contrast(0.7);
 }
 
 legend {
-  color: #fafafa;
+  color: #b1ada7;
 }
 
 hr {
@@ -95,7 +99,7 @@ input[type="datetime"], input[type="datetime-local"], input[type="date"],
 input[type="month"], input[type="time"], input[type="week"], input[type="number"],
 input[type="email"], input[type="url"], input[type="search"], input[type="tel"],
 input[type="color"], .uneditable-input {
-  color: white;
+  color: #b1ada7;
 }
 
 select {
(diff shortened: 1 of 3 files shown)


SFisGOD:
@ppigazzini Please update DEV :)

ppigazzini:
[DEV](https://dfts-0.pigazzini.it) update, problem still there.

SFisGOD:
Okay let me check

SFisGOD:
@ppigazzini Why was this altered?
https://github.com/glinscott/fishtest/blob/67518594c0867626aef3aec062c5b9fac5ab7a9a/server/fishtest/templates/elo_results.mak

ppigazzini:
To add a pink border to the widget of "time odds" test.
DEV updated, something is improved.

SFisGOD:
Ahh to indicate that it's a time odds test.

SFisGOD:
I will push, but need to update integrity again, wait ....

SFisGOD:
@ppigazzini Please update DEV

ppigazzini:
DEV updated, looks good to me. Please double check, if it's all ok squash.

SFisGOD:
Maybe I should darken the pink border in dark mode

ppigazzini:
Feel free to experiment :)

SFisGOD:
Is this fine?
image

ppigazzini:
Yes.

On DEV I set `border-radius:5px;'` here, seems fine:
https://github.com/glinscott/fishtest/blob/9ef5981bdcbcd7e0d7aa5390aa2d2752d38ca02f/server/fishtest/templates/elo_results.mak#L12

ppigazzini:
More simple, delete `border-radius`:
```python
ret+='border-style:solid;border-color:Pink;border-width:medium;' 
```

@vdbergh do you squared the widget on purpose?

ppigazzini:
DEV updated, IMO better to delete `border-radius:` and keep the default.

SFisGOD:
image
`border-radius:10px` adds a rounded corner effect to better distinguish the time odds tests

SFisGOD:
> DEV updated, IMO better to delete `border-radius:` from both light/dark themes and keep the default corner rounding.

Okay wait

vdbergh:
@ppigazzini No not really. I actually had not noticed that the normal widgets are also rounded.  I am fine with rounding the border.

I also did not think of dark mode since I never use it. If different versions of the border will be used then perhaps it would be better to use a proper css class for it, instead of inline styling. 

I must confess I have not looked how dark mode works exactly.

SFisGOD:
@ppigazzini Is it fine now?

ppigazzini:
Yes.

SFisGOD:
@ppigazzini Squashed and force pushed :)

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @SFisGOD :)

PR #1014, opened 2021-09-04
Report the name of the worker when stopping a run

### LINKED ISSUE #1013: add worker ID on actions page
On the action page https://tests.stockfishchess.org/actions it would be useful to have the identifier of the worker that generates the wrong bench. Right now a noob worker seems to generate wrong benches from time to time and I suspect it is some hardware issue. Having the ID would help.

ppigazzini:
Fix #1013

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -9,7 +9,7 @@ from pyramid.httpexceptions import HTTPFound, HTTPUnauthorized, exception_respon
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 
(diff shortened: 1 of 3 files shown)


ppigazzini:
@vondele let me know if this message is fine for you:
"stop run: user00-3cores-ba6f8f1f : Wrong bench in stockfish_3bda3256056556d316d96d03ca7e2e2d97e010cb Expected: 5600615 Got: 56006151"

vondele:
that would probably be an improvement.

ppigazzini:
Tested ok on DEV with Linux/WSL2/msys2 workers.

PR #1017, opened 2021-09-06
Improve "requests" package import

ppigazzini:
The local package "requests" (v 2.11.1) imports the module "collections",
deprecated since Python 3.3 and dropped in Python 3.10, see
https://docs.python.org/3/library/collections.html

This patch:

- updates the "requests" package and his required ones to the newer version
  compatible with "collections.abc" and tested working with the old Python 3.4
- moves the local packages in a "packages" folder to lower their precedence
  respect to an eventually "requests" package installed by pip
- imports the local "requests" package only after an import exception

| Package  | Version   |
| -------- | --------- |
| certifi  | 2021.5.30 |
| chardet  | 4.0.0     |
| idna     | 2.10      |
| requests | 2.25.1    |
| urllib3  | 1.26.7    |


ppigazzini:
This is the relevant diff:

```diff
diff --git a/worker/worker.py b/worker/worker.py
index 41c82b2..bc3c86e 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -16,8 +16,6 @@ import time
 import traceback
 import uuid

-import requests
-
 IS_WINDOWS = "windows" in platform.system().lower()

 try:
@@ -34,6 +32,12 @@ from datetime import datetime
 from optparse import OptionParser
 from os import path

+try:
+    import requests
+except ModuleNotFoundError:
+    sys.path.append(path.join(path.dirname(path.realpath(__file__)), "pkgs"))
+    import requests
+
 from games import run_games
 from updater import update
```

tomtor:
Now that pip is a standard part of python3, it should be possible to do a "pip install requests" either from python code and restart or in a wrapping script?

It would be nice if we could use the in tree "requests" code only as a last resort.

ppigazzini:
I agree with the `python3 -m pip install requests` way.

This is the status quo:

- windows worker with msys2: has `requests` installed in a virtual environment, see
https://github.com/ppigazzini/worker-setup/blob/c925ea8593233f994706f634cbfa7081b9c46834/02_install_worker.sh#L44-L49

- linux worker or windows wsl worker installed with the suggested [wiki ubuntu setup script](https://github.com/glinscott/fishtest/wiki/Running-the-worker-on-Linux): has `requests` installed in a virtual environment
```bash
 # fishtest worker setup and first start only to write the "fishtest.cfg" configuration file
sudo -i -u ${worker_user} << EOF
python3 -m venv \${VENV}
\${VENV}/bin/python3 -m pip install --upgrade pip setuptools wheel
\${VENV}/bin/python3 -m pip install requests

\${VENV}/bin/python3 \${HOME}/fishtest/worker/worker.py --concurrency ${n_cores} ${usr_name} ${usr_pwd} --only_config True && echo "concurrency successfully set" || echo "Restart the script using a proper concurrency value"
EOF
```
- linux worker custom install, we should install `requests` based on the started python type:
  - with system python the preferred way is to install with `python3 -m pip install requests --user`
  - with virtual environment python the ways should be `python3 -m pip install requests`, the user mode don't work, see https://stackoverflow.com/questions/33412974/how-to-uninstall-a-package-installed-with-pip-install-user/56948334#56948334
At the moment I found only this info to run `pip` from a python script, https://www.activestate.com/resources/quick-reads/how-to-install-python-packages-using-a-script/

Please mind that now our local `requests` has the precedence respect to the `pip` installed `requests`, if we want to still provide that fall back solution, we must move the local `requests` in a folder as in this PR.

tomtor:
> if we still want to provide that fall back solution, we must move the local `requests` in a folder like in this PR.

In my opinion we should drop the requests fallback and tree, it was needed in the Python 2 era, but now every one is running Python 3 workers (which all have the "pip" module by default?) we should/could drop it.

ppigazzini:
>  now every one is running Python 3 workers (which all have the "pip" module by default?)

@tomtor the system python in ubuntu/debian lacks the modules `pip`, `ensurepip`, `venv`, that must be installed with the packet manager. So we could have some ubuntu/debian machines running the worker with the system python and without pip (e.g. the wiki AWS EC2 script).

ppigazzini:
Tested working fine with several OS/Python version:
 - Linux: from Python 3.4.3 to Python 3.10.rc2
 - WSL: from Python 3.8.12 to Python 3.10.rc2
 - Windows msys2: Python 3.9.6

PR #1018, opened 2021-09-06
Use multiples of 100 for freq.

xoto10:
The frequency modified here determines the scaling on the x-axis of graphs of tuning tests. Currently the scaling can be any multiple of 25, e.g. 375 or 425, which is not very intuitive for the user. Also, the last data point from the test is often not displayed on the graph as numbers like these tend not to divide exactly into the number of games in a test. This change ensures that the frequency is a multiple of 100 and divides neatly into 10,000, making the graph more intuitive for users.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1056,10 +1056,14 @@ class RunDb:
         if "param_history" not in spsa:
             spsa["param_history"] = []
         L = len(spsa["params"])
-        freq = L * 25
-        if freq < 100:
-            freq = 100
-        maxlen = 250000 / freq
+        freq_max = 250000
+        freq = max(100, min(L * 25, freq_max))
+
+        # adjust freq to multiples of 100 that divide exactly into freq_max
+        freq_div = (i for i in range(100, freq_max + 1, 100) if not freq_max % i)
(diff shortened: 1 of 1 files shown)


ppigazzini:
[DEV](https://dfts-0.pigazzini.it/tests) updated, scheduled some SPSAs.

ppigazzini:
@xoto10 the code breaks for `L<4`. If you chose 10000 to limit the computation, then a pythonic way to get the result is:
```python
freq_max = 250000
freq = max(100, min(L * 25, freq_max))

# adjust freq to multiples of 100 that divide exactly into freq_max
base = (i for i in range(100, freq_max + 1, 100) if not freq_max % i)
freq = [i for i in base if i <= freq][-1]

maxlen = freq_max / freq
```

By the way the original code need a check, IMO it was based on the python2 division by integer:

https://github.com/glinscott/fishtest/blob/58d3b95500d8eec145592589ed5e8b7f73c20fe7/server/fishtest/rundb.py#L1059-L1063

xoto10:
@ppigazzini 
I should have spotted the problem with L<4, my bad.
I'm not a regular python programmer nowadays, and was never an expert anyway, so no problem changing the code style as you prefer. I have updated the PR.

ppigazzini:

PR #1021, opened 2021-09-08
Dynamic tasks

vdbergh:
We start with an empty task list and create new tasks as needed.
    
This incorporates #1020. 
    
There is now an attempt to create tasks which take about 15 minutes to complete.

I did some limited local testing (randomingly starting and stopping workers).

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -41,16 +41,22 @@ def get_worker_key(task):
 
 def get_chi2(tasks, bad_users):
     """Perform chi^2 test on the stats from each worker"""
-    results = {"chi2": 0.0, "dof": 0, "p": 0.0, "residual": {}}
+
+    default_results = {
+        "chi2": float("nan"),
+        "dof": 0,
+        "p": float("nan"),
+        "residual": {},
+    }
 
     # Aggregate results by worker
     users = {}
     has_pentanomial = None
     for task in tasks:
-        task["worker_key"] = get_worker_key(task)
         if "worker_info" not in task:
             continue
         key = get_worker_key(task)
+        task["worker_key"] = key
         if key in bad_users:
             continue
         stats = task.get("stats", {})
@@ -63,61 +69,63 @@ def get_chi2(tasks, bad_users):
                 float(stats.get("draws", 0)),
             ]
         else:
-            p = stats["pentanomial"]
+            p = stats.get("pentanomial", 5 * [0])  # there was a small window
+            # in time where we could have both trinomial and pentanomial
+            # workers
+
             # The ww and ll frequencies will typically be too small for
             # the full pentanomial chi2 test to be valid. See e.g. the last page of
             # https://www.open.ac.uk/socialsciences/spsstutorial/files/tutorials/chi-square.pdf.
             # So we combine the ww and ll frequencies with the wd and ld frequencies.
             wld = [float(p[4] + p[3]), float(p[0] + p[1]), float(p[2])]
-        if wld == [0.0, 0.0, 0.0]:
-            continue
         if key in users:
             for idx in range(len(wld)):
                 users[key][idx] += wld[idx]
         else:
             users[key] = wld
 
-    if len(users) == 0:
-        return results
-
-    observed = numpy.array(list(users.values()))
-    rows, columns = observed.shape
-    # Results only from one worker: skip the test for workers homogeneity
-    if rows == 1:
-        return {"chi2": float("nan"), "dof": 0, "p": float("nan"), "residual": {}}
-    column_sums = numpy.sum(observed, axis=0)
-    columns_not_zero = numpy.count_nonzero(column_sums)
-    df = (rows - 1) * (columns - 1)
-
-    if columns_not_zero == 0:
-        return results
-    # Results only of one type: workers are homogeneous
-    elif columns_not_zero == 1:
-        return {"chi2": 0.0, "dof": df, "p": 1.0, "residual": {}}
-    # Results only of two types: drop the column of zeros to avoid divide by zero
-    elif columns_not_zero == 2:
-        observed = observed[:, ~numpy.all(observed == 0, axis=0)]
+    # We filter out the workers whose expected frequences are <= 5 as
+    # they break the chi2 test.
+    filtering_done = False
+    while not filtering_done:
+        # Wheneve less than two qualifying workers are left,
+        # we bail out and just return "something".
+        if len(users) <= 1:
+            return default_results
+        # Now do the matrix computations with numpy.
+        observed = numpy.array(list(users.values()))
+        rows, columns = observed.shape
         column_sums = numpy.sum(observed, axis=0)
+        row_sums = numpy.sum(observed, axis=1)
+        grand_total = numpy.sum(column_sums)
+        # if no games have been received, we cannot continue
+        if grand_total == 0:
+            return default_results
+        expected = numpy.outer(row_sums, column_sums) / grand_total
+        keys = list(users)
+        filtering_done = True
+        for idx in range(len(keys)):
+            if min(expected[idx]) <= 5:
+                del users[keys[idx]]
+                filtering_done = False
 
-    row_sums = numpy.sum(observed, axis=1)
-    grand_total = numpy.sum(column_sums)
-
-    expected = numpy.outer(row_sums, column_sums) / grand_total
+    # Now we do the basic chi2 computation.
+    df = (rows - 1) * (columns - 1)
     raw_residual = observed - expected
+    chi2 = numpy.sum(raw_residual * raw_residual / expected)
+    p_value = 1 - scipy.stats.chi2.cdf(chi2, df)
+
+    # Finally we also compute for each qualifying worker a residual
+    # indicating how badly it deviates from the norm.
     std_error = numpy.sqrt(
         expected
         * numpy.outer((1 - row_sums / grand_total), (1 - column_sums / grand_total))
     )
     adj_residual = raw_residual / std_error
-    for idx in range(len(users)):
-        users[list(users.keys())[idx]] = numpy.max(numpy.abs(adj_residual[idx]))
-    chi2 = numpy.sum(raw_residual * raw_residual / expected)
-    return {
-        "chi2": chi2,
-        "dof": df,
-        "p": 1 - scipy.stats.chi2.cdf(chi2, df),
-        "residual": users,
-    }
+    for idx in range(len(keys)):
+        users[keys[idx]] = numpy.max(numpy.abs(adj_residual[idx]))
+
+    return {"chi2": chi2, "dof": df, "p": p_value, "residual": users}
 
 
 def calculate_residuals(run):
@@ -131,7 +139,7 @@ def calculate_residuals(run):
         for task in run["tasks"]:
             if task["worker_key"] in bad_users:
                 continue
-            task["residual"] = residuals.get(task["worker_key"], 0.0)
+            task["residual"] = residuals.get(task["worker_key"], float("inf"))
 
             # Special case crashes or time losses
             stats = task.get("stats", {})
--- a/worker/games.py
+++ b/worker/games.py
@@ -339,7 +339,9 @@ def find_arch_string():
             and "x86-64-avx512" in targets
         ):
             res = "x86-64-avx512"
-            res = "x86-64-bmi2"  # use bmi2 until avx512 performance becomes actually better
+            res = (
+                "x86-64-bmi2"
+            )  # use bmi2 until avx512 performance becomes actually better
         elif (
             "-mbmi2" in props["flags"]
             and "x86-64-bmi2" in targets
@@ -840,7 +842,10 @@ def run_games(worker_info, password, remote, run, task_id):
     repo_url = run["args"].get("tests_repo", REPO_URL)
     games_concurrency = int(worker_info["concurrency"]) // threads
 
-    start_game_index = task_id * task["num_games"] + input_total_games
+    opening_offset = task.get("start", task_id * task["num_games"])
+    if "start" in task:
+        print("Variable task sizes used. Opening offset = {}".format(opening_offset))
+    start_game_index = opening_offset + input_total_games
     run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2 ** 30)
 
     # Format options according to cutechess syntax
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -154,6 +154,7 @@ class ApiView(object):
             min_task = {"num_games": task["num_games"]}
             if "stats" in task:
                 min_task["stats"] = task["stats"]
+            min_task["start"]=task["start"]
             min_run["my_task"] = min_task
         else:
             for task in run["tasks"]:
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -47,18 +47,13 @@ class RunDb:
         self.deltas = self.db["deltas"]
 
         self.chunk_size = 200
+        self.task_duration = 900 # 15 minutes
 
         global last_rundb
         last_rundb = self
 
     def generate_tasks(self, num_games):
-        tasks = []
-        remaining = num_games
-        while remaining > 0:
-            task_size = min(self.chunk_size, remaining)
-            tasks.append({"num_games": task_size, "pending": True, "active": False})
-            remaining -= task_size
-        return tasks
+        return []
 
     def new_run(
         self,
@@ -142,7 +137,9 @@ class RunDb:
             "tc_base": tc_base,
             "new_tc": new_tc_,
             "base_same_as_master": base_same_as_master,
-            # Will be filled in by tasks, indexed by task-id
+            # Will be filled in by tasks, indexed by task-id.
+            # Starts as an empty list.
+            # generate_tasks() is currently a dummy
             "tasks": self.generate_tasks(num_games),
             # Aggregated results
             "results": {"wins": 0, "losses": 0, "draws": 0, "pentanomial": 5 * [0]},
@@ -533,6 +530,18 @@ class RunDb:
 
     worker_runs = {}
 
+    def worker_cap(self, run, worker_info):
+        game_time = estimate_game_duration(run["args"]["tc"])
+        concurrency = worker_info["concurrency"] // run["args"]["threads"]
+        assert concurrency >= 1
+        games = self.task_duration / game_time * concurrency
+        if "sprt" in run["args"]:
+            batch_size = 2 * run["args"]["sprt"].get("batch_size", 1)
+            games = max(batch_size, batch_size * int(games / batch_size + 1 / 2))
+        else:
+            games = max(2, 2 * int(games / 2 + 1 / 2))
+        return games
+
     def request_task(self, worker_info):
         if self.task_semaphore.acquire(False):
             try:
@@ -635,38 +644,63 @@ class RunDb:
             ):
                 task_id = -1
                 cores = 0
+
+                committed_games = 0
+                run_size = 0
+                for task in run["tasks"]:
+                    run_size += task["num_games"]
+                    if not task["active"]:
+                        if "stats" in task:
+                            stats = task["stats"]
+                            committed_games += (
+                                stats["wins"] + stats["losses"] + stats["draws"]
+                            )
+                    else:
+                        committed_games += task["num_games"]
+
+                remaining = run["args"]["num_games"] - committed_games
+                if remaining <= 0:
+                    continue
+
                 if "spsa" in run["args"]:
                     limit_cores = 40000 / math.sqrt(len(run["args"]["spsa"]["params"]))
                 else:
-                    limit_cores = 1000000  # No limit for SPRT
+                    limit_cores = 1000000  # No limit for SPRT and num_games.
+
+                run_found = True
                 for task in run["tasks"]:
                     if task["active"]:
                         cores += task["worker_info"]["concurrency"]
                         if cores > limit_cores:
+                            run_found = False
                             break
-                    task_id = task_id + 1
-                    if not task["active"] and task["pending"]:
-                        task["worker_info"] = worker_info
-                        task["last_updated"] = datetime.utcnow()
-                        task["active"] = True
-                        run_found = True
-                        break
+
             if run_found:
                 break
 
         if not run_found:
             return {"task_waiting": False}
 
-        self.sum_cores(run)
-        self.task_runs.sort(
-            key=lambda r: (
-                -r["args"]["priority"],
-                r["cores"] / r["args"]["itp"] * 100.0,
-                -r["args"]["itp"],
-                r["_id"],
-            )
+        task_size = min(self.worker_cap(run, worker_info), remaining)
+        run["tasks"].append(
+            {
+                "num_games": task_size,
+                "pending": True,
+                "active": True,
+                "worker_info": worker_info,
+                "last_updated": datetime.utcnow(),
+                "start": run_size,
+            }
         )
+        task_id = len(run["tasks"]) - 1
+        run["tasks"][task_id]["stats"] = {
+            "wins": 0,
+            "losses": 0,
+            "draws": 0,
+            "pentanomial": 5 * [0],
+        }
 
+        self.sum_cores(run)
         self.buffer(run, False)
 
         # Update worker_runs (compiled tests)
@@ -675,13 +709,6 @@ class RunDb:
         if run["_id"] not in self.worker_runs[worker_key]:
             self.worker_runs[worker_key][run["_id"]] = True
 
-        if "stats" not in run["tasks"][task_id]:
-            run["tasks"][task_id]["stats"] = {
-                "wins": 0,
-                "losses": 0,
-                "draws": 0,
-                "pentanomial": 5 * [0],
-            }
         if DEBUG:
             print(
                 "Allocate run: https://tests.stockfishchess.org/tests/view/{} task_id: {} to {}/{} Stats: {}".format(
@@ -730,11 +757,31 @@ class RunDb:
         self, run_id, task_id, stats, nps, ARCH, spsa, username, unique_key
     ):
         run = self.get_run(run_id)
+        update_time = datetime.utcnow()
+
+        # First some sanity checks on the update
+        # If something is wrong we return early.
         if task_id >= len(run["tasks"]):
+            print("Update_task: task_id {} is invalid".format(task_id), flush=True)
             return {"task_alive": False}
 
         task = run["tasks"][task_id]
-        if not task["active"] or not task["pending"]:
+
+        # "pending": the task is not finished;
+        # "active" : a worker is working on this task.
+        # Currently tasks are created "pending" and "active"
+        # since they are created as a result of a worker request.
+        # If the worker walks away before the task is finished
+        # then the task becomes "pending" and "not active".
+        # It is possible to migrate such tasks to a different worker,
+        # but currently we do not do this.
+        if not task["pending"]:
+            print("Update_task: task {} is finished".format(task_id), flush=True)
+            task["active"] = False  # should be true already
+            return {"task_alive": False}
+
+        if not task["active"]:
+            print("Update_task: task {} is not active".format(task_id), flush=True)
             return {"task_alive": False}
 
         if task["worker_info"]["unique_key"] != unique_key:
@@ -745,6 +792,7 @@ class RunDb:
                 ),
                 flush=True,
             )
+            task["active"] = False
             return {"task_alive": False}
         if task["worker_info"]["username"] != username:
             print(
@@ -754,6 +802,7 @@ class RunDb:
                 ),
                 flush=True,
             )
+            task["active"] = False
             return {"task_alive": False}
 
         # Guard against incorrect results
@@ -768,58 +817,88 @@ class RunDb:
             or (spsa_games > 0 and num_games <= 0)
             or (spsa_games > 0 and "stats" in task and num_games <= old_num_games)
         ):
+            print("Update_task: task {} has incompatible stats".format(task_id))
+            print(
+                "Before {}. Now {}. SPSA_games {}.".format(
+                    old_num_games, num_games, spsa_games
+                ),
+                flush=True,
+            )
+            task["active"] = False
             return {"task_alive": False}
         if (
             num_games - old_num_games
-        ) % 2 != 0:  # the worker should only runs game pairs
+        ) % 2 != 0:  # the worker should only return game pairs
+            print(
+                "Update_task: odd number of games received for task {}. Before {}. Now {}.".format(
+                    task_id, old_num_games, num_games
+                )
+            )
+            task["active"] = False
             return {"task_alive": False}
         if "sprt" in run["args"]:
             batch_size = 2 * run["args"]["sprt"].get("batch_size", 1)
             if num_games % batch_size != 0:
+                print(
+                    "Update_task: the number of games received for task {} is incompatible with the SPRT batch size"
+                ).format(task_id)
+                task["active"] = False
                 return {"task_alive": False}
 
-        all_tasks_finished = False
+        # The update seems fine. Update run["tasks"][task_id] (=task).
 
         task["stats"] = stats
         task["nps"] = nps
         task["ARCH"] = ARCH
+        task["last_updated"] = update_time
+
+        task_finished = False
         if num_games >= task["num_games"]:
             # This task is now finished
-            if "cores" in run:
-                run["cores"] -= task["worker_info"]["concurrency"]
+            task_finished = True
             task["pending"] = False  # Make pending False before making active false
             # to prevent race in request_task
             task["active"] = False
-            # Check if all tasks in the run have been finished
-            if not any([t["pending"] or t["active"] for t in run["tasks"]]):
-                all_tasks_finished = True
 
-        update_time = datetime.utcnow()
-        task["last_updated"] = update_time
+        # Now update the current run.
+
         run["last_updated"] = update_time
-        run["results_stale"] = True
 
-        # Update SPSA results
+        if task_finished:
+            run["cores"] -= task["worker_info"]["concurrency"]
+
+        run["results_stale"] = True  # force recalculation of results
+        updated_results = self.get_results(
+            run, False
+        )  # computed from run["tasks"] which
+        # has just been updated. Sets run["results_stale"]=False.
+
+        if "sprt" in run["args"]:
+            sprt = run["args"]["sprt"]
+            fishtest.stats.stat_util.update_SPRT(updated_results, sprt)
+
         if "spsa" in run["args"] and spsa_games == spsa["num_games"]:
             self.update_spsa(task["worker_info"]["unique_key"], run, spsa)
 
-        # Check SPRT state to decide whether to stop the run
-        if "sprt" in run["args"]:
-            sprt = run["args"]["sprt"]
-            fishtest.stats.stat_util.update_SPRT(self.get_results(run, False), sprt)
-            if sprt["state"] != "":
-                # If SPRT is accepted or rejected, stop the run
-                self.buffer(run, True)
-                self.stop_run(run_id)
-                return {"task_alive": False}
+        # Check if the run is finished.
 
-        if all_tasks_finished:
-            # If all tasks are finished, stop the run
+        run_finished = False
+        if count_games(updated_results) >= run["args"]["num_games"]:
+            run_finished = True
+        elif "sprt" in run["args"] and sprt["state"] != "":
+            run_finished = True
+
+        # Return.
+
+        if run_finished:
             self.buffer(run, True)
             self.stop_run(run_id)
+            ret = {"task_alive": False}
         else:
             self.buffer(run, False)
-        return {"task_alive": task["active"]}
+            ret = {"task_alive": task["active"]}
+
+        return ret
 
     def upload_pgn(self, run_id, pgn_zip):
         self.pgndb.insert_one({"run_id": run_id, "pgn_zip": Binary(pgn_zip)})
@@ -837,14 +916,11 @@ class RunDb:
             )
         if task_id >= len(run["tasks"]):
             return {"task_alive": False}
-
         if not task["active"] or not task["pending"]:
             return {"task_alive": False}
-
         # Test if the task is reallocated to a different worker.
         if task["worker_info"]["unique_key"] != unique_key:
             return {"task_alive": False}
-
         # Mark the task as inactive: it will be rescheduled
         print(
             "Inactive: https://tests.stockfishchess.org/tests/view/{} {} {}".format(
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -306,7 +306,12 @@
    <td>${stats.get('time_losses', '-')}</td>
 
    %if 'spsa' not in run['args']:
-   <td style="background-color:${task['residual_color']}">${'%.3f' % (task['residual'])}</td>
+   %if task['residual']!=float("inf"):
+       <td style="background-color:${task['residual_color']}">${'%.3f' % (task['residual'])}</td>
+   %else:
+##       <td>${'%.3f' % (task['residual'])}</td>
+         <td>-</td>
+   %endif
    %endif
   </tr>
   %endfor
--- a/server/tests/test_api.py
+++ b/server/tests/test_api.py
@@ -38,6 +38,16 @@ class TestApi(unittest.TestCase):
 
         # Set up a task
         self.task_id = 0
+
+        task = {
+            "num_games": self.rundb.chunk_size,
+            "stats": {"wins": 0, "draws": 0, "losses": 0, "crashes": 0},
+            "pending": True,
+            "active": True,
+        }
+
+        run["tasks"].append(task)
+
         for i, task in enumerate(run["tasks"]):
             if i is not self.task_id:
                 run["tasks"][i]["pending"] = False
@@ -59,6 +69,7 @@ class TestApi(unittest.TestCase):
             "unique_key": "unique key",
             "version": WORKER_VERSION,
         }
+        run["tasks"][0]["worker_info"] = self.worker_info
         self.rundb.userdb.create_user(self.username, self.password, "email@email.email")
         user = self.rundb.userdb.get_user(self.username)
         user["blocked"] = False
@@ -139,11 +150,11 @@ class TestApi(unittest.TestCase):
         request = self.correct_password_request({"worker_info": self.worker_info})
         response = ApiView(request).request_task()
         self.assertEqual(self.run_id, response["run"]["_id"])
-        self.assertEqual(self.task_id, response["task_id"])
+        self.assertNotEqual(self.task_id, response["task_id"])
 
         run = self.rundb.get_run(self.run_id)
         self.assertEqual(run["cores"], self.concurrency)
-        task = run["tasks"][self.task_id]
+        task = run["tasks"][response["task_id"]]
         self.assertTrue(task["pending"])
         self.assertTrue(task["active"])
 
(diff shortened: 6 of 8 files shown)


ppigazzini:
DEV updated.

ppigazzini:
Dev updated.

ppigazzini:
DEV Updated.

ppigazzini:
Dev updated.

vdbergh:
One test to do at some point would be to start some tests under master and then switch to this branch. Normally this should be seamless. The current branch should simply ignore the pregenerated tasks and append new ones at the end of the task list.

ppigazzini:
Created some tests with DEV/master, running for a while, switched to DEV/PR.

If I'm not wrong with the PR workflow we could have some low counting frequencies (LTC batches, stopped worker etc.), the chi squared test code could need an update.

vdbergh:
First of all I would set `task_duration=3600`. Then almost all workers will have a reasonable amount of games in normal circumstances, even at LTC. 

Put it may happen that people start the worker to try it out and then quickly kill it. Without task migration this will lead to workers with very few games, or even none at all.

Probably we should filter such workers in the chi2 test. I have not considered how yet.

It would not be very difficult to re-enable task migration but people don't like this very much. It solves the chi2 problem by cheating.

vondele:
btw, thanks for this, I think that's the right direction.

ppigazzini:
> Probably we should filter such workers in the chi2 test. I have not considered how yet.

This was my first thought, but IMO is unnecessary: the "low count" workers shouldn't affect so much the p-value and in any case they should be the candidates to be purged.
 
> It would not be very difficult to re-enable task migration but people don't like this very much. It solves the chi2 problem by cheating.

I dislike task migration too.

vdbergh:
> > Probably we should filter such workers in the chi2 test. I have not considered how yet.
> 
> This was my first thought, but IMO is unnecessary: the "low count" workers shouldn't affect so much the p-value and in any case they should be the candidates to be purged.
> 

The low count workers have little effect on the outcome of the test (if there aren’t too many). However they contribute on average the same amount to the chi2 value as any other worker (but their contribution does not follow a chi2 distribution, so they break the statistics).

Therefore it makes sense I think to filter such workers. I looked at the code and the code to filter is already there. In fact we are already filtering workers with 0 games

> > It would not be very difficult to re-enable task migration but people don't like this very much. It solves the chi2 problem by cheating.
> 
> I dislike task migration too.

server/fishtest/util.py
```diff
@@ -63,20 +64,28 @@ def get_chi2(tasks, bad_users):
...
         else:
             users[key] = wld
 
+    users_ = copy.deepcopy(users)
+    for key in users_:
+        if min(users[key]) < 10:  # the normal rule is expected >= 5
+            # but this is inconvenient since we haven't computed the
+            # expected frequencies at this point
+            del users[key]
+
```
ppigazzini, line 88: IMO this filtering is helpful If we want only to use chi2 only like a meter.
This filtering is problematic if we want to use chi2 to auto purge a tests because (if I read correctly the code) in this way we don't compute the residuals for the low count workers, keeping all the low count workers results (low total of course).
In a test with few workers at high concurrency and many workers at very low concurrency with low cell count, the former will dominate the expected cell count (so under H0 they will have low residuals and low chi2 contribution), the latter could have high residuals lowering/polluting the p-value (if used like a meter). But in that case the low concurrency workers are the candidates to be purged if the p-value is less than the critical one.

ppigazzini:
> The low count workers have little effect on the outcome of the test (if there aren’t too many). However they contribute on average the same amount to the chi2 value as any other worker (but their contribution does not follow a chi2 distribution, so they break the statistics).

We should decide the purpose of the chi2 test statistic under H0 of workers uniformity:
 - a trigger to auto purge the workers if the test rejects H0: in that case the low count workers should have high residuals
 - a test meter, filtering out the low count workers but saving them form the auto purge if the test rejects H0
 
By the way we can also compute the two different chi2 test statistics.

server/fishtest/util.py
```diff
@@ -87,7 +96,7 @@ def get_chi2(tasks, bad_users):
         return {"chi2": float("nan"), "dof": 0, "p": float("nan"), "residual": {}}
     column_sums = numpy.sum(observed, axis=0)
     columns_not_zero = numpy.count_nonzero(column_sums)
-    df = (rows - 1) * (columns - 1)
+    df = (rows - 1) * (columns_not_zero - 1)  # we filter zero columns
```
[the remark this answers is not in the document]
vdbergh: I do not understand this comment. If we delete zero columns, then the df needs to be adapted.... The df can change during the test if a zero column becomes non-zero.

server/fishtest/util.py
vdbergh: I don't think that low count workers need to have bad data. Also they don't have to have low core count. They may be high core count workers that have been stopped prematurely. 

I have been looking at earlier tests and with the new book there are very few workers that don't have enough games to be included in the chi2 test, even at LTC.

I don't think a standardized residual for a worker with low cell count makes a lot of sense since the distribution is not normal. Once could try to be more accurate but since such workers contribute little to the outcome of the test I think it is not worth it.

vdbergh:
The chi2 test was created for people with misconfigured machines. I.e. people that contribute large amounts of games whose outcome distribution differs substantially from the rest of the workers (probably because the machines are overloaded). This can skew the outcome of a test. After the chi2 test was implemented it turned out that a large contributor had residuals of 7.

IMHO workers that produce only a few games are not the problem.

vdbergh:
In the latest commit I am filtering workers on their expected frequencies (they should be >=5). This is the recommended practice. It is much less drastic than the earlier quick hack. At LTC a worker should contribute >=50 games to qualify (approximately). That is a quite low bar IMHO. A worker that contributes <= 50 games to a test of 100K games should have little impact.

I am sorry to continue in this direction, but I really don't believe that incorporating workers with too few games in the  chi2 test (including the residuals) makes sense statistically. 

One thing we could possibly do is to pool all low count workers in one pseudo worker.

ppigazzini:
DEV updated.
I'm wondering if we can assign a LCT task only to workers with a minimum throughput (based on base_nps and concurrency).

ppigazzini:
@vdbergh something strange happened, see https://dfts-0.pigazzini.it/tests/view/613b86fced182e1ef94a3cd4

vdbergh:
> @vdbergh something strange happened, see https://dfts-0.pigazzini.it/tests/view/613b86fced182e1ef94a3cd4
> 
I don't see what's wrong...

> EDIT_000: ok, tasks grouped by worker, the change it's a bit extreme, though, and we loose the grouping after ordering the table according to one of the table header (clicking on the table header, eg Idx, Worker, Info)

Ah ok. I was not aware of the sorting when clicking on headers. So we can already sort by worker. Still it's a pity that the table contains so much repeated information in that case.

ppigazzini:
DEV updated.

vdbergh:
I have reverted the change. Concerning further tweaking of the task size/selection. I think it might be best to wait a little bit to see what issues there are in practice. The request_task code is somewhat messy right now and it should be cleaned up. But I don't feel like tackling this right now.

PS. IIUC the base_nps is only known if the worker has already completed a task.

vdbergh:
Perhaps the table should be sorted by worker by default. Task id is just some internal index which has no meaning to the user.

vdbergh:
I wonder by what magic the table sorting happens...

ppigazzini:
> Perhaps the table should be sorted by worker by default. Task id is just some internal index which has no meaning to the user.

I have no preference here, the sorting by header is a click away. IMO the only practical use of Idx sorting is to group the running workers.

> PS. IIUC the base_nps is only known if the worker has already completed a task.

Right, I need some coffee :)

> I wonder by what magic the table sorting happens...

My knowledge of server code is very poor :(

server/fishtest/util.py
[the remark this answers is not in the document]
ppigazzini: I was asking if the dof is a property of the experiment or a property of the contingency table.
We never expect to have a zero column for a complete NumGames test master vs master at LTC. If this should happen I'm wondering if is correct to change the dof.

vdbergh:
Instead of sorting per worker, displaying summary table per worker would be a good idea. Then we can also summarize the stats. This would allow for some auditing in case of large residuals. But that's a bit more work, so it better be a separate PR.

ppigazzini:
Ok, drop the WIP when the PR is ready to be merged (@vondele suggestions?)

vdbergh:
The latest version also contains a tiny worker change that should fix the issue raised by @vondele.

If it working then the worker should print "Variable task sizes used.." after starting to work on a new task.

vdbergh:
And [WIP] removed.

BTW. I changed `task_duration` to 3600. I hope that's ok.

vondele:
I would suggest to make task_duration somewhat smaller, that should guarantee that new jobs are picked up quickly. I liked the 900s. But it is no big deal.

I have another question though, what happens on worker failure? Imagine a worker that fails to compile and repeatedly gets assigned the same test, would the task list grow indefinitely, and at high rate? Do we need some mechanism to protect against this case?

ppigazzini:
DEV and workers updated.

vdbergh:
> I would suggest to make task_duration somewhat smaller, that should guarantee that new jobs are picked up quickly. I liked the 900s. But it is no big deal.
> 

Even with 1 hour I think this would still happen. Since the different calls to request_task will be more or less homogeneously spread over 1 hour.

> I have another question though, what happens on worker failure? Imagine a worker that fails to compile and repeatedly gets assigned the same test, would the task list grow indefinitely, and at high rate? Do we need some mechanism to protect against this case?

That might be a genuine issue.  What to do when someone submits a test with a non compiling stockfish?

ppigazzini:
> That might be a genuine issue. What to do when someone submits a test with a non compiling stockfish?

My solution https://github.com/glinscott/fishtest/pull/992 backfired because there was a big contributor with also 2 not working workers on raspberry pi. We should stop a test only if at least N different users report the problem.

vondele:
a simple guard would be to have a time delay that increases on failure. Something like:
```diff
diff --git a/worker/worker.py b/worker/worker.py
index 41c82b2..e837fe3 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -508,7 +508,10 @@ def main():
         if not success:
             if FLEET:
                 worker_exit()
-            time.sleep(HTTP_TIMEOUT)
+            time.sleep(delay)
+            delay = delay * 2
+        else:
+            delay = HTTP_TIMEOUT;
         success = worker(worker_info, password, remote)
```

vondele:
without the task migration, these failing workers will show up as empty tasks for the same worker (I guess), so that will simplify identification.

vdbergh:
```
index 41c82b2..e837fe3 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -508,7 +508,10 @@ def main():
         if not success:
             if FLEET:
                 worker_exit()
-            time.sleep(HTTP_TIMEOUT)
+            time.sleep(delay)
+            delay = delay * 2
+        else:
+            delay = HTTP_TIMEOUT;
         success = worker(worker_info, password, remote)
```
Thanks. That simple indeed! I put it in.

ppigazzini:
behavior with a worker instrumented to fail the compile:

<html>
<body>


133 | user00-24cores-5-dea504a8 | Linux 5.4.0-81-generic ARCH=64bit SSE41 SSSE3 SSE2 POPCNT | 2021-09-11 13:02:33 | 1344 / 5590 | [6, 151, 351, 155, 9] | 0 | 0 | 0.623
-- | -- | -- | -- | -- | -- | -- | -- | --
134 | user00-7cores-6-5f308027 | Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=? | 2021-09-11 13:02:03 | 000 / 1630 | [0, 0, 0, 0, 0] | - | - | -
135 | user00-7cores-7-de9633a5 | Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=? | 2021-09-11 13:02:13 | 000 / 1630 | [0, 0, 0, 0, 0] | - | - | -
136 | user00-7cores-7-de9633a5 | Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=? | 2021-09-11 13:02:33 | 000 / 1630 | [0, 0, 0, 0, 0] | - | - | -
137 | user00-7cores-7-de9633a5 | Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=? | 2021-09-11 13:02:51 | 000 / 1630 | [0, 0, 0, 0, 0] | - | - | -


</body>
</html>

vdbergh:
@ppigazzini The latest push has vondele's fix.

ppigazzini:
> @ppigazzini The latest push has vondele's fix.

Ok, updated, this is the progression of the failing worker at 8 cores:

153	user00-8cores-9-28dd786b	Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=?	2021-09-11 13:08:52	000 / 1864	[0, 0, 0, 0, 0]	-	-	-
154	user00-8cores-10-8f9c79fa	Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=?	2021-09-11 13:09:02	000 / 1864	[0, 0, 0, 0, 0]	-	-	-
155	user00-8cores-10-8f9c79fa	Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=?	2021-09-11 13:09:22	000 / 1864	[0, 0, 0, 0, 0]	-	-	-
156	user00-8cores-10-8f9c79fa	Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=?	2021-09-11 13:09:55	000 / 1864	[0, 0, 0, 0, 0]	-	-	-
157	user00-8cores-10-8f9c79fa	Linux 5.10.16.3-microsoft-standard-WSL2 ARCH=?	2021-09-11 13:10:58	000 / 1864	[0, 0, 0, 0, 0]	-	-	-

vdbergh:
This seems ok right? Given that we are dealing with a situation that hopefully is exceptional.

vondele:
I think we can start with this, I suspect it is less exceptional than we think. Previously this behavior was hidden, now it will become visible, which is good, as it is easier to fix once it is visible.

ppigazzini:
If I'm not wrong a test with a non compiling stockfish could shutdown all the workers, so #992 should be improved adding here that at least N different users have failed building stockfish:

https://github.com/glinscott/fishtest/blob/6b2ad289693b9b451479296ce69aa569dc1c0719/server/fishtest/api.py#L226-L240

vondele:
a non-compiling test would be noticed quite quickly, and could be stopped by hand. 

But yes, I agree that the alternative is to stop it as soon as it doesn't compile on N different workers. This should be independent from this PR, I suggest.

ppigazzini:
> a non-compiling test would be noticed quite quickly, and could be stopped by hand. 

Some weeks ago we had some SPSA tests with code failing only for linux workers because the "pedantic" flag (missing on windows workers), never noticed for years I suppose :)

> But yes, I agree that the alternative is to stop it as soon as it doesn't compile on N different workers. This should be independent from this PR, I suggest.

I agree.

vondele:
> > a non-compiling test would be noticed quite quickly, and could be stopped by hand.
> 
> Some weeks ago we had some SPSA tests with code failing only for linux workers because the "pedantic" flag (missing on windows workers), never noticed for years I suppose :)

yes, now they will become visible as 0-sized tasks.

ppigazzini:
DEV on MASTER and workers on PR, then switch DEV on PR looks good, see https://dfts-0.pigazzini.it/tests/view/613cb5c3db27ed0516f88f53

IMO the safest updated is:
- merge the PR
- trigger the workers update
- when most/all workers are running the new version, switch PROD on PR

ppigazzini:
> BTW. I changed `task_duration` to 3600. I hope that's ok.

I will update the script to lower `task_duration` on DEV ;)

vdbergh:
I do not feel strongly about 3600, and the added interactivity of 900 is indeed appealing. 

But I was a bit afraid that with 900 we might have too many small tasks at LTC (in depends on the number of low core workers). The task list of course takes disk space (and remember that now tasks are not recycled). 

Of course the pre-generated tasks (2000 for SPRT) also took disk space, but perhaps less as these tasks were not initialized. 

Anyway we could try with 900 for a while and if the disk space is an issue then we can switch to 3600 (or something else). Let me know what you think.

vdbergh:
> DEV on MASTER and workers on PR, then switch DEV on PR looks good, see https://dfts-0.pigazzini.it/tests/view/613cb5c3db27ed0516f88f53
> 
> IMO the safest updated is:
> 
> * merge the PR
> * trigger the workers update
> * when most/all workers are running the new version, switch PROD on PR

Yes that sounds like a good plan!

ppigazzini:
> I do not feel strongly about 3600, and the added interactivity of 900 is indeed appealing.
> 
> But I was a bit afraid that with 900 we might have too many small tasks at LTC (in depends on the number of low core workers). The task list of course takes disk space (and remember that now tasks are not recycled).

I was joking! On DEV I use USTC to speedup the framework, so in any case I will lower to something like 360 :)

vondele:
sounds like a plan... let's see. I think 900s will be fine, that might, for STC, already lead to actually fewer tasks being created.

ppigazzini:
@vdbergh please raise the worker version:
https://github.com/glinscott/fishtest/blob/6b2ad289693b9b451479296ce69aa569dc1c0719/worker/worker.py#L40
https://github.com/glinscott/fishtest/blob/6b2ad289693b9b451479296ce69aa569dc1c0719/server/fishtest/api.py#L12

worker/games.py
```diff
@@ -840,7 +840,10 @@ def run_games(worker_info, password, remote, run, task_id):
     repo_url = run["args"].get("tests_repo", REPO_URL)
     games_concurrency = int(worker_info["concurrency"]) // threads
 
-    start_game_index = task_id * task["num_games"] + input_total_games
+    opening_offset=task.get("start",task_id * task["num_games"])
```
ppigazzini, line 843: space before comma,

vdbergh:
Ok worker version raised to 107, and task_duration lowered to 900.

worker/games.py
vdbergh: Ok. I think black fixed it. Contained in latest push.

ppigazzini: Thank you

ppigazzini (approved):

ppigazzini:
Triggered the workers update.

ppigazzini:
@linrock This worker is illegal :)
<html>
<body>


Machine | Cores | UUID | MNps | System | Version | Running on | Last updated
-- | -- | -- | -- | -- | -- | -- | --
linrock | 100 | 66b14234 | 0.00 | Linux 5.4.0-80-generic | 107:3.8.10 | newdatad8^ | seconds ago


</body>
</html>

vondele:
let's see how large the batch size will be on that one :-)

vondele:
it will definitely be used more efficiently.

linrock:
> @linrock This worker is illegal :)

oops what's the issue with this one? i'll shut it down in the meantime

vdbergh:
Oops. I did not raise the worker version on the server. Yet workers are updating. What's happening?

vondele:
> > @linrock This worker is illegal :)
> 
> oops what's the issue with this one? i'll shut it down in the meantime

I guess @ppigazzini was joking... it is the largest machine on fishtest.

vdbergh:
Ok I guess @ppigazzini raised the worker version manually on the previous master (still running on PROD). So should I make a separate PR to raise the worker version server side on the new master?

linrock:
> I guess @ppigazzini was joking... it is the largest machine on fishtest.

lol in that case, i'll bring it back and we'll see how it performs!

linrock:
seeing each task crash with these errors though. the worker is on latest master as of now:

image

vondele:
do you have a larger part of the log? Seemingly a problem starting the engine.

ppigazzini:
PROD updated to master.

vdbergh:
@ppigazzini But what about the worker version on the server still being 106 (cfr my comments above)?

ppigazzini:
> @ppigazzini But what about the worker version on the server still being 106 (cfr my comments above)?

Oh, I thought that you raised that too. No problem, raised by hand on the server, now I'll write a pr, thank you for the info.

linrock:
> do you have a larger part of the log? Seemingly a problem starting the engine.

here's a log of an entire failed task from start task until the error:
```
CPU factor : 1.1228653861149378 - tc adjusted to 67.372+0.674
CPU factor : 1.1228653861149378 - tc adjusted to 67.372+0.674                                                                                                                           Running srchsmalltune1 vs master                                                                                                                                                        ['/root/bench/fishtest/worker/testing/cutechess-cli', '-repeat', '-games', '488', '-tournament', 'gauntlet', '-pgnout', 'results-673edace-cd7e-4d80-991f-19ff7a5b9e2c.pgn', '-site', 'https://tests.stockfishchess.org/tests/view/613c6a12689039fce12e12d2', '-event', 'Batch 2000: srchsmalltune1 vs master', '-srand', '278824882', '-resign', 'movecount=3', 'score=400', '-draw', 'movenumber=34', 'movecount=8', 'score=20', '-concurrency', '100', '-openings', 'file=UHO_XXL_+0.90_+1.19.epd', 'format=epd', 'order=random', 'plies=16', 'start=200001', '-engine', 'name=New-df74cad847', 'tc=67.372+0.674', 'cmd=stockfish_df74cad8475f023a071637861e182e3238d66cb8', 'option.Hash=64', 'option.Use NNUE=true', 'option.EvalFile=/root/bench/fishtest/w
orker/testing/nn-6762d36ad265.nnue', '-engine', 'name=Base-30fdbf4328', 'tc=67.372+0.674', 'cmd=stockfish_30fdbf43283eb0d2f8e5ca325877d3e514758804', 'option.Hash=64', 'option.Use NNUE=
true', 'option.EvalFile=/root/bench/fishtest/worker/testing/nn-6762d36ad265.nnue', '-each', 'proto=uci', 'option.Threads=1']
TC limit 1643.8749252722691 End time: 2021-09-11 16:41:26.764856
Indexing opening suite...
Started game 25 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 26 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 1 of 488 (New-df74cad847 vs Base-30fdbf4328)                                                                                                                               Started game 3 of 488 (New-df74cad847 vs Base-30fdbf4328)                                                                                                                               Started game 2 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 5 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 4 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 9 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 10 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 20 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 8 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 6 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 7 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 15 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 12 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 11 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 19 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 30 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 18 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 17 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 16 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 22 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 13 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 21 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 32 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 14 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 34 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 23 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 27 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 29 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 33 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 31 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 24 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 35 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 28 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 37 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 41 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 43 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 36 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 38 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 40 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 42 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 39 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 44 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 56 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 46 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 47 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 55 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 45 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 57 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 51 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 48 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 50 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 60 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 49 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 52 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 53 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 54 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 59 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 58 of 488 (Base-30fdbf4328 vs New-df74cad847)
Started game 61 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 67 of 488 (New-df74cad847 vs Base-30fdbf4328)
Started game 69 of 488 (New-df74cad847 vs Base-30fdbf4328)
Warning: QProcessPrivate::createPipe: Cannot create pipe 0x7f5a20004768: Too many open files
Finished game 66 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 62 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 63 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 73 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 72 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 68 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 87 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 84 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 88 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 89 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 86 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 80 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 65 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 83 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 85 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 79 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 82 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 78 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 90 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 64 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 75 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 91 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 76 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 77 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 92 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 74 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 8 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 71 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 81 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 38 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 35 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 17 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 70 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 48 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 53 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 56 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 3 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 52 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 18 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 22 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 6 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 30 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 26 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 28 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 9 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 37 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 40 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 33 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 49 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 61 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 21 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 5 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 25 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 60 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 34 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 31 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 44 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 45 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 50 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 32 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 39 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 57 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 46 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 19 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 10 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 41 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 15 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 55 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 54 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 29 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 24 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 20 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 2 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 1 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 43 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 58 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 11 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 67 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 14 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 23 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 36 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 12 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 7 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 59 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 13 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 4 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 16 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 47 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 27 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 51 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 42 (Base-30fdbf4328 vs New-df74cad847): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Finished game 69 (New-df74cad847 vs Base-30fdbf4328): * {No result}
Score of New-df74cad847 vs Base-30fdbf4328: 0 - 0 - 0  [nan] 0
Elo difference: nan +/- nan, LOS: nan %, DrawRatio: nan %
Warning: Cannot start engine New-df74cad847:
Cannot execute command: stockfish_df74cad8475f023a071637861e182e3238d66cb8
Finished match
Exception running games
  File "/root/bench/fishtest/worker/games.py", line 783, in launch_cutechess
    task_state = parse_cutechess_output(
  File "/root/bench/fishtest/worker/games.py", line 620, in parse_cutechess_output
    assert num_games_updated == games_to_play
AssertionError
Wait 7 seconds before upload of PGN...
Uploading compressed PGN of 3316 bytes
Task exited
Fetch task...

API call rate limits: {'limit': 60, 'remaining': 56, 'reset': 1631378779, 'used': 4, 'resource': 'core'}
```

linrock:
ah i just saw this in the logs, so maybe a ulimit issue on the worker side:
`Warning: QProcessPrivate::createPipe: Cannot create pipe 0x7f5a20004768: Too many open files`

vondele:
@linrock one reason could be that with 100 concurrent games some of the limits are exceeded.

Indeed `ulimit -n` is too small

linrock:
yep just confirmed `ulimit -n 999999` fixes the error!

vondele:
I've mentioned these limits on the wiki page now..

vondele:
as far as I can see, things run smoothly now, and the dynamic sizes we have now look like they work. Thanks @vdbergh and @ppigazzini !

ppigazzini:
Great job @vdbergh :)

tomtor:
Great job! @vdbergh and @ppigazzini !

vdbergh:
One thing that should be stressed is that without the DEV server and the army of workers that populates it, it would be almost impossible to make non-trivial changes to Fishtest. Thanks @ppigazzini  !

ppigazzini:
@vdbergh only to point out the open issues solved by this PR :D

PR #1026, opened 2021-09-12
Resort the task queue after a task has been assigned.

vdbergh:
Bugfix! I broke Joona's algorithm :frowning_face: 

Resorting is important because the task queue is only regenerated every minute.

The resorting was accidentally deleted in the commit "dynamic_tasks".

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -47,7 +47,7 @@ class RunDb:
         self.deltas = self.db["deltas"]
 
         self.chunk_size = 200
-        self.task_duration = 900 # 15 minutes
+        self.task_duration = 900  # 15 minutes
 
         global last_rundb
         last_rundb = self
@@ -554,6 +554,13 @@ class RunDb:
             return {"task_waiting": False}
 
     def sync_request_task(self, worker_info):
+        priority = lambda r: (
+            -r["args"]["priority"],
+            r["cores"] / r["args"]["itp"] * 100.0,
+            -r["args"]["itp"],
+            r["_id"],
+        )
+
         if time.time() > self.task_time + 60:
             if DEBUG:
                 print("Refresh queue", flush=True)
@@ -563,14 +570,7 @@ class RunDb:
                 self.sum_cores(run)
                 self.calc_itp(run)
                 self.task_runs.append(run)
-            self.task_runs.sort(
-                key=lambda r: (
-                    -r["args"]["priority"],
-                    r["cores"] / r["args"]["itp"] * 100.0,
-                    -r["args"]["itp"],
-                    r["_id"],
-                )
-            )
+            self.task_runs.sort(key=priority)
             self.task_time = time.time()
 
         max_threads = int(worker_info["concurrency"])
@@ -693,14 +693,18 @@ class RunDb:
             }
         )
         task_id = len(run["tasks"]) - 1
-        run["tasks"][task_id]["stats"] = {
-            "wins": 0,
-            "losses": 0,
-            "draws": 0,
-            "pentanomial": 5 * [0],
-        }
+        task = run["tasks"][task_id]
+        task["stats"] = {"wins": 0, "losses": 0, "draws": 0, "pentanomial": 5 * [0]}
 
         self.sum_cores(run)
+
+        # Resort the queue to take into account
+        # the alterned number of cores of "run".
+        # Resorting is important since the queue
+        # is only regenerated every 60 seconds.
+
+        self.task_runs.sort(key=priority)
+
         self.buffer(run, False)
 
         # Update worker_runs (compiled tests)


ppigazzini:
DEV updated, joined some workers, started few test with throughput 25-50-100%

ppigazzini:
All tests with priority 9, throughput 25-50-100% (100% not indicated)
<html>
<body>
<h3 style="margin: 10px 0px; font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-weight: bold; line-height: 40px; color: rgb(177, 173, 167); text-rendering: optimizelegibility; font-size: 24.5px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(17, 17, 17); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Active - 9 tests</h3>

21-09-12 | use | master | diff | LLR: 0.77 (-2.94,2.94) <-0.50,2.50>Total: 1312 W: 440 L: 375 D: 497Ptnml(0-2): 34, 130, 282, 157, 53 | sprt @ 0.5+0.005 th 1cores: 4 | #1026 SPRT - throughput 25%
-- | -- | -- | -- | -- | -- | --
21-09-12 | use | master | diff | LLR: 0.22 (-2.94,2.94) <-0.50,2.50>Total: 1136 W: 365 L: 344 D: 427Ptnml(0-2): 35, 134, 219, 135, 45 | sprt @ 0.5+0.005 th 1cores: 5 | #1026 NumGames - Throughput 25%
21-09-12 | use | master | diff | LLR: -0.06 (-2.94,2.94) <-0.50,2.50>Total: 2080 W: 652 L: 653 D: 775Ptnml(0-2): 61, 242, 437, 237, 63 | sprt @ 0.5+0.005 th 1cores: 6 | #1026 NumGames - Throughput 50%
21-09-12 | use | master | diff | LLR: -0.14 (-2.94,2.94) <-0.50,2.50>Total: 2016 W: 640 L: 647 D: 729Ptnml(0-2): 74, 211, 434, 226, 63 | sprt @ 0.5+0.005 th 1cores: 6 | #1026 SPRT - throughput 50%
21-09-12 | use | master | diff | LLR: -0.20 (-2.94,2.94) <-0.50,2.50>Total: 5048 W: 1567 L: 1573 D: 1908Ptnml(0-2): 144, 570, 1106, 556, 148 | sprt @ 0.5+0.005 th 1cores: 12 | #1026 SPRT
21-09-12 | use | master | diff | LLR: -0.80 (-2.94,2.94) <-0.50,2.50>Total: 4512 W: 1395 L: 1451 D: 1666Ptnml(0-2): 165, 512, 938, 496, 145 | sprt @ 0.5+0.005 th 1cores: 9 | #1026 NumGames
21-09-12 | use | moves1 | diff | 916/30000 iterations1832/60000 games played | 60000 @ 0.5+0.005 th 1cores: 8 | #1026 SPSA
21-09-12 | use | moves1 | diff | 414/30000 iterations828/60000 games played | 60000 @ 0.5+0.005 th 1cores: 6 | #1026 SPSA - Throughput 50%
21-09-12 | use | moves1 | diff | 272/30000 iterations544/60000 games played | 60000 @ 0.5+0.005 th 1cores: 3 | #1026 SPSA - Throughput 25%


</body>
</html>

vdbergh:
Thanks for testing. It is hard to conclude anything on the short term. The throughput algorithm should keep the average throughput constant (roughly speaking), but on the sort term there can be a lot of variation.

The bug I introduced is that when a run is in front of the queue for a new task, the queue is not resorted for some time, and so the same run stays in front, hence it will also be dispensing the next task, and so on. I noticed this when I saw a run suddenly grabbing a lot of cores.

ppigazzini (approved):

ppigazzini:
The queue on DEV looks good, so merged and PROD updated.
Thank you @vdbergh :)

vdbergh:
Thanks. Unfortunately it is still happening.

cores

The second task is getting far too many cores compared to its itp value. Some of those should have gone to the third task (both itp values are ~ 100, since the third task is running at throughput 200).

vondele:
@vdbergh now I follow the discussion, and realize the itp algorithm you talk about is mine ;-)

The second task gets much more resources because not all workers can take all tasks. Certain workers (noob's fleet) can right now only do STC tasks because of memory constraints, other workers (more than 32 threads) can only take LTC tasks. So while the itp is providing the order, the task assigned might be different from the top-most one.

vdbergh:
@vondele Ah sorry. Throughput was introduced by Joona. But I did not know that the current algorithm is yours. Very nice.

I thought about your explanation, but I do not really see why the 2d task would get more cores than the 3d (they are both 10+0.1 so the same machines qualify). For SPSA there is a core limit but as I said I do not think it kicks in.

vondele:
non-sprt tasks get a 0.5 reduction of their itp, and LLR positive tasks an additional small boost. Not sure if that fully explains it?

vdbergh:
@vondele 

Yes but that is already incorporated in the itp values. The 2d and 3d task both had itp value ~ 100 (the SPSA job is running at throughput 200).

vdbergh:
I also checked priorities (both 0).

I was thinking of the following scenario: the 2d task had a high llr at one point and grabbed a bunch of servers from the 3d task (so the effect is doubled). Then its llr came down (and the itp value with it) so that its core count became too high with respect to its itp value. If that's the case then the algorithm is working correctly (after < 15 min the situation should clear).

ppigazzini:
Just added a SPRT with time odds on DEV

<html>
<body>
<h3 style="margin: 10px 0px; font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-weight: bold; line-height: 40px; color: rgb(177, 173, 167); text-rendering: optimizelegibility; font-size: 24.5px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(17, 17, 17); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Active - 3 tests</h3>

21-09-13 | use | master | diff | LLR: 0.05 (-2.94,2.94) <-0.50,2.50>Total: 1488 W: 476 L: 469 D: 543Ptnml(0-2): 54, 167, 290, 184, 49 | sprt @ 0.5+0.005 th 1cores: 26 | #1027 SPRT - priority 5 - throughput 100%
-- | -- | -- | -- | -- | -- | --
21-09-13 | use | master | diff | Pending... (0.2 hrs)<-0.50,2.50> | sprt @ 0.5+0.005 th 1cores: 2 | #1027 SPRT - priority 5 - throughput 100% - time odds 10%
21-09-13 | use | moves1 | diff | 684/30000 iterations1368/60000 games played | 60000 @ 0.5+0.005 th 1cores: 24 | #1027 SPSA - priority 5 - throughput 200%


</body>
</html>

vondele:
yes, the second scenario can also happen, and indeed should clear up after 15min. (unless something stops this process, like reassigning tasks to the last run task automatically).

vdbergh:
> Just added a SPRT with time odds on DEV
> 
> ### Active - 3 tests
> 21-09-13	use	master	diff	Pending... (0.2 hrs)<-0.50,2.50>	sprt @ 0.5+0.005 th 1cores: 2	#1027 SPRT - priority 5 - throughput 100% - time odds 0.5+0.005 vs 0.55+0.005
> 21-09-13	use	master	diff	LLR: -0.08 (-2.94,2.94) <-0.50,2.50>Total: 21104 W: 6661 L: 6625 D: 7818Ptnml(0-2): 695, 2387, 4376, 2375, 719	sprt @ 0.5+0.005 th 1cores: 26	#1027 SPRT - priority 5 - throughput 100%
> 21-09-13	use	moves1	diff	8152/30000 iterations16304/60000 games played	60000 @ 0.5+0.005 th 1cores: 22	#1027 SPSA - priority 5 - throughput 200%

@ppigazzini I just looked on DEV and currently the 3 runs have similar itp values and similar numbers of cores. So things seems to be correct...

ppigazzini:
The time odds one (with higher abs(LLR)), always with few cores added:

<html>
<body>
<h3 style="margin: 10px 0px; font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-weight: bold; line-height: 40px; color: rgb(177, 173, 167); text-rendering: optimizelegibility; font-size: 24.5px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(17, 17, 17); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Active - 3 tests</h3>

21-09-13 | use | master | diff | LLR: 2.53 (-2.94,2.94) <-0.50,2.50>Total: 3528 W: 1193 L: 991 D: 1344Ptnml(0-2): 71, 350, 770, 452, 121 | sprt @ 0.5+0.005 th 1cores: 20 | #1027 SPRT - priority 5 - throughput 100% - time odds 0.5+0.005 vs 0.55+0.005
-- | -- | -- | -- | -- | -- | --
21-09-13 | use | master | diff | LLR: 0.28 (-2.94,2.94) <-0.50,2.50>Total: 24848 W: 7848 L: 7775 D: 9225Ptnml(0-2): 814, 2786, 5178, 2805, 841 | sprt @ 0.5+0.005 th 1cores: 16 | #1027 SPRT - priority 5 - throughput 100%
21-09-13 | use | moves1 | diff | 9380/30000 iterations18760/60000 games played | 60000 @ 0.5+0.005 th 1cores: 14 | #1027 SPSA - priority 5 - throughput 200%


</body>
</html>

vdbergh:
Actually I think that there _is_ an issue with the current code. `run["cores"]` for a `run` in `task_run` (the queue of unfinished runs) is updated when a worker gets a task (i.e. in `request_task()`), but not when a worker finishes a task (i.e. in `update_task()`). This, combined with the fact that `task_run` is only updated every 60 seconds, would allow a task to grab (many?) more cores than it is entitled to.

vdbergh:
I think I discovered one possible issue. I do not know how serious it is. See #1029.

vdbergh:
> yes, the second scenario can also happen, and indeed should clear up after 15min. (unless something stops this process, like reassigning tasks to the last run task automatically).

@vondele There seems to be indeed such a mechanism. If a worker is near its github rate limit then it will only accept tasks for runs it has seen before (since then it already has the binaries). I guess workers should be strongly encouraged to create a .netrc file to raise their api limit.

vdbergh:
In fact the mechanism isn't perfect since if the worker is restarted its unique key changes and then it will not be assigned any tasks until the rate limit clears.

I have been thinking that it would also be nice to have a worker key that it preserved across restarts. It could the md5sum of a combination of `uuid.getnode()` and the path name of the worker script.

ppigazzini:
> I guess workers should be strongly encouraged to create a .netrc file to raise their api limit.

With `task_duration=900` the GitHub rate limit should be hit rarely.
Before that PR a fast Zen3 CPU at 31 cores finished a 200 games STC batch in 2 minutes (see [here](https://github.com/glinscott/fishtest/pull/988#issuecomment-894648592)), so now it will finish a 872 games STC batch in 9 minutes.

IIRC the API calls should be:
- 4 : worker first setup or worker version raise (base, test, book, cutechess-cli)
- 2 : after a master change (base, test)
- 1 : frequent case (test, base cached)
- 0 : frequent case (base and test cached)

vdbergh:
I guess it would be nice if the server the knew the base_nps before computing the task size. The server caching it seems to be the easiest solution (and then again having "unique key" which is partially stable under worker restarts would be nice).

vondele:
I think caching nps is a simple solution. 

On the other hand, I don't think that some variance in the task times is bad, it will make it a bit smoother to load balance.

vdbergh:
Well we could also add a random component to the worker nps. That would create some controlled variance. Now the variance seems to be quite large.

ppigazzini:
> and then again having "unique key" which is partially stable under worker restarts would be nice

The "unique key" should change according to change of "--concurrency" and "--use_all_cores" flags (and perhaps "username" from a DEV tester point of view).

vdbergh:
I think the unique key should be built up from several parts: getnode() (=MAC address), the path of the worker script, the options (the cfg file, and command line), the time and a random component. 

The unique key should be the concatenation of these components (after they have been obscured by hashing).

ppigazzini:
If I'm not wrong our scheduler seems to keep stuck to the same run a high cores count worker.
The linkrock 100 cores worker contributed to only 3 runs in the whole set in homepage (btw always with high residuals):
- 56 tasks - https://tests.stockfishchess.org/tests/view/613f5ab0f29dda16fcca86cf
- 16 tasks - https://tests.stockfishchess.org/tests/view/614053e07315e7c73204a4bb

vdbergh:
I think that is explainable. When linrock-100cores finishes, the core count of the run drops dramatically, so the run moves to the front of the queue. Since linrock-100cores is at that moment looking for a new task, it is likely to get the same run.

vdbergh:
Maybe we should have an additional sorting criterion that takes into account the last run of a worker.

vdbergh:
> In the finished run the 100 cores worker contributed 27k/102k games, I'm wondering if these (biased) data fulfill the chi2 test requirements.

I think it is bad. But the scheduling algorithm can probably be adapted to avoid such lopsidedness.

locutus2:
Looking at the current running tests it seems something is broken. Currently it seems LTC tests has now only around 50% more cores than STC tests assigned (see screenshot) unlike in the past were it was around factor 3. Is this intended or have i overlooked something? Perhaps its only an unusual situation at the framwork.
fishtest

ppigazzini:
@locutus2 with #1033 now a worker changes test after having completed a run. LCT tests can now have a high variability in core counts with so many workers with 16-100 cores switching tests, but in the long run the games played should be proportional to the itp of the tests. The previous behavior was smoother because a big worker was always stuck to the same test, but this was bad for chi2 and residuals.

vdbergh:
There is also the issue that the noob fleet can currently only do STC. So when it comes online, STC runs may get disproportionately many cores.

vdbergh:
`@locutus2 with #1033 now a worker changes test after having completed a run. LCT tests can now have a high variability in core counts with so many workers with 16-100 cores switching tests, but in the long run the total games played should be the usual function of the itp of the tests. `

I thought about it but I think this is not strictly true. Assume we have only one worker and 2 runs. Then with the new algorithm the worker will alternate between the two runs, spending 50% of the time on each of them, independently
of the itp values.

But with the original algorithm the worker would spend all its time on the run with the highest itp value.So that is even less fair.

Presumably with (many) more workers and runs things will smooth out...

ppigazzini:
DEV running with 2 unbalanced itp tests and 2 unbalanced cores workers.

Previous tests on DEV with more tests/workers ran smooth games/itp wise.

ppigazzini:
2 unbalanced itp tests with 2 unbalanced cores workers have the same games in long run (as expected)

Started 3 unbalanced test with 2 unbalanced workers.

PR #1029, opened 2021-09-13
Sort "task_runs" at the correct place.

vdbergh:
Bugfix: the current code (like the old code) may not pick up changes of `run["cores"]` due to finished tasks.
    
More precisely: if a cached version of `task_runs` is not sorted at the beginning of `request_task()` its first entry may be a run which does not have the lowest value of `run["cores"]/run["itp"]` because it may be that since the last call to `request_task()` one or more tasks in other runs finished (handled in `update_task()`).
    
Also some cosmetic changes around the `request_task()/update_task()` code.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -571,8 +571,16 @@ class RunDb:
                 self.sum_cores(run)
                 self.calc_itp(run)
                 self.task_runs.append(run)
-            self.task_runs.sort(key=priority)
             self.task_time = time.time()
+
+        # Note that a cached task_run may have runs with
+        # a changed number of cores.
+        # Either by the code below or else by finished tasks
+        # detected in update_task().
+        # Note that update_task() uses the same objects as here
+        # (they are not copies).
+
+        self.task_runs.sort(key=priority)
 
         max_threads = int(worker_info["concurrency"])
         min_threads = int(worker_info.get("min_threads", 1))
@@ -683,29 +691,20 @@ class RunDb:
             return {"task_waiting": False}
 
         task_size = min(self.worker_cap(run, worker_info), remaining)
-        run["tasks"].append(
-            {
-                "num_games": task_size,
-                "pending": True,
-                "active": True,
-                "worker_info": worker_info,
-                "last_updated": datetime.utcnow(),
-                "start": run_size,
-            }
-        )
+        task = {
+            "num_games": task_size,
+            "pending": True,
+            "active": True,
+            "worker_info": worker_info,
+            "last_updated": datetime.utcnow(),
+            "start": run_size,
+            "stats": {"wins": 0, "losses": 0, "draws": 0, "pentanomial": 5 * [0]},
+        }
+        run["tasks"].append(task)
+
         task_id = len(run["tasks"]) - 1
-        task = run["tasks"][task_id]
-        task["stats"] = {"wins": 0, "losses": 0, "draws": 0, "pentanomial": 5 * [0]}
 
-        self.sum_cores(run)
-
-        # Resort the queue to take into account
-        # the alterned number of cores of "run".
-        # Resorting is important since the queue
-        # is only regenerated every 60 seconds.
-
-        self.task_runs.sort(key=priority)
-
+        run["cores"] += task["worker_info"]["concurrency"]
         self.buffer(run, False)
 
         # Update worker_runs (compiled tests)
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated.

ppigazzini:
DEV updated.

Please mind that in #1028 I replaced the lambda with a function for readability, rebase if you need to change that code in this PR.

vdbergh:
I already rebased the PR.

vdbergh:
Nothing obviously wrong on DEV...

tomtor:
> Bugfix: the current code (like the old code) may not immediately pick up changes in core counts due to finished tasks because a cached `task_runs` is currently only sorted at the end of request_task. We move the sorting to the beginning.

It's not clear to me what the difference is between sorting in the beginning of the synchronized request_task function or at the end? The result would be the same because timing is not important. Only one thread can enter the request_task function at a time.

Your code changes look good, so please proceed anyway.

vdbergh:
@tomtor  This is what I thought: if a cached version of `task_runs` is not sorted at the beginning or `request_task()` its first entry may be a run which does not have the lowest value of `run["cores"]/run["itp"]` because it may be that since the last call to `request_task()`, one or more tasks in other runs finished (handled in `update_task()`).

kad-vijlbc:
Ah I see, I think you are right.

Note that access to cores from update_task is not synchronized with the updates in request_task:

https://github.com/glinscott/fishtest/blob/28ebad9fc7b1dec4b7f7f85c5d6f0a71ddda70a8/server/fishtest/rundb.py#L873

It would be better if they shared the same lock, eg the existing lock:

https://github.com/glinscott/fishtest/blob/988e2fb7f1a45d1e54838f063a7a8fa1408f408f/server/fishtest/rundb.py#L525

or a dedicated one for this core counter update. It's not important, but if we have the ambition to get it really right....

vdbergh:
@tomtor Thanks. I think that in this case the global interpreter lock will take care of serializing the access to `run["cores"]`.  But maybe it's better to apply defensive programming. 

So I updated the PR with
```
with self.task_lock:
     run["cores"] -= task["worker_info"]["concurrency"] 
```
 (in `update_task()`).

ppigazzini:
DEV updated. AFK for 2 hours.

ppigazzini:
DEV updated, my tests in the last couple of hours look good.
@vdbergh let me know if I can merge.

vdbergh:
@ppigazzini As far as I am concerned it’s ready…

ppigazzini:

vdbergh:
@tomtor I have a question. Sometimes the code uses `self.buffer(run,False)` and sometimes `self.buffer(run,True)`.

I understand the difference between the two, but I wonder if it is ever _really_ necessary to use `self.buffer(run,True)`? 

I am assuming here that all run objects are created by `self.get_run(r_id)` (they should be...), so that they are singletons at any given moment in time.

vdbergh:
@tomtor Wouldn't that be only for a _dirty_ restart, where buffers aren't flushed? Personally I would think that 1 second of data loss is acceptable in that situation (since it should be exceptional), but I guess opinions may differ. The same goes for the delay fishtest server -> webserver.

Anyway I am noticing that `flush_buffers()` may delete run object from the cache. So I wonder if my claim that run objects created by `get_run(r_id)` are singletons at any given moment in time is always true (I find it difficult to understand under which conditions the deletion occurs).

vdbergh:
BTW thanks for mentioning that the webserver is a different instance of fishtest on PROD. I was not aware of this.

kad-vijlbc:
@vdbergh Note that request_task also reads from MongoDb because the tasks are created in de web instance. A delay from a few seconds is ok for syncing between the instances. 5 minutes would be ok for a dirty restart for me, but the current short delay creates just a low load.

vdbergh:
@tomtor `request_task()` reads the ids from the db but creates the run objects with `get_run(r_id)` so they should normally be singletons.

If `update_task()` did not flush finished runs, then `task_runs` could contains runs which in reality are already finished. But this is harmless since they will not be assigned tasks now (since `request_task()` and `update_task()` see identical run objects).

Anyway I am not against `self.buffer(run,True)` but it would be nice if there was a clear criterion for when to use it.

kad-vijlbc:
@vdbergh Regarding the 1s sync, every 1s the oldest run is flushed, so with 20 runs active the oldest flush is 20s.

vdbergh:
@tomtor Ah ok. I had not understood this. 20s is maybe indeed already somewhat long.

kad-vijlbc:
> but it would be nice if there was a clear criterion for when to use it.

Task creation/termination is the most important to sync to/from the web instance

vdbergh:
> > but it would be nice if there was a clear criterion for when to use it.
> 
> Task creation/termination is the most important to sync to/from the web instance

But ... `request_task()` actually uses `self.buffer(run,False)` (I did not change it). It seemed perfectly appropriate. 

I was thinking that it would be more  run creation, finishing, termination that would make flushing appropriate (these are relatively infrequent events).

tomtor:
> I was thinking that it would be more run creation, finishing, termination that would make flushing appropriate (these are relatively infrequent events).

Yes, I agree. I mentioned tasks but I should have mentioned runs.

PR #1032, opened 2021-09-15
Worker fingerprint

vdbergh:
This should make it possible to more easily track workers that are frequently restarted.
    
The fingerprint is shown on the machines table instead of the UUID (but it contains the former UUID field).
    
The fingerprint consists of 3 fixed parts which are respectively hashed versions of:
    
    - The output of uuid.getnode() (normally the MAC address).
    - The pathname of the worker script.
    - The config options (even when modified by the command line).
    
The fourth part is the first field of the UUID (which is itself randomly generated).

### DIFF
--- a/server/fishtest/templates/machines_table.mak
+++ b/server/fishtest/templates/machines_table.mak
@@ -2,9 +2,9 @@
        style="max-width: 960px;">
   <thead>
     <tr>
-      <th>Machine</th>
+      <th>User</th>
       <th>Cores</th>
-      <th>UUID</th>
+      <th>Fingerprint (machine-path-config-key)</th>
       <th>MNps</th>
       <th>System</th>
       <th>Version</th>
@@ -23,7 +23,7 @@
             % endif
             ${machine['concurrency']}
           </td>
-          <td>${machine['unique_key'].split('-')[0]}</td>
+	  <td>${machine.get('fingerprint',machine['unique_key'].split('-')[0])}</td>
           <td>${f"{machine['nps'] / 1000000:.2f}"}</td>
           <td>${machine['uname']}</td>
           <td>${machine['version']}</td>
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 from __future__ import print_function
 
+import hashlib
 import json
 import math
 import multiprocessing
@@ -340,6 +341,19 @@ def gcc_version():
     print("Found g++ version {}.{}".format(major, minor))
 
 
+def config_to_dict(config):
+    dictionary = {}
+    for section in config.sections():
+        dictionary[section] = {}
+        for option in config.options(section):
+            dictionary[section][option] = config.get(section, option)
+    return dictionary
+
+
+def hash_string(s):
+    return hashlib.md5(s.encode("utf-8")).hexdigest()[:8]
+
+
 def main():
     worker_dir = path.dirname(path.realpath(__file__))
     print("Worker started in " + worker_dir + " ...\n")
@@ -490,6 +504,28 @@ def main():
     print("UUID:", worker_info["unique_key"])
     with open(path.join(worker_dir, "uuid.txt"), "w") as f:
         print(worker_info["unique_key"], file=f)
+
+    fingerprint_node = hash_string(str(uuid.getnode()))
+    path_worker = os.path.abspath(sys.argv[0])
+    fingerprint_path = hash_string(str(path_worker))
+    config_dict = config_to_dict(config)
+    del config_dict["login"]["password"]
+    fingerprint_config = hash_string(str(config_dict))
+    fingerprint_uuid = worker_info["unique_key"].split("-")[0]
+    fingerprint_worker = (
+        fingerprint_node
+        + "-"
+        + fingerprint_path
+        + "-"
+        + fingerprint_config
+        + "-"
+        + fingerprint_uuid
+    )
+    worker_info["fingerprint"] = fingerprint_worker
+
+    print("Fingerprint:", worker_info["fingerprint"])
+    with open(path.join(worker_dir, "fingerprint.txt"), "w") as f:
+        print(worker_info["fingerprint"], file=f)
 
     # Start heartbeat
     threading.Thread(target=heartbeat, args=(worker_info, password, remote)).start()


vdbergh:
Something went wrong with this PR.

vdbergh:
One case that is not recognized is the same worker being started multiple times from the same directory with the same config file (I assume that is what is shown on DEV). Not sure if there is a good solution for this.

ppigazzini:
DEV and workers updated (some restarts changing something).
You have shortened the random uuid to only the first digit group, this not seem to be an improvement.

vdbergh:
Well in master only the first part of the UUID is shown. So the amount of information shown now has strictly increased.

vdbergh:
To avoid misunderstanding. The fingerprint is not a replacement for the UUID. It is just a way to keep track of workers across restarts.

ppigazzini:
> To avoid misunderstanding. The fingerprint is not a replacement for the UUID. It is just a way to keep track of workers across restarts.

Thank you, I havn't time now to review the code.

ppigazzini:
DEV and workers updated.

vdbergh:
In this test

https://tests.stockfishchess.org/tests/view/61421a3c7315e7c73204a624

there were 37 noob workers who played zero games. It would be nice to understand this. With this pr one would be able to see if it is always the same machine.

vondele:
additionally, for example this:
```
21-09-16 03:34:33 | ChessDBCN | flPvCnRed | stop run: ChessDBCN-16cores-01dd5949: Wrong bench in  stockfish_3405f9f0f64c9169ba74ff2d24af8670121426ba Expected: 7045401  Got: 6809864
```
is a noob worker that generates a wrong bench... likely hardware related (IMO). Can we use that info to track which worker that is? (For this the real MAC address or similar would probably be useful).

vondele:
maybe a simple approach would be to add a command line option to the worker --identifyer=.... so that the user can pick a string to show as an id ? --identifyer=`hostname` or whatever user string? There would of course be no guarantee a user picks sane or unique ids.

vdbergh:
> additionally, for example this:
> 
> ```
> 21-09-16 03:34:33 | ChessDBCN | flPvCnRed | stop run: ChessDBCN-16cores-01dd5949: Wrong bench in  stockfish_3405f9f0f64c9169ba74ff2d24af8670121426ba Expected: 7045401  Got: 6809864
> ```
> 
> is a noob worker that generates a wrong bench... likely hardware related (IMO). Can we use that info to track which worker that is? (For this the real MAC address or similar would probably be useful).

The first 8 hexadecimal digits of the fingerprint are a hash of the MAC address. The worker writes out the fingerprint to a file fingerprint.txt. In that way the machine can be identified by the owner.

vdbergh:
> maybe a simple approach would be to add a command line option to the worker --identifyer=.... so that the user can pick a string to show as an id ? --identifyer=`hostname` or whatever user string? There would of course be no guarantee a user picks sane or unique ids.

I thought of that. It might be useful for workers that are started in parallel in the same directory with the same config. Those are not distinguished by the current scheme. But on the other hand: workers that are started in the same directory with the same config are really the same... (they will still get a different UUID of course).

vdbergh:
@vondele Not sure you looked on the DEV server but the fingerprint looks like this

(machine-path-config-key)
5bdc0618-0f70f3cf-bf375bfd-654aeb18

vondele:
> > additionally, for example this:
> > ```
> > 21-09-16 03:34:33 | ChessDBCN | flPvCnRed | stop run: ChessDBCN-16cores-01dd5949: Wrong bench in  stockfish_3405f9f0f64c9169ba74ff2d24af8670121426ba Expected: 7045401  Got: 6809864
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is a noob worker that generates a wrong bench... likely hardware related (IMO). Can we use that info to track which worker that is? (For this the real MAC address or similar would probably be useful).
> 
> The first 8 hexadecimal digits of the fingerprint are a hash of the MAC address. The worker writes out the fingerprint to a file fingerprint.txt. In that way the machine can be identified by the owner.

I believe one issue is that upon failure @noobpwnftw worker quits, and the cloud worker is shutdown, and the file is deleted.  Maybe noob can suggest what he would see as a suitable solution to identify those workers?

vdbergh:
@vondele 

If the virtual machine continues to exist (or gets resurrected with the same MAC address) after the worker quits one can just rerun the worker. The first three fields of the fingerprint will be the same.

vdbergh:
Contrary to what I wrote above I now think it would be a good idea to let the worker "unique key" be equal to the fingerprint in this PR (then we don't need a separate fingerprint anymore). The fingerprint contains a large part which is stable under worker restarts.

To reduce anxiety one may increase the random part of the fingerprint (the 4th field) to 6 or 8 bytes, but I don't think that even be critical.

ppigazzini:
@vdbergh  I lack the vm/containers know how to made a statement about an unique worker identifier that is stable under worker restart.

The 15 instances on colab (started form different colab accounts and identified with different IP/national flags), share the first 3 fields (the first one is unexpected). If this should happen also for some of the fleets vm/container based (noob, mlang, OIVAS7572, oz, pemo etc etc) this implementation IMO will miss the goal.

Untitled

vdbergh:
@ppigazzini Thanks. I am a bit surprised by this. Can you check that they indeed all have the same MAC address? I am just using python's `uuid.getnode()`. 

Anyway I made the unique_key now equal to the former fingerprint (so there is no separate fingerprint any longer). 
I extended the random part to 6 bytes.

So the new unique keys contain strictly more information than the old ones. So there would be no loss in using them. And we could further refine them for specific situations where python's getnode doesn't cut it (the mechanism is in place now).

ppigazzini:
The colab instances seems to share the same net configuration:
```
ip add
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
7: eth0@if8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:1c:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.28.0.2/16 brd 172.28.255.255 scope global eth0
       valid_lft forever preferred_lft forever
```

They can have different HW configurations:
```
cat /proc/cpu
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 85
model name	: Intel(R) Xeon(R) CPU @ 2.00GHz
stepping	: 3
microcode	: 0x1
cpu MHz		: 2000.138
cache size	: 39424 KB
...
```
```
cat /proc/cpu
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 63
model name	: Intel(R) Xeon(R) CPU @ 2.30GHz
stepping	: 0
microcode	: 0x1
cpu MHz		: 2299.998
cache size	: 46080 KB
...
```

ppigazzini:
> Anyway I made the unique_key now equal to the former fingerprint (so there is no separate fingerprint any longer).
I extended the random part to 6 bytes.

@vdbergh I don't know the whole server code so in #980 I preferred to publish only the first uuid digit group to keep secret the `unique_key`

ppigazzini:
> additionally, for example this:
> 
> ```
> 21-09-16 03:34:33 | ChessDBCN | flPvCnRed | stop run: ChessDBCN-16cores-01dd5949: Wrong bench in  stockfish_3405f9f0f64c9169ba74ff2d24af8670121426ba Expected: 7045401  Got: 6809864
> ```
> 
> is a noob worker that generates a wrong bench... likely hardware related (IMO). Can we use that info to track which worker that is? (For this the real MAC address or similar would probably be useful).

For `ChessDBCN-16cores-01dd5949` noob can check the uuid.txt to find the worker instance.
Real HW info (e.g. MAC address) in vm/container/cloud is a lost war, and this for our provider https://cpl.thalesgroup.com/software-monetization

vdbergh:
> > Anyway I made the unique_key now equal to the former fingerprint (so there is no separate fingerprint any longer).
> > I extended the random part to 6 bytes.
> 
> @vdbergh I don't know the whole server code so in #980 I preferred to publish only the first uuid digit group to keep secret the `unique_key`

Hmmm I see. I guess there is a risk that a worker impersonates another worker (I don't really see how since the worker has to log in also). But I can see your point so I will revert the last commit for now (reviving the separate fingerprint).

vdbergh:
> > additionally, for example this:
> > ```
> > 21-09-16 03:34:33 | ChessDBCN | flPvCnRed | stop run: ChessDBCN-16cores-01dd5949: Wrong bench in  stockfish_3405f9f0f64c9169ba74ff2d24af8670121426ba Expected: 7045401  Got: 6809864
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is a noob worker that generates a wrong bench... likely hardware related (IMO). Can we use that info to track which worker that is? (For this the real MAC address or similar would probably be useful).
> 
> For `ChessDBCN-16cores-01dd5949` noob can check the uuid.txt to find the worker instance.

The problem is that if the worker is restarted then uuid.txt is overwritten. But the fingerprint will simply be recomputed by the worker. 

> Real HW info (e.g. MAC address) in vm/container/cloud is a lost war, and this for our provider https://cpl.thalesgroup.com/software-monetization

One does not need real HW information. Just something that is preserved across a worker restart. Since I do not know now what a worker restart means for noob I cannot really comment...

ppigazzini:
> Hmmm I see. I guess there is a risk that a worker impersonates another worker (I don't really see how since the worker has to log in also). But I can see your point so I will revert the last commit for now (reviving the separate fingerprint).

Don't knowing well the server code I was worried for db pollution with a legit user/password and a copied `unique_key`. We have already so much code in place to try to deal with old/bad data in db.

E.g. testing your last PRs I found that the server crashes on the older tests, a bug introduced in the last months :(
In the we I will instrument the code to write a fix:
```
Sep 16 11:15:05 vps1572612706 pserve[24101]:   File "/home/usr00/fishtest/server/fishtest/util.py", line 140, in calculate_residuals
Sep 16 11:15:05 vps1572612706 pserve[24101]:     if task["worker_key"] in bad_users:
Sep 16 11:15:05 vps1572612706 pserve[24101]: KeyError: 'worker_key'

```

ppigazzini:
> The problem is that if the worker is restarted then uuid.txt is overwritten. But the fingerprint will simply be recomputed by the worker.

With a worker home pc based it's simple to keep the last n UUID with some date info in uuid.txt, but we don't know if a worker cloud/vm/container based is built on the fly at any restart (I do this for my test workers).

vdbergh:
```
Sep 16 11:15:05 vps1572612706 pserve[24101]:   File "/home/usr00/fishtest/server/fishtest/util.py", line 140, in calculate_residuals
Sep 16 11:15:05 vps1572612706 pserve[24101]:     if task["worker_key"] in bad_users:
Sep 16 11:15:05 vps1572612706 pserve[24101]: KeyError: 'worker_key'
```

Ah. I rearranged the order of some of the statements in get_chi2().

Presumably this (line 56 in `util.py`)
```
        if "worker_info" not in task:
            continue
        key = get_worker_key(task)
        task["worker_key"] = key
```

should be changed to

```
        key = get_worker_key(task)
        task["worker_key"] = key
        if "worker_info" not in task:
            continue
```

Anyway this will be resolved more cleanly once we get rid of sequence numbers.

vdbergh:
Ah no that does not work either since `get_worker_key(task)` needs the worker_info. Now I wonder how it was done before.

ppigazzini:
BTW here is how I start a worker with colab for this PR:
```
!git clone --single-branch --branch master https://github.com/glinscott/fishtest.git
!cd fishtest && git config user.email "you@example.com" && git config user.name "your_name"
!cd fishtest && git pull --no-edit origin pull/1032/head
!python3 fishtest/worker/worker.py user01 user01 --host dfts-0.pigazzini.it --use_all_cores True
```

ppigazzini:
> Ah no that does not work either since `get_worker_key(task)` needs the worker_info. Now I wonder how it was done before.

The bug was here before your PRs, I found it on DEV but PROD crashed too.

ppigazzini:
> Anyway this will be resolved more cleanly once we get rid of sequence numbers.

If I'm not wrong we can drop the sequence number already now, DEV is running with these 2 lines commented out:
https://github.com/glinscott/fishtest/blob/4231b1134f27f5d865a648597f76c2dea779d4c4/server/fishtest/util.py#L34-L35

We should have possible collisions for very old tests before the introduction of `unique_key`, though.

These are respectively the first test ran with all the workers with `unique_key` and the last test ran without workers `unique_key` (both date "2013-09-11",  they both were at test page 4309/4387 when I lloked for)
https://tests.stockfishchess.org/tests/view/523080240ebc5957c3a1cd09
https://tests.stockfishchess.org/tests/view/522dc6d20ebc593dd6bc6efa

vdbergh:
But what is the point of sequence numbers if there is no unique_key? It seems to me that then the only thing you can do is to group the tasks by username+cores. So I do not see the issue for very old tests either.

Perhaps you can make a quick PR to get rid of sequence numbers?

ppigazzini:
> But what is the point of sequence numbers if there is no unique_key? It seems to me that then the only thing you can do is to group the tasks by username+cores. So I do not see the issue for very old tests either.
> 
> Perhaps you can make a quick PR to get rid of sequence numbers?

I commented out by hand these 2 lines in `util.py`, this seems to preserve the test info (residuals, p-value, statistics etc.) respect to PROD, I don't know if there is a side effect in other pages and if this is what you intended for "get rid of sequence number", though. If I comment out also the uuid group (leaving only username and cores count) the info on DEV go crazy.

https://tests.stockfishchess.org/tests/view/613e4b5015591e7c9ebc4000
https://dfts-0.pigazzini.it/tests/view/613e4b5015591e7c9ebc4000

ppigazzini:
This works also for the first test with `unique_key` ("2013-09-11"), of course 

https://tests.stockfishchess.org/tests/view/523080240ebc5957c3a1cd09
https://dfts-0.pigazzini.it/tests/view/523080240ebc5957c3a1cd09

Dark42ed:
Wouldn't it be better to use
```python
fingerprint_worker = "-".join([
        fingerprint_node,
        fingerprint_path,
        fingerprint_config,
        fingerprint_uuid
        ])
```
instead of manually concatenating it?

vdbergh:
> Wouldn't it be better to use
> 
> ```python
> fingerprint_worker = "-".join([
>         fingerprint_node,
>         fingerprint_path,
>         fingerprint_config,
>         fingerprint_uuid
>         ])
> ```
> 
> instead of manually concatenating it?

That seems to be indeed a bit more readable.

vdbergh:
This code is no longer up to date. Closing.

PR #1033, opened 2021-09-15
Try to find a new run for a worker in request_task().

vdbergh:
Currently there is the issue that high core workers always get the same run while their residual keeps steadily rising.

This is not good. We want a good hardware mix for every test.

This PR moves the last run seen by the worker to the end of the queue.

Not live tested.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -554,9 +554,14 @@ class RunDb:
             return {"task_waiting": False}
 
     def sync_request_task(self, worker_info):
-        def priority(run):
+        worker_key = worker_info["unique_key"]
+        last_run_id = self.worker_runs.get(worker_key, {}).get("last_run", None)
+
+        def priority(run):  # lower is better
             return (
                 -run["args"]["priority"],
+                # Try to find a new run for this worker.
+                run["_id"] == last_run_id,
                 run["cores"] / run["args"]["itp"] * 100.0,
                 -run["args"]["itp"],
                 run["_id"],
@@ -607,7 +612,6 @@ class RunDb:
             limit = rate["remaining"] <= 2 * math.sqrt(rate["limit"])
         else:
             limit = False
-        worker_key = worker_info["unique_key"]
 
         # Get a new task that matches the worker requirements
         run_found = False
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, started a 31 cores worker.

ppigazzini:
Started 10 tests (1 NumGames with 200% throughput, the others with equal itp), a 31 worker and several 2 workers. I'm expecting than the 31 worker switches between tests,

ppigazzini:
The 31 cores worker switches tests. I stopped it and started 2 workers at 15 cores to have more balanced core counts.

vdbergh:
I think there is another bug. I think it was there before also but invisible because of the task migration.

I think that with the current code request_task() can return a task for a finished SPRT test (because of the caching of task_runs). Of course the worker will not be able to do any updates on this task so it will become a zero games task.

vdbergh:
I added the fix to the PR.

ppigazzini:
I'm commuting now, I will able to update in 2 hours. Feel free to suggest ways to test the PR.

ppigazzini:
DEV updated

vdbergh:
Went to see the new Dune movie :)

The PR seems to be behaving as advertised. The two 15 core workers where migrating between the different runs, with the higher priority runs (SPRT) generally seeing more of them. This is what one would want I think.

ppigazzini:
Yes, and it works fine also with only 1 test at high priority (other tests stay paused).

ppigazzini (approved):

ppigazzini:

Hoping you liked the movie :)

ppigazzini:
linrock-100cores is stuck to this test, no it should switch
https://tests.stockfishchess.org/tests/view/6142172c7315e7c73204a614

vdbergh:
> linrock-100cores is stuck to this test, no it should switch
> https://tests.stockfishchess.org/tests/view/6142172c7315e7c73204a614

I think so…

vdbergh:
It did not, but that’s because of the server restart. The dictionary keeping track of the worker runs is now empty again. Next time it should really switch.

ppigazzini:
Switched !
https://tests.stockfishchess.org/tests/view/614279aa7315e7c73204a6af

ppigazzini:
technologov-56cores-30-43fe85ce switched too :)

vdbergh:
:)

PR #1034, opened 2021-09-16
Get rid of sequence number in the worker key.

vdbergh:
In this way the worker key becomes a property of the worker and not of the sequence of tasks used to compute it.

This what I want to say about getting rid of the sequence numbers. After that further simplifications can be made.

This was not live tested.

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -9,8 +9,6 @@ import numpy
 import scipy.stats
 from zxcvbn import zxcvbn
 
-UUID_MAP = defaultdict(dict)
-key_lock = threading.Lock()
 
 FISH_URL = "https://tests.stockfishchess.org/tests/view/"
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, inspecting the same tests on PROD/DEV the PR looks good.

Very old tests without `unique_key` should not be a problem:
- `unique_key` was introduced by https://github.com/glinscott/fishtest/commit/90bf3e2016e5855a1547ee6d0e311f9da6712b13 in September 11, 2013
- `chi2` was introduced by https://github.com/glinscott/fishtest/commit/f8707ede14e2df90228b4a0446abd3542bf10a1a in October 10, 2013
- before September 11, 2013 the tests were short (40k games) and with task size = 1000, so with a lower probability of worker restart
- such old tests have only a historic value :)

ppigazzini (approved):

ppigazzini:

PR #1036, opened 2021-09-18
Document and refactor request_task.

vdbergh:
In master "request_task()" is a large for loop which contains a large if. We get rid of the if. This makes the code much more readable.

The new structure is
```
run_found=False
for run in self.task_runs:
     # comment1
     if excluding_condition1:
         continue
     # comment2
     if excluding_condition2:
         continue
     ...
     run_found=True
     break

if not run_found:
     return

(now create a new task for the run that was found)
```
This should be a non-functional change.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -531,6 +531,10 @@ class RunDb:
     worker_runs = {}
 
     def worker_cap(self, run, worker_info):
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
DEV has been running fine for over 90 minutes.

ppigazzini:

PR #1037, opened 2021-09-18
Reformat all mako template files

ppigazzini:


vdbergh:
How did you do the reformatting?

ppigazzini:
By hand :(

Changing the .mak in .html destroyed the javascript formatting.
I formatted with `black` the first 200 lines of python in `tests_stats.mak` using a temporary .py file.

ppigazzini:
The indentation was a mix of spaces and tabs :(

PR #1038, opened 2021-09-18
Write "-" for the worker name for workers without worker_info.

vdbergh:
Not live tested (I don't have the old tests).

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -259,8 +259,6 @@ from fishtest.util import worker_name
   <tbody>
     % for idx, task in enumerate(run['tasks'] + run.get('bad_tasks', [])):
         <%
-          if not "worker_info" in task:
-            continue
           stats = task.get('stats', {})
           if 'stats' in task:
             total = stats['wins'] + stats['losses'] + stats['draws']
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV is running fine after fixing also the line 316 in this way:
```html
315           % if 'spsa' not in run['args']:
316               % if 'worker_info' in task and task['residual']!=float("inf"):
317                   <td style="background-color:${task['residual_color']}">${f"{task['residual']:.3f}"}</td>
318               % else:
319                   <td>-</td>
320               % endif
321           % endif
```

Here are some old tests:
https://dfts-0.pigazzini.it/tests/view/511c69688f0c3e39b3913639
https://dfts-0.pigazzini.it/tests/view/511c69688f0c3e39b3913638
https://dfts-0.pigazzini.it/tests/view/512044248f0c3e26308243e5

ppigazzini (approved):

ppigazzini:

ppigazzini:
@vdbergh I'm wondering if it's right that the GitHub repo "Insight -> Contributors" doesn't show your contribution
https://github.com/glinscott/fishtest/graphs/contributors

vondele:
probably because the email used in the author field is not registered with github? Can be changed https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/adding-an-email-address-to-your-github-account

vdbergh:
@vondele Thanks. That was indeed the cause. I added my commit email address to my profile.

PR #1039, opened 2021-09-19
Get rid of the flag "task['pending']".

vdbergh:
The pending flag is True when a task is not finished. It was always redundant since it can be recomputed from the number of games received on a task and the task size.

However, now that there is no task migration any longer, the pending flag has become even more redundant since, if a worker goes offline, the task it was working on will never receive any further updates, no matter what the pending flag says.

Another issue with the pending flag is that the word "pending" is also used in other contexts (e.g. a run may be pending). This creates confusion.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -22,7 +22,7 @@ from fishtest.util import (
     format_results,
     post_in_fishcooking_results,
     remaining_hours,
-    unique_key,
+    unique_key as unique_key_,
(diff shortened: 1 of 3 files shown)


ppigazzini:
DEV updated.

ppigazzini (approved):

ppigazzini:
DEV and workers look good, logs clean.

vdbergh:
:smile:

PR #1042, opened 2021-09-20
Use f-string in tests_stats.mak

ppigazzini:


ppigazzini:
DEV test stats page look like the PROD one.

PR #1044, opened 2021-09-22
Upgrade bootstrap from 2.3.2 to 5.1.1

linrock:
Makes the site feel more modern since bootstrap 2.3.2 is 8+ years old (Jul 2013) and 5.1.1 was released a few weeks ago. Most of the site should still look similar.

Bootstrap 5.1 docs:
https://getbootstrap.com/docs/5.1

There will likely be issues since I didn't exhaustively check every page and don't have much test data to do so.

issues found by ppigazzini:
- [x] switch from light to dark theme is broken: clicking on the moon icon seems to switch only the icon (moon->sun->moon...)
- [x] gauge test box in light theme: lacks color (red, yellow) in homepage and in test page
- [x] test page (both themes): missing button for "Show/Hide" diff code snippet
- [x] NN stat page (both themes): missing button for "Show/Hide non default net"
- [x] test raw stats page: still to bootstrap 2.3.2
- [x] Chess SPRT Calculator page): still to bootstrap 3.4.2

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -7,7 +7,27 @@
 }
 
 body {
+  font-size: 14px;
   transition: background 0.2s ease-in-out;
+}
+
+a {
+  color: #0088cc;
+  text-decoration: none;
+}
+
+a:hover {
+  color: #0088cc;
+}
+
+h2, h3, h4 {
+  font-weight: bold;
+}
+
+legend {
+  padding-bottom: 5px;
+  border-bottom: 1px solid #e5e5e5;
+  margin-bottom: 20px;
 }
 
 .flash-message {
@@ -15,15 +35,45 @@ body {
 }
 
 .elo-results {
+  background: #f5f5f5;
   font-size: 12px;
   margin: 2px;
   padding: 1px;
   line-height: 13px;
+  padding: 3px;
 }
 
 .checkbox.inline {
   margin-left: 20px;
   margin-right: 0;
+}
+
+/** Left navigation */
+.nav-list {
+  margin-bottom: 0;
+}
+
+.nav-header {
+  color: #999999;
+  font-size: 11px;
+  font-weight: bold;
+  text-transform: uppercase;
+  margin-bottom: 3px;
+  padding-left: 15px;
+}
+
+.nav li + .nav-header {
+  margin-top: 12px;
+}
+
+/* Links in the left navigation */
+.nav li a {
+  display: block;
+  padding: 3px 15px;
+}
+
+.nav li a:hover {
+  background: #eeeeee;
 }
 
 th {
@@ -40,6 +90,28 @@ th {
   text-align: left;
 }
 
+#change-color-theme {
+  margin-left: 15px;
+}
+
 #change-color-theme:hover {
   cursor: pointer;
+}
+
+/* Pagination links */
+.page-link {
+  color: #0088cc;
+}
+
+.pagination li.disabled .page-link {
+  color: rgb(153, 153, 153);
+}
+
+.pagination li.active .page-link {
+  color: rgb(153, 153, 153);
+  background: rgb(245, 245, 245);
+}
+
+.pagination li:not(.active) .page-link:hover {
+  background-color: #f5f5f5;
 }
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -66,7 +66,16 @@ hr {
 }
 
 .table th, .table td {
+  color: #b1ada7;
   border-top: 1px solid #444;
+}
+
+.table th {
+  border-bottom-color: #444 !important;
+}
+
+tbody, td, tfoot, th, thead, tr {
+  border-color: #444;
 }
 
 /* Background colors for running tasks on test run pages */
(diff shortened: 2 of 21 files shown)


ppigazzini:
DEV updated.

- [ ] light/dark theme switch seems broken: icon moon does nothing and disappears

vondele:
Might be time for an upgrade indeed. 

I'll post a screenshot, colors are gone on the front page for me (firefox):

image

vdbergh:
I'm a bit confused. The commit message also mentions bootstrap 4.5.3. Why sometimes 4.5.3 and sometimes 5.1.1? Just curious.

Another thing: It's not really important but the raw stats page still uses 2.3.2. I tried to upgrade it once but it looked awful. But perhaps you know what needs to be done?

vondele:
The button 'hide non default nets' doesn't look like a button:

image

vdbergh:
@ppigazzini It is out of the scope of this PR but IMHO dark/light mode would need an overhaul. IIUC dark mode is implemented by looking for certain strings and replacing these (the replacement marked with !important to override even inline styling). This means that each time there is a small change in light mode, dark mode breaks. Dark/light modes would need proper css implementations.

ppigazzini:
DEV updated, updated also the [bugs list](https://github.com/glinscott/fishtest/pull/1044#issuecomment-924605640) with the fixes :)

ppigazzini:
DEV updated.

Could you fix the vertical alignment at right border when using time odds? (e.g. reducing the "Thread" and "Priority" boxes)
time_odds

ppigazzini:
DEV updated. All bugs fixed but the color for the test gauges in homepage/testpage with light theme, see https://github.com/glinscott/fishtest/pull/1044#issuecomment-924611957

ppigazzini:
DEV updated, colors in elo boxes OK :)

linrock:
> I'm a bit confused. The commit message also mentions bootstrap 4.5.3. Why sometimes 4.5.3 and sometimes 5.1.1? Just curious.

I'd started upgrading to bootstrap 4.5.3 last year when it was the latest version, didn't finish the upgrade, and only picked up the upgrade again recently. By now, the latest version is 5.1.1 !

> Another thing: It's not really important but the raw stats page still uses 2.3.2. I tried to upgrade it once but it looked awful. But perhaps you know what needs to be done?

I upgraded the raw stats page to use 5.1.1 and it should look ok now, but could use some spacing adjustments. The main thing that needed changing were the class names for the new [bootstrap grid system](https://getbootstrap.com/docs/5.1/layout/grid/) (i.e. need to rename `span8` to `col-8`)

> Could you fix the vertical alignment at right border when using time odds? (e.g. reducing the "Thread" and "Priority" boxes or scaling all the boxes)

sure, could we open an issue for this and revisit in a separate PR? i see this is also an issue in current master (bootstrap 2.3.2). would have to look more closely to figure out a fix

linrock:
> DEV updated, colors in elo boxes OK :)

nice, lmk if you guys find more issues, big or small. happy to fix more things in this PR, or squash to prep for merge when you feel it's ready enough

ppigazzini:
> > Could you fix the vertical alignment at right border when using time odds? (e.g. reducing the "Thread" and "Priority" boxes or scaling all the boxes)
> 
> sure, could we open an issue for this and revisit in a separate PR? i see this is also an issue in current master (bootstrap 2.3.2). would have to look more closely to figure out a fix

Sure :)

vdbergh:
> > Another thing: It's not really important but the raw stats page still uses 2.3.2. I tried to upgrade it once but it looked awful. But perhaps you know what needs to be done?
> 
> I upgraded the raw stats page to use 5.1.1 and it should look ok now, but could use some spacing adjustments. The main thing that needed changing were the class names for the new [bootstrap grid system](https://getbootstrap.com/docs/5.1/layout/grid/) (i.e. need to rename `span8` to `col-8`)
> 

Thanks! It looks much better now! Don't worry about fine cosmetics. This page is not normally visited by users.

vondele:
another minor nit on https://dfts-0.pigazzini.it/upload some space would look better between the browse and agree buttons:

image

ppigazzini (approved):
DEV looks good to me, please squash :)

ppigazzini:

snicolet:
On my Mac after the merge, all links are underlined text (both Safari and Chrome):

image

ppigazzini:
Try a page refresh.

snicolet:
I did it in Safari, and of course it reloaded the css file.

But I wonder if for such big changes to the css, it wouldn't be a good idea to *rename* the css in the page to force a fresh load of the new version by default.

vdbergh:
Maybe add a cache buster? E.g. https://css-tricks.com/strategies-for-cache-busting-css/ .

vdbergh:
I do not know how to do a hard page refresh on my iPad. However IIRC static content is only cached for 1 hour . So it is not a big deal.

QuackQuackBlah:
I really like this change! Makes it feel a lot newer! One small issue that I've found, on a Mac with Chrome I can no longer select the "view individual parameter" option for the SPSA graphs. Unclear if that's a browser issue or a fishtest issue. (should I be making a separate issue for this, or is this a fine place to say it?)

ppigazzini:
"View Individual Parameters" for SPSA graphs doesn't work also on Windows :(

https://tests.stockfishchess.org/tests/view/615193040d0a3569cbc641bd

ppigazzini:
The drop down menu works on DEV after reverting this commit.

ppigazzini:
@QuackQuackBlah fixed, refresh a couple of times the browser.

AndrewTCEC:
I think this upgrade also ruined dark theme:
Screenshot_2021-09-27-21-37-52-578_com android chrome

dav1312:
@AndrewTCEC 
Follow the instructions here: https://github.com/glinscott/fishtest/pull/1044#issuecomment-927376972

AndrewTCEC:
This is Chrome Mobile @dav1312

dav1312:
> This is Chrome Mobile @dav1312 What to do in this case?

The underlying issue is the same one, a cache one. Check if it persists in incognito mode.

AndrewTCEC:
Already fixed it by clearing cache in my browser. Thanks for help @dav1312

PR #898, opened 2021-03-23
Bug in ordering users in website: fixing attempt

G-Lorenz:
Ordering users in https://tests.stockfishchess.org/users using "last active" filter led to a wrong output: the program order alphabetically the strings in the column instead of the time passed.
immagine

So I try to introduce an user-defined html attribute whose value is the time difference used by delta_date to write the corresponding strings. In order to do this I split the function delta_date in two parts creating a new function called diff_date whose output is the desired time difference in datetime.datetime format. I rewrite every delta_date call letting them to be compliant with the new definition. Hence I converted diff in isoformat and I stored it in user["diff"]. In application.js I introduced another "or" in the first line: my aim is to let the program search "data-diff" attribute, and only if it doesn't exist it would search innerText.

As I write on discord commits are not tested.

PR #1045, opened 2021-09-22
[Wip] More sane residual(s)

vdbergh:
We compute for each worker two residuals. One sensitive to the draw ratio and one sensitive to Elo. Both are standard normal under the null hypothesis that all workers are equal.

I suspect that the many  cores servers are more noisy which may lead to a lower draw ratio. This should be picked up by the first residual. However a lower draw ratio does not necessarily have to distort the Elo contribution. Distortions which affect Elo are indicated in the second residual.

### DIFF
--- a/server/fishtest/stats/stat_util.py
+++ b/server/fishtest/stats/stat_util.py
@@ -311,43 +311,41 @@ def update_SPRT(R, sprt):
     elif sprt["llr"] > sprt["upper_bound"] - o1:
         sprt["state"] = "accepted"
 
-def residuals(coeffs, table):
-    # If k1j,..,ktj is a row in a contingency table
-    # then we associate a statistic to the sum of ai*kij
-    # which is standard normal under the null hypothesis
-    # which states that the rows of the contingency table are
-    # obtained by sampling from the same distribution.
+
+def residuals(a, k):
+    # Let a_1,...,a_t be a fixed set of numbers and let k
+    # be a contingency table. We assume that all rows of
+    # k are drawn from the same discrete probability
+    # distribution. Let k_i1,..,k_it be the i'th row in k.
+    # and let  X_i be the random variable sum_j k_ij*a_j,
+    # conditioned on the column sums of k.
     #
-    # Let Ki be the column sums, let nj be the sum of the
-    # kij and let N be the grand total. Then after conditioning
-    # on nj and K1,..,Kt, (k1j,...,ktj) follows a multivariate
-    # hypergeometric distribution. See
+    # In this function we compute (E(X_i)/StDev(X_i))_i.
+    # This calculation depends on knowing E(k_ij) and
+    # Cov(k_ij,k_ij') (conditioned on the column sums).
+    # To this end we observe that (k_ij)_j follows a
+    # multivariate hypergeometric distribrution
     # https://en.wikipedia.org/wiki/Hypergeometric_distribution
-    # We assume that the hypergeometric distribution is
-    # asymptotically normal (reference needed).
-    # According to the formulas in loc cit the covariance matrix
-    # (k1j,...,ktj) is obtained by multiplying by (N-nj)/(N-1) the
-    # covariance matrix from a multinomial distribution with
-    # probabilities Ki/N.
-    #
-    # It follows that sum_i ai*kij is asymptotically normal
-    # with expected value mu=sum_i ai*(Ki/N) per draw, expected
-    # value nj*mu and variance
-    # nj*(N-nj)/(N-1)*sum_i (ai-mu)^2*(Ki/N)
+    # and we use the expectation value and covariance
+    # matrix given in loc. cit.
 
-    Ki = numpy.sum(table, axis=0)
-    nj = numpy.sum(table, axis=1)
-    N = numpy.sum(Ki)
-    pi = Ki / N
-    mu = numpy.sum(coeffs * pi)  # expected value per draw
-    mj = nj * mu  # expected values per row
-    varj = (
-        numpy.sum((coeffs - mu) * (coeffs - mu) * pi) * nj * (N - nj) / (N - 1)
-    )  # variances per row
-    mmj = numpy.matmul(table, coeffs)
-    residuals = (mmj - mj) / numpy.sqrt(varj)
-    return residuals
+    # i=row index, j=column_index
 
+    k_dot_j = numpy.sum(k, axis=0)  # column sums
+    k_i_dot = numpy.sum(k, axis=1)  # row sums
+    k_dot_dot = numpy.sum(k_dot_j)  # grand total
+    p_j = k_dot_j / k_dot_dot  # row independent part of the expression for the
+    # expectation value of (k_ij)_j
+    cov_j_j = -numpy.outer(p_j, p_j) + numpy.diag(p_j)  # row independent part
+    # of the expression for the covariance matrix of (k_ij)_j.
+    E = a.dot(p_j)  # row independent part of the expectation
+    # value of X_i.
+    Var = a.dot(numpy.matmul(cov_j_j, a))  # row independent part of the variance of
+    # X_i
+    E_i = E * k_i_dot
+    Var_i = Var * k_i_dot * (k_dot_dot - k_i_dot) / (k_dot_dot - 1)
+    observed_i = numpy.matmul(k, a)
+    return (observed_i - E_i) / numpy.sqrt(Var_i)
 
 
 if __name__ == "__main__":
--- a/server/fishtest/stats/stat_util.py
+++ b/server/fishtest/stats/stat_util.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 import math
 
+import numpy
 import scipy.stats
 from fishtest.stats import LLRcalc, sprt
 
@@ -309,6 +310,44 @@ def update_SPRT(R, sprt):
         sprt["state"] = "rejected"
     elif sprt["llr"] > sprt["upper_bound"] - o1:
         sprt["state"] = "accepted"
+
+def residuals(coeffs, table):
+    # If k1j,..,ktj is a row in a contingency table
+    # then we associate a statistic to the sum of ai*kij
+    # which is standard normal under the null hypothesis
+    # which states that the rows of the contingency table are
+    # obtained by sampling from the same distribution.
+    #
+    # Let Ki be the column sums, let nj be the sum of the
+    # kij and let N be the grand total. Then after conditioning
+    # on nj and K1,..,Kt, (k1j,...,ktj) follows a multivariate
+    # hypergeometric distribution. See
+    # https://en.wikipedia.org/wiki/Hypergeometric_distribution
+    # We assume that the hypergeometric distribution is
+    # asymptotically normal (reference needed).
+    # According to the formulas in loc cit the covariance matrix
+    # (k1j,...,ktj) is obtained by multiplying by (N-nj)/(N-1) the
+    # covariance matrix from a multinomial distribution with
+    # probabilities Ki/N.
+    #
+    # It follows that sum_i ai*kij is asymptotically normal
+    # with expected value mu=sum_i ai*(Ki/N) per draw, expected
+    # value nj*mu and variance
+    # nj*(N-nj)/(N-1)*sum_i (ai-mu)^2*(Ki/N)
+
+    Ki = numpy.sum(table, axis=0)
+    nj = numpy.sum(table, axis=1)
+    N = numpy.sum(Ki)
+    pi = Ki / N
+    mu = numpy.sum(coeffs * pi)  # expected value per draw
+    mj = nj * mu  # expected values per row
+    varj = (
+        numpy.sum((coeffs - mu) * (coeffs - mu) * pi) * nj * (N - nj) / (N - 1)
+    )  # variances per row
+    mmj = numpy.matmul(table, coeffs)
+    residuals = (mmj - mj) / numpy.sqrt(varj)
+    return residuals
+
 
 
 if __name__ == "__main__":
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -311,12 +311,12 @@ from fishtest.util import worker_name
 
           % if 'spsa' not in run['args']:
               % if 'residual' in task and task['residual']['res_draw']!=float("inf"):
-                  <td style="background-color:${task['residual_color']['res_draw']}">${f"{task['residual']['res_draw']:.3f}"}</td>
+                  <td data-diff=${abs(task['residual']['res_draw'])} style="background-color:${task['residual_color']['res_draw']}">${f"{task['residual']['res_draw']:.3f}"}</td>
               % else:
                   <td>-</td>
               % endif
               % if 'residual' in task and task['residual']['res_elo']!=float("inf"):
-                  <td style="background-color:${task['residual_color']['res_elo']}">${f"{task['residual']['res_elo']:.3f}"}</td>
+                  <td data-diff=${abs(task['residual']['res_elo'])} style="background-color:${task['residual_color']['res_elo']}">${f"{task['residual']['res_elo']:.3f}"}</td>
               % else:
                   <td>-</td>
               % endif
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -251,9 +251,10 @@ from fishtest.util import worker_name
       <th>Crashes</th>
       <th>Time</th>
 
-      % if 'spsa' not in run['args']:
-          <th>Residual</th>
-      % endif
+      % if 'spsa' not in run['args']: 
+          <th>Residual (draws)</th>
+          <th>Residual (Elo)</th>
+     % endif
     </tr>
   </thead>
   <tbody>
@@ -309,8 +310,13 @@ from fishtest.util import worker_name
           <td>${stats.get('time_losses', '-')}</td>
 
           % if 'spsa' not in run['args']:
-              % if 'residual' in task and task['residual']!=float("inf"):
-                  <td style="background-color:${task['residual_color']}">${f"{task['residual']:.3f}"}</td>
+              % if 'residual' in task and task['residual']['res_draw']!=float("inf"):
+                  <td style="background-color:${task['residual_color']['res_draw']}">${f"{task['residual']['res_draw']:.3f}"}</td>
+              % else:
+                  <td>-</td>
+              % endif
+              % if 'residual' in task and task['residual']['res_elo']!=float("inf"):
+                  <td style="background-color:${task['residual_color']['res_elo']}">${f"{task['residual']['res_elo']:.3f}"}</td>
               % else:
                   <td>-</td>
               % endif
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1045,7 +1045,10 @@ class RunDb:
         if "bad_tasks" not in run:
             run["bad_tasks"] = []
         for task in run["tasks"]:
-            if unique_key_(task["worker_info"]) in chi2["bad_users"]:
+            if (
+                "worker_info" in task
+                and unique_key_(task["worker_info"]) in chi2["bad_users"]
+            ):
                 purged = True
                 task["bad"] = True
                 run["bad_tasks"].append(task)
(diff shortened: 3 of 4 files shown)


server/fishtest/templates/tests_view.mak
```diff
@@ -251,9 +251,10 @@ from fishtest.util import worker_name
       <th>Crashes</th>
       <th>Time</th>
 
-      % if 'spsa' not in run['args']:
-          <th>Residual</th>
-      % endif
+      % if 'spsa' not in run['args']: 
+          <th>Residual (draws)</th>
+          <th>Residual (Elo)</th>
+     % endif
```
ppigazzini, line 257: missing a space in indentation

ppigazzini:
DEV updated

vdbergh:
Hmm in 
https://dfts-0.pigazzini.it/tests/view/614053e07315e7c73204a4bb

linrock-100cores has a strongly positive draw residual and a strongly negative Elo residual. Different from what I expected.

vdbergh:
This appears to be a somewhat common pattern for linrock-100cores.

vdbergh:
Last push is just a rebase.

server/fishtest/stats/stat_util.py
```diff
@@ -310,6 +311,44 @@ def update_SPRT(R, sprt):
     elif sprt["llr"] > sprt["upper_bound"] - o1:
         sprt["state"] = "accepted"
 
+def residuals(coeffs, table):
+    # If k1j,..,ktj is a row in a contingency table
```
ppigazzini, line 315: Textbooks usually use this convention for contingency table:
 - table indices [row, column]: `Kij`
 - row sum `Ki.`
 - column sum `K.j`
 - total sum `K..`

See [this handout](http://fisher.stats.uwo.ca/faculty/kulperger/SS3858/Handouts/ContingencyTables.pdf) that follows "Categorical Data Analysis" by Agresti.

ppigazzini:
DEV updated.

server/fishtest/stats/stat_util.py
vdbergh: Makes sense! I will fix it.

vdbergh:
When sorting on residual it would be nice if it was done by absolute value... But I do not know how the table sorting works... I assume that somewhere there is some javascript that is responsible but I can't find it.

ppigazzini:
> When sorting on residual it would be nice if it was done by absolute value... But I do not know how the table sorting works... I assume that somewhere there is some javascript that is responsible but I can't find it.

See:
#295 original PR 
#898 improved the ordering by date

vdbergh:
Thanks! That works indeed. I haven't done the other things yet since I would like to see if it possible to get the basic functionality right.

ppigazzini:
DEV updated

vdbergh:
Hmm purging doesn't work.

ppigazzini:
DEV running only with this PR, try now (if I'm not wrong I was able to purge an old test).

vdbergh:
I did not recognize a large negative residual as bad. Forgotten `abs(..)`.

ppigazzini:
DEV updated, loaded a recent db.

vdbergh:
Still not working. It's purging the wrong users.

ppigazzini:
I only checked if a worker was purged: day in :)

vdbergh:
```
if chi2["p"] < 0.001 or abs(task["residual"]["res_elo"]) > 7.0:
```
This is in util.py to select "bad users". It seems purging will almost never do anything these days.

vdbergh:
Ah. Because I set the residuals to infinity for workers that do not have enough games, a poor 1 core workers will get booted... This is also a bug in master.

ppigazzini:
Set by hand on DEV:
`if chi2["p"] < 0.05 or abs(task["residual"]["res_elo"]) > 3.0:`

vdbergh:
Interesting fact
```
>>> a= float("nan")
>>> a==a
False
>>> a is a
True
>>> 
```

ppigazzini:
https://towardsdatascience.com/navigating-the-hell-of-nans-in-python-71b12558895b

vdbergh:
> https://towardsdatascience.com/navigating-the-hell-of-nans-in-python-71b12558895b
> https://towardsdatascience.com/5-methods-to-check-for-nan-values-in-in-python-3f21ddd17eed
> 
> ```python
> def isNaN(num):
>     return num != num
> ```

:smile:

ppigazzini:
I will be at the keyboard in 90 minutes.

ppigazzini:
DEV updated

vdbergh:
Here one can see the effect of purging linrock-100cores

https://dfts-0.pigazzini.it/tests/view/6146e73efb044eb7f5c1701e

The LLR drops to -4.50 so the test should have finished earlier. We should try to understand this better.

Unfortunately the residuals were incorrect (they were genuine residuals but they did not correspond to draws and Elo).

I am not sure if they will be recomputed for the purged tasks.

ppigazzini:
DEV updated, db reloaded :)

vdbergh:
linrock-100cores has now a benign Elo residual. So in the latest PR I use draw residual for purging. This needs a better solution but the code has to rewritten a bit.

ppigazzini:
DEV updated, db reloaded

vdbergh:
Alas, purging is now again not working. I don't understand why. Tomorrow....

vdbergh:
I wonder if there were exceptions. The website seemed sluggish sometimes.

ppigazzini:
`journalctl -u fishtest@6543 --since "2 hours ago"` shows this, neve saw before, could be due to the update of the db

```
Sep 22 20:19:26 vps1572612706 pserve[25291]: Overview exception: error processing query: ns=fishtest_new.runsTree: $and
Sep 22 20:19:26 vps1572612706 pserve[25291]: Sort: {}
Sep 22 20:19:26 vps1572612706 pserve[25291]: Proj: {}
Sep 22 20:19:26 vps1572612706 pserve[25291]:  planner returned error :: caused by :: hint provided does not correspond to an existing index, full error: {'ok': 0.0, 'errmsg': 'error processing query: ns=fishtest_new.runsTree: $and\nSort:
```

ppigazzini:
DEV updated, DB reloaded.
I will be at the keyboard in 2 hours.

vdbergh:
Purging works again. Will try to clean it up.

ppigazzini:
DEV updated, DB reloaded.

server/fishtest/stats/stat_util.py
```diff
@@ -310,6 +311,44 @@ def update_SPRT(R, sprt):
...
+    # then we associate a statistic to the sum of ai*kij
+    # which is standard normal under the null hypothesis
+    # which states that the rows of the contingency table are
+    # obtained by sampling from the same distribution.
+    #
+    # Let Ki be the column sums, let nj be the sum of the
+    # kij and let N be the grand total. Then after conditioning
+    # on nj and K1,..,Kt, (k1j,...,ktj) follows a multivariate
+    # hypergeometric distribution. See
+    # https://en.wikipedia.org/wiki/Hypergeometric_distribution
```
ppigazzini, line 325: I'm confused by the reference to the hypergeometric distribution :(

Data in contingency table should follow:
 - poisson distribution if the marginal totals are not fixed by the sampling process
 - multinomial distribution if the grand total is fixed by the sampling process (NumGames test) or if one of the marginal total (in our case the games played by a worker) is fixed by the sampling process (this should be SPRT test analyzed by tasks) or we have explanatory variables (the workers) so we can conditioning on those marginal totals (poisson distribution conditioned on a total is a multinomial distribution, this should be SPRT test analyzed by workers)
 - hypergeometric distribution if both marginal totals are fixed by the sampling process

I mean: we can design a random test with the explanatory variables (workers) playing those numbers of games, but we can't design a random process with the workers having those results.

vdbergh:
This is how I see it. You cannot fix the columns sums before hand but you can still condition on them. This gives you a valid probability distribution of the type `P(X|Y)` for random variables `X` and `Y`).

The Fisher exact test is based on conditioning on both row and column sums in a 2x2 table (so there is only 1 free variable).

The controversy is not over the fact if the Fisher exact test is mathematically correct (it is), but whether it sacrifices power.

vdbergh:
Trying to purge this one 

https://dfts-0.pigazzini.it/tests/view/51565cdd8f0c3e218b0b82a0

gives an exception.

vdbergh:
It seems we are getting an exception when there are no bad workers.

vdbergh:
Hmmm I can't reproduce it locally.

ppigazzini:
```
Sep 23 14:00:30 vps1572612706 pserve[6791]:     purged = request.rundb.purge_run(run)
Sep 23 14:00:30 vps1572612706 pserve[6791]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 1045, in purge_run
Sep 23 14:00:30 vps1572612706 pserve[6791]:     if unique_key_(task["worker_info"]) in chi2["bad_users"]:
Sep 23 14:00:30 vps1572612706 pserve[6791]: KeyError: 'worker_info'
```

ppigazzini:
I don't understand what that `residuals()` is, what is "ai", because everything starts from a hypergeometric distribution and where the correlation matrix based on multinomial distribution comes from. :(

The Pearson adjusted residuals come from the paper ["The analysis of residuals in cross-classification tables" by JS Haberman](https://www.jstor.org/stable/2529686?seq=2#metadata_info_tab_contents) with data from poisson or multinomial distribution (they share the same ML estimation of Eij = Ni.N.j/N..). Our tests (NumGames and SPRT) use a sampling with multinomial distribution, so we use that formula in master:
https://github.com/glinscott/fishtest/blob/a53918da5fc767fbaafcd7cf3ed38cbf59b91975/server/fishtest/util.py#L107-L117

vdbergh:
What I define is a generalisation.

Let a row be k_i and let a_i be fixed numbers. I compute the expected value and standard deviation for the random variable sum_i a_i*k_i (conditioned on the row and column sums).

If there is one a_i that is 1 and the others are all zero then one obtains the usual adjusted residual (up to a factor sqrt((N-1)/N) where N is the grand total). I just checked this.

I am applying this to the pentanomial draw ratio (this is just a standard adjusted residual) and to -p[0]-p[1]+p[3]+p[4] where the p’s are pentanomial frequencies.

vdbergh:
Note that -2*p[0]-p[1]+p[3]+2*p[4] would be better but I wanted to keep it simple by recycling the table used for the chi2 computation.

vdbergh:
I missed your question about the multinomial distribution.

Let the notations be as in my email above. When conditioning  on the column sums K_i and the row sum n, the k_i (a row) follow a multivariate hypergeometric distribution (under the null hypothesis that all rows are drawn from the same distribution).

To calculate the expected value of sum_i a_i*k one needs to know the expected values of the k_i and the covariance matrix of the k_i. The formulas are in Wikipedia.

If I did not make a mistake, then the formulas are the same as for a multinomial distribution with probabilities p_i=K_i/N (N is the grand total) except that the covariance matrix must be multiplied with (N-n)/(N-1). This is what I am using.

vdbergh:
Maybe an implementation using the covariance matrix directly would be clearer….

ppigazzini:
DEV updated only with your PR and DB reloaded.

After purging (old or new test) the server reload the home page, now we got this exception in home page:
```
Sep 24 08:41:46 vps1572612706 pserve[15027]:   File "/home/usr00/fishtest/server/fishtest/templates/run_tables.mak", line 2, in render_body
Sep 24 08:41:46 vps1572612706 pserve[15027]:     <% pending_approval_runs = [run for run in runs['pending'] if not run['approved']] %>
Sep 24 08:41:46 vps1572612706 pserve[15027]:   File "/home/usr00/fishtest/server/fishtest/templates/run_tables.mak", line 2, in <listcomp>
Sep 24 08:41:46 vps1572612706 pserve[15027]:     <% pending_approval_runs = [run for run in runs['pending'] if not run['approved']] %>
Sep 24 08:41:46 vps1572612706 pserve[15027]: KeyError: 'approved'
```

We got this exception after reloading an old page after purging
https://dfts-0.pigazzini.it/tests/view/51565cdd8f0c3e218b0b82a0

```
Sep 24 08:51:54 vps1572612706 pserve[15027]:   File "/home/usr00/fishtest/server/fishtest/templates/base.mak", line 121, in render_body
Sep 24 08:51:54 vps1572612706 pserve[15027]:     <main>${self.body()}</main>
Sep 24 08:51:54 vps1572612706 pserve[15027]:   File "/home/usr00/fishtest/server/fishtest/templates/tests_view.mak", line 112, in render_body
Sep 24 08:51:54 vps1572612706 pserve[15027]:     class="btn ${'btn-success' if run['base_same_as_master'] else 'btn-warning'}">
Sep 24 08:51:54 vps1572612706 pserve[15027]: KeyError: 'base_same_as_master'
```

A new test opens correctly after purging:
https://dfts-0.pigazzini.it/tests/view/614a300e7bdc23e77ceb885c

Loading master doesn't solve the home page problem, reloading the db of course yes.

ppigazzini:
DEV now with master and fresh DB (lowered by hand the bad worker requirements).

vdbergh:
These errors seem completely unrelated. Very weird.

vdbergh:
So after a db update there must be an exception with the current pr (probably after purging).

After this I think the db state becomes inconsistent.

ppigazzini:
Yes, the two error are unrelated, the most important error is the first one.

Purging works fine on DEV with master and a fresh reloaded DB.
I just purged a little worker (residual "inf") from a finished 40k games test, the test is now queued in home page to play the deleted games:
https://dfts-0.pigazzini.it/tests/view/61489ef77bdc23e77ceb8748

ppigazzini:
The two errors are related.

DEV with PR and a fresh DB purges correctly a new test.

After purging an old test, the old test is added in the queue, but the old test lacks some required data (eg the key "approved")

In master the purge of an old test stops with exception before moving the old test in the queue.

IMHO we should disable active action (purge, reschedule) on very old tests.

vdbergh:
> IMHO we should simply disable the purge on very old tests.

Yes that makes a lot of sense. Old tests should be viewed for historical interest. But not altered. It is too messy.

ppigazzini:
This lazy solution didn't worked :(
```python
    def purge_run(self, run):
        for task in run["tasks"]:
            if "worker_info" not in task:
               return {}
```

I poisoned the DB with this old test that has "worker_info" but lacks "throughput"
https://dfts-0.pigazzini.it/tests/view/513231858f0c3e478e2ebebd
```
Sep 24 11:05:52 vps1572612706 pserve[24068]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 576, in sync_request_task
Sep 24 11:05:52 vps1572612706 pserve[24068]:     self.calc_itp(run)
Sep 24 11:05:52 vps1572612706 pserve[24068]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 502, in calc_itp
Sep 24 11:05:52 vps1572612706 pserve[24068]:     itp = run["args"]["throughput"]
```

ppigazzini:
We want to skip old test, so perhaps better to return in `purge_run()` for a test older than X days.

vdbergh:
> This lazy solution didn't worked :(
> 
> ```python
>     def purge_run(self, run):
>         for task in run["tasks"]:
>             if "worker_info" not in task:
>                return {}
> ```
> 

Maybe it is sufficient to immediately return False in purge_run (or True, it only matters for the message that is displayed)? Initially when worker_info is not in task. But we could refine it later.


> I poisoned the DB with this old test that has "worker_info" but lacks "throughput"
> https://dfts-0.pigazzini.it/tests/view/513231858f0c3e478e2ebebd
> 
> ```
> Sep 24 11:05:52 vps1572612706 pserve[24068]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 576, in sync_request_task
> Sep 24 11:05:52 vps1572612706 pserve[24068]:     self.calc_itp(run)
> Sep 24 11:05:52 vps1572612706 pserve[24068]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 502, in calc_itp
> Sep 24 11:05:52 vps1572612706 pserve[24068]:     itp = run["args"]["throughput"]
> ```

ppigazzini:
This should works, I'm reloading a clean DB:
```python
    def purge_run(self, run):
        now = datetime.utcnow()
        if (now - run["start_time"]).days > 30:
            return {}
        # Remove bad tasks
        purged = False
```

vdbergh:
> This should works, I'm reloading a clean DB:
> 
> ```python
>     def purge_run(self, run):
>         now = datetime.utcnow()
>         if (now - run["start_time"]).days > 30:
>             return {}
>         # Remove bad tasks
>         purged = False
> ```

I think you need to return a boolean, not an empty dictionary.

ppigazzini:
It works, thank you for the fix :)

DEV is running with bad worker p-value<0.5 and residual >2.0

ppigazzini:
Feel free to chose the day delta.

ppigazzini:
Please mind that I fenced also the "start_time" key :)
```python
    def purge_run(self, run):
        now = datetime.utcnow()
        if "start_time" not in run or (now - run["start_time"]).days > 30:
            return False
        # Remove bad tasks
        purged = False
```

ppigazzini:
Fast fix for PROD with #1046

vdbergh:
I am having second thoughts about this. I had hoped to see a clearer pattern in what causes high residuals, but it seems the signal to noise ratio is too high. So I am going to close this PR for now.

ppigazzini:
I missed your latest push!
I updated DEV a couple of times today. To be sure double check DEV, I just started the updated.

vdbergh:
The latest push was nonfunctional (an attempt to more clearly describe the procedure). However I think the two columns with residuals (with signs) are really non-information overload.

ppigazzini:
Ok :)

vdbergh:
I had another idea for a more sane residual.

The current residual is the maximum of the absolute value 3 standard normal variables. Unfortunately these variables are not independent so I do not know how to compute the distribution of the residual that is shown. So we do not know if all the reds we are seeing are real or just statistical fluctuations.

What we can do however is to compute for every row the chi2 of the 2x3 contingency table consisting of that row and the sum of all the other rows. Then we get a residual with a chi2 distribution with two degrees of freedom (under the null hypothesis). So easy to understand statistically.

Instead of the residual it would be better to show the p-value, and to use the Bonferroni correction to color it.

With the refactoring of the purging logic, this should be much easier.

ppigazzini:
I viewed this toy example for data analysis with chi2:
 - counts of some fixed words taken a random chapter from few books of one author, same counts on a random chapter of another author's book 
 - chi2 to check for homogeneity the data from only the first author (table: number of books x number of words). If data are homogeneous we can sum the word counts
 - chi2 to check for homogeneity the totals of the first author and the counts of the second one (table: number of authors x number of words)
 - find the cells with the higher chi2 contributions (we use abs(adjusted residual))
 
I'm wondering if in your proposition it's correct to sum up all the other rows (the data couldn't be homogeneous).

Partitioning table seems to have to fulfill some requirements:
https://people.musc.edu/~bandyopd/bmtry711.11/lecture_10.pdf

vdbergh:
> I viewed this toy example for data analysis with chi2:
> 
> * counts of some fixed words taken a random chapter from few books of one author, same counts on a random chapter of another author's book
> * chi2 to check for homogeneity the data from only the first author (table: number of books x number of words). If data are homogeneous we can sum the word counts
> * chi2 to check for homogeneity the totals of the first author and the counts of the second one (table: number of authors x number of words)
> * find the cells with the higher chi2 contributions (we use abs(adjusted residual))
> 
> I'm wondering if in your proposition it's correct to sum up all the other rows (the data couldn't be homogeneous).

I think it is correct under the null hypothesis (which says that the data is homogeneous).

The principle is quite elegant. Obviously the global `chi2` is the sum of the `chi2` of the rows. Now what seems to be the case is that each `chi2[row]/(1-row_sum/grand_total)` is distributed as `chi2` with `df=#columns-1` (under the null hypothesis). This is a kind of "adjusted `chi2`". 

If the table has two columns one gets the square of the usual Pearson adjusted residual (the two cells in a row have the same adjusted residual if I computed correctly). So in general we can use the adjusted `chi2[row]` as a residual for the row (possibly converted to a `p`-value or a `z`-value for cosmetics) instead of the current ugly `max(abs(..),..,abs(..))` (with unknown distribution).




> 
> Partitioning table seems to have to fulfill some requirements:
> https://people.musc.edu/~bandyopd/bmtry711.11/lecture_10.pdf

PR #1046, opened 2021-09-24
Purge a test only if not older than 30 days

ppigazzini:
After the purge a test is automatically queued to play the purged games.
An old test could lack some dictionary keys making unsafe the queueing.

Thanks to @vdbergh for reporting and debugging the problem.


ppigazzini:
@vdbergh I fixed PROD with this PR, I was a bit nervous with the DB poisoning :)
Feel free to improve the fix.

PR #1050, opened 2021-09-27
Fix toggling individual spsa parameters dropdown

linrock:
The problem was newer bootstrap uses a new `data-bs-toggle` attribute to mark a button as a dropdown toggle.

Fixes https://github.com/glinscott/fishtest/pull/1044#issuecomment-927500322


ppigazzini (approved):
DEV looks good!

Untitled

ppigazzini:

PR #1051, opened 2021-09-27
Enable using dynamic cachebusters for static files

linrock:
This enables auto-updating cache buster query params when static files change. The goal is to:
- not require users to manually hard refresh the page when css/js files change
- not require devs to manually update cache buster query params when css/js files change

This implementation:
- calculates the md5sum of the listed static files when the server starts
- uses the md5sum of the static file as a query param when serving the file

Prompted by https://github.com/glinscott/fishtest/pull/1044#issuecomment-927375281

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -1,4 +1,6 @@
+import hashlib
 import os
+from pathlib import Path
 
 from fishtest.rundb import RunDb
 from pyramid.authentication import AuthTktAuthenticationPolicy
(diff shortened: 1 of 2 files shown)


ppigazzini (dismissed):
Nice :)

Untitled

linrock:
> I'm not able to logout also after some CRTL+F5. I'm wondering if can due to this patch.

weird, does that not happen in master? according to the code, logout will not work if the `application.js` doesn't finish loading before clicking logout. this is possible since this JS file is loaded with `defer`.

i don't see how this patch could directly cause logout to sometimes stop working, and i can't repro in my local dev

ppigazzini (approved):

PR #1057, opened 2021-10-01
Automatically set dark.theme.css sha256 during server init

linrock:
This makes edits to `dark.theme.css` more convenient, since we'll no longer have to manually calculate the sha256sum and update `application.js` to fix the dark/light toggle.

The expected behavior is now:
- make `dark.theme.css` changes
- restart the server, and the toggle will automatically work

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -1,6 +1,7 @@
 import hashlib
 import os
 from pathlib import Path
+import subprocess
 
 from fishtest.rundb import RunDb
 from pyramid.authentication import AuthTktAuthenticationPolicy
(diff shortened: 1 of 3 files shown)


server/fishtest/__init__.py
```diff
@@ -23,15 +24,25 @@ def main(global_config, **settings):
...
+        return Path(__file__).parent / "./static/{}".format(static_file_path)
+
+    def static_file_hash(static_file_path):
+        with open(static_file_full_path(static_file_path), "r") as f:
             return hashlib.md5(f.read().encode("utf-8")).hexdigest()
 
+    # the same hash calculated by browser for sub-resource integrity checks:
+    # https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
+    dark_theme_sha256_integrity = subprocess.run(
+        "sha256sum {} | awk '{{ print $1 }}' | xxd -r -p | base64".format(
```
ppigazzini, line 37: `"openssl {} dgst -sha256 -binary | openssl base64 -A".format(`

this Mozilla suggestion uses only `openssl` (simpler to add the requirement in the server setup script), already tested working with your code

server/fishtest/static/js/application.js
```diff
@@ -55,12 +55,13 @@ $(() => {
     let theme = $.cookie('theme') || 'light';
     $("#change-color-theme").click(function() {
       if (theme === 'light') {
+        const darkThemeSha256 = $("meta[name='dark-theme-sha256']").attr("content");
```
ppigazzini, line 58: Please delete these lines in the comment or use the same command eg `openssl theme.dark.css dgst -sha256 -binary | openssl base64 -A`

https://github.com/glinscott/fishtest/blob/4c0987966e0e9f09d7c73f632b2500de0d8114ba/server/fishtest/static/js/application.js#L53-L54

ppigazzini (commented):
Nice!
Works on DEV.

ppigazzini (approved):

ppigazzini:
PROD updated, great patch, thank you @linrock :)

PR #1059, opened 2021-10-02
Suggested fix for #1058.

### LINKED ISSUE #1058: Explicitly setting residuals for tasks does not work anymore (bug)
The new purging logic purges workers (based on their residuals), not tasks. So the current code which sets the residual to 8 for a task with several crashes has no effect anymore (since it does not change the residual of the worker).

The most easy solution I think would be to start purge_run with a loop which filters tasks with crashes and time losses, and then to compute the worker based residuals (since these will be affected by the preliminary filtering).

vdbergh:
The new purging logic purges workers (based on their residuals), not tasks. So the current code which sets the residual to 8 for a task with several crashes has no effect anymore (since it does not change the residual of the worker).
    
The implemented solution is to start `purge_run()` with a loop which filters tasks with crashes and time losses, and then to compute the worker based residuals (since these will be affected by the preliminary filtering).
    
Fixes #1058.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -16,7 +16,7 @@ from bson.objectid import ObjectId
 from fishtest.actiondb import ActionDb
 from fishtest.userdb import UserDb
 from fishtest.util import (
-    update_residuals,
+    crash_or_time,
     estimate_game_duration,
     format_bounds,
     format_results,
@@ -25,6 +25,7 @@ from fishtest.util import (
     post_in_fishcooking_results,
     remaining_hours,
     unique_key as unique_key_,
+    update_residuals,
     worker_name,
 )
 from pymongo import DESCENDING, MongoClient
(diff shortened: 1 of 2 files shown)


ppigazzini:
Tonight I dirty fixed PROD by hand with this to filter out the time losses:
```python
        # Special case crashes or time losses
        stats = task.get("stats", {})
        total = stats.get("wins", 0) + stats.get("losses", 0) + stats.get("draws", 0)
        crashes = stats.get("crashes", 0)
        time_losses = stats.get("time_losses", 0)
        if crashes > 3:
            task["residual"] = 8.0
        if total > 20 and time_losses / total > 0.1:
            task["residual"] = 8.0
```

ppigazzini:
I successfully purged and DEV thsi test:
https://dfts-0.pigazzini.it/tests/view/615750702d02f48db3961b00
Here the same test on PROD:
https://tests.stockfishchess.org/tests/view/615750702d02f48db3961b00
The worker@71cores with 0 games played now finally hasn't a residual

vdbergh:
I have squashed the PR.

ppigazzini:
DEV updated, DB reloaded, looks good to me, both tests successfully purged with no residual for the 0 games worker@71
https://dfts-0.pigazzini.it/tests/view/6157676e2d02f48db3961b1a
https://dfts-0.pigazzini.it/tests/view/615750702d02f48db3961b00

DB reloaded for your inspection.

Let me known when this is ready.

vdbergh:
It is weird that the residuals of the tasks that were purged for time losses were not set to 10. I thought I fixed this in the
latest commit.
```
tasks = copy.copy(run["tasks"])
   for task in tasks:
      if crash_or_time(task):
                purged = True
                # The next two lines are a bit hacky but
                # the correct residual and color may not have
                # been set yet.
                task["residual"] = 10.0
                task["residual_color"] = "#FF6A6A"
                task["bad"] = True
                run["bad_tasks"].append(task)
                run["tasks"].remove(task)
```

vdbergh:
I guess the two links you gave 

https://dfts-0.pigazzini.it/tests/view/6157676e2d02f48db3961b1a
https://dfts-0.pigazzini.it/tests/view/615750702d02f48db3961b00

were already purged on master. I purged them again and the new purge appeared to work correctly.

ppigazzini:
Yes, purged on master last night, then the worker@71cores started new tasks.

ppigazzini:
On master the purge worked only the first time.

vdbergh:
I purged this one.
 
https://dfts-0.pigazzini.it/tests/view/6156ec812d02f48db3961ac0

But now it purged the worker with time losses and linrock-100cores in one go. I think this is not what one wants.

ppigazzini:
No, we want to purge only one worker for one button push.

In the ideal word time losses and crashes should be dropped without statistical test.

vdbergh:
I think I fixed it. If it is possible to reload the db....

ppigazzini:
DB reloading, it takes 5 minutes...

ppigazzini:
DEV updated, DB reloaded, serve yourself :)

vdbergh:
Yes now this 

https://dfts-0.pigazzini.it/tests/view/6156ec812d02f48db3961ac0

worked correctly. The first purge purged the worker with time losses. The second purge purged linrock-100cores.

The other two tests you did also worked correctly. So I think is PR is ready.

It should not be difficult to enable autopurge in case of crashes or time losses, but first let's relax a bit.

ppigazzini (approved):

ppigazzini:

PR #1061, opened 2021-10-02
Raise worker version to 110: drop old "requests"

ppigazzini:
Raise also server side.

First update to rename the "requests" directory to "_requests".


ppigazzini:
See #1017 #1060 #1057

PR #1063, opened 2021-10-03
Only LTC task to Windows high concurrency worker

ppigazzini:
A Windows 10 worker running with a modern CPU (it should be a 36 cores + HT)
had many time losses with concurrency 64 or 71 in SMPT test at 8 threads 20+0.2
With the same configuration the worker ran without problems all his LTC task.
The Windows 10 Pro seems to be suboptimal for high count threads, see
https://www.anandtech.com/show/15483/amd-threadripper-3990x-review/3

Thanks to: @Vizvezdenec (bug report) @vondele (fix idea)

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -697,11 +697,17 @@ class RunDb:
 
             # To avoid time losses in the case of large concurrency and short TC,
             # probably due to cutechess-cli as discussed in issue #822,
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, looks good to me.
- windows worker at 47 cores got only LTC.
- linux worker at 32 cores got LTC and SMP(LTC).

PR #1072, opened 2021-10-11
Eliminate the global variable "rate" in the worker.

vdbergh:
It may be a good idea to try to eliminate some global variables in the worker as they make the logic hard to follow. Here a small edit that removes the global variable `rate`.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -108,11 +108,7 @@ def worker_exit(val=1):
     os._exit(val)
 
 
-rate = None
-
-
 def get_rate():
-    global rate
     try:
         rate = requests.get(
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
DEV test passed.
```
Worker version 111 connecting to https://dfts-0.pigazzini.it:443
UUID: 248d9d49-06b4-4a72-8bc8-14bac8c2e9f7
Start heartbeat
API call rate limits: {'limit': 60, 'remaining': 0, 'reset': 1633983154, 'used': 60, 'resource': 'core'}
Exception accessing host:
Near API limit
  Send heartbeat for 248d9d49-06b4-4a72-8bc8-14bac8c2e9f7 ... Pleased to hear from you...
```

PR #1073, opened 2021-10-11
Use same template for exceptions

ppigazzini:


ppigazzini:
Please let me know the lines where switch to stdout, I will fix it and merge in a couple of hours.

ppigazzini:
https://github.com/glinscott/fishtest/blob/f87debf2516df269e3500d3a23cbb658c2aee63e/worker/worker.py#L217-L227

I'm testing the Exceptions, I'm wondering if that `finally` clause should be `else:` ...

PR #1076, opened 2021-10-14
Daemonize the heartbeat thread

ppigazzini:
And use the more appropriate sys.exit()


vdbergh:
So the exit code carries essentially no information.

PR #1078, opened 2021-10-14
More refactorings.

vdbergh:
Get rid of all calls to worker_exit().

Eliminate global variable "config".

Move all parsing of parameters to setup_config_file().

Rename main() to worker().

Refactor adjust_tc.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -19,13 +19,10 @@ from functools import partial
 IS_WINDOWS = "windows" in platform.system().lower()
 
 try:
-    from ConfigParser import SafeConfigParser
-
-    config = SafeConfigParser()
+    from ConfigParser import SafeConfigParser as ConfigParser
 except ImportError:
-    from configparser import ConfigParser  # Python3
+    from configparser import ConfigParser  # Python 3
 
-    config = ConfigParser(interpolation=None)
 import base64
 import zlib
 from datetime import datetime
@@ -38,17 +35,21 @@ except ImportError:
     sys.path.append(path.join(path.dirname(path.realpath(__file__)), "packages"))
     import requests
 
-from games import run_games
+from games import run_games, FatalException
 from updater import update
 
 WORKER_VERSION = 112
 HTTP_TIMEOUT = 15.0
 
 
-def setup_config_file(config_file):
-    """Config file setup, adds defaults if not existing"""
+def setup_parameters(config_file):
+
+    # Step 1: read the config file if it exists.
+    config = ConfigParser()
     config.read(config_file)
 
+    # Step 2: replace missing config options
+    # by defaults,
     mem = 0
     system_type = platform.system().lower()
     try:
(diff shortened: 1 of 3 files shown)


vdbergh:
Ok I see it I think.

vdbergh:
3 cpus is the default config option. I set it to 1 now. But maybe it is better to set the default to max(1,multiprocessing.cpu_count()-1)? What do you think?

vdbergh:
Or `min(3,max(1,multiprocessing.cpu_count()-1))`.

ppigazzini:
I agree.

worker/worker.py
```diff
@@ -196,23 +359,32 @@ def fetch_and_handle_task(worker_info, password, remote, current_state):
...
 
-    success = True
     run, task_id = req["run"], req["task_id"]
     current_state["run"] = run
     current_state["task_id"] = task_id
+
+    success = False
     try:
         pgn_file = run_games(worker_info, password, remote, run, task_id)
+        succes = True
```
ppigazzini, line 377: Typo

worker/worker.py
```diff
@@ -340,115 +460,11 @@ def main():
...
-
-    with open(config_file, "w") as f:
-        config.write(f)
-    if options.only_config == "True":
-        worker_exit(0)
-    # We are now finished with handling command line parameters and the config file
+    options = setup_parameters(config_file)
+    if options is None:
+        return 1
+    if options.only_config==True:
```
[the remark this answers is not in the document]
vdbergh: Yep :(

ppigazzini:
Started a couple of workers on DEV.

vdbergh:
I guess strictly speaking, `fish.exit` should also have return code `0`. I think this was the old behaviour since in that case `worker_exit()` wasn't called.

worker/worker.py
```diff
@@ -522,8 +539,10 @@ def main():
...
     print("Waiting for the heartbeat thread to finish...")
     heartbeat_thread.join()
-    print("Finishing the worker normally")
-
+    
+    return 0 if fish_exit else 1
 
 if __name__ == "__main__":
-    main()
+    ret=worker()
```
ppigazzini, line 546: If I'm not wrong `sys.exit(worker())` should made the work, see
https://softwareengineering.stackexchange.com/questions/418600/best-practice-for-python-main-function-definition-and-program-start-exit

vdbergh:
I mean if something goes wrong.

ppigazzini:
We had so many naked except :(

I'm wondering if we can retire:
```python
def worker_exit(val=1):
    os._exit(val)
```

ppigazzini:
`sys.exit()` is the standard way to exit, I'm to retire `worker_exit()`

vdbergh:
Calling it a day now.

ppigazzini (approved):

PR #1081, opened 2021-10-16
Give server feedback for failed tasks.

vdbergh:
Post a message in the server log.
    
Post a message in the actiondb.
    
Introduce "WorkerException" for standard exceptions.
    
One does not want to record all exceptions on the server since this would include things like syntax errors, potentially even leading to privacy issues.
    
If one raises WorkerException(reason) in run_games() then the reason will be recorded on the server using the "failed_task" api. Exceptions caused by syntax errors etc... are of course not of the type WorkerException. Such exceptions will be recorded on the server with the file name and the line number where the exception occurred.
    
Also: fix a bug in pgn uploading created by the last commit.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -32,6 +32,10 @@ class FatalException(Exception):
     pass
 
 
+class WorkerException(Exception):
+    pass
+
+
 def is_windows_64bit():
     if "PROCESSOR_ARCHITEW6432" in os.environ:
         return True
@@ -83,9 +87,8 @@ def required_net(engine):
 
     p.wait()
     p.stdout.close()
-
     if p.returncode != 0:
-        raise Exception("UCI exited with non-zero code {}".format(p.returncode))
+        raise WorkerException("UCI exited with non-zero code {}".format(p.returncode))
 
     return net
 
@@ -114,10 +117,12 @@ def verify_required_cutechess(cutechess):
     p.stdout.close()
 
     if p.returncode != 0:
-        raise Exception("Failed to find cutechess version info")
+        raise WorkerException("Failed to find cutechess version info")
 
     if (major, minor) < (1, 2):
-        raise Exception("Requires cutechess 1.2 or higher, found version doesn't match")
+        raise WorkerException(
+            "Requires cutechess 1.2 or higher, found version doesn't match"
+        )
 
 
 def required_net_from_source():
@@ -215,12 +220,13 @@ def verify_signature(engine, signature, remote, payload, concurrency, worker_inf
         p2.stdout.close()
 
         if p.returncode:
-            raise Exception("Bench exited with non-zero code {}".format(p.returncode))
+            raise WorkerException(
+                "Bench exited with non-zero code {}".format(p.returncode)
+            )
         if p2.returncode:
-            raise Exception(
+            raise WorkerException(
                 "Compiler info exited with non-zero code {}".format(p2.returncode)
             )
-
         if int(bench_sig) != int(signature):
             message = "{}-{}cores-{}: Wrong bench in {} Expected: {} Got: {}".format(
                 worker_info["username"],
@@ -232,7 +238,11 @@ def verify_signature(engine, signature, remote, payload, concurrency, worker_inf
             )
             payload["message"] = message
             send_api_post_request(remote + "/api/stop_run", payload)
-            raise Exception(message)
+            # more compact message for "/api/failed_task"
+            message = "Wrong bench in {}... Expected: {} Got: {}".format(
+                os.path.basename(engine)[:16], signature, bench_sig
+            )
+            raise WorkerException(message)
 
     finally:
         if concurrency > 1:
@@ -255,7 +265,7 @@ def setup(item, testing_dir):
                 f.write(b64decode(blob_json["content"]))
             break
     else:
-        raise Exception("Item {} not found".format(item))
+        raise WorkerException("Item {} not found".format(item))
 
 
 def gcc_props():
@@ -280,7 +290,7 @@ def gcc_props():
     p.stdout.close()
 
     if p.returncode != 0:
-        raise Exception(
+        raise WorkerException(
             "g++ target query failed with return code {}".format(p.returncode)
         )
 
@@ -312,7 +322,9 @@ def make_targets():
     p.stdout.close()
 
     if p.returncode != 0:
-        raise Exception("make help failed with return code {}".format(p.returncode))
+        raise WorkerException(
+            "make help failed with return code {}".format(p.returncode)
+        )
 
     return targets
 
@@ -340,9 +352,7 @@ def find_arch_string():
             and "x86-64-avx512" in targets
         ):
             res = "x86-64-avx512"
-            res = (
-                "x86-64-bmi2"
-            )  # use bmi2 until avx512 performance becomes actually better
+            res = "x86-64-bmi2"  # use bmi2 until avx512 performance becomes actually better
         elif (
             "-mbmi2" in props["flags"]
             and "x86-64-bmi2" in targets
@@ -416,7 +426,9 @@ def setup_engine(
             ):
                 download_net(remote, testing_dir, net)
                 if not validate_net(testing_dir, net):
-                    raise Exception("Failed to validate the network: {}".format(net))
+                    raise WorkerException(
+                        "Failed to validate the network: {}".format(net)
+                    )
             shutil.copyfile(os.path.join(testing_dir, net), net)
 
         ARCH = find_arch_string()
@@ -434,10 +446,12 @@ def setup_engine(
             )
         except Exception as e:
             print("Exception striping binary:\n", e, sep="", file=sys.stderr)
-
-        shutil.move("stockfish" + EXE_SUFFIX, destination)
-    except:
-        raise Exception("Failed to setup engine for {}".format(sha))
+        try:
+            shutil.move("stockfish" + EXE_SUFFIX, destination)
+        except Exception:
+            raise WorkerException(
+                "Unable to move the stockfish binary to its destination"
+            )
     finally:
         os.chdir(worker_dir)
         shutil.rmtree(tmp_dir)
@@ -581,7 +595,6 @@ def validate_pentanomial(wld, rounds):
 def parse_cutechess_output(
     p, remote, result, spsa, spsa_tuning, games_to_play, batch_size, tc_limit
 ):
-
     saved_stats = copy.deepcopy(result["stats"])
     rounds = {}
 
@@ -618,7 +631,8 @@ def parse_cutechess_output(
             message = r'Cutechess-cli says: "{}"'.format(line)
             result["message"] = message
             send_api_post_request(remote + "/api/stop_run", result)
-            raise Exception(message)
+            message = r'Cutechess-cli says: "{}"'.format(line)
+            raise WorkerException(message)
 
         # Parse line like this:
         # Finished game 1 (stockfish vs base): 0-1 {White disconnects}
@@ -704,13 +718,13 @@ def parse_cutechess_output(
                         if not response["task_alive"]:
                             # This task is no longer necessary
                             print("Server told us task is no longer needed")
-                            return response
+                            return False
                         update_succeeded = True
                         num_games_updated = num_games_finished
                         break
                     time.sleep(HTTP_TIMEOUT)
                 if not update_succeeded:
-                    print("Too many failed update attempts")
+                    raise WorkerException("Too many failed update attempts")
                     break
 
         # act on line like this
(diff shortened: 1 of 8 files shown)


vdbergh:
There is an issue with spsa.

vdbergh:
Should be fixed. Removing WIP label.

vdbergh:
Oops I am seeing on DEV that I should have supplied the username instead of the worker name.

ppigazzini:
DEV updated to https://github.com/glinscott/fishtest/pull/1081/commits/b7abe1e86dd50fc04667145d8530d101d98751ca

ppigazzini:
DEV updated

vdbergh:
Hmm... how are you creating messages like this:
```
failed task: task_id: 214, worker: user00-3cores-e54411c4, reason: '('Failed to setup engine for 
c27b5be9e015bd9f9f76d471a9904e5ad83bae15', WorkerException('Failed to validate the network: 
nn-13406b1dcbe0.nnue'))'
```
That should not be possible....

vdbergh:
> GitHib dropped my whole review because I forgot to add the comment....

:frowning_face:

worker/games.py
```diff
@@ -439,7 +446,7 @@ def setup_engine(
 
         shutil.move("stockfish" + EXE_SUFFIX, destination)
     except:
-        raise Exception("Failed to setup engine for {}".format(sha))
+        raise WorkerException("Failed to setup engine for {}".format(sha))
```
ppigazzini, line 449: This is already bad in master because masks all the inner exceptions, I was working on this problem with a test `raise`:
```python
raise WorkerException("Failed to validate the network: {}".format(net)) # TEST
```
```
Build uses default net:  nn-13406b1dcbe0.nnue

Exception running games:
Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15
Informing the server

```
Solutions:
1. forward the inner exception and print a message:
```python
    except Exception as e:
        print ("Failed to setup engine for {}".format(sha))
        raise
```

```
Build uses default net:  nn-13406b1dcbe0.nnue
Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15

Exception running games:
Failed to validate the network: nn-13406b1dcbe0.nnue
Informing the server
Task exited
```

2. print the inner exception and raise the outer exception:
```python
    except Exception as e:
        print ("Exception during the engine setup:\n", e, sep="", file=sys.stderr)
        raise WorkerException("Failed to setup engine for {}".format(sha))
```
```
Build uses default net:  nn-13406b1dcbe0.nnue
Exception during engine setup:
Failed to validate the network: nn-13406b1dcbe0.nnue

Exception running games:
Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15
Informing the server
Task exited
```

3. raise the tuple with outer and inner exceptions:
```python
    except Exception as e:
        raise WorkerException("Failed to setup engine for {}".format(sha), e)
```
```
Build uses default net:  nn-13406b1dcbe0.nnue

Exception running games:
('Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15', WorkerException('Failed to validate the network: nn-13406b1dcbe0.nnue'))
Informing the server
Task exited
```
This is the "Actions" entry for the last point
<html>
<body>


21-10-16 14:40:13 | user00 | sbvals1 | failed task: task_id: 215, worker: user00-3cores-e54411c4, reason: '('Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15', WorkerException('Terminated by signal'))'
-- | -- | -- | --
21-10-16 14:39:57 | user00 | sbvals1 | failed task: task_id: 214, worker: user00-3cores-e54411c4, reason: '('Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15', WorkerException('Failed to validate the network: nn-13406b1dcbe0.nnue'))'


</body>
</html>

worker/worker.py
```diff
@@ -372,26 +373,40 @@ def fetch_and_handle_task(worker_info, password, remote, current_state):
...
-        payload = {
-            "username": worker_info["username"],
-            "password": password,
-            "run_id": str(run["_id"]),
-            "task_id": task_id,
-            "unique_key": worker_info["unique_key"],
-        }
+        print("\nException running games:\n", message, sep="", file=sys.stderr)
+        print("Informing the server")
+        payload["message"]=server_message
```
ppigazzini, line 409: Before the final squash format with:
```
black --exclude=worker/packages .
isort --profile black --skip worker/packages .
```

worker/games.py
```diff
@@ -116,10 +118,10 @@ def verify_required_cutechess(cutechess):
     p.stdout.close()
 
     if p.returncode != 0:
-        raise Exception("Failed to find cutechess version info")
+        raise WorkerException("Failed to find cutechess version info")
```
ppigazzini, line 121: What about `FatalException` for `cutechess-cli` not meeting the version requirement?
In this case should be modified the comment.
```python
    # raise "FatalException". Currently this is only
    # done when the worker is too slow.
```

ppigazzini:
> > GitHib dropped my whole review because I forgot to add the comment....
> 
> ☹️

I clicked on the + mark of the lines to recover my review comments.

worker/games.py
vdbergh: My preference would be for the first solution. I am a bit worried about cluttering the action page too much....

worker/games.py
vdbergh: Yes that is a good idea. I have not looked at the logic of how the cutechess binary is downloaded/installed.

worker/games.py
ppigazzini: Also this works and perhaps is the best solution (the string can be formatted in any way):
```python
    except Exception as e:
        raise WorkerException("Failed to setup engine for {}:\n{}".format(sha, e))
```
```
Build uses default net:  nn-13406b1dcbe0.nnue

Exception running games:
Failed to setup engine for c27b5be9e015bd9f9f76d471a9904e5ad83bae15:
Failed to validate the network: nn-13406b1dcbe0.nnue
Informing the server
```

ppigazzini: > My preference would be for the first solution. I am a bit worried about cluttering the action page too much....

OK, lets go!

vdbergh: But then I am again worried about leaking information if `e` is e.g. a syntax error.

vdbergh: I think the inner exceptions are the most important ones since they indicate where the error is. So I agree that those should be preserved. The outer exceptions have a more cosmetic role.

ppigazzini: With 1. an error from other code within the `try:` is reported at the `setup_engine()` call `failed task: task_id: 262, worker: user00-3cores-92f1b2f4, reason: 'Exception at games.py: 922'`

vdbergh:
I think I am just going to remove the outer except clause.

vdbergh:
Keeping the finally of course.

vdbergh:
Putting in some more inner exceptions.

vdbergh:
We could move the cutechess check outside the main loop. Not now....

vdbergh:
I used the commands 
```
black --exclude=worker/packages .
isort --profile black --skip worker/packages .
```
but I think we have slightly different version of these utilities since there were quite a few changes in things I did not touch.

ppigazzini:
DEV and workers updated.

vdbergh:
(env) pipx upgrade-all

Versions did not change after running 'pip upgrade' for each package 😴

vdbergh:
```
(env) $ black --version
black, version 19.3b0

(env) $ isort --version

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.9.3
```

ppigazzini:
DEV and workers updated.

I have a newer version for `black` for `pipx` or `venv` install.

```
$ black --version
black, version 21.9b0

$ isort --version

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.9.3
```

vdbergh:
```
$ command -v black
/usr/bin/black
$ which black
/home/fishtest/fishtest/server/env/bin/black
```
So it seems `which` is broken in this virtual environment.

vdbergh:
It seems that black was incorrectly hashed in by bash shell. Executing `command -V` seems to have fixed it.

ppigazzini:
@vdbergh unfortunately isort has not skipped the "packages" folder. Please restore that folder from master.

vdbergh:
> @vdbergh unfortunately isort has not skipped the "packages" folder. Please restore that folder from master.

I did...

ppigazzini:
LOL, I though that was intended!
I reloaded the DB and reading the DEV log.

vdbergh:
> LOL, I though that was intended! I reloaded the DB and I'm reading the DEV log to test.
> 
> BTW the "packages" directory could be fixed copying the folder from master (another git, downloading the zip) and adding the commit.

That's what I thought. I did this several times and it did not work (I still don't know what's wrong). So I create a new branch, copying the changed files. But I forgot about the server changes.

ppigazzini:
DEV updated, updating workers.

Perhaps a two steps change to avoid conflict: delete the folder `git rm -r worker/packages` and commit, copy `cp -r <somewhere>/packages worker` and add the folder `git add .` and commit.

Trying to find the info I started the pyramid debug toolbar, never used before, can be useful to easily inspect some info from db.

vdbergh:
> 
> Trying to find the info I started the pyramid debug toolbar, never used before, can be useful to easily inspect some info from db.

:+1:

vdbergh:
About the db: e.g. this code
```
def stop_run(self, username, run):
	self._new_action(username, "stop_run", run)
```
invokes
```
        self.actions.insert_one(
            {
                "username": username,
                "action": action,
		"data": data,
                "time": datetime.utcnow(),
            }
	)
````
I hope this does not mean that the full run object is stored in the db (with its attached task list). I hope that mongodb is smart enough to replace the run object by a link.

**Comment by tomtor** [comment]
> hope this does not mean that the full run object is stored in the db (with its attached task list).

I remember that I commited a patch in the past because exactly that was happening. I will try to find the PR.

**Comment by ppigazzini** [comment]
I'm exporting actions collection in json to inspect.

**Comment by tomtor** [comment]
commit 851180c84e65387f13b1e5c7c00c29f2487cbd32
Author: Tom Vijlbrief <tvijlbrief@gmail.com>
Date:   Mon Jun 22 17:46:58 2020 +0200

    Remove tasks from run data in actions
    
    to save a LOT of space

**Comment by tomtor** [comment]
@vdbergh Note that we used very long pre-allocated task lists before you introduced the dynamic task lists.

The api call to `stop_run` will now have a much shorter task list, but adding a `del_tasks()` is probably best.

**Comment by ppigazzini** [comment]
@tomtor writing #902 I had the impression that some util scripts can be safely deleted, please have a look at your convenient time.

**Comment by ppigazzini** [comment]
I fear that `compact_actions.py` is not up to date, I got this running it on DEV:
```
$ ${VENV}/bin/python3 compact_actions.py
  File "compact_actions.py", line 40, in <module>
    main()
  File "compact_actions.py", line 36, in main
    compact_actions()
  File "compact_actions.py", line 23, in compact_actions
    del a["data"]["before"]["tasks"]
KeyError: 'tasks'
```

BTW I will be away from the keyboard for a couple of hours.

vdbergh:
@ppigazzini I think there is a bug with the added del_tasks in the stop_run api in api.py. This bug hasn’t been triggered because of the CPU hours restriction. 

The problem is that rundown.stop_run is called after del_tasks. I think this is wrong. I would quickly delete the del_task line now, do that we can think about it.

vdbergh:
This is the culprit:
‘’’
            run = del_tasks(run)   <—-
            self.request.actiondb.stop_run(username, run)
            self.request.rundb.stop_run(self.run_id())
‘’’

vdbergh:
It seems that there are many exceptions…

ppigazzini:
Dantist has a not working "cutechess-cli", exception at line 99 of games.py

ppigazzini:
What about a FatalException for this call?

https://github.com/glinscott/fishtest/blob/c329f922491799e49cfcca69ac20704352fdb9dd/worker/games.py#L99-L105

ppigazzini:
Ok, I'll do now.

vdbergh:
Dantist seems to have several workers running. All 7 cores.

ppigazzini:
Opened #1084 , I'm starting some tests.

ppigazzini:
Dantist run Darwin (MacOS) workers?

ppigazzini:
BTW #1084 has stopped the flooding.

PR #1085, opened 2021-10-18
Drop Python2 support from worker

ppigazzini:
Also format the project files with:
```
black --exclude=worker/packages .
isort --profile black --skip worker/packages .
```


ppigazzini:
Tested working with python 3.4 - 3.10 on linux, WSL, msys2.

ppigazzini:
Thank you for the review @tomtor

PR #1087, opened 2021-10-18
Increase the maximum retry time from 4 minutes to 4 hours.

vdbergh:
To reduce possible flooding by fleets...

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -39,6 +39,7 @@ from updater import update
 
 WORKER_VERSION = 115
 HTTP_TIMEOUT = 15.0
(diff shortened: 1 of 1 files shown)


ppigazzini:
The fllooding "cw" windows worker restarted somehow after exiting with FatalException for low Nps. I limited its messages lowering the required Nps from 0.45 MNps to 0.35 MNps, but sometimes also this seems to be too much high.

vdbergh:
Unfortunately cw is running the worker in a loop...

vdbergh:
Running the worker in a loop should only be done by experienced users who understand the implications.

PR #1089, opened 2021-10-18
Deal with "Unable to move the stockfish binary to its destination".

vdbergh:


ppigazzini:
To be tested but at first sight lgtm.

vdbergh:
That's a very good idea (I mean a file lock). I any case I will close this PR.

PR #1092, opened 2021-10-18
Deal with an item download failure.

vdbergh:
Not tested since I do not have an unrestricted network right now.


vdbergh:
I agree. There is already `send_api_post_request()`. We could also have `send_api_get_request()`.

vdbergh:
I mean `requests.get()` calls in `games.py`.

PR #1093, opened 2021-10-18
Stop worker if another is running in the folder

ppigazzini:
There is still a little probability of race condition.

Build the source code path in a robust way.


vdbergh:
I agree with this change. In the future we can try the solution with the file lock.

PR #1096, opened 2021-10-18
Raise worker version to 116 (also server side)

ppigazzini:


vondele:
I think the new actions log will be very useful to make the worker code more robust... many thanks all!

PR #1099, opened 2021-10-19
For a failed get_request use the exception in the error message.

vdbergh:
For a failed get_request use the exception in the error message.
    
I don't want to do this for post requests since I am afraid it may expose passwords. Even if it does not now it may do so in future versions of the requests module.
    
Raise an exception on http status codes >= 400.
    
This is apparently not done by default by the requests module. See
    
    https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module
    
for background.
    
See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes for http status codes.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -51,18 +51,17 @@ EXE_SUFFIX = ".exe" if IS_WINDOWS else ""
 MAKE_CMD = "make COMP=mingw " if IS_WINDOWS else "make COMP=gcc "
 
 
+# See https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module
+# for background.
+# It may be useful to introduce more refined http exception handling in the future.
+
 def requests_get(remote, *args, **kw):
     # A lightweight wrapper around requests.get()
     try:
         result = requests.get(remote, *args, **kw)
+        result.raise_for_status()  # also catch return codes >= 400
     except Exception as e:
-        print(
-            "Exception in requests.get():\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
-        raise WorkerException("Get request to {} failed".format(remote))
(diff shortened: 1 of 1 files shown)


vdbergh:
I think this will be useful to understand the many get request failures that occur, e.g. currently by noob.

ppigazzini:
Github has changed the API rate policy (and deprecated the `rate` object), see:
https://docs.github.com/en/rest/reference/rate-limit
https://docs.github.com/en/rest/reference/search

```
{
  "resources": {
    "core": {
      "limit": 5000,
      "remaining": 4999,
      "reset": 1372700873,
      "used": 1
    },
    "search": {
      "limit": 30,
      "remaining": 18,
      "reset": 1372697452,
      "used": 12
    },
    "graphql": {
      "limit": 5000,
      "remaining": 4993,
      "reset": 1372700389,
      "used": 7
    },
    "integration_manifest": {
      "limit": 5000,
      "remaining": 4999,
      "reset": 1551806725,
      "used": 1
    },
    "code_scanning_upload": {
      "limit": 500,
      "remaining": 499,
      "reset": 1551806725,
      "used": 1
    }
  },
  "rate": {
    "limit": 5000,
    "remaining": 4999,
    "reset": 1372700873,
    "used": 1
  }
}
```

ppigazzini:
Tested the new API, the worker is consuming only the "core" object.

vdbergh:
The two requests small requests #1100, #1106 are independent.

Otherwise the first request that should be merged is this one #1099.

It currently conflicts slightly with #1104. I will rebase #1104 after #1099 is merged.

vdbergh:
Looking at the current action log there does not seem to be anything funky going on with the failed get requests. Just regular network issues. If this persists for a few days then we could go back to displaying a more generic error message in the action log.

PR #1108, opened 2021-10-21
Use the subprocess context manager

ppigazzini:
The context manager in the `.__exit__()` automatically:
 - closes the standard file descriptors
 - waits the process

Use `OSerror` and `SubprocessError` to catch an exception
from the context `.__enter__()`:
```
try:
  with Popen() as p:
     <body>
except (OSError, SubprocessError) as e:
   <handle exception in Popen()>
```

Use `ExitStack()` to deal with conditional context managers.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -14,6 +14,7 @@ import tempfile
 import threading
 import time
 from base64 import b64decode
+from contextlib import ExitStack
 from queue import Empty, Queue
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


vdbergh:
It seems a possibility to get functional equivalence would be to write
```
try:
  with Popen() as p:
     <body>
except SubprocessError as e:
   <handle exception>
```
See 
https://docs.python.org/3/library/subprocess.html

PR #1132, opened 2021-10-30
Try to send errors after failure of the main make command.

vdbergh:
Failure of the main make command happens surprisingly often. Perhaps it will now be clearer why.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -470,15 +470,18 @@ def setup_engine(
 
         cmd = MAKE_CMD + ARCH + " -j {}".format(concurrency) + " profile-build"
         env = dict(os.environ, CXXFLAGS="-DNNUE_EMBEDDING_OFF")
-        try:
-            subprocess.check_call(
-                cmd,
-                shell=True,
-                env=env,
-            )
-        except Exception as e:
-            print("Exception during main make command:\n", e, sep="", file=sys.stderr)
-            raise WorkerException("Executing {} failed".format(cmd), e=e)
+        with subprocess.Popen(
+            cmd,
+            shell=True,
+            env=env,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            bufsize=1,
+            close_fds=not IS_WINDOWS,
+        ) as p:
+            errors = p.stderr.readlines()
+        if p.returncode:
+            raise WorkerException("Executing {} failed. Error: {}".format(cmd, errors))
 
         # Try needed for backwards compatibility with older stockfish,
         # where 'make strip' fails under mingw. TODO: check if still needed.


ppigazzini:
Tested fine with all workers, triggered the workers update, thank you @vdbergh :)

vdbergh:
So we had a first trigger of this code on one of noobs machines working on `https://tests.stockfishchess.org/tests/view/617e4ae52094e7ab11a7b533` :
```
failed task: task_id: 49, worker: ChessDBCN-16cores-de17dd62, reason: 
'Executing make COMP=gcc ARCH=x86-64-avx2 -j 16 profile-build failed. 
Error: ['material.cpp:229:1: internal compiler error: 段错误\n', ' 229 | } // namespace Stockfish\n', \
' | ^\n', 'Please submit a full bug report,\n', 'with preprocessed source if appro'
```
The Chinese characters mean "Segmentation fault". So the compiler segfaults... But all noobs machines seem identical and many of his machines have been working on this run, without issues.  So this seems to point to a HW issue of this particular machine...

ppigazzini:
ping @noobpwnftw :)

vondele:
well that ought to be a compiler bug (internal compiler errors should always be that, except for weird hardware errors). If we know the compiler version used, it should be reproducible in principle. Now the strange thing, unlikely that material.cpp was touched in that patch.

vdbergh:
If it is a compiler bug then it must be a buffer overflow in the compiler, or something like that. Otherwise it would be deterministic. But yes. The compiler version should be part of the machine architecture.

vdbergh:
Actually profiling probably makes it intrinsically non-deterministic.

vondele:
no, or depth based bench for PGO makes the process deterministic.

vdbergh:
> depth based bench for PGO makes the process deterministic.

This seems to be true (I did three profile builds and they yielded identical executables). I seem to recall this was not always the case.

vondele:
yes, one of my first patches for the project a couple of years back :-)

ppigazzini:
Once upon a time, when I didn't know how compute a standard error :D 
https://groups.google.com/g/fishcooking/c/4_ausUwMXP0/m/0vEHOrhmAAAJ

vdbergh:
Another one:
```
failed task: task_id: 180, worker: tolkki963-1cores-eec8dbcb, reason: 'Executing make COMP=gcc ARCH=x86-64-bmi2 -j 1 profile-build failed. Error: ['\n', 'Position: 1/47 (rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1)\n', '\n', 'Position: 2/47 (r3k2r/p1ppqpb1/bn2pnp1/3PN3/1p2P3/2N2Q1p/PPPBBPPP/R3K2R w KQkq - 0 10)\n'
```
gcc 7.5.0. 
It seems that SF writes the bench positions to stderr. So the actual error was not recorded... 

I wonder if in the makefile we could suppress the SF output. It doesn't seem to be needed there.

vdbergh:
Patching the makefile like this seems to fix it.
```
diff --git a/src/Makefile b/src/Makefile
index fea597e76..9308d0f39 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -751,7 +751,7 @@ profile-build: net config-sanity objclean profileclean
        $(MAKE) ARCH=$(ARCH) COMP=$(COMP) $(profile_make)
        @echo ""
        @echo "Step 2/4. Running benchmark for pgo-build ..."
-       $(PGOBENCH) > /dev/null
+       $(PGOBENCH) > /dev/null 2>&1
        @echo ""
        @echo "Step 3/4. Building optimized executable ..."
        $(MAKE) ARCH=$(ARCH) COMP=$(COMP) objclean
```

vondele:
that would be an OK change for me.

PR #1133, opened 2021-10-30
Show worker RAM in home and test pages

### LINKED ISSUE #1131: show memory limit for machines.
Maybe the machine overview could show the memory limit ? 

Right now we know noob's workers can't run LTC tests because of that, but it would be nice to see the actual amount of mem available.

ppigazzini:
closes #1131

### DIFF
--- a/server/fishtest/templates/machines_table.mak
+++ b/server/fishtest/templates/machines_table.mak
@@ -3,10 +3,11 @@
     <tr>
       <th>Machine</th>
       <th>Cores</th>
+      <th>RAM</th>
       <th>UUID</th>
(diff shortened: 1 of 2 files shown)


ppigazzini:
@vondele DEV updated, let me know your preferred message format.

vondele:
to me that looks good. 

The alternative is to show it in the test machine list where the architecture flags are shown. I don't have a strong preference.

ppigazzini:
Added in both pages, PROD updated.

vondele:
Thanks! You're on a super active sprint :-)

PR #1135, opened 2021-10-30
New bounds for SPRT tests

vondele:
slightly narrower interval to avoid flukes. Slightly more difficult to pass Elo gains STC, somewhat more easy to pass Elo gains LTC.


vondele:
This is based on the current observations (large number of LTC, and possibly a bit easy to regress).

ppigazzini (approved):

ppigazzini:
Looks good on DEV, PROD updated, thank you @vondele :)

PR #1136, opened 2021-10-30
Fix missing icon

vondele:
should fix a missing icon on prod right now for the SPRT calculator


vondele:
closing in favor of #1138

dav1312:
Oh sorry I didn't see that there was already a PR about it 😥

vondele:
no problem, literally made PRs at the same time... and yours is the better solution.

PR #1137, opened 2021-10-30
Live elo links and indentation

dav1312:


dav1312:
This change didn't work as expected 😕

PR #1138, opened 2021-10-30
SPRT Calc icon fix

dav1312:


ppigazzini:

PR #1140, opened 2021-10-31
Send the gcc version to the server as part of the worker_info.

vdbergh:
On the server it can then be included in the info string of the workers.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -255,28 +255,23 @@ def gcc_version():
                 major = line.split()[2]
             if "__GNUC_MINOR__" in line:
                 minor = line.split()[2]
+            if "__GNUC_PATCHLEVEL__" in line:
+                patchlevel = line.split()[2]
 
     if p.returncode != 0:
         print("g++ version query failed with return code {}".format(p.returncode))
-        return False
+        return None
 
     try:
         major = int(major)
         minor = int(minor)
+        patchlevel = int(patchlevel)
     except:
         print("Failed to parse g++ version.")
-        return False
+        return None
 
-    if (major, minor) < (7, 3):
-        print(
-            "Found g++ version {}.{}: please update to g++ version 7.3 or later.".format(
-                major, minor
-            )
-        )
-        return False
-
-    print("Found g++ version {}.{}".format(major, minor))
-    return True
+    print("Found g++ version {}.{}.{}".format(major, minor, patchlevel))
+    return (major, minor, patchlevel)
 
 
 def get_exception(files):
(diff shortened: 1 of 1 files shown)


vdbergh:
https://gcc.gnu.org/releases.html

It seems that there have been releases x.y.z (but not recently). If that happens again the code has to be adapted a bit.

vondele:
I think the .z release are not happening any longer, but distros might have patched their gcc in a way that is not obvious from the version number. like `gcc -version` might display `gcc (SUSE Linux) 7.5.0`  which is different from the gcc 7.5.0 one might find on ubuntu. Anyway, it is already a step forward I think.

vdbergh:
If a worker systematically has problems with a certain `g++` version and it is the same version as used by other workers then we can try to refine the version string.

Once the server side is in place (I am hoping @ppigazzini will do this since it is similar as the recent addition of memory) it is easy to make the worker side more or less specific.

PS. It seems the worker uses `g++ -E -dM - < /dev/null` to find the `g++` version.

ppigazzini:
I started working for show the info on the web. My workers have also the patch info>
```
gcc version 11.1.0 (Ubuntu 11.1.0-1ubuntu1~20.04)
gcc version 11.2.0 (Rev1, Built by MSYS2 project)
```

vdbergh:
I will add the patch level, although it is alway zero I think.

ppigazzini:
DEV updated to show the gcc info on the home page and on the test page.
- home:
image
- test:
image

vdbergh:
:+1:

vondele:
LGTM

ppigazzini:
OK, I will trigger the worker update, when all workers will run the new version I will updated the server to show the gcc version (otherwise the home crashes).

ppigazzini:
We have some patch level > 0  in recent gcc:

Machine | Cores | UUID | MNps | RAM | System | GCC | Python | Worker | Running on | Last updated
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
kdave | 2 | ed31513a | 1.51 | 7958 | Linux 5.13.12-6.gd66b4c0-default | 11.1.1 | 3.8.2 | 127 | lowDeltaRed3 | seconds ago
okrout | 30 | 08acf884 | 1.31 | 7992 | Linux 5.14.12-200.fc34.x86_64 | 11.2.1 | 3.9.7 | 127 | evalchoice1 | seconds ago
okrout | 30 | eba7ad7d | 1.46 | 7992 | Linux 5.14.12-200.fc34.x86_64 | 11.2.1 | 3.9.7 | 127 | Test174 | seconds ago
okrout | 30 | 040e0196 | 1.24 | 7992 | Linux 5.14.12-200.fc34.x86_64 | 11.2.1 | 3.9.7 | 127 | lowDeltaRed2 | seconds ago

ppigazzini:
noob's workers-16cores using gcc 9.3.1

vdbergh:
> noob's workers-16cores using gcc 9.3.1

Good that we exposed the patch level then... This seems to be a non-official version. Maybe from CentOS

https://centos.pkgs.org/7/centos-sclo-rh-testing-x86_64/devtoolset-9-gcc-9.3.1-2.el7.x86_64.rpm.html ?

vdbergh:
Ok the non-zero patch levels seems to be development releases. E.g.
 
https://gcc.gnu.org/pipermail/gcc/2021-April/235923.html

PR #1142, opened 2021-11-01
Sort the table rows for dotted version header

ppigazzini:
The original table sorting code (first reference) sorts for:
 - numerical order: for strings that are a number
 - lexygraphical order: for other strings

Using the second reference add also:
 - numerical order: for strings that are a dotted version (eg 3.10.12)

Also add a comment with the references.

https://stackoverflow.com/questions/14267781/sorting-html-table-with-javascript
https://stackoverflow.com/questions/40201533/sort-version-dotted-number-strings-in-javascript

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -1,7 +1,14 @@
+// https://stackoverflow.com/questions/14267781/sorting-html-table-with-javascript
+// https://stackoverflow.com/questions/40201533/sort-version-dotted-number-strings-in-javascript
 const getCellValue = (tr, idx) => tr.children[idx].dataset.diff || tr.children[idx].innerText || tr.children[idx].textContent;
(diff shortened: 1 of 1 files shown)


vdbergh:
So is it correct that the custom sorting mechanism based on the dataset attribute is nowhere used anymore? Some time ago this mechanism was used to sort the tasks when the header of the `Last Updated` column is clicked but I guess it was obsoleted by the fact that clock times are now written with zeros (like `08:08:08`).

ppigazzini:
If you are talking about this code `const getCellValue = (tr, idx) => tr.children[idx].dataset.diff || ...` it's used to proper sort the "Last active" column in "Contributors" and "Top month" pages.

Otherwise let me know the code change to try on the fly on DEV.
My first attempt to write js, though.

vdbergh:
Ok. So it is still being used. Thanks for clarifying.

PR #1145, opened 2021-11-01
Do the check for a working cutechess_cli before building SF.

vdbergh:
This will catch unsupported architectures.


ppigazzini (approved):

ppigazzini:
Tested good on all workers, triggered workers update, thank you @vdbergh :)

PR #1147, opened 2021-11-01
Increase the maximum length of an action message to 1024.

vdbergh:
256 appears to be insufficient, in some cases, to capture  a representative part of the error messages associated with a failed build.


ppigazzini (approved):

ppigazzini:
LGTM, PROD updated.

PR #1148, opened 2021-11-02
Change "Actions" to "Events".

vdbergh:
With the logging of api calls from the worker this seems more accurate.


ppigazzini (approved):

ppigazzini:
LGTM, PROD updated, thank you @vdbergh :)

PR #1149, opened 2021-11-02
Catch duplicate workers early on.

vdbergh:
The start of the worker is racy. So after a small delay we verify that our lock file has not been overwritten by another worker. This will immediately detect duplicate workers, except on extremely slow systems.

This can be tested with the script
```
python worker.py &
python worker.py
```
where the line `python worker.py &` can be repeated as many times as desired.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -381,8 +381,8 @@ def locked_by_others(check_stale=False):
             pid = int(f.read())
             if pid != os.getpid() and (not check_stale or pid_valid(pid, "worker.py")):
                 print(
-                    "\n*** Current worker (PID={}) stopped! ***\n"
-                    "A worker (PID={}) is already running in this directory, "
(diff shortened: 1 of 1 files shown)


ppigazzini:
I thought that this code was meant to deal (one run late) with the racy workers start. If I'm not wrong after this PR it should be never triggered, a sort of assert, so perhaps better to add a comment.

https://github.com/glinscott/fishtest/blob/33e6c3f54c9c133844d8e470063031d5ce5f0369/worker/worker.py#L392-L400

vdbergh:
In theory it can still be triggered if for one of the workers the time between first checking the lock file and writing the lock file takes more than 0.5 seconds.

ppigazzini:
LGTM, tonight I will test it with some workers.

ppigazzini:
Triggered workers update, thank you @vdbergh :)

PR #1150, opened 2021-11-03
Include timestamp in event log query URLs.

vdbergh:
The aim of this PR is to make it possible to refer to events recorded in the event log with an URL. This is useful for discussing such events. It is done by including a time stamp and a count in the query URLs. More specifically:

(1) The URL returned by the <select> button always yields the same result.
(2) The timestamps in the event log have been turned into links which can be used to link to a particular event.


ppigazzini:
DEV updated, LGTM

server/fishtest/views.py
```diff
@@ -246,9 +246,23 @@ def nns(request):
 def actions(request):
     search_action = request.params.get("action", "")
     search_user = request.params.get("user", "")
+    search_before = request.params.get("before", None)
+    if search_before is not None:
```
ppigazzini, line 250: The `try ... except` works fine for `search_before is None`, so if I'm not wrong we could avoid the outer `if .. else`

vdbergh: I removed the outer try except. Not tested since my local fishtest is a bit broken now.

vdbergh: An example of an URL

https://dfts-0.pigazzini.it/actions?count=1&before=1635700115.683

vdbergh:
Een example of an URL

https://dfts-0.pigazzini.it/actions?count=1&before=1635667832.801

server/fishtest/views.py
ppigazzini:
```
>>> float(None)
TypeError: float() argument must be a string or a real number, not 'NoneType'
```

ppigazzini:
DEV updated.

ppigazzini:
The format of the default page is different from the other (in master too).

vdbergh:
This may be hard to avoid. Html looks at all the rows before formatting a table. That’s why a subtable may look different from the full table. I think one may try to tweak it by giving each column a fixed fraction of the total width but this probably requires quite a bit of fiddling to get it right. I would prefer to keep it out of this PR (as you say it is in master too).

vdbergh:
But I guess it would be good to make the hyphen in the date non breaking.

vdbergh:
I replaced the hyphens in the date by non-breaking hyphens.

ppigazzini:
DEV updated. Let me know if you want to keep 2 separated commits.

vdbergh:
I squashed the two commits into one!

ppigazzini (approved):

ppigazzini:

PR #1152, opened 2021-11-05
Send errors when the cutechess version info is not found.

vdbergh:
Maybe we can understand the error.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -8,6 +8,7 @@ import os
 import platform
 import re
 import shutil
+import signal
 import stat
 import subprocess
 import sys
@@ -61,6 +62,20 @@ HTTP_TIMEOUT = 15.0
 REPO_URL = "https://github.com/official-stockfish/books"
 EXE_SUFFIX = ".exe" if IS_WINDOWS else ""
 MAKE_CMD = "make COMP=mingw " if IS_WINDOWS else "make COMP=gcc "
+
+
+def str_signal(signal_):
+    try:
+        return str(signal.Signals(signal_)).split(".")[-1]
+    except ValueError:
+        return "SIG<{}>".format(signal_)
+
+
+def format_return_code(r):
+    if r < 0:
+        return str_signal(-r)
+    else:
+        return str(r)
 
 
 # See https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module
@@ -137,17 +152,23 @@ def required_net(engine):
     return net
 
 
-def verify_required_cutechess(cutechess):
-    print("Obtaining version info for {} ...".format(os.path.basename(cutechess)))
+def verify_required_cutechess(testing_dir, cutechess):
+    print(
+        "Obtaining version info for {} ...".format(os.path.join(testing_dir, cutechess))
+    )
+    os.chdir(testing_dir)
     try:
         with subprocess.Popen(
-            [cutechess, "--version"],
+            [os.path.join(".", cutechess), "--version"],
             stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
             universal_newlines=True,
             bufsize=1,
             close_fds=not IS_WINDOWS,
         ) as p:
+            errors = p.stderr.read()
             pattern = re.compile("cutechess-cli ([0-9]*).([0-9]*).([0-9]*)")
+            major, minor, patch = 0, 0, 0
             for line in iter(p.stdout.readline, ""):
                 m = pattern.search(line)
                 if m:
@@ -156,11 +177,17 @@ def verify_required_cutechess(cutechess):
                     minor = int(m.group(2))
                     patch = int(m.group(3))
     except (OSError, subprocess.SubprocessError) as e:
-        print("Exception running cutechess-cli:\n", e, sep="", file=sys.stderr)
-        raise FatalException("Not working cutechess-cli - sorry!")
+        raise FatalException("Unable to run cutechess-cl. Error: {}".format(str(e)))
 
     if p.returncode != 0:
-        raise FatalException("Failed to find cutechess version info")
+        raise FatalException(
+            "Unable to run cutechess-cli. Return code: {}. Error: {}".format(
+                format_return_code(p.returncode), errors
+            )
+        )
+
+    if major + minor + patch == 0:
+        raise FatalException("Unable to find the version of cutechess-cli.")
 
     if (major, minor) < (1, 2):
         raise FatalException(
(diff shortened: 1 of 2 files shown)


vdbergh:
It could be that cutechess-cli just segfaults. Then there would probably be no error message. So I am also sending the return code. E.g.

https://dfts-0.pigazzini.it/actions?count=1&before=1636179433.06

vdbergh:
BTW here is a short segfaulting program (I used it for testing).
```
#include <signal.h>
int main(){
  raise(SIGSEGV);
}
```

vdbergh:
In the latest commit I run `cutechess-cli --version` in a shell. Then the negative return codes are translated into actual error messages. Like this:  https://dfts-0.pigazzini.it/actions?count=1&before=1636183219.458

vdbergh:
For a command running in a shell catching exceptions is unnecessary. I also changed the error message: https://dfts-0.pigazzini.it/actions?count=1&before=1636184598.201 ("Not working cutechess - sorry" cannot be triggered anymore).

vdbergh:
The error message https://dfts-0.pigazzini.it/actions?count=1&before=1636184948.626 exposed a path on the worker's system. That is not good. So it has now been changed to https://dfts-0.pigazzini.it/actions?count=1&before=1636268543.56 .

vdbergh:
It turns out that in case of a corrupt executable, `p.stderr.readline()` may fail. So we catch that too: https://dfts-0.pigazzini.it/actions?count=1&before=1636270451.778 .

worker/games.py
```diff
@@ -137,30 +137,37 @@ def required_net(engine):
...
-                    patch = int(m.group(3))
-    except (OSError, subprocess.SubprocessError) as e:
-        print("Exception running cutechess-cli:\n", e, sep="", file=sys.stderr)
-        raise FatalException("Not working cutechess-cli - sorry!")
+def verify_required_cutechess(testing_dir, cutechess):
+    print("Obtaining version info for {} ...".format(os.path.join(testing_dir,cutechess)))
+    os.chdir(testing_dir)
+    with subprocess.Popen(
+        [os.path.join(".",cutechess) + " --version"],
+        shell=True,
```
ppigazzini, line 145: It seems that the best practice is to use `shell=True` only when a shell feature is required
https://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

Without `shell=True` the command must a list of tokens.

ppigazzini:
Error on Windows msys2:
```
Unable to run cutechess-cli. Error: ['\'"./cutechess-cli.exe --version"\' is not recognized as an internal or external command,\n', 'operable program or batch file.\n']
```

vdbergh:
I really like `shell=True` in this case . It means we leave the job of getting a meaningful error message to the shell (e.g. Segmentation Fault). It also means we do not have to catch exceptions in popen which simplifies the code. 

Note we also use `shell=True` for the main make command.

I am surprised by the error message of windows msys. It seems the msys shell expects .\cutechess.exe and for some reason Python gives it ./cutechess.exe.... This seems like a bug in the msys python....

vdbergh:
It seems the signal module can give some kind of string representation of negative return codes (which are signals).

```
>>> import signal
>>> print(signal.Signals(11))                                               
Signals.SIGSEGV
```

vondele:
Would that also be useful to enhance the 'failed task: task_id: 1225, worker: ctoks-3cores-87998a76, reason: 'Terminated by signal'' kind of error messages?

vdbergh:
@vondele. I think `Terminated by signal` always means `SIGINT` (ctrl-C) or perhaps `SIGTERM` since it is generated by the corresponding signal handler. I agree this may be confusing.

ppigazzini:
Reading the official documentation with `shell=True` args should ba a string, not a string in a list. I'm commuting so not able to test if this solves the msys2 problem. I started the worker in mingw64 terminal (bash) that uses the Linux path convention.

vdbergh:
> Reading the official documentation with `shell=True` args should ba a string, not a string in a list. I'm commuting so not able to test if this solves the msys2 problem. I started the worker in mingw64 terminal (bash) that uses the Linux path convention.

You must be correct. Strange that this was not caught on Linux. Anyway I now made a version which does not use `shell=True`.

vdbergh:
Here are some errors
https://dfts-0.pigazzini.it/actions?count=1&before=1636483831.198
https://dfts-0.pigazzini.it/actions?count=1&before=1636483914.872
https://dfts-0.pigazzini.it/actions?count=1&before=1636484015.863
https://dfts-0.pigazzini.it/actions?count=1&before=1636484223.075

vdbergh:
I am now also catching SIGBREAK on windows. Not sure if it has any effect since I don't have windows....

ppigazzini:
On Windows ctrl-c works like on Linux:
```

Exception running games:
Terminated by signal SIGINT
Informing the server
```

vdbergh:
Can you try ctrl-break?

vdbergh:
Hmm it seems modern laptops do no longer have a dedicated `Break` key: https://en.wikipedia.org/wiki/Break_key .

ppigazzini:
I will try as soon at the keyboard.

ppigazzini:
ctrl-break:
- msys2: works as intended
```

Exception running games:
Terminated by signal SIGINT
Informing the server
```

- wsl2: stops suddenly without message on the server
```
Started game 10 of 174 (Base-c4a1390f4e vs New-c4a1390f4e)

[process exited with code 3221225786]
```
- linux ssh with putty: nothing happens

ppigazzini (approved):

vdbergh:
> ctrl-break:
> 
> * msys2: works as intended
> 
> ```
> 
> Exception running games:
> Terminated by signal SIGINT
> Informing the server
> ```
> 
> * wsl2: stops suddenly without message on the server
> 
> ```
> Started game 10 of 174 (Base-c4a1390f4e vs New-c4a1390f4e)
> 
> [process exited with code 3221225786]
> ```
> 
> * linux ssh with putty: nothing happens

So none of these actually works as intended (printing SIGBREAK) :smile: .  But it seems ctrl-BREAK is EOL anyway.

ppigazzini:
Wait, I copied from the wrong terminal, msys2 works as intended either in console and server side:
```
Obtaining EvalFile of stockfish_c4a1390f4e485b8b6a2cf43504a968a3c8a4317d.exe ...

Exception running games:
Terminated by signal SIGBREAK
Informing the server
Task exited
```


Time | Username | Run/User | Event
-- | -- | -- | --
21‑11‑13 08:18:47 | user00 | master | failed task: task_id: 318, worker: user00-3cores-e2b8d800, reason: 'Terminated by signal SIGBREAK'

ppigazzini:
In WSL2 ctrl-break kills immediately the whole WSL2 instance.

vdbergh:
> In WSL2 ctrl-break kills immediately the whole WSL2 instance.

It seems so: https://issueexplorer.com/issue/microsoft/WSL/7596 . ctrl-break is sent to the WSL2 process. It kind of makes sense since the Linux running inside WSL2 doesn't know about ctrl-break which is a strictly windows thing.

PR #1153, opened 2021-11-07
Checkbox to disable adjudication.

vdbergh:
Lightly tested.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -597,7 +597,10 @@ def validate_form(request):
         if len(v) == 0:
             raise Exception("Missing required option: {}".format(k))
 
+    # Handle boolean options
     data["auto_purge"] = request.POST.get("auto-purge") is not None
+    # checkbox is to _disable_ adjudication
+    data["adjudication"] = request.POST.get("adjudication") is None
 
     # In case of reschedule use old data,
     # otherwise resolve sha and update user's tests_repo
@@ -1013,6 +1016,7 @@ def tests_view(request):
         "itp",
         "username",
         "tests_repo",
+        "adjudication",
         "info",
     ]:
 
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -90,6 +90,7 @@ class RunDb:
         auto_purge=False,
         throughput=100,
         priority=0,
+        adjudication=True,
     ):
         if start_time is None:
             start_time = datetime.utcnow()
@@ -120,6 +121,7 @@ class RunDb:
             "throughput": throughput,
             "itp": 100,  # internal throughput
             "priority": priority,
+            "adjudication": adjudication,
         }
 
         if sprt is not None:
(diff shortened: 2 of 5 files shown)


vdbergh:
Lightly tested.

vondele:
Thanks... I think it would be interesting to see how much longer games take if we get rid of adjudication altogether. I might do a local measurement.

vdbergh:
Long ago somebody tested it. IIRC it wasn't dramatically more resource intensive.

vondele:
so, local testing at 10+0.1s with and without adjudication:
```
average game duration adj.pgn :  27.6601 s Elo:  Elo difference: 26.1 +/- 2.6, LOS: 100.0 %, DrawRatio: 80.3 %
average game duration noadj.pgn :  31.4449 s Elo:  Elo difference: 25.6 +/- 2.8, LOS: 100.0 %, DrawRatio: 80.8 %
```
about 15% longer games without adjucation, but both Elo and drawratio seem very similar. 

So probably no reason to change the default, but having a checkbox is not too bad.

server/fishtest/views.py
```diff
@@ -594,7 +597,7 @@ def strip_message(m):
         data["base_signature"] = data["new_signature"]
 
     for k, v in data.items():
-        if len(v) == 0:
+        if not k in ['adjudication'] and len(v) == 0:
```
ppigazzini, line 600: `if k not in ['adjudication'] and len(v) == 0:`

ppigazzini:
DEV updated, started a test disabling the adjudication.

server/fishtest/views.py
vdbergh: I guess they are equivalent but `not in` looks indeed nicer.

ppigazzini: `not in` is an operator, always wondered why not a single word as `notin` or `nin`

vdbergh:
It seems it's not working. The pgns contain adjudicate games despite Adjudication showing False on the test page https://dfts-0.pigazzini.it/tests/view/618aac4cfc741c9c217d13c9. Are you also running the updated worker? I did test this locally.

vdbergh:
Tried it with my version of the worker. The pgn https://dfts-0.pigazzini.it/api/pgn/618aac4cfc741c9c217d13c9-382.pgn does not contain adjudicated games.

ppigazzini:
Sorry, I missed the worker changes :(
Label added now.

ppigazzini:
DEV and workers updated, started a couple of tests disabling adjudication.

vdbergh:
Looked at some .pdfs. Now it appears to be working!

ppigazzini:
Great! Unfortunately I'm a bit under pression with my job, very little time for testing/reviewing :(

vdbergh:
> Great! Unfortunately I'm a bit under pression with my job, very little time for testing/reviewing :(

I recognize the situation :) However I think this PR isn't very risky. It is just passing a flag from the "new test page" to the worker along the standard path. We only have to make sure the flag arrives at its destination and is not inverted along the way.

ppigazzini (approved):

ppigazzini:
All PRs merged, triggered the workers update, thank you @vdbergh :)

vdbergh:
Thanks!

vondele:
thank you both!

vondele:
btw those changes would be roughly similar to what would be needed to support things like ponder or FRC.

PR #1155, opened 2021-11-12
Make lock file system more robust.

vdbergh:
We may have several processes reading from and writing to the lock file at the same time with potentially unpredictable consequences (especially on windows). We now refactor the file IO operations and put them in a try: except: bracket.

We also change the logic:
    
At the start of the worker we tolerate an invalid or non-existing lock file since we intend to replace it with one of our own. Once  we have started, we now only accept a valid lock file containing our own PID.
    
Also: the name of the lock file is not a global variable anymore.
    
Also: send stderr to devnull on Linux. Like it is done on Windows.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -38,7 +38,6 @@ WORKER_VERSION = 129
 HTTP_TIMEOUT = 15.0
 MAX_RETRY_TIME = 14400.0  # four hours
 IS_WINDOWS = "windows" in platform.system().lower()
-lock_file = path.join(path.dirname(path.realpath(__file__)), "worker.lock")
 
 """
 Bird's eye view of the worker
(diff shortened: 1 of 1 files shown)


worker/worker.py
```diff
@@ -322,20 +321,40 @@ def heartbeat(worker_info, password, remote, current_state):
         print("Heartbeat stopped")
 
 
-def create_lock_file():
+def read_int(file):
+    try:
+        if path.exists(file):
```
ppigazzini, line 326: `path.exists()` could be dropped, there is already at work the exception `FileNotFoundError`.

vdbergh: You are right! I fixed it.

ppigazzini:
Nice clean up, looks good to me!

ppigazzini (approved):

PR #1161, opened 2021-11-22
Improve the upload requests error handling

ppigazzini:


ppigazzini:
@ppigazzini ping test 2

PR #1160, opened 2021-11-22
Add timeouts

vondele:
might be needed if there is a problem uploading to the server.

PR #1162, opened 2021-11-23
Do more validation of api arguments in api.py.

vdbergh:
This allows for simplification of downstream code. Notably `failed_task()`, `update_task()` and `request_spsa()`.

We do the validation when parsing the arguments. For example `self.task_id()` will throw a `HTTPBadRequest()` exception if it does not succeed in returning a `task_id` corresponding to an existing task. So downstream code may assume that a `task_id` argument is valid.

The same mechanism has allowed us to simplify `api_stop_run` which had recently ballooned because of argument validation.

To check that the api arguments are syntactically correct we provide a rudimentary, but generic, "schema" implementation in `utils.py:validate()`.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -4,15 +4,63 @@ from datetime import datetime
 
 import requests
 from fishtest.stats.stat_util import SPRT_elo
-from fishtest.util import worker_name
+from fishtest.util import optional_key, validate, worker_name
 from fishtest.views import del_tasks
-from pyramid.httpexceptions import HTTPFound, HTTPUnauthorized, exception_response
+from pyramid.httpexceptions import (
+    HTTPBadRequest,
+    HTTPFound,
+    HTTPUnauthorized,
+    exception_response,
+)
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 
 WORKER_VERSION = 131
 
 flag_cache = {}
+
+def validate_request(request):
+    schema = {
+        "password" : str,
+        optional_key("username") : str,
+        optional_key("run_id") : str,
+        optional_key("task_id") : int,
+        optional_key("unique_key") : str,
+        optional_key("pgn") : str,
+        optional_key("message") : str,
+        optional_key("ARCH") : str,
+        optional_key("nps") : float,
+        optional_key("worker_info") : {
+            "uname": str,
+            "architecture": [str, str],
+            "concurrency": int,
+            "max_memory": int,
+            "min_threads": int,
+            "username": str,
+            "version": str,
+            "gcc_version": str,
+            "unique_key": str,
+            "rate": {'limit': int, 'remaining': int}
+        },
+        optional_key("spsa") : {
+            "wins" : int,
+            "losses" : int,
+            "draws" : int,
+            "num_games" : int,
+        },
+        optional_key("stats") : {
+            "wins" : int,
+            "losses" : int,
+            "draws" : int,
+            "crashes" : int,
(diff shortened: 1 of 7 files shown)


ppigazzini:
DEV updated.

We should decide how to deal with the print in the server code, discussion already started in #1160 with @vondele 

I collect here some information found about the matter:
- stdout buffering
https://realpython.com/python-print/#buffering-print-calls
https://eklitzke.org/stdout-buffering

The big delay in `journal `entries when running `pserve` with `systemd` should be due to the switch to a fully buffered write.
We should have a line buffered write running `pserve` in TTY during development/test .

- ways to get unbuffered python print
https://stackoverflow.com/questions/230751/how-can-i-flush-the-output-of-the-print-function-unbuffer-python-output

A simple way is to use the switch `-u` during development/debug, eg `python3 -u pserve`
Another simple way is to use a partial function in each server module: `print = partial(print, flush=True)`

- pyramid logging system
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/logging.html
 
Pyramid has already the configuration for the standard `logging` library (that by default should write to the unbuffered `sys.stderr`), this require a refactoring of the server code, though.
https://docs.python.org/3/library/logging.html
https://docs.python.org/3/howto/logging.html

vdbergh:
I have been using `print(..., flush=True)` but I think it would be best to have a generic "logging" function that does the right thing.

ppigazzini:
Of course we (surely I) need to study a bit the `logging` best practices before decide if and how make the switch.

For development/debug at the moment is easier to use the `-u` switch to have all the prints unbuffered than rewrite all the `print()` instances. The `partial` ways is another simple option.

BTW I verified that  with systemd stdout & stderr are fully buffered.
- test program:
```python
# test_print.py
import time
import sys
for i in range(10)
    print(i, file=sys.stderr)
    time.sleep(1)
```
- systemd unit:
```
# /ets/systemd/system/test_print.service
[Unit]
Description=test print() to journal
After=network.target

[Service]
Type=simple
ExecStart=/home/usr00/env/bin/python3 /home/usr00/test_print.py
User=usr00
WorkingDirectory=/home/usr00

[Install]
WantedBy=multi-user.target
```
- commands:
```bash
sudo systemctl daemon-reload
sudo systemctl start test_print
# check in another terminal
sudo journalctl -f -u test_print
```

vdbergh:
CI fails for now because of the stricter input validation. I will fix this tonight.

ppigazzini:
DEV updated, started some workers.

ppigazzini:
DEV updated.

ppigazzini:
DEV updated.

ppigazzini:
DEV updated, started some workers.

vdbergh:
Thanks! Could perhaps also revive an old SPSA tests? SPSA was not tested at all since it is such a hassle to create SPSA tests locally.

ppigazzini:
SPSA scheduled.

vdbergh:
Thanks!

vdbergh:
Luckily SPSA seems to be proceeding normally.

ppigazzini:
All workers updated.

ppigazzini:
All workers updated after https://github.com/glinscott/fishtest/pull/1162/commits/47352d2ef81fa5265964d2d203d4aecd163321a7

ppigazzini (approved):

ppigazzini:

vdbergh:
Thanks. Next step is to include worker_info in all worker apis. Once this done, we can enforce it on the server and further greatly simplify the server side of the api. In the end it will become quite transparent.

PR #1167, opened 2021-11-26
Temporary fix to make the update 131-132 go through.

vdbergh:


ppigazzini (approved):
Tested working fixing PROD by hand, thank you @vdbergh :)

PR #1168, opened 2021-11-28
Prevent scroll wheel from changing bench for firefox users.

TopoIogist:
Firefox apparently has this weird behavior to change number inputs when the scroll wheel is used. This triggered some unintended off-by-one errors when submitting tests.


ppigazzini:
DEV updated. Waiting the wheel test by @vondele :)

vondele:
works.

ppigazzini:
PROD updated, tahnk you @TopoIogist :)

PR #1170, opened 2021-11-29
Many small cosmetic fixes.

vdbergh:
More generic error handling in `api.py`. In particular the error message will now always include `/api/****`.

The worker now uses `send_api_post_request()` for all POST api requests.

A comment in `api.py` about fishtest instances.

Adapt `test_api.py` to dynamic tasks.

`HTTPUnauthorized` exception when `stop_run()` fails.

Make the beat api the same as the other apis.

Create an "info" field in api when the worker tries to do something with an inactive task.

Make the heartbeat thread a daemon thread.

Get rid of the many "Error from remote" messages in the worker. Print the error message in `send_api_post_request()` instead. As a bonus we also print info messages now.
    
Only check username/password for /api/request_version. By being less strict here we can be more strict elsewhere since the worker will upgrade.
    
Make `upload_pgn` a bit more robust.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -15,6 +15,18 @@ from pyramid.httpexceptions import (
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 
+"""
+Important note
+==============
+
+All apis that are relying on get_run() should be served from a single
+Fishtest instance.
+
+If other instances need information about runs they should query the
+db directly. However this information may be slightly outdated, depending
+on how frequently the main instance flushes its run cache.
+"""
+
 WORKER_VERSION = 133
 
 flag_cache = {}
@@ -39,7 +51,7 @@ def validate_request(request):
             "version": str,
             "gcc_version": str,
             "unique_key": str,
-            optional_key("rate"): {"limit": int, "remaining": int},
+            "rate": {"limit": int, "remaining": int},
         },
         optional_key("spsa"): {
             "wins": int,
@@ -56,10 +68,7 @@ def validate_request(request):
             "pentanomial": [int, int, int, int, int],
         },
     }
-    error = validate(schema, request, "request")
-    if error != "":
-        print(error, flush=True)
-        raise HTTPBadRequest({"error": error})
+    return validate(schema, request, "request")
 
 
 def strip_run(run):
@@ -76,17 +85,17 @@ def strip_run(run):
     return run
 
 
-@exception_view_config(HTTPUnauthorized)
-def authentication_failed(error, request):
-    response = Response(json_body=error.detail)
-    response.status_int = 401
-    return response
-
-
 @exception_view_config(HTTPBadRequest)
 def badrequest_failed(error, request):
     response = Response(json_body=error.detail)
     response.status_int = 400
+    return response
+
+
+@exception_view_config(HTTPUnauthorized)
+def authentication_failed(error, request):
+    response = Response(json_body=error.detail)
+    response.status_int = 401
     return response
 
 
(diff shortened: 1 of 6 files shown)


server/fishtest/api.py
```diff
@@ -15,6 +15,18 @@
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 
+"""
+Important note
+==============
+
+All apis that are relying on get_run() should be served from a single
```
ppigazzini, line 22: Thank you, very appreciated :)

server/fishtest/rundb.py
```diff
@@ -617,7 +626,11 @@ def priority(run):  # lower is better
                     connections = connections + 1
 
         if connections >= self.userdb.get_machine_limit(worker_info["username"]):
-            return {"task_waiting": False, "hit_machine_limit": True}
+            error = "Request_task: Too many connections for user {}".format(
```
ppigazzini, line 629: What about this?
"Request_task: already reached the machine limit for user {}"

worker/worker.py
```diff
@@ -727,6 +715,8 @@ def worker():
     heartbeat_thread = threading.Thread(
         target=heartbeat, args=(worker_info, options.password, remote, current_state)
     )
+    heartbeat_thread.setDaemon(True)  # The should not be necessary,
```
ppigazzini, line 718: "# It should ..." ?

ppigazzini:
PROD and workers updated.

vondele:
dev ?

ppigazzini:
DEV of course :)

ppigazzini:
Workers restarted.

vdbergh:
For https://dfts-0.pigazzini.it:443/tests/view/61a550c37347b974ba95a02f I am getting
```
Warning: Invalid value for option C: 623
Warning: Invalid value for option P: 287
Warning: Invalid value for option x0: -312
Warning: Invalid value for option y0: 282
Warning: Invalid value for option P: -396
Warning: Invalid value for option x0: 371
Warning: Invalid value for option y0: -400
```
(also with master).

vdbergh:
Ok the actual uci options are
```
option name x0 type spin default 0 min -128 max 128
option name y0 type spin default 0 min -128 max 128
option name C type spin default 0 min -128 max 128
option name P type spin default 0 min -128 max 128
```
So the intervals `[-128,128]` are smaller than `[-1024,1024]` which is used by the test. Did @snicolet manually change the intervals on the new test page?

vdbergh:
I am now only verifying username/password in `/api/request_version`. I checked that very old workers can now upgrade again.

ppigazzini:
DEV updated, great improvement :)

ppigazzini:
DEV updated.

ppigazzini:
DEV updated.

vdbergh:
Squashed.

snicolet:
> Ok the actual uci options are
> 
> ```
> option name x0 type spin default 0 min -128 max 128
> option name y0 type spin default 0 min -128 max 128
> option name C type spin default 0 min -128 max 128
> option name P type spin default 0 min -128 max 128
> ```
> 
> So the intervals `[-128,128]` are smaller than `[-1024,1024]` which is used by the test. Did @snicolet manually change the intervals on the new test page?

For my SPSA run in fishtest ( https://tests.stockfishchess.org/tests/view/619e2c93c0a4ea18ba95a450 ) I intended to use the interval [-128..128] for the four variables. So I followed the instructions in the wiki:

a) use `SetRange` for each variable, for instance `TUNE(SetRange(-128, 128), C);`
b) launched SF, copied-pasted the lines in "SPSA parameters" box in "New test" page
c) clicked on "autoselect" check box

My understanding (maybe wrong) was that "autoselect" would change the hyper-parameters of SPSA but that it would not change the range?

ppigazzini:
It's me that scheduled on DEV the test with range [-1024, +1024]. I had no time to review the code and scheduled on the fly the test to verify that SPSA with "Autoselect" choses the "c" value according the parameter range.

vdbergh:
> It's me that scheduled on DEV the test with range [-1024, +1024]. I had no time to review the SPSA code and so I scheduled on the fly the test to verify that SPSA with "Autoselect" choses the "c" value according the parameter range.

Ok! After posting I notices that this might be the cause. In any case the worker will now heed the warning.

vdbergh:
> > Ok the actual uci options are
> > ```
> > option name x0 type spin default 0 min -128 max 128
> > option name y0 type spin default 0 min -128 max 128
> > option name C type spin default 0 min -128 max 128
> > option name P type spin default 0 min -128 max 128
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the intervals `[-128,128]` are smaller than `[-1024,1024]` which is used by the test. Did @snicolet manually change the intervals on the new test page?
> 
> For my SPSA run in fishtest ( https://tests.stockfishchess.org/tests/view/619e2c93c0a4ea18ba95a450 ) I intended to use the interval [-128..128] for the four variables. So I followed the instructions in the wiki:
> 
> a) use `SetRange` for each variable, for instance `TUNE(SetRange(-128, 128), C);` b) launched SF, copied-pasted the lines in "SPSA parameters" box in "New test" page c) clicked on "autoselect" check box
> 
> My understanding (maybe wrong) was that "autoselect" would change the hyper-parameters of SPSA but that it would not change the range?

@snicolet It was a misunderstanding. Sorry for the noise.

ppigazzini (approved):

snicolet:
No pb :-)
And thanks to everybody for the astonishing work you do on fishtest code!

ppigazzini:

PR #1172, opened 2021-12-01
Send a SIGINT to cutechess-cli to give it a chance to exit nicely (Linux only).

vdbergh:
If the server finishes a SPRT test then it takes a long time before the worker actually kills cutechess-cli.
We try to fix this by sending a SIGINT first. This works correctly on Linux. cutechess-cli will write the pgn and then quit.

I cannot test if it works correctly on Windows since I don't have windows.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -82,6 +82,14 @@ def format_return_code(r):
         return str(r)
 
 
+def send_sigint(p):
+    if IS_WINDOWS:
+        # patches welcome...
+        pass
+    else:
+        p.send_signal(signal.SIGINT)
+
+
 # See https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module
 # for background.
 # It may be useful to introduce more refined http exception handling in the future.
@@ -897,6 +905,9 @@ def launch_cutechess(
                 tc_limit,
             )
         finally:
+            # We nicely ask cutechess-cli to stop.
+            # Unfortunately this only works under Linux right now.
+            send_sigint(p)
             print("\nWaiting for cutechess-cli to finish ... ", end="", flush=True)
             try:
                 p.wait(timeout=HTTP_TIMEOUT)


worker/games.py
```diff
@@ -81,6 +81,16 @@ def format_return_code(r):
     else:
         return str(r)
 
+def send_sigint(p):
+    try:
+        if p.poll() is None: # slight race condition
```
vondele, line 86: what's a 'slight race condition' ? Might need some clearer comment.

ppigazzini: IMO the condition checks if the process is still running (exit status `None`), then send the signal, but the process could be exited just after the check.
The EAFP way could have the (low probability) risk to send the signal to a new process with the same pid.

vondele: OK, now I understand.

vondele: can't we use https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal

vdbergh: Ah... I looked for send_signal, but in the signal module it is only in 3.10. I overlooked the subprocess module. I will change it.

vdbergh: Ok I updated it. The code is now much simpler and robust obviously.

vdbergh:
Starting cutechess-cli now with 
```
creationflags=subprocess.CREATE_NEW_PROCESS_GROUP if IS_WINDOWS else 0
```
According to https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal this is the necessary for CTRL_C_EVENT to work.

worker/games.py
```diff
@@ -884,6 +891,9 @@ def launch_cutechess(
         stderr=subprocess.STDOUT,
         universal_newlines=True,
         bufsize=1,
+        # The next option is necessary to able be to send CTRL_C_EVENTs.
```
ppigazzini, line 894: typo "to be able"?

vdbergh: Yep.... Luckily these are the easiest bugs to fix.

vdbergh:
It seems on windows we need to create the process with CREATE_NEW_PROCESS_GROUP https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NEW_PROCESS_GROUP

ppigazzini:
Not good news for Windows msys2 I fear.

Test code:
```python
    while datetime.datetime.utcnow() < end_time:
        try:
            line = q.get_nowait().strip()
        except Empty:
            if p.poll() is not None:
                break
            time.sleep(1)
            continue

        print(line, flush=True)
        # TEST
        time.sleep(30)
        break

        # Have we reached the end of the match? Then just exit.
```

- WSL2: cutechess stops on the fly, pgn uploaded, the worker get another task:
```
CPU factor : 1.0299216374466633 - tc adjusted to 1.030+0.010
CPU factor : 1.0299216374466633 - tc adjusted to 1.030+0.010
TC limit 302.1103469843546 End time: 2021-12-01 21:27:29.807288
Indexing opening suite...

Wait 2 seconds before uploading PGN...
Uploading PGN...
Uploading compressed PGN of 36652 bytes
Task exited
```

- msys2: cutechess needs taskkill, however there is the pgn upload and the worker get another task:
```
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
TC limit 284.75357465086466 End time: 2021-12-01 21:25:59.048861
Indexing opening suite...

Waiting for cutechess-cli to finish ... timeout
Killing cutechess-cli.exe with pid 6060 ... killed
Wait 3 seconds before uploading PGN...
Uploading PGN...
Uploading compressed PGN of 41860 bytes
Task exited
```


- msys2: without the test code cutechess needs taskkill after CTRL+C:
```
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
TC limit 284.75357465086466 End time: 2021-12-01 21:25:59.048861
Indexing opening suite...

Waiting for cutechess-cli to finish ... timeout
Killing cutechess-cli.exe with pid 6060 ... killed
Wait 3 seconds before uploading PGN...
Uploading PGN...
Uploading compressed PGN of 41860 bytes
Task exited
```

- msys2: with master cutechess stops on the fly with CTRL-C:
```
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
TC limit 284.75357465086466 End time: 2021-12-01 21:25:59.048861
Indexing opening suite...

Waiting for cutechess-cli to finish ... timeout
Killing cutechess-cli.exe with pid 6060 ... killed
Wait 3 seconds before uploading PGN...
Uploading PGN...
Uploading compressed PGN of 41860 bytes
Task exited
```


- msys2: commenting out  ` #creationflags=subprocess.CREATE_NEW_PROCESS_GROUP if IS_WINDOWS else 0,`

cutechess stops on the fly with CTRL-C, but tith the test code the workers always exit instead of taking a new task.
One time I even got this exception:
```
CPU factor : 1.100723984001665 - tc adjusted to 1.101+0.011
TC limit 322.879035307155 End time: 2021-12-01 21:48:14.498425
Indexing opening suite...

Waiting for cutechess-cli to finish ... Exception in thread Thread-2:
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "C:/tools\msys64\home\Pasquale Pigazzini\_git\__test_folder00\fishtest\worker/games.py", line 619, in enqueue_output
    for line in iter(out.readline, ""):
ValueError: I/O operation on closed file

Exception running games:
Terminated by signal SIGINT
Informing the server
Uploading PGN...
Uploading compressed PGN of 34172 bytes
Task exited
Heartbeat stopped
Deleting lock file C:/tools/msys64/home/Pasquale Pigazzini/_git/__test_folder00/fishtest/worker/worker.lock
```

vdbergh:
Thanks for the elaborate testing!
> 
> Test code:
> 
> ```python
>     while datetime.datetime.utcnow() < end_time:
>         try:
>             line = q.get_nowait().strip()
>         except Empty:
>             if p.poll() is not None:
>                 break
>             time.sleep(1)
>             continue
> 
>         print(line, flush=True)
>         ### TEST CODE
>         time.sleep(30)
>         break
>         ###
> 
>         # Have we reached the end of the match? Then just exit.
> ```
> 
> * WSL2: cutechess stops on the fly, pgn uploaded, the worker get another task:
> 
> ```
> CPU factor : 1.0299216374466633 - tc adjusted to 1.030+0.010
> CPU factor : 1.0299216374466633 - tc adjusted to 1.030+0.010
> TC limit 302.1103469843546 End time: 2021-12-01 21:27:29.807288
> Indexing opening suite...

Ok. This is expected. WSL is just Linux so I assume IS_WINDOWS returns False. So the Windows code isn't touched.

> 
> Wait 2 seconds before uploading PGN...
> Uploading PGN...
> Uploading compressed PGN of 36652 bytes
> Task exited
> ```
> 
The tests below are all compatible with the hypothesis that `p.send_signal(signal.CTRL_C_EVENT)` does not actually send the signal. Or at least that `cutechess-cli` does not receive it.


> * msys2: cutechess needs taskkill, however pgn uploaded and the worker get another task:
> 
> ```
> CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
> CPU factor : 0.9707508226734023 - tc adjusted to 0.971+0.010
> TC limit 284.75357465086466 End time: 2021-12-01 21:25:59.048861
> Indexing opening suite...
> 
> Waiting for cutechess-cli to finish ... timeout
> Killing cutechess-cli.exe with pid 6060 ... killed
> Wait 3 seconds before uploading PGN...
> Uploading PGN...
> Uploading compressed PGN of 41860 bytes
> Task exited
> ```
> 
Apparently `cutechess-cli` did not receive the CTRL_C_EVENT. 

> * msys2: without the test code cutechess needs taskkill after CTRL+C:
> 
> ```
> Started game 32 of 176 (Base-c1f9a359e8 vs New-c1f9a359e8)
> Finished game 30 (Base-c1f9a359e8 vs New-c1f9a359e8): 1/2-1/2 {Draw by 3-fold repetition}
> Score of New-c1f9a359e8 vs Base-c1f9a359e8: 8 - 12 - 10  [0.433] 30
> Started game 33 of 176 (New-c1f9a359e8 vs Base-c1f9a359e8)
> 
> Waiting for cutechess-cli to finish ... Heartbeat stopped
> timeout
> Killing cutechess-cli.exe with pid 20768 ... killed
> 
> Exception running games:
> Terminated by signal SIGINT
> Informing the server
> Uploading PGN...
> Uploading compressed PGN of 53528 bytes
> Task exited
> Waiting for the heartbeat thread to finish...
> Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
> ```
> 

Yes. Since the creationflags put cutechess-cli inside a new process group, it no longer receives the Ctrl-C from the shell. This is expected behaviour. If `p.send_signal(signal.CTRL_C_EVENT)` _had worked_ then this would have stopped `cutechess-cli` instead.

> * msys2: instead with master cutechess stops on the fly with CTRL+C:
> 
> ```
> Score of New-c1f9a359e8 vs Base-c1f9a359e8: 5 - 2 - 8  [0.600] 15
> Started game 18 of 176 (Base-c1f9a359e8 vs New-c1f9a359e8)
> 
> 
> Exception running games:
> Terminated by signal SIGINT
> Informing the server
> Uploading PGN...
> Uploading compressed PGN of 15372 bytes
> Task exited
> Waiting for the heartbeat thread to finish...
> Heartbeat stopped
> Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
> ```
> 

Yes. If `cutechess-cli` is in the main process group then it gets killed by Ctrl-C (all members of the main process group receive that signal, this is similar on Linux).

> * msys2: commenting out  ` #creationflags=subprocess.CREATE_NEW_PROCESS_GROUP if IS_WINDOWS else 0,`
> 
> cutechess stops on the fly with CTRL+C, but with the test code the workers always exits instead of taking a new task. One time I even got this exception:
> 
> ```
> CPU factor : 1.100723984001665 - tc adjusted to 1.101+0.011
> TC limit 322.879035307155 End time: 2021-12-01 21:48:14.498425
> Indexing opening suite...
> 
> Waiting for cutechess-cli to finish ... Exception in thread Thread-2:
> Traceback (most recent call last):
>   File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 973, in _bootstrap_inner
>     self.run()
>   File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 910, in run
>     self._target(*self._args, **self._kwargs)
>   File "C:/tools\msys64\home\usr00\_git\__test_folder00\fishtest\worker/games.py", line 619, in enqueue_output
>     for line in iter(out.readline, ""):
> ValueError: I/O operation on closed file
> 
> Exception running games:
> Terminated by signal SIGINT
> Informing the server
> Uploading PGN...
> Uploading compressed PGN of 34172 bytes
> Task exited
> Waiting for the heartbeat thread to finish...
> Heartbeat stopped
> Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
> ```

Yes this is also expected. If `cutechess-cli` is not in its own process group then `p.send_signal(signal.CTRL_C_EVENT)` kills all processes attached to the console (this has to do with the fact that  `signal.CTRL_C_EVENT` is equal to 0, a special value).

vdbergh:
I did a bit of experimenting with wine. It seems that a Keyboard Event `Ctrl-C` is translated into a signal `SIGINT` on windows. However this seems to be not the case for a `CTRL_C_EVENT`. To receive this the application would have to install a `ConsoleCtrlCHandler`. 

Code:
```
// test.c
#include<stdio.h>
#include<signal.h>
#include<unistd.h>
#include<stdlib.h>

void sig_handler(int signo) {
  if (signo == SIGINT){
    printf("received SIGINT\n");
  }
}

int main(void) {
  if (signal(SIGINT, sig_handler) == SIG_ERR) {
    printf("\ncan't catch SIGINT\n");
    exit(0);
  }
  while(1) {
    sleep(1);
  }
  return 0;
}
```
```
# test.py
from subprocess import Popen, CREATE_NEW_PROCESS_GROUP
import signal, os, sys

print("version=",sys.version)

with Popen(
        ['./test.exe'],
        creationflags=CREATE_NEW_PROCESS_GROUP,
    ) as p:
    print("pid=",p.pid)
    print("Sending CTRL_C_EVENT")
    p.send_signal(signal.CTRL_C_EVENT)

print("After Popen")
print(hex(p.returncode))
```

vdbergh:
So it seems sending SIGINT is beneficial under Linux (and hence under WSL) and harmful on Msys. I propose that we do it under Linux and leave a comment in the code inviting help for Windows.

vdbergh:
I tried the official sample code from https://docs.microsoft.com/en-us/windows/console/registering-a-control-handler-function

This c program just ignores the CTRL_C_EVENT sent by python (under wine) so it behaves similarly to `cutechess-cli`.

I tried ctypes but this does not work either
```
import ctypes
    ctypes.windll.kernel32.GenerateConsoleCtrlEvent(0, p.pid)
```

vdbergh:
So there is a package console_ctrl (installable with pip) that can do it. It is based on the principles exhibited here
https://stackoverflow.com/questions/7085604/sending-c-to-python-subprocess-objects-on-windows

```
// test.c
#include<stdio.h>
#include<signal.h>
#include<unistd.h>
#include<stdlib.h>

void sig_handler(int signo) {
  if (signo == SIGINT){
    printf("received SIGINT\n");
    fflush(stdout);
  }
}

int main(void) {
  if (signal(SIGINT, sig_handler) == SIG_ERR) {
    printf("\ncan't catch SIGINT\n");
    exit(1);
  }
  while(1) {
    sleep(1);
  }
}

```
```
#test.py
import console_ctrl
import subprocess
from subprocess import Popen, CREATE_NEW_PROCESS_GROUP, CREATE_NEW_CONSOLE, PIPE
import signal, os, sys, time

print("version=",sys.version)

si = subprocess.STARTUPINFO(dwFlags=subprocess.STARTF_USESHOWWINDOW, wShowWindow=subprocess.SW_HIDE)

with Popen(
        ['./test.exe'],
        creationflags=CREATE_NEW_CONSOLE,
        startupinfo=si,
        stdout = PIPE,
        stderr = PIPE,
        universal_newlines=True,
    ) as p:
    print("pid=",p.pid)
    console_ctrl.send_ctrl_c(p.pid)
        print(line)
        console_ctrl.send_ctrl_c(p.pid)

print("After Popen")
print("returncode=",hex(p.returncode))

```

vdbergh:
The package is only a few lines
```
# __init__.py
import sys
import subprocess


def send_ctrl_c(pid):
    """Send CTRL-C signal to the given process."""
    subprocess.check_call([sys.executable, '-m', 'console_ctrl', str(pid)])

```
```
# __main__.py
import ctypes
import sys


kernel = ctypes.windll.kernel32
pid = int(sys.argv[1])

kernel.FreeConsole()

# TODO: what if attach failed
kernel.AttachConsole(pid)

# TODO: improve readability, use constants
# Disable Ctrl-C handling for our program
kernel.SetConsoleCtrlHandler(None, 1)
kernel.GenerateConsoleCtrlEvent(0, 0)
```
I don't think we want this kind of hackery in the worker.

vdbergh:
Actually maybe it is not that ugly. It does not change how cutechess-cli would be started.

vdbergh:
I tried to include and use the `console_ctrl` package. Of course it completely untested (I guess I could install msys in wine...).

ppigazzini:
Using CTRL+C in cli
- first try
```
Started game 18 of 176 (Base-c1f9a359e8 vs New-c1f9a359e8)
Finished game 15 (New-c1f9a359e8 vs Base-c1f9a359e8): 1-0 {White wins by adjudication}
Score of New-c1f9a359e8 vs Base-c1f9a359e8: 4 - 7 - 5  [0.406] 16
  Task updated successfully in 0.109828s
Started game 19 of 176 (New-c1f9a359e8 vs Base-c1f9a359e8)
Exception in thread Thread-3:
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "C:/tools\msys64\home\Pasquale Pigazzini\_git\__test_folder00\fishtest\worker/games.py", line 621, in enqueue_output
    for line in iter(out.readline, ""):
ValueError: I/O operation on closed file

Exception running games:
Exception at games.py:89 (Command '['C:/tools/msys64/home/Pasquale Pigazzini/_git/__test_folder00/fishtest/worker/env/bin/python3.exe', '-m', 'console_ctrl', '7980']' returned non-zero exit status 3221225781.)
Informing the server
Exception in requests.post():
Terminated by signal SIGINT
Exception posting failed_task:
Terminated by signal SIGINT
Uploading PGN...
Uploading compressed PGN of 22484 bytes
Task exited
Heartbeat stopped
Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
```

- second try: console stuck for several seconds
```
Score of New-c1f9a359e8 vs Base-c1f9a359e8: 0 - 1 - 1  [0.250] 2
Started game 5 of 176 (New-c1f9a359e8 vs Base-c1f9a359e8)
Finished game 1 (New-c1f9a359e8 vs Base-c1f9a359e8): 1-0 {White wins by adjudication}
Score of New-c1f9a359e8 vs Base-c1f9a359e8: 1 - 1 - 1  [0.500] 3
Started game 6 of 176 (Base-c1f9a359e8 vs New-c1f9a359e8)
Heartbeat stopped
Exception in thread Thread-2:
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "C:/tools/msys64/mingw64/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "C:/tools\msys64\home\Pasquale Pigazzini\_git\__test_folder00\fishtest\worker/games.py", line 621, in enqueue_output
    for line in iter(out.readline, ""):
ValueError: I/O operation on closed file

Exception running games:
Terminated by signal SIGINT
Informing the server
Uploading PGN...
Uploading compressed PGN of 173212 bytes
Task exited
Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
```

vdbergh:
Yek...

ppigazzini:
I'm wondering why it is necessary to start a new python process only to pass a pid...

vdbergh:
> I'm wondering why it is necessary to start a new python process only to pass a pid...

That's windows :)

ppigazzini:
Using this kill the worker too.
```python
def send_sigint(p):
    if IS_WINDOWS:
        if p.poll() is None:
            kernel = ctypes.windll.kernel32
            kernel.FreeConsole()
            # TODO: what if attach failed
            kernel.AttachConsole(p.pid)
            # TODO: improve readability, use constants
            # Disable Ctrl-C handling for our program
            kernel.SetConsoleCtrlHandler(None, 1)
            kernel.GenerateConsoleCtrlEvent(0, 0)
```

So if I'm not wrong we are using directly the windows kernel functions, attaching a console to the pid, and signaling to all the console group.

vdbergh:
The culprit seems to be 
```
C:/tools/msys64/home/usr00/_git/__test_folder00/fishtest/worker/env/bin/python3.exe', '-m', 'console_ctrl', '7980']' returned non-zero exit status 3221225781
```
The error code is 0xC0000135 which means DLL not found according to
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55

I had to change the PYTHONPATH in console_ctrl to include the current directory.

ppigazzini:
But using directly the kernel32.ddl we could try to follow one of the tutorial in other languages, without starting a new python process, this is the first found:
https://blog.codetitans.pl/post/sending-ctrl-c-signal-to-another-application-on-windows/

vdbergh:
> But using directly the kernel32.ddl we could try to follow one of the tutorial in other languages, without starting a new python process, this is the first found: https://blog.codetitans.pl/post/sending-ctrl-c-signal-to-another-application-on-windows/

I'll have a look....

vondele:
If it gets to messy, I think a linux - only solution is reasonable. This is just an optimization, and 80% of resources are on linux these days.

vdbergh:
Well I tried and could not get it to work under wine...

vdbergh:
I now went for Linux only. Maybe some windows specialist will know what to do. Since I don't have windows I cannot seriously look into this.

ppigazzini:
This works on windows with `multiprocessing` library (with `threading` the worker stops too)
https://docs.python.org/3/library/multiprocessing.html

```python
import ctypes
import multiprocessing
...

def send_ctrl_c(pid):
    kernel = ctypes.windll.kernel32
    kernel.FreeConsole()
    kernel.AttachConsole(pid)
    kernel.SetConsoleCtrlHandler(None, True) # IMO could be safely dropped
    kernel.GenerateConsoleCtrlEvent(0, 0)


def send_sigint(p):
    if IS_WINDOWS:
        proc = multiprocessing.Process(target = send_ctrl_c, args = (p.pid, ))
        proc.start()
        proc.join()
    else:
        p.send_signal(signal.SIGINT)

...
    with subprocess.Popen(
        cmd,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        universal_newlines=True,
        bufsize=1,
        # The next options are necessary to be able to send a CTRL_C_EVENT to this process.
        # https://stackoverflow.com/questions/7085604/sending-c-to-python-subprocess-objects-on-windows
        startupinfo=subprocess.STARTUPINFO(
            dwFlags=subprocess.STARTF_USESHOWWINDOW, wShowWindow=subprocess.SW_HIDE
        )
        if IS_WINDOWS
        else None,
        creationflags=subprocess.CREATE_NEW_CONSOLE if IS_WINDOWS else 0,
        close_fds=not IS_WINDOWS,
    ) as p:
```

vdbergh:
Perhaps you can merge the Linux version and then you can do the Windows version, This is more efficient since you can test it right away,...

ppigazzini:
Ok, commuting now, I will merge tonight.

ppigazzini (approved):

ppigazzini:
Triggered workers update, thank you @vdbergh :)

vdbergh:
Hmm...

https://tests.stockfishchess.org/actions?count=1&before=1638517068.296

vdbergh:
So it seems that `send_signal()` can fail.

vdbergh:
I learnt something interesting. Consider
```
try:
    try:
        raise Exception("a")
    finally:
        raise Exception("b")
except Exception as e:
    print(str(e))
```
versus
```
try:
    try:
        raise Exception("a")
    finally:
        pass
except Exception as e:
    print(str(e))
```
In the first case `"b"` is printed and in the second case `"a"`.  Somehow the last exception prevails it seems.

vdbergh:
This is the source code of `send_signal()` (https://github.com/python/cpython/blob/main/Lib/subprocess.py).
```
        def send_signal(self, sig):
            """Send a signal to the process."""
            # Don't signal a process that we know has already died.
            if self.returncode is not None:
                return
            if sig == signal.SIGTERM:
                self.terminate()
            elif sig == signal.CTRL_C_EVENT:
                os.kill(self.pid, signal.CTRL_C_EVENT)
            elif sig == signal.CTRL_BREAK_EVENT:
                os.kill(self.pid, signal.CTRL_BREAK_EVENT)
            else:
                raise ValueError("Unsupported signal: {}".format(sig))
```
It seems the only way it can throw an exception is by the inherent race condition triggering.

vdbergh:
I deleted my last comment since I mistakenly posted the code of `send_signal()` for `windows`.

vdbergh:
Ok this is the code of `subprocess.send_signal()`. It can only throw an exception if the race condition triggers. But how could dew had it 3 times in a row?
https://tests.stockfishchess.org/actions?action=&user=dew&before=1638527972.126&count=3
```
        def send_signal(self, sig):
            """Send a signal to the process."""
            # bpo-38630: Polling reduces the risk of sending a signal to the
            # wrong process if the process completed, the Popen.returncode
            # attribute is still None, and the pid has been reassigned
            # (recycled) to a new different process. This race condition can
            # happens in two cases.
            #
            # Case 1. Thread A calls Popen.poll(), thread B calls
            # Popen.send_signal(). In thread A, waitpid() succeed and returns
            # the exit status. Thread B calls kill() because poll() in thread A
            # did not set returncode yet. Calling poll() in thread B prevents
            # the race condition thanks to Popen._waitpid_lock.
            #
            # Case 2. waitpid(pid, 0) has been called directly, without
            # using Popen methods: returncode is still None is this case.
            # Calling Popen.poll() will set returncode to a default value,
            # since waitpid() fails with ProcessLookupError.
            self.poll()
            if self.returncode is not None:
                # Skip signalling a process that we know has already died.
                return

            # The race condition can still happen if the race condition
            # described above happens between the returncode test
            # and the kill() call.
            try:
                os.kill(self.pid, sig)
            except ProcessLookupError:
                # Suppress the race condition error; bpo-40550.
                pass
```

vdbergh:
Dew has some windows machines as well. If they are not correctly identified as such then this may cause an exception (on windows SIGINT is not recognized as an exception).

vdbergh:
Perhaps we should use os.name == 'nt' to recognize windows. See here https://docs.python.org/3/library/os.html .

ppigazzini:
The `finally:` clause it used to clean up some resource also in case of exception, so perhaps it's not expected to have there some dangerous code.

I tried to simplify the exceptions handling dropping some nested `try` clauses and keeping only the "dangerous" code in `try` clause.

Here is how to re-raise nested exceptions:
https://stackoverflow.com/questions/18188563/how-to-re-raise-an-exception-in-nested-try-except-blocks

Here is how to get exception info in `finally` clause:
https://stackoverflow.com/questions/1611561/can-i-get-the-exception-from-the-finally-block-in-python

ppigazzini:
Mmmm, the windows code has not the `self.poll()` ...

- windows:
```python
        def send_signal(self, sig):
            """Send a signal to the process."""
            # Don't signal a process that we know has already died.
            if self.returncode is not None:
                return
```
- linux:
```python
        def send_signal(self, sig):
            """Send a signal to the process."""
            self.poll()
            if self.returncode is not None:
                # Skip signalling a process that we know has already died.
                return
```

ppigazzini:
dew-12cores in https://tests.stockfishchess.org/actions?count=1&before=1638517068.296

https://tests.stockfishchess.org/tests/view/61a4271360afd064f2d72128
330 | dew-12cores-6a73657f | os: Linux 3.12.13-gentoo; ram: 24147MiB; gcc: 10.2.0; arch: 64bit SSE41 SSSE3 SSE2 POPCNT | 2021-12-03 07:37:48 | 048 / 334 | [0, 1, 7, 13, 3] | 0 | 0
-- | -- | -- | -- | -- | -- | -- | --

ppigazzini:
> Ok this is the code of `subprocess.send_signal()`. It can only throw an exception if the race condition triggers. But how could dew have it 3 times in a row? https://tests.stockfishchess.org/actions?action=&user=dew&before=1638527972.126&count=3

This should be the last implementation of `send_signal()`, that dew workers could uses python 3.3.3

vdbergh:
> > Ok this is the code of `subprocess.send_signal()`. It can only throw an exception if the race condition triggers. But how could dew have it 3 times in a row? https://tests.stockfishchess.org/actions?action=&user=dew&before=1638527972.126&count=3
> 
> This should be the last implementation of `send_signal()`, that dew workers could uses python 3.3.3

Perhaps the original implementation was more fragile. The new version cannot throw an exception at all I think (I said it incorrectly).

BTW From our experience it seems the windows `subprocess.send_signal()` is unusable anyway. So it does not really matter that the implementation is wrong.

vdbergh:
Perhaps we should also add the python version to the info lines for the tasks. Otherwise that information is gone afterwards.

ppigazzini:
> Perhaps we should also add the python version to the info lines for the tasks. Otherwise that information is gone afterwards.

I agree, the python version needs also to be separated by the worker version, something like this:
```python
        "version": "{}".format(WORKER_VERSION)
        "python_version": "{}.{}.{}".format(sys.version_info.major,  sys.version_info.minor, sys.version_info.micro,),
        "gcc_version": "{}.{}.{}".format(major, minor, patchlevel),
```

I can do in the weekend.

vdbergh:
Well if we do this then we could once and for all change switch to the schema (see `api.py`)
```
version: int,
python_version: [int,int,int],
gcc_version: [int,int,int],
```
This would make the validation on the server side more mechanical.

ppigazzini:
I will try :)

vdbergh:
@ppigazzini Do you still want to do this? Otherwise I can do it.  I would like to streamline the upgrade process a bit and for this it is useful if the version number is already available as  a separate field.

ppigazzini:
Do it, I don't know when I will have time to code it.

PR #1173, opened 2021-12-01
Improve account creation messages

ppigazzini:

PR #1174, opened 2021-12-01
Simplify authentication in api.py.

vdbergh:
As a bonus we can now also return "blocked" messages.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -131,9 +131,7 @@ class ApiView(object):
         )
         if "error" in token:
             self.handle_error(
-                "Invalid password for user: {}".format(
-                    self.request_body["worker_info"]["username"],
-                ),
+                token["error"],
                 exception=HTTPUnauthorized,
             )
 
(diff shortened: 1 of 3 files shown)


ppigazzini:
I can merge this and refactor #1173 

Unfortunately on DEV I'm getting this error installing fishtest (also master), perhaps a new pymongo version:

```
  File "<stdin>", line 4, in <module>
  File "/home/usr00/fishtest/server/fishtest/userdb.py", line 77, in add_user_group
    self.users.save(user)
  File "/home/usr00/fishtest/server/env/lib/python3.8/site-packages/pymongo/collection.py", line 2584, in __call__
    raise TypeError("'Collection' object is not callable. If you meant to "
TypeError: 'Collection' object is not callable. If you meant to call the 'save' method on a 'Collection' object it is failing because no such method exists.
```

vdbergh:
This is discussed here https://stackoverflow.com/questions/46866702/save-is-deprecated-for-what-in-pymongo

vdbergh:
So maybe it should be 
```
self.users.replace_one({'_id': user['_id']}, user)
```

ppigazzini:
I'm pushing a fix to use `pymongo 3.x,` better to debug the issue taking the proper time.

ppigazzini:
DEV updated.

PR #1175, opened 2021-12-01
Use pymongo 3.x

ppigazzini:
The code needs some refactoring to use the new pymongo 4.x

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -8,7 +8,7 @@ requires = [
     "pyramid_debugtoolbar",
     "pyramid_mako",
     "waitress",
-    "pymongo",
+    "pymongo < 4",
     "numpy",
     "scipy",
     "requests",


ppigazzini:
I pushed this PR after getting these 2 errors:
1.
```
File "/home/usr00/fishtest/server/fishtest/userdb.py", line 77, in add_user_group
    self.users.save(user)
  File "/home/usr00/fishtest/server/env/lib/python3.8/site-packages/pymongo/collection.py", line 2584, in __call__
    raise TypeError("'Collection' object is not callable. If you meant to "
TypeError: 'Collection' object is not callable. If you meant to call the 'save' method on a 'Collection' object it is failing because no such method exists.
```
2.
```
File "/home/usr00/fishtest/server/fishtest/rundb.py", line 470, in get_finished_runs
    count = c.count()
AttributeError: 'Cursor' object has no attribute 'count'
```

PR #1181, opened 2021-12-03
Send CTRL+C to cutechess-cli on Windows

ppigazzini:
On Windows the signalling of the CTRL+C event requires a disposable process.

References:
https://stackoverflow.com/questions/7085604/sending-c-to-python-subprocess-objects-on-windows
https://stackoverflow.com/questions/813086/can-i-send-a-ctrl-c-sigint-to-an-application-on-windows
https://docs.microsoft.com/en-us/windows/console/freeconsole
https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler
https://docs.microsoft.com/en-us/windows/console/attachconsole
https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent

Added some controls before signalling to minimize
the race condition probability.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -23,6 +23,9 @@ from zipfile import ZipFile
 import requests
 
 IS_WINDOWS = "windows" in platform.system().lower()
+if IS_WINDOWS:
+    import ctypes
+    from multiprocessing import Process
 
 ARCH = "?"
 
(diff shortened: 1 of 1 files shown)


vdbergh:
https://github.com/glinscott/fishtest/blob/748d0307bf7c00fe1285d8f8ac630b4c4844ae04/worker/games.py#L934

This comment should be changed...

PR #1183, opened 2021-12-04
Do not make dead tasks visible by default in the events log.

vdbergh:
This is similar to system events.

A dead task may for example be caused by a user closing their laptop. Or a colab worker being disconnected.

Dead tasks do not carry a lot of information (they have no associated message) and there are a lot of them.

But the fact that a task was declared dead can explain why a pgn file was not uploaded. So it might still be useful to have a record of them.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -13,7 +13,7 @@ class ActionDb:
         if action:
             q["action"] = action
         else:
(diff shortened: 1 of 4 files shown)


ppigazzini:
DEV updated, LGTM, let me know when I can merge it.

vdbergh:
I think it is ok now. It is a very simple patch.

ppigazzini:

PR #1185, opened 2021-12-04
Imply use_all_cores=False if lowering concurrency.

### LINKED ISSUE #1184: use_all_cores silently overrides concurrency
If you launch the client once with ```use_all_cores```, it will save this in the configuration. If you then relaunch it with a lower ```--concurrency```, that will get ignored because the saved ```use_all_cores=True``` overrules it.

I can fix this, but I'm not sure what the desired behavior is. I think the easiest and least mistake prone is to warn and refuse to run (after all, these options directly conflict). But perhaps the intent of ```use_all_cores``` is to serve as an extra warning/unlock to not use concurrency=number of cpus, and simply shouldn't be saved to the config file at all, perhaps renaming it to ```allow_all_cores``` to reflect that.

Thoughts?

gcp:
Currently, if you launch the client once with use_all_cores, it
will save this in the configuration. If you then relaunch it with
a lower --concurrency, that will get ignored because the saved
use_all_cores=True overrules it.

The cleanest fix (that is still backwards compatible and avoids
introducing new flags) appears to be to make a concurrency that's
less than maximum imply use_all_cores=False.

Fixes #1184.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -217,6 +217,12 @@ def setup_parameters(config_file):
     elif protocol == "https" and options.port == "80":
         # Rewrite old port 80 to 443
         options.port = "443"
+    # If we are not going to use all cores, then unset that option.
+    # Note that if we used that option previously, we'd have set
+    # concurrency to the maximum value in the config file, so this will
+    # (only) apply when overruling concurrency via the command line.
+    if cpu_count < multiprocessing.cpu_count():
+        options.use_all_cores = "False"
 
     try:
         if options.use_all_cores == "True":


vdbergh:
I guess it means that we translate `--use_all_cores=True` into `--concurrency=cpu_count`.  `--use_all_cores=False` translates into disallowing `--concurrency>=cpu_count`.

So logically `--use_all_cores` is almost superfluous (but convenient).

vdbergh:
If backward compatibility weren't an issue then the cleanest solution would be the introduce the special value "max" for concurrency.  `--concurrency=max` would then be equivalent to `--use_all_cores`. (I.e. `--concurrency=cpu_count),

worker/worker.py
```diff
@@ -217,6 +217,12 @@ def setup_parameters(config_file):
     elif protocol == "https" and options.port == "80":
         # Rewrite old port 80 to 443
         options.port = "443"
+    # If we are not going to use all cores, then unset that option.
+    # Note that if we used that option previously, we'd have set
+    # concurrency to the maximum value in the config file, so this will
+    # (only) apply when overruling concurrency via the command line.
+    if cpu_count < multiprocessing.cpu_count():
```
ppigazzini, line 224: This is always `True` if we don't have already `concurrency = <Max Threads>` in the config file, so we are not able anymore to use all Threads with `--use_all_cores True`

gcp: You're totally right. We probably can't get around to changing options handling so it can make the distinction between command line options and config file ones

vdbergh: Yes. `--concurrency` and `--use_all_cores` are a unit. If one of them is present on the command line then neither of them should default to the value in the configfile.

PR #1187, opened 2021-12-05
Some tweaks around option handling.

vdbergh:
1) We try to fix
    
https://github.com/glinscott/fishtest/issues/1184
    
The semantics of `--use_all_cores` is now as follows:
    
 `--use_all_cores True`
    
(1) allows the use of all cores;
(2) makes the default concurrency equal to the number of cores.
    
The default can be overridden by supplying an explicit `--concurrency` option.
    
 `--use_all_cores False` disallows the use of all cores.
    
2) We validate config file entries and command line parameters.
    
In particular we make sure that in the option object returned by setup_parameters() all attributes have the expected type. This makes the downstream code simpler.
    
3) We switch from optparse to argparse. optparse is deprecated.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -93,35 +93,15 @@ In that case the corresponding value is an error message.
 """
 
 
-bool_list = ["true", "1", "false", "0"]
+def _bool(x):
+    x = x.strip().lower()
+    if x in {"true", "1"}:
+        return True
+    if x in {"false", "0"}:
+        return False
+    raise ValueError(x)
 
-# A custom boolean type that can be instantiated
-# from a string.
-class BOOL(object):
-    def __init__(self, x):
-        y = x.lower().strip()
-        if y not in bool_list:
-            raise ValueError(x)
-        self.x = y
-
-    def __bool__(self):
-        if self.x in ["0", "1"]:
-            return self.x == "1"
-        else:
-            return self.x.lower() == "true"
-
-    def __eq__(self, w):
-        if not isinstance(w, BOOL) and not isinstance(w, bool):
-            return False
-        return self.__bool__() == w.__bool__()
-
-    def __str__(self):
-        return str(self.__bool__())
-
-    __repr__ = __str__
-
-
-BOOL.__name__ = "bool"
+_bool.__name__ = "bool"
 
 
 def safe_sleep(f):
@@ -162,11 +142,6 @@ def setup_parameters(config_file):
         with os.popen(cmd) as proc:
             mem_str = str(proc.readlines())
         mem = int(re.search(r"\d+", mem_str).group())
-        print(
-            "System memory determined to be: {:.3f}GiB".format(
-                mem / (1024 * 1024 * 1024)
-            )
-        )
     except Exception as e:
         print("Exception checking HW info:\n", e, sep="", file=sys.stderr)
         return None
@@ -197,8 +172,8 @@ def setup_parameters(config_file):
         ("parameters", "concurrency", str(cpu_count), int),
         ("parameters", "max_memory", str(int(mem / 2 / 1024 / 1024)), int),
         ("parameters", "min_threads", "1", int),
-        ("parameters", "fleet", "False", BOOL),
-        ("parameters", "use_all_cores", "False", BOOL),
+        ("parameters", "fleet", "False", _bool),
+        ("parameters", "use_all_cores", "False", _bool),
     ]
 
     for v in defaults:
@@ -209,7 +184,7 @@ def setup_parameters(config_file):
         else:
             o = config.get(v[0], v[1])
             t = v[3]
-            if isinstance(t, type):
+            if callable(t):
                 try:
                     _ = t(o)
                 except:
@@ -229,7 +204,10 @@ def setup_parameters(config_file):
 
     # Step 3: parse the command line. Use the current config options mostly as defaults.
 
-    parser = ArgumentParser()
+    parser = ArgumentParser(
+        description="usage (long version):"
+        " worker.py USERNAME PASSWORD [OPTIONS]"
+    )
     parser.add_argument(
         "-P",
         "--protocol",
@@ -280,17 +258,17 @@ def setup_parameters(config_file):
         "-f",
         "--fleet",
         dest="fleet",
-        default=BOOL(config.get("parameters", "fleet")),
-        type=BOOL,
-        choices=[BOOL("False"), BOOL("True")],  # useful for usage message
+        default=config.getboolean("parameters", "fleet"),
+        type=_bool,
+        choices=[False, True],  # useful for usage message
         help="quit in case of errors or if no task is available",
     )
     parser.add_argument(
         "-a",
         "--use_all_cores",
         dest="use_all_cores",
-        type=BOOL,
-        choices=[BOOL("False"), BOOL("True")],
+        type=_bool,
+        choices=[False, True],
         help="allow the worker to use all cores",
     )
     parser.add_argument(
@@ -298,8 +276,8 @@ def setup_parameters(config_file):
         "--only_config",
         dest="only_config",
         default=False,
-        type=BOOL,
-        choices=[BOOL("False"), BOOL("True")],
+        type=_bool,
+        choices=[False, True],
         help="just write the configfile and quit",
     )
 
@@ -354,7 +332,7 @@ def setup_parameters(config_file):
     ):
         options.concurrency = max_cpu_count
     if options.use_all_cores is None:
-        options.use_all_cores = BOOL(config.get("parameters", "use_all_cores"))
+        options.use_all_cores = config.getboolean("parameters", "use_all_cores")
     if options.concurrency is None:
         options.concurrency = config.getint("parameters", "concurrency")
 
@@ -394,6 +372,11 @@ def setup_parameters(config_file):
     with open(config_file, "w") as f:
         config.write(f)
 
+    print(
+        "System memory determined to be: {:.3f}GiB".format(
+            mem / (1024 * 1024 * 1024)
+        )
+    )
     print(
         "Worker constraints: {{'concurrency': {}, 'max_memory': {}, 'min_threads': {}}}".format(
             options.concurrency, options.max_memory, options.min_threads
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -15,11 +15,11 @@ import time
 import traceback
 import uuid
 import zlib
-from configparser import ConfigParser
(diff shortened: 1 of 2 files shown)


ppigazzini:
Nice!

worker/worker.py
```diff
@@ -95,31 +95,14 @@
...
-        if not isinstance(w, BOOL) and not isinstance(w, bool):
-            return False
-        return self.__bool__() == w.__bool__()
-
-    def __str__(self):
-        return str(self.__bool__())
-
-    __repr__ = __str__
-
+def BOOL(x):
```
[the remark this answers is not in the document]
vdbergh: I saw that too but I preferred not to depend on an external function. That way can tweak the recognition function more easily to our needs (it's very small anyway).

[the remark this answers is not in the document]
ppigazzini: Ok. Notes:
- leftover ? `BOOL.__name__ = "bool"`
- here are suggesting to use `str2bool()` to avoid confusion with the meaning of standard `bool()` https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

```python
def str2bool(x):
    x = x.strip().lower()
    if x in {"true", "1"}:
        return True
    elif x in {"false", "0"}:
        return False
    else:
        raise ValueError(x)
```

BTW `distutils.util.strtobool()` is broken (`return 1` is not `return True`). Fortunately `distutil` is deprecated and will be dropped in python 3.12
https://github.com/python/cpython/blob/98fac8bc183c113903403793ffe411358ee40d8a/Lib/distutils/util.py#L308-L321

```python
import distutils.util
print(f"{distutils.util.strtobool('True') is True}")
```

vdbergh:
> leftover ? BOOL.__name__ = "bool"

No this is important. Otherwise you get some strange output.

I agree that your `str2bool` is better. I will change it (but we would still neet `str2bool.__name__ = bool`)

vdbergh:
This is what you get without `BOOL.__name__ = "bool"`
```
python worker.py --only_config brla
Worker started in /home/fishtest/fishtest/worker ... (PID=312574)
Creating lock file /home/fishtest/fishtest/worker/worker.lock
System memory determined to be: 11.130GiB
argument -w/--only_config: invalid BOOL value: 'brla'                <---------------------
Error parsing options. Config file not written.
Deleting lock file /home/fishtest/fishtest/worker/worker.lock
```

ppigazzini:
see https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse/19233287#19233287 for this alternative:

```python
    parser = ArgumentParser()
    parser.register("type", "bool", BOOL)
...
    parser.add_argument(
        "-w",
        "--only_config",
        dest="only_config",
        default=False,
        type="bool",
        choices=[False, True],
        help="just write the configfile and quit",
    )
```
output:
```
env/bin/python3 worker.py --only_config TTTTT
Worker started in /home/usr00/_git/__test_folder00/fishtest/worker ... (PID=24524)
Creating lock file /home/usr00/_git/__test_folder00/fishtest/worker/worker.lock
System memory determined to be: 24.811GiB
argument -w/--only_config: invalid 'bool' value: 'TTTTT'
Error parsing options. Config file not written.
```

vdbergh:
Actually I think I am not so happy with `str2bool`. We want a name that suggests a type (since it is used as `type=...`).
We cannot use `bool` since that would shadow the builtin `bool`. But perhaps `_bool` (if you don't like `BOOL`)?

Another point of view is that semantically the function is really a constructor, and constructors are usually named after the type of the object they construct. Since we cannot use `bool` we should use something close to it.

Note that `BOOL` (as currently named) is also used for validating config entries (along with `int` and `str`).

ppigazzini:
It's up to you, I thought that `BOOL` was a leftover of the previous class implementation.

vdbergh:
Sorry I missed your earlier post. Will reply after dinner.

vdbergh:
> Sorry I missed your earlier post. Will reply after dinner.

I thought about the `register` solution. But since we also need the `_bool` type for validating config entries, the solution with `_bool.__name__="bool"` seems more convenient in the end.

I will do it later but now I need to do something urgent at home.

vdbergh:
I made a rudimentary password dialog. I think the worker used to have one but not any longer. Probably for a reason...

vdbergh:
I made an improved version in which the worker will first verify the credentials, before writing them to the configfile.

vdbergh:
More robust version I hope. username/password will now always be validated with the server, even if they are coming from the configfile.

vdbergh:
I added an argument to `--validate` to override username/password validation. This is useful in case there is no network access and you just want to create a config file.

worker/worker.py
```diff
@@ -93,18 +95,101 @@
...
 
 
-def setup_parameters(config_file):
+def verify_credential(remote, username, password, cached):
+    # Returns:
+    # True  : username/password are ok
+    # False : username/password are not ok
+    # None  : network error: unable to determine the status of
+    #         username/password
+    try:
```
ppigazzini, line 123: Writing from mobile, this should work:
```python
    req = {}
    if username != "" and password != "":
        print(
            "Confirming {} credentials with {}".format(
                "cached" if cached else "supplied", remote
            )
        )
        payload = {"worker_info": {"username": username}, "password": password}
        try:
            req = send_api_post_request(remote + "/api/request_version", payload)
        except:
            return None  # network problem (unrecoverable)
        if "error" in req:
            return False  # invalid username/password
    else:
        return False  # empty username or password
    print("Credentials ok!")
    return True
```

vdbergh: Thanks! Looks good to me.

ppigazzini: Cleaned up more.

worker/worker.py
```diff
@@ -133,139 +216,262 @@ def setup_parameters(config_file):
...
     config.set("parameters", "protocol", options.protocol)
     config.set("parameters", "host", options.host)
-    config.set("parameters", "port", options.port)
-    config.set("parameters", "concurrency", options.concurrency)
-    config.set("parameters", "max_memory", options.max_memory)
-    config.set("parameters", "min_threads", options.min_threads)
-    config.set("parameters", "fleet", options.fleet)
-    config.set("parameters", "use_all_cores", options.use_all_cores)
+    config.set("parameters", "port", str(options.port))
+    if options.concurrency != max_cpu_count:
```
ppigazzini, line 454: In master wins `use_all_cores = True` when both `concurrency = 5` and `use_all_cores = True` are in configfile, with the proposed code instead wins `concurrency = 5`
With this code we have only one entry between `concurrency = 5` and `use_all_cores = True`. Setting `--use_all_cores = False` without concurrency will use the default number of cores like the proposed code:
```python
    if options.concurrency != max_cpu_count:
        config.set("parameters", "concurrency", str(options.concurrency))
        config.remove_option("parameters", "use_all_cores")
    else:
        assert options.use_all_cores
        config.set("parameters", "use_all_cores", str(options.use_all_cores))
        config.remove_option("parameters", "concurrency")
    config.set("parameters", "max_memory", str(options.max_memory))
    config.set("parameters", "min_threads", str(options.min_threads))
    config.set("parameters", "fleet", str(options.fleet))

    with open(config_file, "w") as f:
```

BTW I agree with you that, without the retro compatibility requirement, it was way better something like `--concurrency max`

vdbergh: I did it intentionally this way (it is written in the announcement). It is much more logical and for the common use cases it makes no difference.

ppigazzini: My suggested change uses your logic but writes only one of the two alternative options in `fishtest.cfg` and not both. IMO this is confusing:
```
[login]
username = user00
password = user00

[parameters]
protocol = https
host = dfts-0.pigazzini.it
port = 443
max_memory = 16224
min_threads = 1
fleet = False
use_all_cores = True
concurrency = 5
```

ppigazzini: Alternative:
```python
    if options.concurrency != max_cpu_count:
        config.set("parameters", "concurrency", str(options.concurrency))
        options.use_all_cores = str(False)
    else:
        assert options.use_all_cores
```

vdbergh:
For me `--use_all_cores true` means that the worker _can_ use all cores if it wants to.

Assume there are 4 cores.
```
python worker.py  --use_all_cores true    <--- now 4 cores are used
python worker.py  --concurrency 2           <--- now 2 cores are used
python worker.py  --concurrency 4           <--- should work in IMHO
```
If we drop `---use_all_cores true` from the config file after the second line (or set use_all_cores to false) then the third line will not work.

vdbergh:
The reason I am dropping `concurrency` from the config file if it is equal to `max_cpu_count` (this appears to be a purely cosmetic operation) is that in this way the config file can be distributed to other machines with an unknown number of cores.

ppigazzini:
> In my mind `--use_all_cores true` means that the worker can use all cores if it wants to.

In this case this line should start or the concurrency already in `fishtest.cfg` or the default concurrency, not max cores:
`python worker.py  --use_all_cores true      <--- now 4 cores can be used`

I know that max cores is set for retro compatibility, but it breaks the logic.

ppigazzini:
BTW I haven't found problems in my tests on WSL2 and msys2.

vdbergh:
> > In my mind `--use_all_cores true` means that the worker can use all cores if it wants to.
> 
> In this case this line should start or the concurrency already in `fishtest.cfg` or the default concurrency, not max cores: `python worker.py --use_all_cores true <--- now 4 cores can be used`
> 
> I know that max cores is set for retro compatibility, but it breaks the logic.

Yes this is for backward compatibility.  In the PR `--use_all_cores` stands for 1) allow the worker to use all cores 2) set the default number of cores to `max_cpu_count`.

vdbergh:
> BTW I haven't found problems in my tests on WSL2 and msys2.

That's good news. I worried a bit that the username/password dialog would not work under msys2.

vdbergh:
BTW It seems that on Linux `getpass` reads from `/dev/tty`. Do it cannot be controlled by `Popen(stdin=PIPE, ...)`.

ppigazzini:
In msys2 `getpass` works fine either with `mintty` or `cmd`.

vdbergh:
Do you think it would be very bad to break backwards compatibility, and to introduce ‘concurrency = max’ ?
We can print a message if —use_all_cores is used on the command line. Something like.

```
—use_all_cores has been replaced by —concurrency=max
```
It would make the code much simpler.

vdbergh:
We could have a type _cores that does all the validation.

ppigazzini:
I'm to go with a better code :)

vdbergh:
replaced by #1189

PR #1189, opened 2021-12-08
Some tweaks around option handling.

vdbergh:
1) This PR started by trying to fix.
    
https://github.com/glinscott/fishtest/issues/1184
    
It turned that there are fundamental issues with the semantics of `--use_all_cores`. So it was decided to deprecate `--use_all_cores` and to introduce the possibility of using `--concurrency max` as a clearly defined substitute for `--use_all_cores True`.
    
Some backward compatibility is temporarily retained, but `--use_all_cores` will be removed in the future to simplify the worker code.
    
2) We now strictly validate config file entries and command line parameters.
    
In particular we make sure that in the options object returned by `setup_parameters()` all attributes have the expected type. This makes the downstream code simpler.
    
3) We switch from `optparse` to `argparse`. `optparse` is deprecated.
    
4) We now check, at the startup of the worker, that the password, supplied via the config file or on the command line, is correct. If it is not then we present a password dialog.
    
This validation step is skipped when the flag `--no_validation` is present on the command line. In combination with `--only_config` this is useful when there is no network access.
    
5) `--only_config` is now a flag. In contrast --fleet remains a boolean since all three possibilities (`True/False` or omitted) are meaningful.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -150,7 +150,23 @@ def send_api_post_request(api_url, payload):
         data=json.dumps(payload),
         headers={"Content-Type": "application/json"},
         timeout=HTTP_TIMEOUT,
-    ).json()
+    )
+    valid_response = True
+    try:
+        response = response.json()
+    except Exception as e:
+        valid_response = False
+    if valid_response and not isinstance(response, dict):
+        valid_response = False
+    if not valid_response:
+        message = "The reply to post request {} was not a json encoded dictionary".format(api_url)
+        print(
+            "Exception in send_api_post_request():\n",
+            message,
+            sep="",
+            file=sys.stderr,
+        )
+        raise WorkerException(message, e=e)
     if "error" in response:
         print("Error from remote: {}".format(response["error"]))
     if "info" in response:
--- a/worker/test_worker.py
+++ b/worker/test_worker.py
@@ -25,8 +25,11 @@ class workerTest(unittest.TestCase):
             pass
 
     def test_config_setup(self):
-        sys.argv = [sys.argv[0], "user", "pass"]
-        worker.setup_parameters("foo.txt")
+        sys.argv = [sys.argv[0], "user", "pass", "--no_validation"]
+        if os.path.exists("foo.txt"):
+            os.remove("foo.txt")
+        worker.CONFIGFILE = "foo.txt"
+        worker.setup_parameters(".")
         from configparser import ConfigParser
 
         config = ConfigParser()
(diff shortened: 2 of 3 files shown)


ppigazzini:
At the moment we need to keep the `_bool` for the `--use_all_cores True` backward compatibility, but I'm wondering if in the long run that could be dropped:
 - `--only_config False` is redundant, we could use simply `--only_config` and update the worker setup scripts
 - I don't know if `--validate False` has an effective use

vdbergh:
Yes at some point we should definitely drop the booleans for the flag options (`--only_config` and `--validate`). `--validate` should be replaced by something like `--no_validation`.

We can do it now if you don't mind breaking backward compatibility (the flags have no effect on the config file). 

I would keep `_bool` for `--fleet` since it is not a flag option.

For `--use_all_cores` I would also keep it since --use_all_cores will be deleted soon.

ppigazzini:
I'm to make the logic right in this moment.
`--only_config` should be used only by me in the worker setup scripts, `--validate` is brand new and I didn't find at the moment an useful use of `--validate False`

vdbergh:
I went ahead and made `--only_config` and `--no_validation` flags.

vdbergh:
> I'm to make the logic right in this moment. `--only_config` should be used only by me in the worker setup scripts, `--validate` is brand new and I didn't find at the moment an useful use of `--validate False`

--validate False is meant to be used with --only_config when there is no network available.

vdbergh:
At least I found `--validate False` (now `--no_validation`) very useful during testing.

ppigazzini:
Understood, so it was `--validate True` without an use.

ppigazzini:
LGTM with all the OS workers. Tested also `python3 worker.py` with `use_all_cores = True` in `fishtest.cfg`, so the update should be issue free.

vdbergh:
Ok I will squash.

ppigazzini (approved):

vdbergh:
Ok done squashing. Sorry for doing so many pushes but this happened when I noticed typos in the rather long commit message.

ppigazzini:
Never mind. I re fetched and re done `black `and `isort`, merged with https://github.com/glinscott/fishtest/commit/8cc3b507d49513234005d4435d9767598dbf947c

ppigazzini:
PROD updated, triggered the workers update, thank you @vdbergh :)

vdbergh:
> * wiki pages updated (worker cli parameters and linux worker setup script)
> * windows worker setup script updated
> * missing 22 (230 vs 208) workers after the update, notably the 10 mlang-16cores workers. I will write an e-mail to mlang

:cry: I wonder what might have caused that.
>   BTW: the `--no_validation` flag is very useful for the setup script

vdbergh:
I wonder if it is some python incompatibility. Like `callable` was removed in 3.0 and 3.1 and then reintroduced in 3.2.

ppigazzini:
@vdbergh mail sent to Mario asking for information. I'm wondering if you want to join the fishtest maintainer group :)

ppigazzini:
Still no reply to my mail but now there are 50 mlang-16-cores :)

Wrote a message on the forum to advertise the deprecated `--use_all_cores True` options, pinged noob on discord, updated the Colab wiki section.

PR #1195, opened 2021-12-11
Adjust network timeouts

vondele:
MAX_RETRY_TIME: speedup recovery time after longer fishtest outages.
HTTP_TIMEOUT: allow a little more time for longer downloads


ppigazzini (approved):

ppigazzini:
PROD updated, triggered workers update, thank you @vondele :)

PR #1196, opened 2021-12-11
Remove a bunch of superfluous arguments from verify_signature().

vdbergh:
Also: make ARCH non-global.

It was actually used for two different things!


vdbergh:
It seems the concurrency argument does not mean what its name suggests. I need to think about this.

vdbergh:
Fixed now.

ppigazzini (approved):

PR #1198, opened 2021-12-13
Introduce a short delay at worker startup

vondele:
to avoid flooding the server when a worker fails early, but executes in a loop,
introduce a small delay on startup. Over the typical time of a worker running,
this introduces essentially no inefficiency.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -974,8 +974,7 @@ def worker():
     # always end with a little delay.
     # The exception is when connecting to the dev server, to aid testing
     print("Thank you for contributing, please wait as the worker finalizes!")
-    if options.host != "dfts-0.pigazzini.it":
-        safe_sleep(STARTUP_TIMEOUT)
+    safe_sleep(STARTUP_TIMEOUT)
 
     return 0 if fish_exit else 1
 
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -44,6 +44,7 @@ from updater import update
 
 WORKER_VERSION = 140
 HTTP_TIMEOUT = 30.0
+STARTUP_TIMEOUT = 15.0
 MAX_RETRY_TIME = 900.0  # 15 minutes
 IS_WINDOWS = "windows" in platform.system().lower()
 CONFIGFILE = "fishtest.cfg"
@@ -924,6 +925,7 @@ def worker():
     # All seems to be well...
     remote = "{}://{}:{}".format(options.protocol, options.host, options.port)
 
+
     # Start heartbeat thread as a daemon (not strictly necessary, but there might be bugs)
     heartbeat_thread = threading.Thread(
         target=heartbeat,
@@ -967,6 +969,13 @@ def worker():
 
     print("Waiting for the heartbeat thread to finish...")
     heartbeat_thread.join(HTTP_TIMEOUT)
+
+    # To avoid flooding the server for a failing worker in a loop,
+    # always end with a little delay.
+    # The exception is when connecting to the dev server, to aid testing
+    print("Thank you for contributing, please wait as the worker finalizes!")
+    if options.host != "dfts-0.pigazzini.it":
+        safe_sleep(STARTUP_TIMEOUT)
 
     return 0 if fish_exit else 1
 


vdbergh:
I am not a fan of this since it makes testing painful.

At best we could do it when the worker quits, and only if —fleet is true.

But —fleet was specifically introduced to eliminate delays. If we reintroduce them, we may as well eliminate —fleet (that would be my preferred option).

vondele:
First, I agree that this will not win a beauty contest, but it is a quite reliable fix for flooding.

just doing this for fleet true doesn't fix it when a normal user puts his failing worker in a loop. Fleet, IMO, was introduced to be able to redirect workers to other work in case there were no patches that needed testing. Doing this redirect with a 15s delay seems no problem.

I also considered doing this at the end, but that seemed less robust (can we make sure we always exit there on all possible error conditions?). It is probably equally annoying (or more so for the typical user). The testing use case is definitely a bit of a problem, but is it worth a more complicated solution (like something server side will likely be), that will also scale with fleet size?

vondele:
maybe a fix for this testing use case is to skip the wait of the URL points to the dev server?

vdbergh:
I think the worker now quits reliably. So putting the delay at the end is really not a problem. 

I also think users should be told not to run the worker in a loop.

noobpwnftw:
In fact, workers has to fail the request in the first place(timeout on the github request) in order to flood the server, so in practice there was already a delay, it is only because of the amount of workers that creates a problem, therefore maybe there should be a general rate limit on the server side or just beef up the server for such scenarios...

vdbergh:
I propose we put the delay at the end.

If ever there is a case of flooding because somehow the delay at the end was missed then we can discuss again. Note that we had only one case of flooding so far…

vondele:
Like so?

vondele:
> In fact, workers have to fail the request in the first place(timeout on the github request) in order to flood the server, so in practice there was already a delay, it is only because of the amount of workers that creates a problem, therefore maybe there should be a general rate limit on the server side or just beef up the server for such scenarios...

unless the failure is immediate and not a timeout (like domain not found or whatever error could be immediate)?

vondele:
Admittedly, it still is strange that there was an overload, as the number of error messages in the event log was on the order of 1 per second, and that doesn't seem so much.

worker/worker.py
```diff
@@ -968,6 +970,13 @@ def worker():
     print("Waiting for the heartbeat thread to finish...")
     heartbeat_thread.join(HTTP_TIMEOUT)
 
+    # To avoid flooding the server for a failing worker in a loop,
+    # always end with a little delay.
+    # The exception is when connecting to the dev server, to aid testing
+    print("Thank you for contributing, please wait as the worker finalizes!")
+    if options.host != "dfts-0.pigazzini.it":
```
ppigazzini, line 977: The developer will test also with his server. If this is necessary better to use
```python
if options.host == "tests.stockfishchess.org":
```

vdbergh:
> Like so?

I will look tomorrow. However I would not exclude the DEV server (one cannot use the DEV server for testing if it behaves differently from the main server…).

Also there should be no delay after Ctrl-c, but that is easily fixable.

vdbergh:
> Admittedly, it still is strange that there was an overload, as the number of error messages in the event log was on the order of 1 per second, and that doesn't seem so much.

I think after each failed task the server writes the (growing) run to the db. This perhaps leads to quadratic behaviour.

vondele:
quadratic behavior could definitely be a good reason.

vdbergh:
Another problem might be that while the server is multithreaded this is useless in Python for CPU bound work loads because of the GIL.

vdbergh:
One possibility to mitigate flooding would be to use `buffer(run,False)` in `failed_task()`. This means that the run is marked as dirty in the cache but it is not written to disk. So there will be no db access for every failed task. It means that the data the webserver sees is a bit more outdated than now, but in exceptional situations this should be ok.

A fleshed out version of the run is still written to the actiondb (without caching). However that is much less data.

Of course it depends whether the db access is the bottleneck. It might since the db has to serialize the data (to bson), which probably depends on the size of the data (i.e.\ the size of the task list). 

Anyway I think it is worth a try.

vondele:
I think this PR can be closed in light of the speedup fix.

PR #1200, opened 2021-12-18
Make `failed_task()` less expensive.

vdbergh:
(1) Avoid making a copy of the task list in `del_tasks()`.
(2) Use `buffer(run,False)`.
(3) Do not grab the active run lock (there seems to be no point).

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -362,12 +362,11 @@ class RunDb:
                     ),
                     flush=True,
                 )
-                with self.active_run_lock(str(run_id)):
-                    run["dead_task"] = "task_id: {}, worker: {}".format(
-                        task_id, worker_name(task["worker_info"])
-                    )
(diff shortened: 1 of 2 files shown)


vdbergh:
On an unscientific test with a small run with `200` tasks on my local fishtest, with this PR `failed_task()` was 3 times faster compared to master: `0.008s` vs `0.026s`. I did the test several times and the results were consistent.

We should probably have some mechanism to benchmark apis on the server side, rather than on the worker side (to account for network overhead).

vdbergh:
With PR #1202 we can check the server cost of the two implementations. Now the difference is even much more spectacular `0.0016s` vs `0.0265s` (a ratio of `15.9`) (for a run with `200` tasks on my local fishtest).

ppigazzini (approved):

vondele:
Thanks!

ppigazzini:
PROD updated,

PR #1204, opened 2021-12-21
Update class and trim whitespaces

dav1312:


dav1312:
There are still some trailing whitespaces here, not sure if I should trim them too
https://github.com/glinscott/fishtest/blob/b658e85530112f2d8dc4ae36241eb9a156bcefc8/worker/worker.py#L52-L57

ppigazzini:
The logo is fine after a trim by hand.

worker/packages/requests/sessions.py
```diff
@@ -506,7 +506,7 @@ def request(self, method, url,
             ``False``, requests will accept any TLS certificate presented by
             the server, and will ignore hostname mismatches and/or expired
             certificates, which will make your application vulnerable to
-            man-in-the-middle (MitM) attacks. Setting verify to ``False`` 
+            man-in-the-middle (MitM) attacks. Setting verify to ``False``
```
ppigazzini, line 509: IMO better to skip this trim, this is a package installed with `pip`

ppigazzini (approved):
DEV looks good with SPSA like this:
https://tests.stockfishchess.org/tests/view/619e2c93c0a4ea18ba95a450

ppigazzini:

ps: logo white spaces already trimmed in https://github.com/glinscott/fishtest/commit/b33f012691e0924f5dcb2b9fb39f6e63d4d98551 to have only a worker update

Issue #879, opened 2021-01-08
Remaining hours wrong calculated?


pb00068:
At the moment the fishtest-page mainpage says 
191 machines 1774 cores 1.12M nps (1987.47M total nps) 1393 games/minute **5.4 hours remaining** 

But from experience I know with almost certainty that in 5,5 hours the tests are by far not finished yet.
How is this remaining hours calculated and what does it mean?

It would be fine when the remaining hours would be a good estimation about when the queue will be empty if no-one submits new tests.

vondele:
I think that's what the estimate is doing.

vondele:
Maybe something is wrong. But it looks not too wrong. in 5.5 hours we have something like 400k-500k games for 10 tests remaining, with only 2LTC.

vondele:
https://github.com/glinscott/fishtest/commit/66c8fd676e65f76530770d0a4d813acc536b127b

pb00067:
Look at the situation now, there are :
- 12 LTC tests running, most of them far from converging.
- 14 STC tests  running
-  2 Tuning sessions 
-  2945 cores running 0.78 MNps (2300.08 total MNps) and 2118 games/minute

Fishtest says remaining hours: 9.4    
I simply don't believe that.

vondele:
The estimate is based on two heuristics that need adjusting:

https://github.com/glinscott/fishtest/blob/16448a4b95c3c6203a5d4b9352df8e84a26a9739/server/fishtest/util.py#L326-L337

* the average number of games for a SPRT test running on fishtest. This needs updating after the SPRT bounds change. I think I just looked at the last N SPRT tests on fishtest and took the average. 
* the estimated duration per game... I extracted that from a couple of pgns downloaded. 

So the estimate might be a bit off these days, but I don't think by a 2x factor. Note that this estimate assumes no further tests are being submitted.... that's of course not reality

pb00067:
Thanks for analysis Joost!

vondele:
the average of the number of games in the last 400 LTC tests is 58000 so actually quite close to the number we have.

PR #1206, opened 2021-12-22
Adjust remaining hours estimate

vondele:
as discussed in the issue, our estimate is probably relatively accurate.

It might 'feel' wrong, as it reports the time needed to empty fishtest
assuming no further patches are submitted.

This updates one of the parameters to the current measured value.

fixes https://github.com/glinscott/fishtest/issues/879


ppigazzini (approved):

ppigazzini:

PR #1208, opened 2021-12-23
Add timestamp to api timings.

vdbergh:
This was suggested by @vondele.

Also: log api calls, with timestamp, to a file "api.log".

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -145,6 +145,7 @@ def requests_post(remote, *args, **kw):
 
 
 def send_api_post_request(api_url, payload, benchmark=True):
+    logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), "api.log")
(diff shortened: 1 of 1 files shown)


worker/games.py
```diff
@@ -25,6 +25,7 @@
 import requests
 
 IS_WINDOWS = "windows" in platform.system().lower()
+LOGFILE = os.path.abspath("api.log")
```
ppigazzini, line 28: If I'm not wrong this can be dropped.

vdbergh: Oops.

vdbergh: Fixed.

ppigazzini (approved):

ppigazzini:
PROD updated to trigger the workers update. Thank you @vdbergh :)

vdbergh:
Thx!

PR #1210, opened 2021-12-23
Fix html tag attributes and auto format css, html, js files

ppigazzini:
```
npx prettier --write "server/fishtest/static/{css/*.css,html/*.html,js/*[!highlight.diff.min].js}"
```

- fix by hand some missing html tags attributes quotes that stop `prettier`
- fix some typos

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -1,4 +1,3 @@
-
 /* Bootstrap responsive override */
 @media (max-width: 850px) {
   body {
@@ -6,7 +5,7 @@
   }
 
   .mainnavbar .nav li {
-    margin-right: 4px
+    margin-right: 4px;
   }
 
   .mainnavbar .nav li a,
@@ -19,7 +18,7 @@
 
   .mainnavbar .nav li a::before,
   .mainnavbar .nav li div::before {
-    font-size: 14px
+    font-size: 14px;
   }
 
   .mainnavbar {
@@ -27,13 +26,14 @@
     width: 56px !important;
   }
 
-  .btn-group-sm>.btn, .btn-sm {
-    padding: 0 .25rem;
+  .btn-group-sm > .btn,
+  .btn-sm {
+    padding: 0 0.25rem;
   }
 
   .nav-header,
   .mainnavbar::-webkit-scrollbar {
-    display: none !important
+    display: none !important;
   }
 
   .nav .nav-header:not(:first-child) + li {
@@ -53,7 +53,7 @@ body {
 }
 
 ::-webkit-scrollbar-corner {
-  background: transparent
+  background: transparent;
 }
 
 ::-webkit-scrollbar-thumb {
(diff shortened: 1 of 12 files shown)


dav1312:
You may also want to add the missing `</li>` tag in line 84 and fix the indent issue

dav1312:
Also, kinda weird to have some classes, IDs, etc without quotes, some with single quotes and some with double quotes.

ppigazzini:
@dav1312 auto formatting with:
```
npx prettier --write "server/fishtest/static/{css/*.css,html/*.html,js/*[!highlight.diff.min].js}"
```
Let me know your opinion.

ppigazzini:
> You may also want to add the missing `</li>` tag in line 84 and fix the indent issue

Fixed by `prettier`

ppigazzini:
> Also, kinda weird to have some classes, IDs, etc without quotes, some with single quotes and some with double quotes.
> 
> https://github.com/glinscott/fishtest/blob/c4d990dbc7421eb2c2999634a3e1a28491ab1377/server/fishtest/static/html/SPRTcalculator.html#L56-L58

Fixed by `prettier`

dav1312:
> @dav1312 auto formatting with:
> 
> ```
> npx prettier --write "server/fishtest/static/{css/*.css,html/*.html,js/*[!highlight.diff.min].js}"
> ```
> 
> Let me know your opinion.

I would change the indentation to 4 spaces to make it the same as in the python scripts

Also it looks like it moved curly braces to a new line in some places but not others (maybe due to print width?) which is not great for consistency.

https://github.com/glinscott/fishtest/pull/1210/files#diff-af68d6bed50f20918ece350ef78b0ec4d3d9accaecbbb2d0365f9695058fff76R282-R283

And (again maybe due to the print width?) it seems to split some `<a>` `<table>` and `<script>` tags into multiple lines which looks weird but that's more personal preference

https://github.com/glinscott/fishtest/pull/1210/files#diff-e01afe0e95fe1be13180e5a73318b337c1cd8d64727f374066ab1379a5e163ffR80-R83

#### But in general the code seems much easier to read and that's great!

---

Also, I noticed that there is some inconsistency with the comments in JS files
Using `/* */` for single-lines instead of `//` 

https://github.com/glinscott/fishtest/blob/master/server/fishtest/static/js/spsa_new.js#L99

Using `//` multiple times for multi-line comments instead of something like starred blocks

https://github.com/glinscott/fishtest/blob/master/server/fishtest/static/js/application.js#L1-L2

https://github.com/glinscott/fishtest/blob/master/server/fishtest/static/js/application.js#L59-L61

```js
/*
 * https://stackoverflow.com/questions/14267781/sorting-html-table-with-javascript
 * https://stackoverflow.com/questions/40201533/sort-version-dotted-number-strings-in-javascript
 */
```

And beginning the comments with a space and uppercase in some files and not doing it in others

https://github.com/glinscott/fishtest/blob/master/server/fishtest/static/js/spsa.js#L42

*Maybe I'm being too picky*

ppigazzini:
@dav1312 thank you for the review. I merged the PR to have a bit improved code format.
TODO: fix the other format issues.

PR #1212, opened 2021-12-23
Introduce generic logging function with timestamp.

vdbergh:
Trying to find a compromise between user friendly terminal messages and machine parsability is a losing battle. So we now separate the two.

For api logging we now print something user friendly on the terminal and for logging we use something more machine parsable.
    
We also log run/task_id since there may be a correlation between timings and the length of the run.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -25,6 +25,9 @@ from zipfile import ZipFile
 import requests
 
 IS_WINDOWS = "windows" in platform.system().lower()
+LOGFILE = "api.log"
+
+LOG_LOCK = threading.Lock()
 
 
 class WorkerException(Exception):
@@ -68,6 +71,30 @@ UPDATE_RETRY_TIME = 15.0
 REPO_URL = "https://github.com/official-stockfish/books"
 EXE_SUFFIX = ".exe" if IS_WINDOWS else ""
 MAKE_CMD = "make COMP=mingw " if IS_WINDOWS else "make COMP=gcc "
+
+
+def log(s):
+    logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), LOGFILE)
+    with LOG_LOCK:
+        with open(logfile, "a") as f:
+            f.write("{} : {}\n".format(datetime.datetime.utcnow(), s))
+
+
+def backup_log():
+    try:
+        logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), LOGFILE)
+        logfile_previous = logfile + ".previous"
+        if os.path.exists(logfile):
+            print("Moving logfile {} to {}".format(logfile, logfile_previous))
+            with LOG_LOCK:
+                os.replace(logfile, logfile_previous)
+    except Exception as e:
+        print(
+            "Exception moving log:\n",
+            e,
+            sep="",
+            file=sys.stderr,
+        )
 
 
 def str_signal(signal_):
@@ -144,8 +171,7 @@ def requests_post(remote, *args, **kw):
     return result
 
 
-def send_api_post_request(api_url, payload, benchmark=True):
-    logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), "api.log")
+def send_api_post_request(api_url, payload, quiet=False):
     t0 = datetime.datetime.utcnow()
     response = requests_post(
         api_url,
@@ -175,19 +201,25 @@ def send_api_post_request(api_url, payload, benchmark=True):
         raise WorkerException(message)
     if "error" in response:
         print("Error from remote: {}".format(response["error"]))
-    if "info" in response:
-        print("Info from remote: {}".format(response["info"]))
-    if benchmark:
-        t1 = datetime.datetime.utcnow()
-        message = "{}: {}: s:{:.2f}ms, w:{:.2f}ms".format(
-            t1,
+
+    t1 = datetime.datetime.utcnow()
+    w = 1000 * (t1 - t0).total_seconds()
+    s = 1000 * response["duration"]
+    log(
+        "{:6.2f} ms (s)  {:7.2f} ms (w)  {}".format(
+            s,
+            w,
             api_url,
-            1000 * response["duration"],
-            1000 * (t1 - t0).total_seconds(),
         )
-        print(message)
-        with open(logfile, "a") as f:
-            f.write(message + "\n")
+    )
+    if not quiet:
+        if "info" in response:
+            print("Info from remote: {}".format(response["info"]))
+        print(
+            "Post request {} handled in {:.2f}ms (server: {:.2f}ms)".format(
+                api_url, w, s
+            )
+        )
     return response
 
 
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import atexit
 import base64
+import datetime
 import getpass
 import math
 import multiprocessing
@@ -35,6 +36,8 @@ from games import (
     FatalException,
     RunException,
     WorkerException,
+    backup_log,
+    log,
     run_games,
     send_api_post_request,
     str_signal,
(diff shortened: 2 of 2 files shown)


ppigazzini:
@vondele please review the new text format :)

vondele:
will work. The easiest for tools like `awk` `cut` `sort` etc, are when there are a fixed number of fields (which we have here), and there is a separation between numbers and 'other stuff'. So this would be convenient:
`"{} : {}  (s) {:.2f} ms  (w) {:.2f} ms"`

worker/games.py
```diff
@@ -70,6 +71,12 @@ def is_64bit():
 MAKE_CMD = "make COMP=mingw " if IS_WINDOWS else "make COMP=gcc "
 
 
+def log(s):
+    logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), LOGFILE)
+    with open(logfile, "a") as f:
+        f.write("{}: {}\n".format(datetime.datetime.utcnow(),s))
```
vondele, line 77: If I want to graph directly with the timestamp having `{} : {}\n` is an advantage

worker/worker.py
```diff
@@ -681,6 +681,13 @@ def locked_by_others(lock_file, require_valid=True):
         return True
     return False
 
+def utcoffset():
```
vondele, line 684: https://docs.python.org/3/library/datetime.html?highlight=utcoffset#datetime.datetime.utcoffset exists?

vdbergh: I tried this but all my attempts failed. E.g.
```
>>> print(datetime.datetime.now().utcoffset())
None
```
It's a bit like java programming. You search the docs until you have finally found a class that seems to do what you want, and then it turns out to be an abstract class.

vdbergh: Actually my implementation is wrong (for offsets that are not multiple of hours).

vondele: I wouldn't try to implement that.... however the python one requires  this:

datetime.tzinfo
The object passed as the tzinfo argument to the datetime constructor, or None if none was passed.

so somehow one needs to init datetime correctly.

vdbergh: Yes, if one can get hold of a `tzinfo` object describing the local timezone. It seems that ought to be possible...

ppigazzini: https://stackoverflow.com/questions/32402502/how-to-change-the-time-zone-in-python-logging/47104004

worker/worker.py
```diff
@@ -714,7 +722,8 @@ def fetch_and_handle_task(worker_info, password, remote, lock_file, current_stat
         if req["version"] > WORKER_VERSION:
             print("Updating worker version to {}".format(req["version"]))
             update()
-        print("Fetch task...")
+        print("Current time is {} {}".format(datetime.datetime.utcnow(), utcoffset()))
```
ppigazzini, line 725: Can help this?
```
>>> print("utc:\t{}\nlocal:\t{}".format(datetime.datetime.utcnow(), datetime.datetime.now()))
utc:    2021-12-24 12:10:00.723139
local:  2021-12-24 13:10:00.723146
```

vdbergh: Yes this is one of the suggestions at SO (for t0 one can take any time). I used another suggestion based on the time module.

Note that most of the code is for converting to the standard stringformat e.g. `+0100`.

vdbergh: Note I am using `utcoffset()` only for qualifying the utctime on the terminal (otherwise users might get confused). The timestamps in the logfile just use utc.

vdbergh: I guess one can replace `t0.timestamp()` by `0`...

vondele: Informing the users could be with a `UTC: ` or `(UTC)` addition.

vdbergh: I changed it this way, since it shaves off one line of code.

vdbergh: @vondele Currently it prints
```
Current time is 2021-12-24 13:16:21.469050 +0100
```
I think this is more or less standard to indicate the utcoffset.

vondele: this is fine with me.... I'm just worried about any code that deals with timezones, daylight saving, and things like that. Somehow I think having it all UTC will simplify things.

ppigazzini: > Yes this is one of the suggestions at SO (for t0 one can take any time). I used another suggestion based on the time module.
> 
> Note that most of the code is for converting to the standard stringformat e.g. `+0100`.

Little OT: I'm relying less and less on SO for python, some posts have outdated, though still working, solutions. After starting to understand the language, I'm founding helpful the python documentation. In this case the https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp referenced also the `time` way.

ppigazzini:
> Actually my implementation is wrong (for offsets that are not multiple of hours).

@vdbergh If I'm not wrong this message is outdated, ie the PR is ready to be merged.

vdbergh:
I am extending the logging statement a bit.

vdbergh:
Now it should in principle be done.

vdbergh:
I did another small tweak. The log function now uses a lock since it is used from two different threads. Although the risk is low, I think that in principle file operations are not thread safe.

vondele:
eventually something should be in place to make sure the log doesn't grow indefinitely. Maybe move api.log to api.log.previous every time the work updates?

vdbergh:
> eventually something should be in place to make sure the log doesn't grow indefinitely. Maybe move api.log to api.log.previous every time the worker updates?

That seems reasonable and simple. No need to over engineer this I think.

vdbergh:
I implemented it.

worker/games.py
```diff
@@ -70,6 +73,28 @@ def is_64bit():
...
+def log(s):
+    logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), LOGFILE)
+    with LOG_LOCK:
+        with open(logfile, "a") as f:
+            f.write("{} : {}\n".format(datetime.datetime.utcnow(), s))
+
+
+def backup_log():
+    try:
+        logfile = os.path.join(os.path.dirname(os.path.realpath(__file__)), LOGFILE)
```
vondele, line 85: `with LOG_LOCK:` ?

vdbergh: Ok (the heartbeat thread isn't running yet when this is invoked, but it is better to be safe than sorry).

vdbergh: Fixed.

vdbergh:
I reverted to the time module based solution for computing the utcoffset. I was wrong that one can compute the utcoffset for `timestamp=0` as this does not take daylight savings time into account. Since one also needs to get the current timestamp (as it was in the solution proposed by @ppigazzini) then the two solutions are the same number of lines, but I think the time module based solution has some pedagogical value since it tells you where the offset actually comes from.

worker/games.py
```diff
@@ -70,6 +73,30 @@ def is_64bit():
...
+        logfile_previous = logfile + ".previous"
+        if os.path.exists(logfile):
+            print("Moving logfile {} to {}".format(logfile, logfile_previous))
+            with LOG_LOCK:
+                os.rename(logfile, logfile_previous)
```
ppigazzini, line 90: This fail on Windows when "api.log.previous" already exists (it seems in WSL2 too in my tests).
https://docs.python.org/3/library/os.html
```
os.rename(src, dst, *, src_dir_fd=None, dst_dir_fd=None)
Rename the file or directory src to dst. If dst exists, the operation will fail with an OSError subclass in a number of cases:

On Windows, if dst exists a FileExistsError is always raised.
```

vdbergh: Hmm. I replaced `os.rename()` by `os.replace()`. That should do the trick.

ppigazzini:
I start commuting, tests will continue tonight. If ok I will merge up to #1214 with a workers update trigger.

vdbergh:
Sounds good!

ppigazzini (approved):
Tests completed.

ppigazzini:
PROD updated, workers update triggered,
Thank you @vdbergh :)

vdbergh:
:+1:

PR #1216, opened 2021-12-27
Introduce `/api/get_task/{id}/{task_id}`

vdbergh:
This makes it convenient to inspect tasks from long ago before the current db schema crystallized. It's another step towards db conversion.

Tweaks in the returned task.

- The `unique_key` is truncated to 8 characters.
- The remote address is replaced by "?.?.?.?".
- If `"task_id"` ends with `"bad"` then the api call returns the corresponding element from `run["bad_tasks"]`.
- `task["last_updated"]` is replaced by a string since json does not know about `datetime`.
- If `task["residual"]` is infinite then it is replaced by the string `"inf"` since JSON does not know about infinity.


ppigazzini (approved):
I'm wondering if this route can be served by the non critical pserve (upstream in nginx config).

ppigazzini:

PR #1225, opened 2022-01-05
Adjust max games limit, and default test limit

vondele:


vondele:
the 400k games limit is reached now more regularly, as well as the 800k limit, which limits the number of params that can be tuned.

ppigazzini:
DEV updated.
The new max games are:
- SPRT 800k
- SPSA 3.2M
- NumGames 3.2M

ppigazzini:

Issue #1224, opened 2022-01-05
Pagination for colored runs not working correctly


vdbergh:
Still trying to wrap my head around pagination in Fishtest.

If we view for example yellow runs, Fishtest currently list 4618 page but from page 509 on, all pages are empty E.g. https://dfts-0.pigazzini.it/tests/finished?page=509&yellow_only=1 .

The culprit I think is the following code
```python
            count = self.runs.estimated_document_count(hint=idx_hint)
```
`estimated_document_count` does not seem to take a `hint` argument. See https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount/

The correct way to handle this IMHO is to create a "yellow" view on the "runs" collections and then to use `estimated_documment_count` on this view. See https://docs.mongodb.com/manual/core/views/.

vondele:
eventually it would be nice if we could somehow merge and make more flexible the list of tests displayed. Like right now, I sometimes would like to see all LTC (also active), or my tests, that are LTC, etc.

vdbergh:
> eventually it would be nice if we could somehow merge and make more flexible the list of tests displayed. Like right now, I sometimes would like to see all LTC (also active), or my tests, that are LTC, etc.

The api almost supports this already. https://tests.stockfishchess.org/tests/user/vdv?ltc_only=1 . Currently it is (or should be) restricted to finish runs, but that is easily fixed I think. However the pagination is broken (4622 pages most of which are empty). This is not a practical problem, but it looks very unprofessional.

ppigazzini:
Joost has 115528 finished LTC ???

vdbergh:
More criteria are also possible

https://tests.stockfishchess.org/tests/user/vdv?ltc_only=1&success_only=1

All finished green LTC tests by vondele.

vdbergh:
Fixed by #1227 .

PR #1226, opened 2022-01-05
[WIP] Retire `idx_hint`. It doesn't work.

vdbergh:
Also: replace `get_finished_runs()` by the more generic `get_runs()` which can also return unfinished runs. Currently this possibility is not used.

Quick fix for #1224 .

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -449,7 +449,7 @@ class RunDb:
                     )
         return (runs, pending_hours, cores, nps)
 
-    def get_finished_runs(
+    def get_runs(
         self,
         skip=0,
         limit=0,
@@ -457,35 +457,39 @@ class RunDb:
         success_only=False,
         yellow_only=False,
         ltc_only=False,
+        finished=True,
     ):
-        q = {"finished": True}
-        idx_hint = "finished_runs"
+        fast_count = True
+        q = {}
+        if finished:
+            q["finished"] = finished
+            fast_count = False
         if username:
             q["args.username"] = username
-            idx_hint = None
+            fast_count = False
         if ltc_only:
             q["tc_base"] = {"$gte": 40}
-            idx_hint = "finished_ltc_runs"
+            fast_count = False
         if success_only:
             q["is_green"] = True
-            idx_hint = "finished_green_runs"
+            fast_count = False
         if yellow_only:
             q["is_yellow"] = True
-            idx_hint = "finished_yellow_runs"
+            fast_count = False
 
         c = self.runs.find(
             q, skip=skip, limit=limit, sort=[("last_updated", DESCENDING)]
         )
 
-        if idx_hint:
+        if fast_count:
             # Use a fast COUNT_SCAN query when possible
-            count = self.runs.estimated_document_count(hint=idx_hint)
+            count = self.runs.estimated_document_count()
         else:
             # Otherwise, the count is slow
-            count = c.count()
+            count = self.runs.count_documents(q)
         # Don't show runs that were deleted
         runs_list = [run for run in c if not run.get("deleted")]
-        return [runs_list, count]
+        return runs_list, count
 
     def get_results(self, run, save_run=True):
         if not run["results_stale"]:
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1162,13 +1162,14 @@ def get_paginated_finished_runs(request):
 
     page_idx = max(0, int(request.params.get("page", 1)) - 1)
     page_size = 25
-    finished_runs, num_finished_runs = request.rundb.get_finished_runs(
+    finished_runs, num_finished_runs = request.rundb.get_runs(
         username=username,
         success_only=success_only,
         yellow_only=yellow_only,
         ltc_only=ltc_only,
         skip=page_idx * page_size,
         limit=page_size,
+        finished=True,
     )
 
     pages = [
--- a/server/tests/test_rundb.py
+++ b/server/tests/test_rundb.py
@@ -189,7 +189,7 @@ class CreateRunDBTest(unittest.TestCase):
         self.rundb.buffer(run, True)
 
     def test_40_list_LTC(self):
-        finished_runs = self.rundb.get_finished_runs(limit=3, ltc_only=True)[0]
+        finished_runs = self.rundb.get_runs(limit=3, ltc_only=True)[0]
         for run in finished_runs:
             print(run["args"]["tc"])
 


ppigazzini:
DEV updated after the fix of the test.

ppigazzini:
#574 PR that introduced the `idx_hint`

vdbergh:
> DEV is very slow opening "Greens", "Yellows", "LTC"

Yes that's because the code in master does not work (1000s of empty pages). Anyway I changed this to WIP.

The issue seems hard to fix. Maybe be the simplest run a cron job to get the counts of common queries and put them in a "registry" type collection. It probably doesn't matter if the count is only approximative at any given time (I would need to think about this).

ppigazzini:
Perhaps it was a hack by linrock to get a speedup at the price of a wrong total count.

vdbergh:
Could you try on DEV

```javascript
use fishtest_new
db.createView("runs_green","runs", {"$match" : { "is_green": true, "finished": true }} )
```
and if successful then
```javascript
db.runs_green.estimatedDocumentCount()
```

vdbergh:
> Perhaps it was a hack by linrock to get a speedup at the price of a wrong total count.

No it was not a hack. The code simply doesn't work (estimated_document_count() cannot use an index).

ppigazzini:
https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.count_documents
> - hint (string or list of tuples): The index to use. Specify either the index name as a string or the index specification as a list of tuples (e.g. [(‘a’, pymongo.ASCENDING), (‘b’, pymongo.ASCENDING)]). This option is only supported on MongoDB 3.6 and above.

ppigazzini:
> ```js
> db.createView("runs_green","runs", {"$match" : { "is_green": true, "finished": true }} )
> ```

```
db.createView("runs_green","runs", {"$match" : { "is_green": true, "finished": true }} )
MongoServerError: BSON field 'create.pipeline' is the wrong type 'object', expected type 'array'
```

vdbergh:
`estimatedDocumentCount()` does not take a hint argument but `countDocuments()` does!
Can you try:
```javascript
db.runs.countDocuments({"is_green" : true, "finished" : true}, {"hint" : "finished_green_runs"} )
```

ppigazzini:
```
db.runs.countDocuments({"is_green" : true, "finished" : true}, {"hint" : "finished_green_runs"} )
10471
 db.runs_green.estimatedDocumentCount()
0
```

vdbergh:
The issue that estimatedDocumentCount does not take a query argument. 

```
db.runs.countDocuments({"is_green" : true, "finished" : true}, {"hint" : "finished_green_runs"} )
10471
```
So was it faster than just
```
db.runs.countDocuments({"is_green" : true, "finished" : true})
```
Now I thinking it should probably not be since mongodb would be smart enough to figure out that it can use the index.

ppigazzini:
PROD patched by hand with:
```python
            #count = self.runs.estimated_document_count(hint=idx_hint)
            count = self.runs.count_documents(q, hint=idx_hint)
```

vdbergh:
> PROD patched by hand with:
> 
> ```python
>             #count = self.runs.estimated_document_count(hint=idx_hint)
>             count = self.runs.count_documents(q, hint=idx_hint)
> ```

In case there are no arguments it could still use `self.runs.estimated_document_count()`. I think that would still be faster.

ppigazzini:
It seems working fine also with:
https://tests.stockfishchess.org/tests/user/vdv?ltc_only=1
https://tests.stockfishchess.org/tests/user/vdv?ltc_only=1&success_only=1

vdbergh:
I was thinking something along the lines of 
```
	if idx_hint == "finished_runs":
            # Use a fast COUNT_SCAN query when possible                                                                                      
            count = self.runs.estimated_document_count()
        else:
            # Otherwise, the count is slow                                                                                                   
            count = self.runs.document_count(q, hint=idx_hint)
```

ppigazzini:
If we haven't an index "finished_runs" I prefer something like this:
```python
	if idx_hint is None:
            # Use a fast COUNT_SCAN query when possible                                                                                      
            count = self.runs.estimated_document_count()
        else:
            # Otherwise, the count is slow                                                                                                   
            count = self.runs.document_count(q, hint=idx_hint)
```

vdbergh:
> If we haven't an index "finished_runs" I prefer something like this:
> 
> ```python
> 	if idx_hint is None:
>             # Use a fast COUNT_SCAN query when possible                                                                                      
>             count = self.runs.estimated_document_count()
>         else:
>             # Otherwise, the count is slow                                                                                                   
>             count = self.runs.document_count(q, hint=idx_hint)
> ```

That would be wrong :) since `idx_hint is None` means a user query.

But it was just a quick hack to be able to test if `estimated_document_count()` would be faster.

vdbergh:
There is an indexed "finished_runs". The question is: which ones is faster
```javascript
db.runs.estimatedDocumentCount()
```
or
```
db.runs.countDocuments({}, {"hint" : "finished_runs"} )
```

ppigazzini:
PROD on master, DEV with:
```python
        if idx_hint == "finished_runs":
            # Use a fast COUNT_SCAN query when possible
            count = self.runs.estimated_document_count()
        else:
            # Otherwise, the count is slow
            count = self.runs.count_documents(q, hint=idx_hint)
```

vdbergh:
It seems hard to tell the difference (if any). So perhaps let's go for `count = self.runs.count_documents(q, hint=idx_hint)`.

Since you were going to update this code anyway, perhaps you can do it?

vdbergh:
I will close this PR then.

ppigazzini:
DEV updated to deal with a "username" query:
```python
        if idx_hint == "finished_runs":
            # Use a fast COUNT_SCAN query when possible
            count = self.runs.estimated_document_count()
        elif idx_hint is None:
        else:
           # Otherwise, the count is slow
            count = self.runs.count_documents(q, hint=idx_hint)
```

ppigazzini:
> I will close this PR then.

I use pymongo monkey style, please write the new PR :)

vdbergh:
Ok.

ppigazzini:
Thanks, I will not able to write an useful commit message :)

I rewrote in this way (`hint` parameter cannot be `None`):
```python
        if idx_hint == "finished_runs":
            # Use a fast COUNT_SCAN query when possible
            count = self.runs.estimated_document_count()
        # Otherwise, the count is slow
        elif idx_hint:
            count = self.runs.count_documents(q, hint=idx_hint)
        else:
```

PR #1227, opened 2022-01-05
Fix issue with incorrectly reported document counts.

vdbergh:


ppigazzini:
DEV updated, `hint` cannot be `None`:

https://dfts-0.pigazzini.it/tests/user/vdv
```
Jan 05 17:12:39 vps1572612706 pserve[1465]:   File "/home/usr00/fishtest/server/fishtest/rundb.py", line 480, in get_finished_runs
Jan 05 17:12:39 vps1572612706 pserve[1465]:     count = self.runs.count_documents(q, hint=idx_hint)
Jan 05 17:12:39 vps1572612706 pserve[1465]:   File "/home/usr00/fishtest/server/env/lib/python3.8/site-packages/pymongo/collection.py", line 1816, in count_documents
Jan 05 17:12:39 vps1572612706 pserve[1465]:     kwargs["hint"] = helpers._index_document(kwargs["hint"])
Jan 05 17:12:39 vps1572612706 pserve[1465]:   File "/home/usr00/fishtest/server/env/lib/python3.8/site-packages/pymongo/helpers.py", line 91, in _index_document
Jan 05 17:12:39 vps1572612706 pserve[1465]:     raise TypeError("must use a list of (key, direction) pairs, "
Jan 05 17:12:39 vps1572612706 pserve[1465]: TypeError: must use a list of (key, direction) pairs, not: None
```

ppigazzini:
Fixed on DEV with:
```python
        if idx_hint:
            count = self.runs.count_documents(q, hint=idx_hint)
        else:
```

vdbergh:
Ok! Copied in the latest commit.

ppigazzini:
DEV update.

The home page is slow, the best solution seems to be your first suggestion:
```python
        if idx_hint == "finished_runs":
            # Use a fast COUNT_SCAN query when possible
            count = self.runs.estimated_document_count()
           # Otherwise, the count is slow
        elif idx_hint is not None:
            count = self.runs.count_documents(q, hint=idx_hint)
        else:
```

ppigazzini:
DEV is running this PR.
PROD is running your first proposed solution.

vdbergh:
Ok. Updated. 

For very fast document counts we should count runs before a certain date (e.g. in a cron job, say before 1-1-2021) and cache it. Then we can count the results of queries with small result sets. The question is if the added complexity would be worth it.

vdbergh:
I now subtract the unfinished runs from the total count or "runs". This requires an extra query but it it should be very fast since it returns only a few documents and there is an index.

ppigazzini:
DEV updated.

vdbergh:
It seems to be fine. The extra query seems to make no difference for the loading of the main page.

ppigazzini:
Looks good to me.
MongoDB needs the `hint` to use the index. LTC, Greens, Yellows are painfully slow with:
```python
        elif idx_hint is not None:
```

vdbergh:
I will squash the PR.

ppigazzini (approved):

ppigazzini:

PR #994, opened 2021-08-14
Revert "Stop the test after failing to setup an engine"

ppigazzini:
Reverts glinscott/fishtest#992

PR #1229, opened 2022-01-07
Create `failed`, `deleted`, `is_green`, `is_yellow` flags in new_run.

vdbergh:
Queries in mongodb for missing flags are a bit messy. This is the first step at fixing this. The next step is to loop through the runs collection and set/unset these flags as needed. **Both steps should be mostly non functional for now** (except for a few bug fixes along the way).

Also: we document where the flags

`finished`, `failed`, `deleted`, `is_green`, `is_yellow` are set and unset.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -161,10 +161,35 @@ class RunDb:
                 "pentanomial": 5 * [0],
             },
             "results_stale": False,
-            "finished": False,
             "approved": False,
             "approver": "",
         }
+
+        # administrative flags
+
+        # "finished"
+        # set in stop_run(), /api/stop_run
+        # cleared in purge_run(), /tests/modify
+        new_run["finished"] = False
+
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated

ppigazzini:
It's a functional (and an enhancement one). Before this PR a "failed" run can be restarted with "modify" but it was still flagged as "failed" after stopping.
I reverted (#994) a my commit  for this issue.

vdbergh:
> It's a functional (and an enhancement one) PR. Before this PR a "failed" run can be restarted with "modify" but it was still flagged as "failed" after stopping. I reverted (#994) a my commit for this issue.

I updated to description...

We should have some kind of voting system for stopping a run. E.g. bench failure 3 votes. Make failure 1 vote. Stop the run after 10 votes.

ppigazzini:

PR #1231, opened 2022-01-07
Do not use absolute pathname for net.

vdbergh:
SF cannot handle paths with non-ascii characters on windows.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1196,14 +1196,10 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
     # Add EvalFile with full path to cutechess options, and download the networks if missimg.
     net_base = required_net(base_engine)
     if net_base:
-        base_options = base_options + [
-            "option.EvalFile={}".format(os.path.join(testing_dir, net_base))
-        ]
+        base_options = base_options + ["option.EvalFile={}".format(net_base)]
     net_new = required_net(new_engine)
     if net_new:
-        new_options = new_options + [
-            "option.EvalFile={}".format(os.path.join(testing_dir, net_new))
-        ]
+        new_options = new_options + ["option.EvalFile={}".format(net_new)]
 
     for net in [net_base, net_new]:
         if net:
@@ -1349,6 +1345,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
                 "name=New-" + run["args"]["resolved_new"][:10],
                 "tc={}".format(scaled_new_tc),
                 "cmd={}".format(new_engine_name),
+                'dir=.',
             ]
             + new_options
             + ["_spsa_"]
@@ -1357,6 +1354,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
                 "name=Base-" + run["args"]["resolved_base"][:10],
                 "tc={}".format(scaled_tc),
                 "cmd={}".format(base_engine_name),
+                'dir=.',
             ]
             + base_options
             + ["_spsa_"]


vondele:
actually, this might not work cutechess-cli requires the absolute path, since the working directory of the binary is $TEMP or similar. (IIRC).

vdbergh:
It does work on Linux though... Of course I do not really know if SF picked up the net from the working instead of from the executable's directory.

vdbergh:
But I did check that when cutechess-cli is started, the current directory is the testing directory, if that is what you mean.

vondele:
when cutechess-cli is started the current directory is the testing directory, but cutechess itself executes the stockfish process in a different directory ($TEMP).

vdbergh:
Ok. I will have to take your work for it. A more robust solution then is to embed the net.

vdbergh:
Embedding the net works fine but this blows up the size of the executable of course.

vdbergh:
It seems one can specify the working directory of the engine on the command line.

ppigazzini:
Actually all the workers are running fine with this PR and cutechess-cli runs in "testing" folder, see:

https://github.com/glinscott/fishtest/blob/dc7b9bf13875eb08479f5dcb3fffdb4843128f96/worker/games.py#L1166-L1168

https://github.com/glinscott/fishtest/blob/dc7b9bf13875eb08479f5dcb3fffdb4843128f96/worker/updater.py#L61-L64

vdbergh:
@ppigazzini  What @vondele says is that cutechess-cli changes the working directory to a temporary directory before starting the engine. I checked in the cutechess-cli source that it is correct.

However this can be overridden by specifying the working directory on the command line. This is what I did in the latest push.

vdbergh:
@ppigazzini Perhaps you can try to run this in a directory with a non-ascii name? Here is how to enter unicode characters in windows 

> To insert a Unicode character, type the character code, press ALT, and then press X. For example, to type a dollar symbol ($), type 0024, press ALT, and then press X. For more Unicode character codes, see Unicode character code charts by script.

I assume you can also enter special characters with the usual tool.

ppigazzini:
Restarted all the workers, queued fresh tests.

vdbergh:
I did check that it works on Linux with a non-ascii directory. But I guess SF supports UTF-8 out of the box.

ppigazzini:
I'm trying to enter a UTF-8 char on Windows, these instructions don't work.
I enter ASCII code with the numeric pad, eg ALT+96 for `

ppigazzini:
Ok, I used the character map, now the msys2 worker is running this PR in `/home/usr00/_git/__test_folder00/fishtestжӖӚӋѼѾ/worker`

vdbergh:
> Ok, I used the character map, now the msys2 worker is running this PR in `/home/usr00/_git/__test_folder00/fishtestжӖӚӋѼѾ/worker`

Great!!! Can you also check that master does **not** work?

ppigazzini:
Master doesn't work in `/home/usr00/_git/__test_folder00/fishtestݴڞۚٸۚ۳۩ݾڣڝکڰۋ۝۬ۙې/worker` :

```
Finished game 98 (Base-ad926d34c0 vs New-ad926d34c0): 0-1 {White disconnects}
Score of New-ad926d34c0 vs Base-ad926d34c0: 49 - 49 - 0  [0.500] 98
Started game 99 of 176 (New-ad926d34c0 vs Base-ad926d34c0)
Started game 100 of 176 (Base-ad926d34c0 vs New-ad926d34c0)
Terminating process of engine New-ad926d34c0(98)
Finished game 99 (New-ad926d34c0 vs Base-ad926d34c0): 0-1 {White disconnects}
Score of New-ad926d34c0 vs Base-ad926d34c0: 49 - 50 - 0  [0.495] 99
Terminating process of engine Base-ad926d34c0(99)
Finished game 100 (Base-ad926d34c0 vs New-ad926d34c0): 0-1 {White disconnects}
Score of New-ad926d34c0 vs Base-ad926d34c0: 50 - 50 - 0  [0.500] 100
```

ppigazzini (approved):

ppigazzini:
Triggered the worker update, thank you @vdbergh @vondele :)

PR #1234, opened 2022-01-08
Further cleanups in the db.

vdbergh:
1) We replace a `unique_key` `xxxxxxxx` by a random `unique key` depending on `worker_name`.

Making the `unique_key` equal to `xxxxxxxx` if it was missing, messes up the residual calculation for old tests.

After this replacement we can retire `unique_key` in `utils.py`.

2) We replace the fields in `run["args"]` which are "?" by "". This corresponds to the defaults in `new_run()`.

3) Set the flags `finished`, `failed`, `deleted`, `is_green`, `is_yellow`, if they are missing.

I see no issues running this against a live system. It changes the `runs` collection in place.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -168,7 +168,7 @@ class RunDb:
         # administrative flags
 
         # "finished"
-        # set in stop_run(), /api/stop_run
+        # set in stop_run(), /api/stop_run, /tests/delete
         # cleared in purge_run(), /tests/modify
         new_run["finished"] = False
 
--- a/server/utils/upgrade.py
+++ b/server/utils/upgrade.py
@@ -1,18 +1,15 @@
-import copy
-import datetime
-import math
-import os
-import pprint
+import copy, datetime, math, pprint, os, uuid
 
 import pymongo
 from fishtest.stats import stat_util
-from fishtest.util import format_results
+from fishtest.util import format_results, worker_name
 
 
 def show(p):
     pprint.pprint(p)
 
 
+# for documentation
 run_default = {
     "_id": "?",
     "args": {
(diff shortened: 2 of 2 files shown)


ppigazzini:
Ran on DEV
```bash
$ time sudo bash mongodb_runs_upgrade.sh
stop cron and fishtest ...
upgrade runs collection ...
Starting conversion...
Runs converted: 115476.
Conversion finished in 2282.86 seconds. Time per run: 19.77ms.
start fishtest and cron ...

real    38m10.522s
user    26m2.097s
sys     1m32.889s
```

ppigazzini:
Script ran on DEV with a live system (few workers):
```
$ ${VENV}/bin/python3 fishtest/server/utils/upgrade.py
Starting conversion...
Runs converted: 115648.
Conversion finished in 2429.93 seconds. Time per run: 21.01ms.
```

vdbergh:
Unfortunately there was a bug (only affecting very old tests). I updated `run` instead of `run["args"]`. This is fixed in the latest push. It needs to be run on a fresh copy of the db.

ppigazzini:
Started the upgrade of a fresh db on a live system.

ppigazzini:
ISP connection dropped :(
Reloaded a fresh db, just started the upgrade script using `tmux`

vdbergh:
Not sure now if PROD would be able to handle it. The duration of api calls becomes quite long. I think if we want to do it live then we have to build in some delays.

ppigazzini:
IMO the simple solution is to stop fishtest server on PROD for 40 minutes.

vdbergh:
> IMO the simple solution is to stop fishtest server on PROD for 40 minutes.

Maybe that's indeed the easiest.

ppigazzini:
`runs` collection upgraded, thnak you @vdbergh :)
```
$ time sudo bash mongodb_runs_upgrade.sh | tee mongodb_runs_upgrade.sh.log
stop cron and fishtest ...
upgrade runs collection ...
Starting conversion...
Runs converted: 115694.
Conversion finished in 1802.04 seconds. Time per run: 15.58ms.
start fishtest and cron ...

real    30m13.392s
user    20m17.968s
sys     1m0.837s
```

PR #1235, opened 2022-01-08
Trivial cleanups2

vdbergh:
Remove glue code at various places.

It should only make a difference for very old tests. So I have not been able to test this.

PR #1236, opened 2022-01-11
List the pentanomial draw ratio on the raw stats page.

vdbergh:
The pentanomial draw ratio is the ratio of `DD+WL+LW` results (i.e. non-decisive game pairs).

When the probability of double wins/losses is low then normalized Elo may be approximated as
```
nElo = sqrt(2)*Elo/sqrt(1-d)
```
Here `Elo` denotes `logistic Elo` and `d` is the pentanomial draw ratio.


ppigazzini (approved):
DEV updated (with also #1235), looks good to me.

PR #1239, opened 2022-01-12
Make `run_table.mak` more parametrizable.

vdbergh:
If we want to implement a general query system then we need a way to present the results. `run_table.mak` almost fits the bill but it is not very self contained. In this PR we fix this.

As an example we make all sections on the first page collapsible. This is now trivial to do.

Also:

Delete obsolete copy of `jquery.cookie.js` and get it from a cdn instead. Added the appropriate SRI hash.

Use `slideToggle()` also for the machines table.

Remove some inline svg and create images instead.


vdbergh:
E.g. the "Finished" section on the first page is not created as follows.
```python
<%include file="run_table.mak" args="runs=finished_runs,
                                     header='Finished',
                                     count= num_finished_runs,
                                     toggle = 'finished' if page_idx==0 else None,
                                     pages=finished_runs_pages"
/>
```

ppigazzini:
DEV updated, new tests queued, workers joined.

worker/games.py
```diff
@@ -396,7 +396,7 @@ def verify_signature(engine, signature, active_cores):
 
         if active_cores > 1:
             busy_process.communicate("quit\n")
-
+    p.returncode=1
```
vondele, line 399: is this an intended change (the if's after this look strange).

vdbergh: Oops. Debug (to generate an `api/stop_run`).

ppigazzini:
DEV updated.

vdbergh:
Hmm the buttons on the user page are not persistent in Chrome (Firefox seems fine). Chrome may sent the same cookie multiple times, the one that applies is the first one (IMHO this is silly but that's the way it is). Unfortunately `pyramid.request` picks the last one...   So the browser and the server disagree about which cookie applies.

The only solution seems to be to parse `pyramid.request.headers` directly.

Another solution would be to avoid using cookies with the same name for different url's. I am thinking about an easy mechanism to do this.

vdbergh:
I solved in by parsing  `pyramid.request.headers` directly.

ppigazzini:
DEV updated.

ppigazzini:
Looks good to me.

vdbergh:
Yes it appears to behave as one would expect.

vondele:
this looks good to me, and it would be cool to be able to query/filter the list (e.g. replacing the links on the left with checkboxes on the main page or so).

dav1312:
We can also use the already imported font-awesome icons instead of the old ones
image
Trash: `<i class="fas fa-trash"></i>`
Trash alt: `<i class="fas fa-trash-alt"></i>`
Thumbs Up (Solid): `<i class="fas fa-thumbs-up"></i>`
Thumbs Up (Regular): `<i class="far fa-thumbs-up"></i>`

---
image
Maybe also changing the warning background color and icon color so it's more visible
image
Question Mark: `<i class="fas fa-question"></i>`

vondele:
admittedly, I don't know what the thumbs-up and question marks really mean. Do they still make sense / have a function with the current setup? If not, I would remove them.

ppigazzini:
"question mark" - run to be approved
"thumb up" - run approved

With the "Pending approval" section now they should be useless.

vondele:
In that case, I think I would just remove them.

ppigazzini:

vondele:
one thing that has changed (and I did not notice on dev) is that we're back to the double scroll bar for active tests. That's not so nice. I would prefer no to have the scroll bar for the active tests, and things that are paginated already (like the finished test).

dav1312:
> I merge this PR. @dav1312 at you convenient time please open a new PR to use the new font awesome and to try to drop the allegedly useless icons :)

I don't understand which ones should be changed and which ones should be removed.
The "paused" ones changed and the "pending approval" ones removed? 🤔

vondele:
> > I merge this PR. @dav1312 at you convenient time please open a new PR to use the new font awesome and to try to drop the allegedly useless icons :)
> 
> I don't understand which ones should be changed and which ones should be removed. The "paused" ones changed and the "pending approval" ones removed? thinking

I think there is no need for the thumbs-up and the question mark symbols, they are redundant with the section where the tests appear.

ppigazzini:
We should fix also the white box size for the "Confirm" button :)

image

vdbergh:
> one thing that has changed (and I did not notice on dev) is that we're back to the double scroll bar for active tests. That's not so nice. I would prefer no to have the scroll bar for the active tests, and things that are paginated already (like the finished test).

Perhaps we should remove all double scroll bars? They seem redundant now that all sections are collapsible.

vondele:
that is reasonable as well

PR #1240, opened 2022-01-13
[RFC] Experiment. Remove all double scrollbars.

vdbergh:


ppigazzini:
DEV updated, it works on "Finished"

ppigazzini:
PROD updated by hand (not easy to test this on DEV).

vondele:
I think this looks good.

vondele:
looks like the double scroll bars are back on prod, definitely not better...

ppigazzini (approved):

dav1312:
> looks like the double scroll bars are back on prod, definitely not better...

I don't have double scroll bars on prod 🤔

ppigazzini:
Dropped line 63 by hand just now.

ppigazzini:

ppigazzini:
BTW I dropped line 63 instead of comment out it.

PR #1244, opened 2022-01-14
Try once again to remove the hacks to count finished runs.

vdbergh:


ppigazzini:
DEV updated, LGTM.

vdbergh:
> DEV updated, LGTM.

Yes I think so too.

This will greatly simplify creating an interface for querying the db (like openbench has).

ppigazzini (approved):

ppigazzini:

Issue #684, opened 2020-05-23
macOS and qt-static cutechess-cli


d3vv:
Under macOS we always have:

`Downloading cutechess-cli-linux-64bit.zip ...`

Could it possible to include my qt5-static cutechess-cli fresh build into fishtest distro?

```
dew@MacBook-Pro cutechess % uname -a
Darwin MacBook-Pro 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
dew@MacBook-Pro cutechess % projects/cli/cutechess-cli -version
cutechess-cli 1.1.0
Using Qt version 5.11.1

Copyright (C) 2008-2018 Ilari Pihlajisto and Arto Jonsson
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

dew@MacBook-Pro cutechess % otool -L projects/cli/cutechess-cli
projects/cli/cutechess-cli:
	/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1675.129.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 1069.22.0)
	/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit (compatibility version 45.0.0, current version 1894.40.150)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59306.101.1)
	/System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices (compatibility version 1.0.0, current version 52.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1675.129.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
[...]

PR #1245, opened 2022-01-15
add_clang_and_m1_support

RaimundHeid:
Additions to support clang and apple-silicon (which is far more efficient on a M1 hardware (factor 1.7 vs. x86-64 arch)).

Tested on Mac with Intel and M1 cpu and on Windows 10 with Intel 64-bit cpu.

In order to make things perfect for Mac users a cutechess-cli with static linked qt5 would be fine. I can provide you with a binary that runs on Intel and M1 macs. If you upload it I can make the changes in games.py to check and download it to the testing dir if it is missing. Please drop me a note.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -444,8 +444,59 @@ def download_from_github(item, testing_dir):
         raise WorkerException("Item {} not found".format(item))
 
 
+def cc_is_clang():
+    """Parse the output of g++ -E -dM -"""
+    with subprocess.Popen(
+        ["g++", "-E", "-dM", "-"],
+        stdin=subprocess.DEVNULL,
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+        bufsize=1,
+        close_fds=not IS_WINDOWS,
+    ) as p:
+        clang = "None"
+        for line in iter(p.stdout.readline, ""):
+            if "__clang__" in line:
+                clang = line.split()[2]
+
+    return clang == "1"
+
+
+def clang_props():
+    """Parse the output of g++ -E -dM - and extract the available clang properties"""
+    with subprocess.Popen(
+        ["g++", "-E", "-dM", "-"],
+        stdin=subprocess.DEVNULL,
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+        bufsize=1,
+        close_fds=not IS_WINDOWS,
+    ) as p:
+        flags = []
+        arch = "None"
+        version = "None"
+        arm64 = "None"
+        for line in iter(p.stdout.readline, ""):
+            if "__VERSION__" in line:
+                version = line.split()[2]
+            if "__arm64__" in line:
+                arm64 = line.split()[2]
+
+    if p.returncode != 0:
+        raise FatalException(
+            "g++ target query failed with return code {}".format(
+                format_return_code(p.returncode)
+            )
+        )
+
+    if "Apple" in version and arm64 == "1":
+        arch = "apple-silicon"
+
+    return {"flags": flags, "arch": arch}
+
+
 def gcc_props():
-    """Parse the output of g++ -Q -march=native --help=target and extract the available properties"""
+    """Parse the output of g++ -Q -march=native --help=target and extract the available gcc properties"""
     with subprocess.Popen(
         ["g++", "-Q", "-march=native", "--help=target"],
         stdout=subprocess.PIPE,
@@ -469,6 +520,15 @@ def gcc_props():
         )
 
     return {"flags": flags, "arch": arch}
+
+
+def cc_props():
+    if cc_is_clang():
+        global MAKE_CMD
+        MAKE_CMD = "make COMP=clang "
+        return clang_props()
+    else:
+        return gcc_props()
 
 
 def make_targets():
@@ -506,7 +566,7 @@ def find_arch():
 
     targets = make_targets()
 
-    props = gcc_props()
+    props = cc_props()
 
     if is_64bit():
         if (
@@ -550,7 +610,10 @@ def find_arch():
         ):
             arch = "x86-64-sse3-popcnt"
         else:
-            arch = "x86-64"
+            if props["arch"] in targets:
+                arch = props["arch"]
+            else:
+                arch = "x86-64"
     else:
         if (
             "-mpopcnt" in props["flags"]
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -518,6 +518,7 @@ def gcc_version():
         bufsize=1,
         close_fds=not IS_WINDOWS,
     ) as p:
+        clang_major = "None"
         for line in iter(p.stdout.readline, ""):
             if "__GNUC__" in line:
                 major = line.split()[2]
@@ -525,15 +526,26 @@ def gcc_version():
                 minor = line.split()[2]
             if "__GNUC_PATCHLEVEL__" in line:
                 patchlevel = line.split()[2]
+            if "__clang_major__" in line:
+                clang_major = line.split()[2]
+            if "__clang_minor__" in line:
+                clang_minor = line.split()[2]
+            if "__clang_patchlevel__" in line:
+                clang_patchlevel = line.split()[2]
 
     if p.returncode != 0:
         print("g++ version query failed with return code {}".format(p.returncode))
         return None
 
     try:
-        major = int(major)
-        minor = int(minor)
-        patchlevel = int(patchlevel)
+        if clang_major == "None":
+            major = int(major)
+            minor = int(minor)
+            patchlevel = int(patchlevel)
+        else:
+            major = int(clang_major)
+            minor = int(clang_minor)
+            patchlevel = int(clang_patchlevel)
     except:
         print("Failed to parse g++ version.")
         return None


ppigazzini:
To upload the `cutechess-cli` binaries open the PR to this repo:
https://github.com/official-stockfish/books

I suggest to check the binaries with VirusTotal, see:
https://github.com/official-stockfish/books/commit/b5b41bc3a6a47a0c895c6fba4d2a35187a71d6e6

If I'm not asking too much it will be wonderful to have also script/instruction on the wiki page:
https://github.com/glinscott/fishtest/wiki/Build-cutechess-with-Qt5-static

worker/worker.py
```diff
@@ -518,22 +518,34 @@ def gcc_version():
         bufsize=1,
         close_fds=not IS_WINDOWS,
     ) as p:
+        clang_major = "None"
```
ppigazzini, line 521: If I'm not wrong this could be simplified:
```python
            if "__GNUC__" in line or "__clang_major__" in line:
                major = line.split()[2]
            if "__GNUC_MINOR__" in line or "__clang_minor__" in line:
                minor = line.split()[2]
            if "__GNUC_PATCHLEVEL__" in line or "__clang_patchlevel__" in line:
                patchlevel = line.split()[2]

    if p.returncode != 0:
        print("g++ version query failed with return code {}".format(p.returncode))
        return None

    try:
        major = int(major)
        minor = int(minor)
        patchlevel = int(patchlevel)
```

ppigazzini:
I'm running a worker on DEV with msys2/clang after changing this:
```python
MAKE_CMD = "make COMP=clang " if IS_WINDOWS else "make COMP=gcc "
```
https://github.com/glinscott/fishtest/blob/e145a5d693332ee59074a4b3402bc53209640270/worker/games.py#L73

worker/games.py
```diff
@@ -471,6 +522,13 @@ def gcc_props():
     return {"flags": flags, "arch": arch}
 
 
+def cc_props():
```
ppigazzini, line 525: Please format with space, not with tab.

ppigazzini:
```python
        if cc_is_clang():
            global MAKE_CMD
            MAKE_CMD = "make COMP=clang "
            return clang_props()
        else:
            return gcc_props()
```

vdbergh:
After this PR we can put an extra field in `worker_info` with the compiler type (the version info is a bit meaningless without it).

ppigazzini:
If I-m not wrong in this way we should be able to check only once the compiler type.

vdbergh:
Yes. It does not have to be done for every task.

vdbergh:
Probably there should also be a worker option. If both compilers are available people may have a preference.

worker/worker.py
RaimundHeid: clang also defines __GNUC__, __GNUC_MINOR__ and __GNUC_PATCHLEVEL__, but with different values than in __clang_major__, ... . In the above suggestion the resulting values would depend on the order of the output, which is not a solid solution although it might work by random.

RaimundHeid: On my system the clang output of "g++ -E -dM -" is:

#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
...
#define __clang_major__ 12
#define __clang_minor__ 0
#define __clang_patchlevel__ 5
#define __clang_version__ "12.0.5 (clang-1205.0.22.9)"

ppigazzini: Oh, you are right, also clang/msys2 had the GNUC define
```
$ g++ -E -dM - < /dev/null | grep GNUC
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

worker/games.py
RaimundHeid: done.

RaimundHeid:
> To upload the `cutechess-cli` binaries open the PR to this repo: https://github.com/official-stockfish/books
> 
> I suggest to check the binaries with VirusTotal, see: [official-stockfish/books@b5b41bc](https://github.com/official-stockfish/books/commit/b5b41bc3a6a47a0c895c6fba4d2a35187a71d6e6)
> 
> If I'm not asking too much it will be wonderful to have also script/instruction on the wiki page: https://github.com/glinscott/fishtest/wiki/Build-cutechess-with-Qt5-static

I tried to build a cutechess-cli 1.2.0 with static qt5 for MacOs quite a while ago and gave up - shame on me ;) - when I found a binary ... somewhere. The TotalVirus check was Ok although and I will create a PR in the books repo. Will also try to build the bin on my own and will add instructions to the wiki page as soon as I succeed.

ppigazzini:
> I tried to build a cutechess-cli 1.2.0 with static qt5 for MacOs quite a while ago and gave up - shame on me ;) - when I found a binary ... somewhere. The TotalVirus check was Ok although and I will create a PR in the books repo. Will also try to build the bin on my own and will add instructions to the wiki page as soon as I succeed.

Try this instructions https://github.com/glinscott/fishtest/issues/684#issuecomment-633247654

RaimundHeid:
> > I tried to build a cutechess-cli 1.2.0 with static qt5 for MacOs quite a while ago and gave up - shame on me ;) - when I found a binary ... somewhere. The TotalVirus check was Ok although and I will create a PR in the books repo. Will also try to build the bin on my own and will add instructions to the wiki page as soon as I succeed.
> 
> Try this instructions [#684 (comment)](https://github.com/glinscott/fishtest/issues/684#issuecomment-633247654)

Well, nowadays it seems to be much more simple than way back when I tried it first. For me it worked out of the box following the instructions on https://github.com/cutechess/cutechess/wiki/Building-from-source. Referring to this page would be my recommendation for the fishtest wiki.

RaimundHeid:
Obviously this was just the easy part of the cutechess story since it only produces an app with dynamic linked qt5 by default :( . In order to get a static linked cutechess I had to build qt5 from the sources - which takes several hours on a reasonable fast machine. I will write down all the steps on the wiki page.

ppigazzini:
> Obviously this was just the easy part of the cutechess story since it only produces an app with dynamic linked qt5 by default :( . In order to get a static linked cutechess I had to build qt5 from the sources - which takes several hours on a reasonable fast machine. I will write down all the steps on the wiki page.

I invested some weeks of work to write the instructions for Linux and msys2 (actually this was fast) :)

RaimundHeid:
:) I peeked into your script for Linux and I assume a lot of it can be copied for MacOs. Scripting seems to be the best option for the documentation, so I will go this way. Moreover I hope the process will be much shorter by excluding some parts of qt5 that are not required for this rather small cutechess-cli ... :)

ppigazzini:
This is the same script for centos7, no need to skip the Qt version check but works only for cutechess-cli 64 bit (no 32 bit devtoolset)

```bash
#!/bin/bash
# sudo bash make_cute.sh
# https://wiki.qt.io/Building_Qt_5_from_Git
# http://doc.qt.io/qt-5/linux-deployment.html

n_jobs=6

# update and install packages
yum check-update
#yum install -y yum-utils
#yum-builddep -y qt5-qtbase

yum install -y centos-release-scl
yum install -y devtoolset-7
source scl_source enable devtoolset-7
yum install -y git
yum install -y libxcb libxcb-devel xcb-util xcb-util-devel mesa-libGL-devel libxkbcommon-devel

# build Qt5 static
git clone https://code.qt.io/qt/qt5.git
cd qt5
git checkout 5.15
perl init-repository --module-subset=default,-qtwebengine
mkdir /opt/qt5-static
./configure -static -release -opensource -confirm-license -prefix /opt/qt5-static -nomake examples -nomake tests -nomake tools -skip qtwebengine -no-icu
make -j ${n_jobs}
make -j ${n_jobs} install

# build cutechess-cli and cutechess
cd
git clone https://github.com/cutechess/cutechess.git
cd cutechess

# checkout the latest tag
git fetch -p --tags --all
tag=$(git describe --tags `git rev-list --tags --max-count=1`)
git checkout ${tag} -b latest

git log --oneline -n 5 > ../cutechess.git.log
/opt/qt5-static/bin/qmake -config release
make -j ${n_jobs}
strip projects/cli/cutechess-cli
strip projects/gui/cutechess
zip -j9 ../cutechess-cli-linux-$(getconf LONG_BIT)bit.zip COPYING projects/cli/cutechess-cli
```

ppigazzini:
BTW linux, WSL, msys2/mingw64 and msys2/clang workers run fine with this PR.

So when you think to have finished, please squash the commits.
```
git checkout master
git fetch -p upstream
git reset --hard upstream master
git checkout apple-silicon
git rebase -i master
# change all the "pick" to "squash" but the first one, save and exit
# write the final commit message, save and exit
git push -f origin apple-silicon
```

Let me know your name for AUTHORS (I suppose Raimund Heid) :)

Example of future TODO:
- `def clang_props():` should return the proper CPU ARCH
- add compiler in `worker_info`
- Stockfish Makefile changes to account clang on windows (the EXE variable is set only by mingw)

vondele:
I can also merge the pull request at the books  repo, but I would prefer if the binary is one built from sources (just to make sure we're not picking up something random).

RaimundHeid:
> I can also merge the pull request at the books repo, but I would prefer if the binary is one built from sources (just to make sure we're not picking up something random).

Fine by me. I wrote a script that currently runs on my MacBook Air 2018 - which is rather slow. But it will produce a binary that runs on Intel and ARM. Will update the PR as soon as it has finished.

ppigazzini:
PROD updated, workers update triggered, thank you @RaimundHeid :)

RaimundHeid:
You're welcome, @ppigazzini :)

Next tasks from my point of view:
1) provide cutechess-cli for MacOs (wip)
2) optimize builds for Apple Silicon + gcc

I assume I can use the current fork for those?

ppigazzini:
For 1. you need to fork https://github.com/official-stockfish/books , see my post https://github.com/glinscott/fishtest/pull/1245#issuecomment-1013695681

RaimundHeid:
> For 1. you need to fork https://github.com/official-stockfish/books , see my post [#1245 (comment)](https://github.com/glinscott/fishtest/pull/1245#issuecomment-1013695681)

ofc, PR already exists (but will be updated as soon as my build is finished). I was asking for the changes in games.py (which seem to be trivial - I hope the chmod will not cause any trouble in MacOs Monterey).

vondele:
the binary has been added to the book repo

PR #1248, opened 2022-01-16
Refactoring `purge_run` - again.

### LINKED ISSUE #1247: completed test not finished
This simpDraw test is not flagged as completed despite an LLR that is -3.81.

image

image

vdbergh:
This is an attempt to fix #1247 by reordering `stop_run()` and enhancing `purge_run()`.

We now let `stop_run()` completely stop the run and _then_ we invoke `purge_run()`, if `auto_purge` is enabled. `purge_run()`, which can now only be invoked on a finished run, will usually revive the run, _except_ when the test is an sprt test and after purging the LLR is outside the bounds (this part is new).

This PR is almost completely untested.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -55,7 +55,6 @@ class RunDb:
         global last_rundb
         last_rundb = self
 
-
     def new_run(
         self,
         base_tag,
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated.
- schedule NumGames test @ 0.05 to have time losses
- stop test
- purge test
- modify test

vdbergh:
This one has become a zombie

https://dfts-0.pigazzini.it/tests/view/61e408813115fe9c7c3ff999

It is listed on the main page as finished, but it isn't. This can be seen via

https://dfts-0.pigazzini.it/api/get_run/61e408813115fe9c7c3ff999

Yet it does not get any new workers.

Any it seems that we have to rebuild the results_info widget after purging. But there is something strange going on.

vdbergh:
Latest commit seems to cure the zombies at least.

vdbergh:
But I haven't been able to recreate the situation from #1247 yet (a purged SPRT test that after purging still has its LLR outside the bounds).

ppigazzini:
Too early :)
I modified the purged test to move it from the "finished" section like my previous test.
DEV updated only now.

ppigazzini:
LGTM:
- stop test
- purge test, the test goes directly in "paused" section

vdbergh:
Nonetheless I think I fixed a serious bug. I was modifying a run object without at least marking it dirty. This means it would hang around but not find its way into the db.

vdbergh:
> LGTM:
> 
> * stop test
> * purge test, the test goes directly in "paused" section
> * the test gets workers and goes in "active" section

Yes but there is a specific situation where the test should _not_ go in the paused section. This was the issue in #1247 . This is what this PR aims to solve.

This situation occurs when an SPRT test is purged, and after purging it is still in a decided state.
I have not been able to recreate this situation yet.

ppigazzini:
That was a SPRT test.

vdbergh:
> That was a SPRT test.

Yes but it went into the paused section, as you say. We should have a case where it should **not** go in the paused section (i.e. when the test is still finished after purging).

ppigazzini:
On DEV I scheduled only NumGames, so after a purge they cannot be still finished.

vdbergh:
I did create some SPRT tests. But I did not manage to create a situation where after purging a finished SPRT test, it was still in a decided state.

ppigazzini:
A strange behavior with this test
https://dfts-0.pigazzini.it/tests/view/61e4163572376dccd1bbb528
 - NumGames @ 0.025 with Autopurge
 - the workers seems not able to play the last 4 games task
 
image

vdbergh:
https://dfts-0.pigazzini.it/actions?count=1&before=1642338313.65

"past end time"

vdbergh:
It's strange that this would only occur for the last 4 games.

vdbergh:
The reason is that 4 games is just a very small amount. This must be an artifact of the very fast time control.

ppigazzini:
The workers have the sleep time before retrying.

ppigazzini:
Autopurge is working, so the problem was due to the small task and very low TC combination.

ppigazzini:
@vdbergh let me know when I can merge this (the commit perhaps not solve #1247 but it's anyway a bug fix).

vdbergh:
I squashed the PR. I am reasonably sure it actually solves #1247, but we'll see.

ppigazzini (approved):

ppigazzini:

We have a big problem, mongodb has raised HD space of 20 GB after the changes of the last weeks, nearly no space left on the VPS :(

vdbergh:
Is it the size of the runs collection?

ppigazzini:
No "pgns" collection:
```
ls -larSh
-rw------- 1 mongodb mongodb  18M Jan 16 15:18 index-9-5266237248956174028.wt
-rw------- 1 mongodb mongodb  25M Jan 16 15:08 WiredTigerHS.wt
-rw------- 1 mongodb mongodb  28M Jan 16 15:18 index-47--419891883545482193.wt
-rw------- 1 mongodb mongodb 779M Jan 16 15:16 collection-38--419891883545482193.wt
-rw------- 1 mongodb mongodb 2.8G Jan 16 15:18 collection-0--6161276535059129678.wt
-rw------- 1 mongodb mongodb  84G Jan 16 15:18 collection-46--419891883545482193.wt

```

ppigazzini:
I revert mongodb to standalone to compact the "pgns" collection, gained some only 2 GB :(

The script doesn't crash:
```bash
${VENV}/bin/python3 purge_pgn.py
PGN runs/tasks saved:      0/      0
PGN runs/tasks purged:     0/      0
```

vdbergh:
But we did not touch the pgns collection...

BTW. One can get the size of a collection in bytes using `db.<collection>.totalSize()` (after switching to the right db of course).

ppigazzini:
```
fishtest_new> db.pgns.totalSize()
8916044595247562752
fishtest_new> db.runs.totalSize()
29146562565967872
```

vdbergh:
These numbers cannot be right...

ppigazzini:
It seems right:
```
db.runCommand( { collStats : "pgns", scale: 1024 } )
{
  ns: 'fishtest_new.pgns',
  size: 84123805,
  count: 599703,
  avgObjSize: 143642,
  storageSize: 87072504,
  freeStorageSize: 561032,
  capped: false,
  wiredTiger: {
    metadata: { formatVersion: 1 }
```

ppigazzini:
The problem is that just after switching to mongodb 5 the pgns collection was 64 GB :(

vondele:
maybe the purging script of the STC pgns doesn't work after the change to the db?

ppigazzini:
Yes, the script doesn't work anymore, I set 1 day (default 2 dasy)
```python
def main():
    purge_pgn(days=1)
```

and no pgns dropped:

```bash
${VENV}/bin/python3 purge_pgn.py
PGN runs/tasks saved:      0/      0
PGN runs/tasks purged:     0/      0
```

Issue #775, opened 2020-08-24
Stockfish Testing after NNUE implementation


d3vv:
I always carefully install workers, check for free CPU/Memory resources, etc.
But after NNUE implementation I have some issues:

1. Some of my workers received "Red-Residual " for tests. So I just try to reduce "concurrency-cores"..

2. Most of the workers became slow - "This machine is too slow to run fishtest effectively - sorry!"

3. Some workers has "This machine is too slow to run fishtest effectively - sorry!" message when I'm using more then 50% of cpu-cores..

4. How to use an alternative compiler (clang for example), if  `custom_make.txt` is no longer used?

vondele:
all related to NNUE merge, and the fact that the implementation is sensitive to the hardware. However, some answers/comments:

4. no custom_make.txt is no longer supported, right now fishtest assumes gcc, in part to be able to autodetect which build is best. Very recently there was a suggestion we could do autodetection also for clang, but it would be some work to implement support for clang in the fishtest worker. The basic input is here: https://github.com/official-stockfish/Stockfish/pull/3045#issuecomment-679126992

3. If you use hyperthreaded workers, this might happen. We have plans to reduce the minimum nps for a worker. seehttps://github.com/glinscott/fishtest/pull/761 you can right now experiment with this using a manual change to the worker code.

2. See 3. Can you share which CPU model these are using (see `cat /proc/cpuinfo | grep 'model name' | uniq`)

1. The red residuals are currently ignored. Right now, they mean nothing more than 'different from most other machines on fishtest'.

d3vv:
@vondele 

1. Are You sure that it is a good practice?
Cause "red residuals"  signaled real problems to me earlier. For example about incorrect software/hardware behavior.

2. These are really very old processors like:
`model name      : Intel(R) Core(TM)2 Quad CPU    Q6600  @ 2.40GHz`
And I don't care about it then. But they worked without complaints before.

3. IMHO "a manual change to the worker code" it is wrong approach that can skew all tests.

4. Thank You for information.

vondele:
@d3vv well valid concerns of course. 
1) Previously we were able to more or less deal with the hardware differences, now it is much more difficult. It should become less of an issue as soon as we stop regression test against SF11. But that's what the chi2 test is measuring: are these machines equivalent. Right now, the answer is no. We can still manually purge tests, it is just not automatic anymore. T
[...]

PR #1251, opened 2022-01-17
download missing cutechess-cli for macos

RaimundHeid:
This seems to work without further ado:

...
Running simpLimitExtension vs master
Variable task sizes used. Opening offset = 108600
Downloading cutechess-cli-macos-64bit.zip ...
Obtaining version info for /Users/raimundheid/Applications/forked_fishtest/worker/testing/cutechess-cli ...
Found:  cutechess-cli 1.2.0
...

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -25,6 +25,7 @@ from zipfile import ZipFile
 import requests
 
 IS_WINDOWS = "windows" in platform.system().lower()
+IS_MACOS = "darwin" in platform.system().lower()
 LOGFILE = "api.log"
 
 LOG_LOCK = threading.Lock()
@@ -1167,6 +1168,8 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
     if not os.path.exists(cutechess):
         if len(EXE_SUFFIX) > 0:
             zipball = "cutechess-cli-win.zip"
+        elif IS_MACOS:
+            zipball = "cutechess-cli-macos-64bit.zip"
         else:
             zipball = "cutechess-cli-linux-{}.zip".format(platform.architecture()[0])
         download_from_github(zipball, testing_dir)


ppigazzini (approved):
Workers run fine with linux, msys2, WSL and the patch looks good to me.

ppigazzini:
PROD updated to trigger the workers update.
Thank you @RaimundHeid :)

vondele:
so, with this patch, the worker downloaded cutechess-cli and runs the worker on a macbook pro (so not m1), compiles with clang++, but doesn't pick the right arch (just a generic x86-64).

ppigazzini:
Yes, this needs the proper parsing:
https://github.com/glinscott/fishtest/blob/9c5a80d2ed7c363a60f9358c70c3ac875f9728d5/worker/games.py#L466-L467

RaimundHeid:
> so, with this patch, the worker downloaded cutechess-cli and runs the worker on a macbook pro (so not m1), compiles with clang++, but doesn't pick the right arch (just a generic x86-64).

You are right, although I assume Mx support for g++-users if of higher practical relevance. Will tackle both issues ofc.

Maybe we will have to get the clang props like it is described in https://stackoverflow.com/questions/35357659/what-flags-does-march-native-activate-with-clang - will investigate.

ppigazzini:
Check #775 where that way to find GCC and clang props was suggested.

ppigazzini:
Only to point out that `msys2` and `cygwin` have `/proc/cpuinfo`, so the code could be:
```python
def cpu_props():
    p = subprocess.Popen(
        ["cat", "/proc/cpuinfo"],
        stdout=subprocess.PIPE,
        universal_newlines=True,
        bufsize=1,
        close_fds=not IS_WINDOWS,
    )

        if "flags" in line:
                return {"flags": line.split()[2:]}
```

Here is the ouput:
```
new:
{'flags': ['fpu', 'vme', 'de', 'pse', 'tsc', 'msr', 'pae', 'mce', 'cx8', 'apic', 'sep', 'mtrr', 'pge', 'mca', 'cmov', 'pat', 'pse36', 'clflush', 'acpi', 'mmx', 'fxsr', 'sse', 'sse2', 'ss', 'ht', 'tm', 'pbe', 'constant_tsc', 'arch_perfmon', 'rep_good', 'nopl', 'xtopology', 'cpuid', 'aperfmperf', 'pni', 'pclmuldq', 'est', 'tm2', 'ssse3', 'cx16', 'xtpr', 'pdcm', 'pcid', 'sse4_1', 'sse4_2', 'popcnt', 'aes', 'xsave', 'avx', 'f16c', 'rdrand', 'hypervisor', 'lahf_lm', 'epb', 'fsgsbase', 'smep', 'erms', 'xsaveopt', 'dtherm', 'ida', 'pln', 'pts', 'md_clear', 'flush_l1d', 'arch_capabilities']}

master:
{'flags': ['-m128bit-long-double', '-m64', '-m80387', '-maccumulate-outgoing-args', '-maes', '-malign-double', '-malign-stringops', '-mavx', '-mavx256-split-unaligned-load', '-mavx256-split-unaligned-store', '-mcx16', '-mf16c', '-mfancy-math-387', '-mfentry', '-mfp-ret-in-387', '-mfsgsbase', '-mfxsr', '-mhard-float', '-mieee-fp', '-mlong-double-80', '-mmmx', '-mms-bitfields', '-mno-align-vector-insn', '-mpclmul', '-mpe-aligned-commons', '-mpopcnt', '-mpush-args', '-mrdrnd', '-mred-zone', '-msahf', '-msse', '-msse2', '-msse3', '-msse4', '-msse4.1', '-msse4.2', '-mssse3', '-mstack-arg-probe', '-mstackrealign', '-mvzeroupper', '-mxsave', '-mxsaveopt'], 'arch': 'ivybridge'}
```

ppigazzini:
@vondele the CPU flags could be parsed in the Makefile.

PR #1252, opened 2022-01-18
Fix issue with usernames containing special characters.

vdbergh:
The current code generates incorrect html for "Unknown user".


ppigazzini:

PR #1253, opened 2022-01-19
support apple-silicon when COMP=gcc

RaimundHeid:
Tested on Mac with M1 and gcc 11.2.0.

   IS_MACOS and props["arch"].startswith("armv8") 

would be a bit speculative but is worth to be considered.


ppigazzini (approved):
Looks good to me, the workers are running fine on DEV with this PR.

ppigazzini:
PROD updated to trigger the workers update.
Thank you @RaimundHeid :)

PR #1258, opened 2022-01-21
Worker code simplication using the compiler info

ppigazzini:


ppigazzini:
Tested that Linux requires `shell=True` to find `make`.

ppigazzini:
Ops, forgot to drop `cc_is_clang()`, I'll do when at the keyboard.

PR #1262, opened 2022-01-23
Refactor option handling a bit.

vdbergh:
This should be a non-functional change.


ppigazzini (approved):
Tested with several workers OS/compiler combinations, LGTM.

ppigazzini:
All workers are running fine, thank you @vdbergh :)

Issue #1265, opened 2022-01-29
[feature] Add ability to setup CPU cores as a percentage of total (i.e. 50pct)


Technologov:
[fishtest-worker-CPU-pct.patch.txt](https://github.com/glinscott/fishtest/files/7962859/fishtest-worker-CPU-pct.patch.txt)

This is particulary useful for cloud usage, where one user could have a lot of different servers to manage with different capabilities.

vdbergh:
A slightly more generic solution would be to allow expressions. E.g. `max-1`. For this one needs an expression parser. This one seems to work https://github.com/lhelwerd/expression-parser. This could also be used for memory. E.g. `max-memory=max/2`.

vdbergh:
With #1266 one could write `--concurrency=0.75*max`. But also for example `--concurrency=max-3`.

ppigazzini:
@Technologov done, you can use elaborated expression for concurrency and memory like:
```bash
env/bin/python3 worker.py --concurrency=max*0.75+1 --max_memory=max*0.53+1024
```

PR #1266, opened 2022-01-29
Allow expressions involving `max` for `--concurrency` and `--max_memory`.

vdbergh:
This is a more generic solution for #1265.
   
Examples:
```    
    --concurrency max-1            # leave 1 core for the OS
    --concurrency min(4,max/2)     # reserve half of the cores for
                                   # Fishtest with a maximum of 4
```    
We compute the defaults for `concurrency` and `max_memory` in the configfile.
    
A freshly minted configfile now contains
```    
    concurrency = 1 if max==1 else min(3, max-1)
    max_memory = max/2
```
If the expression for `concurrency` does not contain "max", `concurrency` is limited to `max_cpu_count-1`. Otherwise it is limited to `max_cpu_count`.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -23,13 +23,17 @@ from functools import partial
 
 # Try to import an user installed package,
 # fall back to the local one in case of error.
+
+packages_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "packages")
 try:
     import requests
 except ImportError:
-    sys.path.append(
-        os.path.join(os.path.dirname(os.path.realpath(__file__)), "packages")
-    )
+    sys.path.append(packages_dir)
     import requests
+
+if packages_dir not in sys.path:
+    sys.path.append(packages_dir)
+import expression
 
 from games import (
     FatalException,
@@ -117,38 +121,6 @@ def _bool(x):
 
 
 _bool.__name__ = "bool"
-
-
-def _cores(max_cpu_count, x):
-    x = x.strip()
-    if x == "max":
-        return max_cpu_count
-    try:
-        x = int(x)
-    except ValueError:
-        raise ValueError(x)
-
-    if x <= 0:
-        raise ValueError(x)
-
-    if x >= max_cpu_count:
-        print(
-            (
-                "\nYou cannot have concurrency {} but at most:\n"
-                "  a) {} with '--concurrency max';\n"
-                "  b) {} otherwise.\n"
-                "Please use option a) only if your computer is very lightly loaded.\n"
-            ).format(
-                x,
-                max_cpu_count,
-                max_cpu_count - 1,
-            )
-        )
-        raise ValueError(x)
-    return x
-
-
-_cores.__name__ = "concurrency"
 
 
 def safe_sleep(f):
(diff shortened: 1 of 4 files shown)


vdbergh:
I am now also computing the default for `concurrency` in the configfile:
```
concurrency = 1 if max==1 else min(3, max-1)
```
We already had
```
max_memory = max/2
```

ppigazzini:
I'm running a worker on DEV with `--max_memory="max*0"`

image

vdbergh:
You could always set `--max_memory` to zero I think. But then you are not supposed to get any tasks.  This looks like a server bug.

vdbergh:
Apparently `max_memory <= 0` means no limit.

vdbergh:
Yes `--max_memory 0` means no limit.

```python
            # We check if the worker has reserved enough memory                                                                              
            need_tt = 0
            need_base = 0

            if max_memory > 0:

                def get_hash(s):
                    h = re.search("Hash=([0-9]+)", s)
                    if h:
                        return int(h.group(1))
                    return 0

                need_tt += get_hash(run["args"]["new_options"])
                need_tt += get_hash(run["args"]["base_options"])
                need_tt *= max_threads // run["args"]["threads"]
                # estime another 70MB per process for net (40) and other things besides hash                                                 
                need_base = 2 * 70 * (max_threads // run["args"]["threads"])

            if need_base + need_tt > max_memory:
                continue
```

The condition `if max_memory > 0` is wrong.

worker/worker.py
```diff
@@ -23,14 +23,17 @@
...
 try:
     import requests
 except ImportError:
-    sys.path.append(
-        os.path.join(os.path.dirname(os.path.realpath(__file__)), "packages")
-    )
+    sys.path.append(packages)
     import requests
 
+sys.path.append(packages)
```
ppigazzini, line 34: Nitpicking :)

```python
if packages not in sys.path:
    sys.path.append(packages)
```

I will prefer `packages_dir`

vdbergh: I changed it. 

I wonder which of these packages are still being used...

ppigazzini: These all the packages required by `requests`, collected installing `requests` in a clean env.

ppigazzini: See https://github.com/glinscott/fishtest/commit/7014a90c37806f282cc9046a1e0eafa4fe7b556d

vdbergh: Ok makes sense.

vdbergh: I missed part of your suggestion, so I updated again.

worker/worker.py
```diff
@@ -461,6 +413,45 @@ def my_error(e):
...
+                max_cpu_count - 1,
+            )
+        )
+        return None
+
+    e = expression.Expression_Parser(
+        variables={"max": mem / 1024 / 1024}, functions={"min": min}
+    )
+    try:
+        options.max_memory = round(e.parse(options.max_memory_))
```
ppigazzini, line 450: Just to avoid to report a worker with TB of ram :)
```python
        options.max_memory = round(
            max(min(e.parse(options.max_memory_), mem / 1024 / 1024), 0)
        )
```

vdbergh: Changed!

ppigazzini (approved):

ppigazzini:

ppigazzini:
wiki updated
https://github.com/glinscott/fishtest/wiki/Running-the-worker%3A-overview#worker-parameters

PR #1267, opened 2022-01-30
Fix bug with `--max_memory 0`.

vdbergh:


ppigazzini (approved):
Tested good with DEV

ppigazzini:

PR #1269, opened 2022-02-01
Earlier validation of `--concurrency` and `--max_memory`.

vdbergh:
The principle is that a broken config file should not prevent a worker to start, since errors in the config file are not obvious to diagnose. This principle was no longer valid, but this PR restores it.

Also:
    
Put actual values for concurrency, max_memory as comments in config file.

It now looks like:
```
concurrency = MAX-2 ; = 2 cores
max_memory = MAX/4 ; = 2849 MiB
```

Also:
    
Rename max to MAX and inject the function max in the parser.
    
As a temporary measure, for backward compatibility "max", when used as a variable, is automatically converted to "MAX", both in the config file and on the command line. When this happens a deprecation warning is given.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -122,6 +122,76 @@ def _bool(x):
 _bool.__name__ = "bool"
 
 
+# For backward compatibility.
+# Will be removed.
+def max_to_MAX(x, quiet=True):
+    y = re.sub(r"max(?=[\s]*([^(\s]|$))", r"MAX", x, count=0)
+    if y != x and not quiet:
+        print(
+            "\n"
+            '   WARNING: the use of "max" as a variable is deprecated\n'
+            '   and support for it will be removed. Use "MAX" instead.\n'
+        )
+    return y
+
+
+class _memory:
+    def __init__(self, MAX):
+        self.MAX = MAX
+        self.__name__ = "memory"
+
+    def __call__(self, x):
+        x = max_to_MAX(x, quiet=False)  # for backward compatibility
+        e = expression.Expression_Parser(
+            variables={"MAX": self.MAX}, functions={"min": min, "max": max}
+        )
+        try:
+            ret = round(max(min(e.parse(x), self.MAX), 0))
+        except:
+            print("Unable to parse expression for max_memory")
+            raise ValueError(x)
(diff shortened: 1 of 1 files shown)


ppigazzini:
Wow, I was still testing the 2 loc change :)

ppigazzini:
What about a breaking change switching to MAX?
```
concurrency = -1 * min(-1 * max/4, -1 * (max/3+1)) if max/3 > 2 else 8 ; = 4 cores
max_memory = (min(max > 1, max < 1 > 1) < max) * 1024 ; = 1024 MiB
```

vdbergh:
You mean replacing the variable “max” by “MAX” and then injecting the _function_ max in the parser?

I would be in favour of that. However it breaks backward compatibility.

ppigazzini:
IMO using MAX the expression will be more readable in documentation, configuration file etc.
The `max()` function will be useful like `min()` and the pattern `-1 * min(-1 * value0, -1 * value1)` it's a bit ugly.

@vondele let us know your opinion about the loss of backward compatibility.

ppigazzini:
BTW I was not able to find a bug in this PR testing with my workers, code still not reviewed though.

vdbergh:
Proactive commit. It looks indeed much nicer.

vdbergh:
With the last commit the backward compatibility problem should be solved.

vdbergh:
It can be done better. Tomorrow.

vdbergh:
I am now quite happy with the state of things. The config file will be silently converted from "max" to "MAX" and "max" on the command line will still work but a deprecation warning is given.

ppigazzini (approved):
LGTM, tested starting:
```
env/bin/python3 worker.py --concurrency="min(max(max/2, max/4), max/3)" --max_memory="min(max(max/2, max/4), max/3)"
```
and with a `fishtest.cfg` with:
```
concurrency = min(max(max/2, max/4), max/3) ; = 3 cores
max_memory = min(max(max/2, max/4), max/3) ; = 10816 MiB
```

ppigazzini:
Triggered workers update, wiki updated.
Thank you @vdbergh :)

PR #1270, opened 2022-02-05
Some cosmetics in option handling.

vdbergh:
1) Show a message when a config option is preprocessed
(e.g. `max->MAX`).

2) Show defaults in the help message.

3) Correct a comment concerning the use of options from the config file as defaults.

4) Bugfix: enable comments when the config file is initialized because of a syntax error.

5) Tweak several help messages.

6) Hide the default values for flags in the help messages.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -16,7 +16,7 @@ import time
 import traceback
 import uuid
 import zlib
-from argparse import ArgumentParser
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
Tests with workers looks good.
EDIT_000: just discovered now `validate()` function.

ppigazzini:
Started workers update, thank you @vdbergh :)

PR #1271, opened 2022-02-08
Fix info not showing in smaller screens

peregrineshahin:
My first Pull Request ever :)

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -67,6 +67,7 @@ body {
   overflow: hidden scroll;
   height: 100vh;
   width: 100%;
+  display: inline-table;
 }
 
 .contentbase main > h2 {


ppigazzini:
[DEV](https://dfts-0.pigazzini.it) updated with this PR, LGTM :)

ppigazzini (approved):

ppigazzini:
Please let me know your real name if you want that too in the AUTHORS file.

vdbergh:
Hmm..  for me the info widget does not wrap (as it should IMHO). This PR creates horizontal scrolling instead in the Active tab (Linux Chrome on laptop) on a small screen.

The Finished tab behaves correctly though. Why are the Active tab and the Finished tab behaving differently...? They have the same source code (as one can see with view source).

vdbergh:
In Firefox the Active tab also behaves correctly (the info widget wraps and no horizontal scrolling).

vdbergh:
Perhaps this is the reason:

https://validator.w3.org/nu/?doc=https%3A%2F%2Ftests.stockfishchess.org%2Ftests

If there are errors in the html then one may expect unpredictable things to happen.

vdbergh:
Ok I think I found the real cause. Chrome does not break links. Some of the current info widgets contain links, preventing the info widgets from wrapping.

vdbergh:
See #1273

PR #1272, opened 2022-02-08
Style changes and W3

dav1312:
Using more Bootstrap classes
Changed pending users icon and text
Changed grid system in test view
Increased contrast ratio of green links in light theme

Added lang (en) attribute
td inside of tr

This also reverts #1271

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -4,21 +4,19 @@
     padding: 0;
   }
 
-  .mainnavbar .nav li {
-    margin-right: 4px;
-  }
-
   .mainnavbar .nav li a,
   .mainnavbar .nav li div {
     font-size: 0;
-    padding-top: 0;
+    padding: 0;
     padding-bottom: 0;
     line-height: 30px;
+    justify-content: center;
   }
 
   .mainnavbar .nav li a::before,
-  .mainnavbar .nav li div::before {
+  .mainnavbar .nav li#change-color-theme div::before {
     font-size: 14px;
+    margin-right: 0;
   }
 
   .mainnavbar {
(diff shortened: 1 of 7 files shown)


ppigazzini:
DEV updated

ppigazzini:
DEV updated, start commuting now.

ppigazzini:
DEV updated

dav1312:
Okay this should be all for this PR

ppigazzini:
PROD - DEV
Is it the intended behavior?

image

dav1312:
Yes, it is. Now the info cell has text wrapping instead of having a scrollbar for all the page and breaking the sidebar.

---

I have been testing some possible implementations to avoid the height problem.
I'll make a draft(?) pr for it soon after. It should be something like this when the screen size is very small:
image

ppigazzini:
Fell free to add any tests on DEV.
You can use your credentials or "user01/user01" "user00/user00" as username/password for normal user and approver.
I view that this solved the weird vertical scrollbar in Contributors for small windows (never noted the problem before).

dav1312:
Thinking about this... why not add a limit for the length of info? It it really necessary to have more than... 400-500 characters?

ppigazzini:
By default the Info is the the commit message, this test with the last SF commit has 5766 chars
https://dfts-0.pigazzini.it/tests/view/6206cab47ec248cffe7a8a60

I'm fine to limit the Info length, @vondele ?

vondele:
yes, reasonable to limit that.

ppigazzini:
@dav1312 please squash the commits when ready to be merged

ppigazzini:
The squash is a message to the maintainer the the wip is finished :)
I never tried the "Squash and merge" button, but in that case the commit message will be the cat of the single messages.

ppigazzini (approved):

ppigazzini:

PR #1273, opened 2022-02-08
Fix for non-wrapping info window.

vdbergh:


peregrineshahin:
I suggest adding `style= "width: max-content;"` to `<div id="active">` as it is still behaving differently from others(eg. finished) in small screens because the width is not calculated properly.

but I can't find the html file in the repo

ppigazzini:
Waitress creates the html pages from the mako templates:
https://github.com/glinscott/fishtest/tree/master/server/fishtest/templates

ppigazzini:
@vdbergh DEV updated, please check if it's OK.

ppigazzini:
If I'm not wrong this fix could be already done in this https://github.com/glinscott/fishtest/pull/1272

vdbergh:
I don’t have Chrome here now, but this page is a good example https://dfts-0.pigazzini.it/tests?page=5 .

It could indeed be fixed by #1272.  Better ask.

vdbergh:
I checked that https://dfts-0.pigazzini.it/tests?page=5 looks ok on Chrome even with a small window. One would have to compare without the PR.

vdbergh:
Actually https://dfts-0.pigazzini.it/tests?page=5 is not a good example since the URLs contain hyphens which allow line breaks anyway.

vdbergh:
To see the problem one needs to load a copy of the current db on PROD.

ppigazzini:
On Windows it seems that chrome wraps the links in the same way on DEV (with the PR) and on PROD.

image

vdbergh:
As I wrote https://dfts-0.pigazzini.it/tests?page=5 is _not a good example_ since the URLs contain hyphens which allow line breaks anyway. Sorry for first suggesting it.

On PROD https://tests.stockfishchess.org/tests/view/62025c6dbf46cb834cbdf533 has an info window `50% of https://tests.stockfishchess.org/tests/view/62007179bf46cb834cbda243`, which does not wrap.

ppigazzini:
In communication the negation is always dropped by default :)

vdbergh:
Please look at the Pending test on DEV. It is has a very long info window 
 AVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOWAVEERYLONGINFOWINDOW

On DEV it wraps properly. I created the same test on PROD (also in Pending) and there it does not wrap.

vdbergh:
I applied to safe fix to the info string on the test page. https://dfts-0.pigazzini.it/tests/view/620344d9402b7da689ecfe37 should now also display properly (not tested).

vdbergh:
Somebody approved the test with a very large info window on PROD :)  It is now under the "Paused" tab (I gave it priory -10).

vdbergh:
I still don’t  understand the point of this PR. By design(!) all tabs are generated from the same template so they behave identically. The reason why the Active tab _appeared_ to behave differently was the presence of very long textualized URLs in info fields which were not wrapped. This is fixed in #1273 .

ppigazzini:
> Somebody approved the test with a very large info window on PROD :) It is now under the "Paused" tab (I gave it priory -10).

Nice patch btw, it seems a sure big Elo gainer,  it should pass STC and LTC on the fly🤞

vdbergh:
Just for information: the version of the PR running on DEV is not the latest one. This page
is still broken on DEV: https://dfts-0.pigazzini.it/tests/view/620344d9402b7da689ecfe37 but it should now be fixed in the PR.

vdbergh:
I see that DEV has been updated and things work as intended! So the wrapping problem is solved.

ppigazzini (approved):

ppigazzini:

PR #1274, opened 2022-02-08
Further fix #active behaves differently from other tabs due to width not set.

peregrineshahin:
in smaller screens, width for #active is not calculated per percentages properly because there is no set width.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -70,6 +70,10 @@ body {
   display: inline-table;
 }
 
+#active {
+  width: max-content; 
+}
+
 .contentbase main > h2 {
   top: 0;
   background: var(--bs-body-bg);


vdbergh:
In what sense is `active` different from `finished`? They are generated from the same template.

peregrineshahin:
best practice is to always have wrappers around both left side and right side, but in this case only left side is set to minimum width, and right side width is set to 100%, once that happens, my guess is calculating width as percentage will not be easily inherited nor figured out by some browsers.

peregrineshahin:
and by width percentages I mean each <td> width percentage of the whole width of <tr>,
basically ,how much every single column takes as a percentage from the width of the full row.

peregrineshahin:
image

image

vdbergh:
I still don’t understand the point of this PR. By design(!) all tabs are generated from the same template so they behave identically. The reason why the Active tab appeared to behave differently was the presence of very long textualized URLs in info fields which were not wrapped. This is fixed in https://github.com/glinscott/fishtest/pull/1273 .

peregrineshahin:
The active tab is not different it just has more text (cores) which is showing the flow of not setting width for them all, they are not different it is that active has more text in the sprt column, width should be maxed out by template which generates them all but the problem only showing in active because it's showing more text.

The calculation of each td in a tr is made in design by percentages, which is not calculated correctly without maxing out the width.

vdbergh:
You added
```css
#active {
  width: max-content; 
}
```
According to https://developer.mozilla.org/en-US/docs/Web/CSS/max-content we have

> The max-content sizing keyword represents the intrinsic maximum width or height of the content. For text content this means that the content will not wrap at all even if it causes overflows.

It seems to me this is precisely the opposite of what one wants.

Anyway I object to the use of the css selector `#active`. The tabs should be identical. To apply styling one should add a class to the divs and add the appropriate css to that class (but I think this is unnecessary here).

PR #1276, opened 2022-02-09
Remember Show/Hide status for 10 years.

vdbergh:


ppigazzini:
The claim is misleading: leap years not taken in account :)

ppigazzini:
DEV updated.

ppigazzini (approved):
No way to test the 10 years claim because SSL is complaining the wrong date :)

PR #1277, opened 2022-02-12
Made some pages responsive

dav1312:
Also update and use more Bootstrap classes

This should make the live elo, SPRTcalculator and raw statistics pages be responsive (look good on mobile devices)

### DIFF
--- a/server/fishtest/static/css/live_elo.css
+++ b/server/fishtest/static/css/live_elo.css
@@ -26,3 +26,7 @@ a:hover {
 a:focus {
   color: #497449;
 }
+
+td:last-child {
+  word-break: break-word;
+}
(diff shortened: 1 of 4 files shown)


ppigazzini (approved):
Looks good to me (pc and mobile). I will merge to get feedbacks from the users.

ppigazzini:

vdbergh:
This PR seems to break the cursor positioning in the SPRT calculator (see screenshot). Strangely this is only on PROD and not on my local fishtest. I have no time to debug it now.

Screenshot from 2022-02-13 17-07-37

vdbergh:
This is how it should be

Screenshot from 2022-02-13 17-14-03

PR #1278, opened 2022-02-13
Fix cursor positioning in the SPRT calculator

dav1312:
Hotfix for https://github.com/glinscott/fishtest/issues/1277#issuecomment-1038223194

### DIFF
--- a/server/fishtest/static/html/SPRTcalculator.html
+++ b/server/fishtest/static/html/SPRTcalculator.html
@@ -73,7 +73,7 @@
           </div>
         </form>
         <hr />
-        <div id="mouse_screen" class="row">
+        <div id="mouse_screen" class="row g-0">
           <div id="pass_prob_chart_div" class="chart-div col-12 col-md-6"></div>
           <div id="expected_chart_div" class="chart-div col-12 col-md-6"></div>
         </div>


vdbergh:
Perhaps the issue is that the javascript code does not take the padding and border widths of the chart divs into account. Perhaps bootstrap sets these to non-zero values by default.

https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

ppigazzini:
DEV updated.
With Chrome on Windows I have the opposite result:
- PROD: point on the graph aligned with the cursor
- DEV: point on the graph not aligned with the cursor

dav1312:
> DEV updated. With Chrome on Windows I have the opposite result:
> 
> * PROD master: point on the graph aligned with the cursor
> * DEV PR: point on the graph not aligned with the cursor
> 
> With Edge on Windows instead PROD is not aligned and DEV is aligned.

I have the same result on both Chrome and Edge... try a force reload of the page with `Shift + F5`

ppigazzini:
Yes, the CRTL F5 changed the situation, I was editing my previous post.

ppigazzini:

vdbergh:
> Perhaps the issue is that the javascript code does not take the padding and border widths of the chart divs into account. Perhaps bootstrap sets these to non-zero values by default.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

Yes this seems to be the explanation. This is css added by bootstrap to the chart divs:
```css
{
    padding-right: calc(var(--bs-gutter-x) * .5);
    padding-left: calc(var(--bs-gutter-x) * .5);
}
```
Currently these expressions evaluate to zero, but probably this wasn't the case before this PR.

dav1312:
> > Perhaps the issue is that the javascript code does not take the padding and border widths of the chart divs into account. Perhaps bootstrap sets these to non-zero values by default.
> > [developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)
> 
> Yes this seems to be the explanation. This is css added by bootstrap to the chart divs:
> 
> ```css
> {
>     padding-right: calc(var(--bs-gutter-x) * .5);
>     padding-left: calc(var(--bs-gutter-x) * .5);
> }
> ```
> 
> Currently these expressions evaluate to zero, but probably this wasn't the case before this PR.

Yeah I basically removed all gutter with `g-0`

vdbergh:
There is still something strange. If I resize the browser window then for the chart divs to resize I have to reload the page (before this series of PRs the chart divs did not resize at all, they simply became stacked vertically).

Perhaps this is a javascript issue. I don't really understand how responsiveness works.

dav1312:
Yeah you need js to redraw the chart when that happens so it's out of the scope of this pr

vdbergh:
> Yeah you need js to redraw the chart when that happens so it's out of the scope of this pr

Well this PR messes up the layout when the browser window resizes (e.g. when rotating a phone or an iPad). So is to some extent a regression. I think the previous behaviour with the vertical stacking was preferable….

dav1312:
> > Yeah you need js to redraw the chart when that happens so it's out of the scope of this pr
> 
> Well this PR messes up the layout when the browser window resizes (e.g. when rotating a phone or an iPad). So it is to some extent a regression. I think the previous behaviour with the vertical stacking was preferable….

I'm no expert on js but ig adding something like this to `calc.js` should work?
```js
var resizeTimeout;
window.onresize = function() {
  clearTimeout(resizeTimeout);
  resizeTimeout = setTimeout(() => {draw_charts()}, 100);
};
```

PR #1282, opened 2022-02-15
Safety check when asking for new spsa parameters.

vdbergh:
It is possible that the task is finished (not tested).

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -981,6 +981,11 @@ def launch_cutechess(
         if "error" in req:
             raise WorkerException(req["error"])
 
+        if not req["task_alive"]:
+            # This task is no longer necessary
+            print("Server told us task is no longer needed")
+            return False
+
         result["spsa"] = {"num_games": games_to_play}
 
         w_params = req["w_params"]


worker/games.py
```diff
@@ -1117,7 +1122,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
     if "start" in task:
         print("Variable task sizes used. Opening offset = {}".format(opening_offset))
     start_game_index = opening_offset + input_total_games
-    run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2**30)
+    run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2 ** 30)
```
ppigazzini, line 1125: The last version of `black` switched from `x ** y` format to `x**y`.
Just as information, I usually auto format in the worker version raising commit.

vdbergh:
I tested it on DEV by stopping an SPSA test while the worker was still compiling. The output was as expected.
```
Info from remote: Request_spsa: task 620635f5d71106ed12a46cd0/294 is not active
Post request https://dfts-0.pigazzini.it:443/api/request_spsa handled in 108.94ms (server: 1.99ms)
Server told us task is no longer needed
Task exited
```

ppigazzini (approved):
Tested working also with windows worker.

ppigazzini:

PR #1283, opened 2022-02-16
Login and register redesign

dav1312:
This also makes those 2 pages responsive by adding the meta directly into them.
If the website someday becomes fully responsive, that should be removed.


ppigazzini:
DEV updating.

dav1312:
I might try a similar style for the "User Administration" tomorrow

ppigazzini (approved):

ppigazzini:
Looks very good, PROD updated, thank you @dav1312 :)

PR #1285, opened 2022-02-17
User Administration redesign

dav1312:
This also makes this page responsive by adding the meta directly into it.
If the website someday becomes fully responsive, that should be removed.


ppigazzini:
DEV updated, looks good to me.
I'm wondering if it's possible to add a comment for the section that could be removed in the future to have a a fully responsive website.

ppigazzini:

PR #1288, opened 2022-02-20
Add "Games pair ratio" and "Points ratio" to stats

ppigazzini:


server/fishtest/templates/tests_stats.mak
```diff
@@ -54,6 +54,16 @@
...
       results5 = run["results"]["pentanomial"]
       results5_ = fishtest.stats.LLRcalc.regularize(results5)
+      results5_pair_ratio = (
+          sum(results5[3:]) / sum(results5[0:2])
+          if sum(results5[0:2]) > 0
+          else float("nan")
+      )
+      results5_point_ratio = (
+          (2 * results5[4] + results5[3]) / (2 * results5[0] + results5[1])
+          if sum(results5[0:2])
```
vondele, line 64: can this be a division by 0?

vdbergh:
Hmmm... I am not really in favor of adding random metrics to the raw stats page.  I would like to keep it clean.

The raw stats page is for auditing the correctness of the statistical methods that are used in Fishtest. It is normally not for users.

I added the pentanomial draw ratio since it can be used to compute an easy approximation for normalized Elo in case of an unbalanced book (the approximation isn't very good it turns out, but it still gives a useful sanity check).

server/fishtest/templates/tests_stats.mak
ppigazzini: No, I used the ternary operator to set "nan", let me write on a single line:
```python
a = b / c if c else float("nan")
```

ppigazzini:
It's a time saver for the wiki Regression Tests page maintenance.
https://github.com/glinscott/fishtest/wiki/Regression-Tests

vondele:
I think it is a useful thing to have, even if not the key thing we optimize for, it is an interesting measure of strength.

vdbergh:
> It's a time saver for the wiki Regression Tests page maintenance. https://github.com/glinscott/fishtest/wiki/Regression-Tests

The regression tests page should be generated automatically...

server/fishtest/templates/tests_stats.mak
vondele: yeah, misread the code... the end index is not part of the array slice in python.

vdbergh:
> I think it is a useful thing to have, even if not the key thing we optimize for, it is an interesting measure of strength.

Well for me the auditing page should only contain stuff that has some serious theoretical basis. 

Perhaps a new page should be created for more ad hoc quantities (I can think of other interesting measures of strength)....

ppigazzini:
I propose to merge it until we have an automatic regression tests page update.
If the new fields position is disrupting a copy+paste workflow I can move in a different place.

dav1312:
I made a spreadsheet that gets all data from fishtest and generates all of the charts automatically a few weeks ago

server/fishtest/templates/tests_stats.mak
ppigazzini: In that line I lost the `> 0` in a pasto, worked for luck, I improved the conditional statement with the shorter/faster `any()`.

It's possible to use also `float("+inf")`

vdbergh:
> I propose to merge it until we have an automatic regression tests page update. If the new fields position is disrupting a copy+paste workflow I can move in a different place.

Well I remain opposed for the reasons I outlined above. 

But in fact _not merging_ it provides a greater incentive for automating the procedure, which is really the way to go!

PR #1289, opened 2022-02-21
Add PairsRatio to elo_result gauge for NumGames

ppigazzini:


ppigazzini:
Creative way to sneak it.
DEV updated.

vondele:
maybe only add this for Elo measurements, not for SPRT (for SPRT this will usually be a number very close to 1 anyway?)

PR #1291, opened 2022-02-21
Add nElo to elo_result gauge for NumGames

ppigazzini:


ppigazzini:
I verified the data with few raw_statistics pages. Reviews are welcome.

ppigazzini:
image
image

PR #1304, opened 2022-03-29
Simplify the update of the SPSA history parameters

ppigazzini:
The SPSA graph x-axis is now the percentage of completion after the recent
commits to "spsa.js".
This companion PR:
- simplify the updates of the SPSA parameters history
- set the max size of the SPSA history parameters list to 10000 entries
- fix the bug that limited the parameters history to a max of 250k iterations

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1236,16 +1236,10 @@ class RunDb:
         # so that the required storage (performance) remains constant.
         if "param_history" not in spsa:
             spsa["param_history"] = []
-        L = len(spsa["params"])
-        freq_max = 250000
-        freq = max(100, min(L * 25, freq_max))
-
-        # adjust freq to multiples of 100 that divide exactly into freq_max
-        freq_div = (i for i in range(100, freq_max + 1, 100) if not freq_max % i)
-        freq = max(i for i in freq_div if i <= freq)
-
-        maxlen = freq_max / freq
-        grow_summary = len(spsa["param_history"]) < min(maxlen, spsa["iter"] / freq)
+        n_params = len(spsa["params"])
+        samples = 101 if n_params < 100 else 10000 / n_params if n_params < 1000 else 1
+        freq = run["args"]["num_games"] / 2 / samples
+        grow_summary = len(spsa["param_history"]) < spsa["iter"] / freq
 
         # Update the current theta based on the results from the worker
         # Worker wins/losses are always in terms of w_params


ppigazzini:
@vondele the SPSA x-axis is the index of the parameters history list, the past attempts to relate it with a nice number of iterations IMO has produced over complex code that works as expected (perfect division of the total number of iterations, nice freq number etc.) only when the SPSA has a nice total number of iterations. By the way the code updates the history when the current iteration values is bigger than a threshold, so the relation at a number of games is fuzzy.

Options:
1. keep the proposed code
1. drop the requirement to have a nice freq (eg the actual: divisor of the total iterations that is multiple of 100, or the previous: 25 * number of parameters).
```diff
diff --git a/server/fishtest/rundb.py b/server/fishtest/rundb.py
index d843261..50783b4 100644
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1236,15 +1236,9 @@ class RunDb:
         # so that the required storage (performance) remains constant.
         if "param_history" not in spsa:
             spsa["param_history"] = []
-        L = len(spsa["params"])
-        freq_max = 250000
-        freq = max(100, min(L * 25, freq_max))
-
-        # adjust freq to multiples of 100 that divide exactly into freq_max
-        freq_div = (i for i in range(100, freq_max + 1, 100) if not freq_max % i)
-        freq = max(i for i in freq_div if i <= freq)
-
-        maxlen = freq_max / freq
+        freq_max = run["args"]["num_games"] / 2
+        maxlen = 10000 / len(spsa["params"])
+        freq = max(100, min(freq_max / maxlen, freq_max))
         grow_summary = len(spsa["param_history"]) < min(maxlen, spsa["iter"] / freq)
```

2. keep only 100 entries for the param history list (deal for big number of params), so the x-axis (for the new SPSA) will be the percentage of completion.
```diff
diff --git a/server/fishtest/rundb.py b/server/fishtest/rundb.py
index 50783b4..6109273 100644
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1236,10 +1236,20 @@ class RunDb:
         # so that the required storage (performance) remains constant.
         if "param_history" not in spsa:
             spsa["param_history"] = []
-        freq_max = run["args"]["num_games"] / 2
-        maxlen = 10000 / len(spsa["params"])
-        freq = max(100, min(freq_max / maxlen, freq_max))
-        grow_summary = len(spsa["param_history"]) < min(maxlen, spsa["iter"] / freq)
+        n_params = len(spsa["params"])
+        samples = (
+            100
+            if n_params <= 100
+            else 50
+            if n_params <= 200
+            else 10
+            if n_params <= 1000
+            else 5
+            if n_params <= 2000
+            else 1
+        )
+        freq = run["args"]["num_games"] / 2 / samples
+        grow_summary = len(spsa["param_history"]) < spsa["iter"] / freq

         # Update the current theta based on the results from the worker
```

vondele:
something simple is good IMO. Having 100 entries makes sense, I like percentage complete. If the graph would have an x-axis label, that would be clear immediately.
At first sight code might need some guards against division by zero in corner cases.

ppigazzini:
Please refresh the browser, I set the SPSA percentage progression in js:
https://github.com/glinscott/fishtest/commit/72bacc0dc162347fce50a3aa555d7da9abd18e0a
https://github.com/glinscott/fishtest/commit/f25a1672d129d19cbb8c86d2efb3aaae680f5c9c

https://github.com/glinscott/fishtest/blob/72bacc0dc162347fce50a3aa555d7da9abd18e0a/server/fishtest/templates/tests_view.mak#L12-L15
The js uses these SPSA info
https://tests.stockfishchess.org/tests/view/6243607355e09600728a049a/spsa_history

ppigazzini:
This is a complete SPSA with 101 entries, no need of the gaussian kernel smoother:

image

PR #1310, opened 2022-04-08
Print exception name

vdbergh:
Print the name of an unclassified exception in the event log.

Example: https://tests.stockfishchess.org/actions?count=1&before=1649400523.77


vdbergh:
Cleaner version: https://tests.stockfishchess.org/actions?count=1&before=1649406017.217

ppigazzini (approved):

ppigazzini:
Triggered the update of the workers, thank you @vdbergh :)

PR #1317, opened 2022-04-14
Normalize the y values in SPSA gaussian regression

ppigazzini:
Our implementation of the gaussian kernel regression works bad when the range
of y values is big, so normalize the y values in [0, 1] before the regression.
This is a fast fix, the gaussian regression code needs a review.


ppigazzini:
Before:
image
After:
image

PR #1318, opened 2022-04-15
Rewrite the Gaussian Kernel Regression

ppigazzini:
The SPSA test history has now 101 samples, use a fixed bandwidth
to smooth the graph. Adjust the bandwidth only for:
- old SPSA tests with more than 101 samples in history
- SPSA tests with many parameters that has less than 101 samples


ppigazzini:
Tested with some SPSA tests:
- many parameters: https://tests.stockfishchess.org/tests/view/61438983171fa5e08b4ec554
- many history samples: https://tests.stockfishchess.org/tests/view/58c3163c0ebc59035df32c61

PR #1319, opened 2022-04-15
Expand the vertical range of data in SPSA chart

ppigazzini:
The default vAxis setting often compresses the range of negative data
in order to insert the x-axis in the chart.
https://developers.google.com/chart/interactive/docs/gallery/linechart#configuration-options


ppigazzini:
https://tests.stockfishchess.org/tests/view/61438983171fa5e08b4ec554
Before:
image
After:
image

PR #1322, opened 2022-04-16
Update the highlight js library to v 11.5.1

ppigazzini:
Load this library and all the other js/css resources from https://cdnjs.com/


ppigazzini:
Before:
image

After:
image

PR #1334, opened 2022-04-27
Integrate Live Elo page in GUI

ppigazzini:
- drop the url parameter and pass the test id with a js const
- simplify the companion javascript
- use cache busters and SRI for js and CSS
- convert in string the mongodb ObjectID "_id"


vdbergh:
Very nice project. Fishtest now has a much more consistent feel.

ppigazzini:
:)

PR #1341, opened 2022-04-30
Delete deprecated code in the worker.

vdbergh:
At one point the variable `max` was renamed as `MAX` and there was a bit of code to automatically convert config files. By now all config files should have been converted.


ppigazzini:
Triggered the workers update. Thank you @vdbergh :)

PR #1349, opened 2022-05-05
Refine memory estimate

vondele:
Might fix some of the crashes we currently see of fishtest for a test with 8 threads and 512MB hash.


ppigazzini:
PROD updated to test the values.

vondele:
thanks, it always estimates more than master, and fixes the underestimation of master for the mentioned case.

Issue #1284, opened 2022-02-17
Allow more than 100 results in "Events" log


ppigazzini:

vdbergh:
The event log is flooded by nps too low messages. It would be nice if this were resolved.

vdbergh:
https://tests.stockfishchess.org/actions?action=&user=ChessDBCN&before=1649518086.6&count=100

I wonder if noob has been contacted about this? This is appears to be a case of a badly configured machine (since his machines usually get around 1Mnps). 

This flooding of the event log makes it unusable.

ppigazzini:
cc @noobpwnftw

vdbergh:
@noobpwnftw It is getting worse and worse. Often the entire event log is filled by “machine too slow messages” by ChessDBCN… https://tests.stockfishchess.org/actions?action=&user=&before=1649836779.341&count=100

vdbergh:
A possible solution might be to change the meaning of `--fleet` slightly. I.e. let `--fleet` quit the worker only if there is no task available. 

Currently `--fleet` also quits the worker when there is an error. If the worker runs in a loop then it immediately restarts, possibly leading to flooding.

ppigazzini:
Smart suggestion.

dav1312:
Can't we use pagination here too?

vdbergh:
> Can't we use pagination here too?

Yep!

vdbergh:
Something I was also planning to implement is a free text search on the comment field (that would be very useful to search for errors of a specific type). This needs a text index on the comment field. I am currently very busy so if somebody else wants to do it... :)

PR #1351, opened 2022-05-20
Make --fleet only quit the worker when no task is available (RFC).

vdbergh:


noobpwnftw:
This will make fleet less efficient. Instead of waiting/looping on some transient failures, the workers can try other tasks and check fishtest again.

vondele:
It would be nice to figure out what is causing this `  | failed task: task_id: 95, worker:  ChessDBCN-16cores-c477cb52, reason: 'This machine is too slow (235163.0  nps / thread) to run fishtest effectively - sorry!'` this nps is about 1/5 of what should be expected.

noobpwnftw:
Probably some hardware error.

vdbergh:
Unfortunately, as pointed out in #1284 the event log has been flooded with such messages for several months now https://tests.stockfishchess.org/actions?action=&user=&before=1653108833.342&count=100 . The event log is an important debugging tool and it is currently unusable. IMHO a solution should be found. Personally I think transient errors are very rare, hence this PR.

noobpwnftw:
This does not prevent log flooding. It only makes the fishtest worker not quit on client side errors.

noobpwnftw:
OTOH, why is it necessary to regenerate UUID for every worker invocation? If they can be persistent(which is already written to a file) then it will become easier to filter/find events from the particular machine.

vdbergh:
> This does not prevent log flooding. It only makes the fishtest worker not quit on client side errors.

Normally it should prevent it as a delay is inserted before the worker asks for a new task. The delay is small in the beginning but it gets larger if the errors repeat (i.e. if the error is non transient).

vdbergh:
> OTOH, why is it necessary to regenerate UUID for every worker invocation? If they can be persistent(which is already written to a file) then it will become easier to filter/find events from the particular machine.

The problem with preserving the UUID is that people create new workers by just copying the worker directory. So we would soon have distinct workers having the same UUID.

Some time ago I tried to push for a "worker fingerprint" (a hash of the configfile, the ethernet address and the directory of the worker) as a means of uniquely identifying workers. It was not perfect (it did not distinguish between google colab workers) but I would still favor that solution.

noobpwnftw:
I found 2 workers that got throttled due to CPU overheat, probably after years of running and their thermal paste dried.
I've stopped those workers, event flooding seems ended.

It does not however reproduce at every bench, so I guess a longer sleep to cool the CPU will also fix the issue.

As for UUID, why is it a problem if someone did copy the entire worker directory and ends up with the same ID for workers? I think it is perfectly fine to tell them to remove not only this file but also the entire cached compiles due to possible different CPU arch.

ppigazzini:
> As for UUID, why is it a problem if someone did copy the entire worker directory and ends up with the same ID for workers?

If I recall correctly fishtest should consider only one worker the workers with the same UUID and cores numbers in several places: log events, chi2, residuals etc.
The `api.log` in worker folder perhaps has info that help to find the faulty worker in a fleet.

image

noobpwnftw:
But it has nothing to do with the necessity of generating UUID every launch, obviously there are more things to clean up for people copying the entire worker folder and run things properly than just the `uuid.txt`.

vdbergh:
> But it has nothing to do with the necessity of generating UUID every launch, obviously there are more things to clean up for people copying the entire worker folder and run things properly than just the `uuid.txt`.

Unfortunately experience has shown that some people simply ignore all good practice instructions. Since multiple workers having the same UUID would be disastrous, we cannot run this risk.

This being said, being able to identify workers across restarts remains an important problem. I am now thinking that preserving only part of UUID might be a solution. In that way we cannot have distinct workers with the same UUID, but we would still be able to identity workers across restarts in the most common scenarios.

vdbergh:
See #1352

vdbergh:
Closing this for now.

PR #1352, opened 2022-05-22
Preserve the first 8 characters of the uuid across restarts and make them user settable.

vdbergh:
Example:
    
```
$ python worker.py --uuid_prefix `hostname -s`
…
^C
$ cat uuid.txt
pl-f91e-4757-950d-392ca20856dc
```
`--uuid_prefix _hw` resets the prefix to one depending on the system (default).

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -18,7 +18,11 @@ def worker_name(worker_info):
     uuid = worker_info["unique_key"]
     name = "{}-{}cores".format(username, cores)
     if len(uuid) != 0:
-        name += "-" + uuid.split("-")[0]
+        uuid_split=uuid.split("-")
+        if len(uuid_split) >= 1:
+            name += "-" + uuid_split[0]
+        if len(uuid_split) >= 2:
+            name += "-" + uuid_split[1]
     return name
 
 
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -3,12 +3,14 @@ import atexit
 import base64
 import datetime
 import getpass
+import hashlib
 import math
 import multiprocessing
 import os
 import platform
 import re
 import signal
+import socket
 import subprocess
 import sys
 import threading
@@ -120,6 +122,21 @@ def _bool(x):
 
 
 _bool.__name__ = "bool"
+
+
+def _alpha_numeric(x):
+    x = x.strip()
+    if x == "_hw":
+        return x
+    if len(x) <= 1:
+        print("The prefix {} is too short".format(x))
+        raise ValueError(x)
+    if not x.isalnum():
+        raise ValueError(x)
+    return x[:8]
+
+
+_alpha_numeric.__name__ = "alphanumeric"
 
 
 class _memory:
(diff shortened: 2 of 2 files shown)


vdbergh:
Hmm `--uuid_prefix=None` does not work as expected.

vdbergh:
Fixed I think.

vdbergh:
Cleaner logic now.

ppigazzini:
I'm a little worried...
image

vondele:
the 8 character constraint is quite restrictive IMO, not too much room for misuse.
If there is misuse, we can always block.
Having said this, probably we should have a 3 state system: not blocked, blocked, perma-blocked. With the later category not showing up in the pending users.

vdbergh:
I am a bit concerned about this too. But I am not sure it is a real problem. People can already trivially do this by slightly hacking the worker.

We could restrict it to using the output of `hostname -s` if you prefer.

vondele:
users can pick offensive user names, branch names, etc etc.. so far that hasn't been a problem.

vdbergh:
It occurred to me that by preserving part of the uuid it will also become possible to block a specific worker if we believe, for example, that it has a hardware problem.

vdbergh:
I didn't really like the previous stateful implementation. So now I simply let the default prefix depend on the hardware. This is much simpler.

Caveat: this will generate the same prefix for all colab workers, but this is not really an issue since the rest of the UUID is still different. Note that colab workers don't have permanent storage so the previous stateful method also did not work well. In the future we may perhaps investigate how we can distinguish colab workers.

vdbergh:
It seems colab workers can be distinguished by hostname.

vdbergh:
I now included the hostname in the `hw_id`. This can be fine tuned.

vdbergh:
@ppigazzini Do have perhaps time to look at this PR? The code is quite simple and the impact should be minimal.

The issue with noob's machines has been solved but it would be nice to have more information in the future when there are misbehaving workers.

ppigazzini:
Tests were fine. Triggered the workers update, thank you @vdbergh :)

ppigazzini:
wiki page updated https://github.com/glinscott/fishtest/wiki/Running-the-worker:-overview

vdbergh:
Thx :+1:

vdbergh:
I came across an interesting file on Linux 
```
/etc/machine-id
```
From the man page:
> The machine ID is usually generated from a random source during system installation or first boot and stays constant for all subsequent boots.

It is apparently create by systemd during installation.

I also discovered `/etc/hostid` (read by `gethostid()`). I do not know how this is generated.

ppigazzini:
`/etc/machine-id` and `/etc/hostid` are missing in WSL and Alpine (both systemd free).

vdbergh:
@ppigazzini 

What about `/var/lib/dbus/machine-id` ?

Anyway if these files are present we could perhaps use them. But maybe it is not necessary if the MAC address has sufficient persistence for cloud instances (cfr PR #1354).

ppigazzini:
`/var/lib/dbus/machine-id` presents on WSL but missing on Alpine 3.16

vdbergh:
It seems to me we could make our own "machine-id" by writing a file in the user's home directory.

But maybe that would create problems with people which run the worker as a service (not sure if there is a concept of a home directory in that case).

PR #1353, opened 2022-05-30
Put the system id as comment in the config file (if prefix == _hw).

vdbergh:
A small cosmetic addition similar to the options concurrency and max_memory.

This is the line in the config file
```
uuid_prefix = _hw ; = 734c4716
```


vdbergh:
I merged this PR in #1354 since it also about the system id.

PR #1354, opened 2022-05-31
Add " (colab)" to the OS string for a colab worker.

vdbergh:
Colab workers are a bit special so it is useful to flag them.


vdbergh:
It seems `uuid.getnode()` has a bug on aws instances. It does not return the MAC address.

```
ubuntu@ip-172-31-12-208:~$ ifconfig 
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 9001
        inet 172.31.12.208  netmask 255.255.240.0  broadcast 172.31.15.255
        inet6 fe80::8b5:7cff:fe35:f6eb  prefixlen 64  scopeid 0x20<link>
        ether 0a:b5:7c:35:f6:eb  txqueuelen 1000  (Ethernet)   <-------------
        RX packets 714  bytes 96200 (96.2 KB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 815  bytes 98002 (98.0 KB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 68  bytes 6683 (6.6 KB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 68  bytes 6683 (6.6 KB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

ubuntu@ip-172-31-12-208:~$ python
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0] on linux
>>> import uuid
>>> hex(uuid.getnode())
'0xff928c4cdf9c'    <--------
```

vdbergh:
I did a bit of debugging. Apparently on posix python uses `uuid_generate_time_safe` from `libuuid`. For some reason this function on aws does not return a uuid containing the mac address (although it seems to be just a standard Ubuntu 20.04 LTS).

ppigazzini:
On my Hyper-V install (official Ubuntu Server ISOs):
- Ubuntu 20.04 LTS has Python 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0] on linux
- Ubuntu 22.04 LTS has Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0] on linux
From the python version AWS seem to use Ubuntu 22.04 LTS

vdbergh:
AWS does have `/etc/machine-id` though. I am thinking of using this file if it is available. If it not available then use `uuid.getnode()`.

vdbergh:
I am now using `/etc/machine-id` or `/var/lib/dbus/machine-id` if one of these files exists.

vdbergh:
I made this PR WIP. I am going to try a different mechanism.

vdbergh:
It remains a bit of a mystery that on your version of Ubuntu 22.04 it works correctly....

vdbergh:
I got a reply on the util-linux issue tracker

https://github.com/util-linux/util-linux/issues/1704#issuecomment-1144962011

So it seems to be a systemd issue.

vdbergh:
I merged PR #1353 with this PR and closed the first one.

vdbergh:
I refactored the workaround for the buggy `uuid.getnode()` into a separate function `_getnode()`. See also

https://github.com/rcaloras/bashhub-client/issues/82

for another reference.

vdbergh:
The libuuid/systemd configuration bug has been fixed upstream... 

https://github.com/util-linux/util-linux/commit/304b4dc4936b115ca33af5325c3b04d0997c1353
https://github.com/util-linux/util-linux/commit/c9671a3cf7738bb81e1cbef2f56485a36c6e7623

vdbergh:
I am now happy with this PR. The aim is to have a `uuid_prefix` which is as constant as possible but which nonetheless distinguishes distinct workers. The changes compared to the current situation are.

1) Only use the hostname for colab workers (the hostname is not very stable).
2) Work around a bug in `uuid.getnode()` for systemd installations.
3) Put the `uuid_prefix` in the config file as a comment.

All this is of course if the user has not specified their own `uuid_prefix`. 

I have written a little program to monitor the list of workers and their `uuid_prefixes` so that I can see if something strange happens.

vdbergh:
@ppigazzini I added some code to get the machine id from the registry under windows but I do not have windows so I cannot really test it (it works under wine though).

vdbergh:
Now MacOS should also work (thanks to the help of Disservin at #1356 ).

ppigazzini (approved):
Tested working on multiple workers started either on linux, wsl, msys2, colab. Not tested on macOS, tough.

vdbergh:
Thanks!

PR #1357, opened 2022-06-11
New shot at generating a stable default uuid_prefix for workers..

vdbergh:
Currently the algorithm for the default `uuid_prefix` uses the MAC address. But the MAC address is unstable on modern OSes. Moreover `uuid.getnode()` seems to be buggy on a number of systems and simply returns a random number. This needs ugly and fragile workarounds.

The new algorithm uses:

1) the `machine_id` (a uuid generated during system install);
2) the absolute path of the worker;
3) a random seed which is saved in a new "private" section of the config file.

This algorithm can be defeated by cloning a cloud instance with an installed worker.

This seems like a niche case. But if it turns out that it does happen then we can work around it by detecting the resulting duplicate prefix and generating a new seed.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -8,9 +8,9 @@ import math
 import multiprocessing
 import os
 import platform
+import random
 import re
 import signal
-import socket
 import subprocess
 import sys
 import threading
@@ -414,6 +414,7 @@ def setup_parameters(worker_dir):
         ("parameters", "min_threads", "1", int, None),
         ("parameters", "fleet", "False", _bool, None),
         ("parameters", "compiler", default_compiler, compiler_names, None),
+        ("private", "hw_seed", str(random.randint(0, 0xffffffff)), int, None),
     ]
 
     validate(config, schema)
@@ -554,7 +555,7 @@ def setup_parameters(worker_dir):
 
     options.compiler = compilers[options.compiler_]
 
-    options.hw_id = hw_id()
+    options.hw_id = hw_id(config.getint("private", "hw_seed"))
     print("Default uuid_prefix: {}".format(options.hw_id))
 
     # Step 6: determine credentials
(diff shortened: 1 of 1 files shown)


vdbergh:
Sorry about this. It is the 3d iteration, using the experience gained from the previous iterations.

vdbergh:
New "private" section in the config file for internal data...

```ini
[login]
username = XXXXXXXXXXXXX
password = YYYYYYYYYYYYY

[parameters]
protocol = https
host = tests.stockfishchess.org
port = 443
concurrency = max(1,min(3,MAX-1)) ; = 3 cores
max_memory = MAX/2 ; = 991 MiB
uuid_prefix = _hw ; = 21c31ada
min_threads = 1
fleet = False
compiler = g++

[private]
hw_seed = 318204269
```

worker/worker.py
```diff
@@ -8,6 +8,7 @@
 import multiprocessing
 import os
 import platform
+import random
 import re
 import signal
 import socket
```
ppigazzini, line 14: If I'm not wrong `socket` can be dropped.

worker/worker.py
```diff
@@ -414,6 +415,7 @@ def setup_parameters(worker_dir):
         ("parameters", "min_threads", "1", int, None),
         ("parameters", "fleet", "False", _bool, None),
         ("parameters", "compiler", default_compiler, compiler_names, None),
+        ("parameters", "hw_seed", str(random.randint(0, 2<<32-1)), int, None),
```
ppigazzini, line 418: `black` formats in this way:
```diff
418c418
<         ("parameters", "hw_seed", str(random.randint(0, 2<<32-1)), int, None),
---
>         ("parameters", "hw_seed", str(random.randint(0, 2 << 32 - 1)), int, None),
```

vdbergh:
@ppigazzini I dropped socket and manually fixed the black formatting (my main laptop broke down and I am using an older laptop on which black does not appear to work).

ppigazzini:
> @ppigazzini I dropped socket and manually fixed the black formatting (my main laptop broke down and I am using an older laptop on which black does not appear to work).

https://www.notebookcheck.net/?&hide_date=1&tagArray[]=16&typeArray[]=1&id=48

vdbergh:
Thanks, but I already ordered a new one from the university. Installing Linux on a new laptop is always a bit of an adventure...

ppigazzini:
> Installing Linux on a new laptop is always a bit of an adventure...

A conscientious fishtest contributor should have a Windows 11 laptop with msys2 WSL and several distro in Hyper-V VMs. Ah, and a Mx Apple toy sitting somewhere :D

vdbergh:
Fixed a small inconsequential bug. `2 << 32 -1` has to be `(1 << 32) -1`. But I replaced it by `0xffffffff`.

vdbergh:
> A conscientious fishtest contributor should have a Windows 11 laptop with msys2 WSL and several distro in Hyper-V VMs. Ah, and a Mx Apple toy sitting somewhere :D

Yeah I know...

ppigazzini (approved):

ppigazzini:
black wrote the hex constant in upper case.

vdbergh:
> Triggered the workers update, thank you @vdbergh :) black wrote the hex constant in upper case.

:eyes: :+1:

PR #1358, opened 2022-06-12
Proposal to relax the  uuid_prefix conditions WIP

vdbergh:
In this PR all printable characters are allowed. However " " and "-" will be replaced by "`_`". Moreover the uuid_prefix cannot start with "`_`" (prefixes starting with "_" are meant to have a special meaning - although currently only _hw is defined).

We also raise the truncation length from 8 to 12 characters.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -133,19 +133,31 @@ def _bool(x):
 _bool.__name__ = "bool"
 
 
-def _alpha_numeric(x):
+def _printable(x):
     x = x.strip()
     if x == "_hw":
         return x
     if len(x) <= 1:
-        print("The prefix {} is too short".format(x))
+        print("The uuid_prefix is too short")
         raise ValueError(x)
-    if not x.isalnum():
+    if x[0] == "_":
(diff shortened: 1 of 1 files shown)


vdbergh:
Maybe all printable characters is too much. Perhaps `[._0-9a-zA-Z]` is better.

vdbergh:
I am going to close this for now. Once you start allowing more characters it is not clear where to stop. At least with the current setup the rules are clear.

ppigazzini:
To be honest you have already made a great jobs with the UUID. I'm expecting that an user running a linux fleet is able to set a proper alphanumeric (only letters and numbers) UUID prefix. Adding symbols could have unexpected side effects in any part of the framework code.

vdbergh:
@ppigazzini I agree. That's why I was having 2nd thoughts. Anyway the current incarnation of the default `uuid_prefix` seems to be stable for the great majority of the workers(*) so there is not much need for custom `uuid_prefixes`. 

(*) The exceptions are given by people that are using cloud workers and are reinstalling the worker each time they run it (e.g. colab workers and people using aws spot instances).

PR #1359, opened 2022-06-12
Restrict uuid_prefixes to alphanumeric _ascii_ strings.

vdbergh:
Currently "Štockfish" is allowed.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -140,7 +140,7 @@ def _alpha_numeric(x):
     if len(x) <= 1:
         print("The prefix {} is too short".format(x))
         raise ValueError(x)
-    if not x.isalnum():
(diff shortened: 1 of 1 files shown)


ppigazzini:
I'm a noobie in unicode ascii etc, only wondering if `isascii()` and `isalphanum()` can do the same work:
```python
"Štockfish".isascii()
False
```

vdbergh:
isascii unfortunately was only introduced in Python 3.7. I think that's a bit recent.

vdbergh:
For now the worker still runs with Python 3.5.

ppigazzini (approved):

ppigazzini:

vdbergh:
:+1:

PR #1361, opened 2022-06-15
Style changes

dav1312:
Redesign the nn upload page and make it mobile friendly
Reorder test inputs and add some placeholders

Note: in the future the tests_run page would need to be entirely redesigned too

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -487,9 +487,7 @@ option.grayedoutoption {
   background-color: LightGray;
 }
 
-#user-administration,
-#login,
-#register {
+.col-limited-size {
   max-width: 450px;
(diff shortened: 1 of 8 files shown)


vondele:
The style changes look good to me, I wonder if this could influence the script that I use to automatically upload the nets 'bypassing' the website? If not the change is fine with me.

ppigazzini:
@vondele Feel free to try to upload a fake net: I can delete the net from the DB and from the net server.

vondele:
later today, I've plenty of not-so-fake nets.

vondele:
works for me

ppigazzini:
PROD updated, thank you @vondele for the review/test and thank you @dav1312 for the patch :)

Issue #1360, opened 2022-06-14
Workers with issues.


vdbergh:
I am creating this issue to report workers with issues. E.g. currently.

https://tests.stockfishchess.org/actions?action=failed_task&user=Oakwen&before=1655180683.077&max_actions=100

vondele:
This worker is probably running in a bug in latest clang. https://github.com/llvm/llvm-project/issues/55377 
I see his workers are based on clang 15.

vondele:
for whatever reason the TLS might be handled differently or the code aligned properly by luck, depending on the OS.

vdbergh:
Oakwen-3cores-8708609b switched to g++ so the problem is solved.

vdbergh:
Another issue: The matches of Dantist-7cores-3e5ab901 always finish with "Finished match uncleanly":

https://tests.stockfishchess.org/actions?action=&user=Dantist&before=1655265828.358&count=100

This has been going on since forever. I have no idea how it is possible,

vondele:
hard to guess, would need a more detailed error message.

silversolver1:
I somehow managed to notice the "Finished match uncleanly" events from the Dantist worker happen ~15-16 minutes apart every time. MAX_RETRY_TIME in worker.py is about 15 minutes. Perhaps this is related?

vdbergh:
The matches with "Finished match uncleanly" have no games but also no crashes. This suggests that cutechess-cli failed to start the engine(s). I would be good to have access to the worker output of  Dantist-7cores-3e5ab901 so that we can see what's going on.

vondele:
@Dantist can you provide such output?

vdbergh:
@noobpwnftw Your worker ChessDBCN-16cores-f3dad03d is now suffering from throttling. See https://tests.stockfishchess.org/actions?action=&user=ChessDBCN&before=1655455764.117&count=18 .

Note: The hexadecimal number f3dad03d is the first 8 characters of the UUID (which are constant). It can be found as a comment in the config file and also in uuid.txt.

Dantist:
@vdbergh @vondele 
> This suggests that cutechess-cli failed to start the engine(s).

That actually looks correct: [run.log](https://github.com/glinscott/fishtest/files/8931082/run.log)

---
I have to say that I have a somewhat unique setup.. I planned to deploy the worker to many servers, so I dockerized it with the `alpine:edge` base image.

[...]

PR #1362, opened 2022-06-18
Stop the task if the cutechess-cli cannot start an engine.

vdbergh:
See #1360

PR #1363, opened 2022-06-19
Call the engine binaries by absolute path.

vdbergh:
This should fix the docker image built by @Dantist. See #1360 .

The instructions for rebuilding the image are in Dockerfile. Since this PR has not been committed yet I could not test the rebuilding. However I applied the PR manually (after starting the container) and it worked!

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1143,7 +1143,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
     if "start" in task:
         print("Variable task sizes used. Opening offset = {}".format(opening_offset))
     start_game_index = opening_offset + input_total_games
-    run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2**30)
+    run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2 ** 30)
 
     # Format options according to cutechess syntax.
     def parse_options(s):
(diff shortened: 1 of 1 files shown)


vdbergh:
One complication with the docker image is, that it apparently has no permanent storage to preserve the config file. So each time it is run it will have a different `uuid_prefix`.

vdbergh:
It seems to be not very difficult to have the worker directory outside the container and map it to a directory inside the container

https://docs.docker.com/storage/bind-mounts/

vdbergh:
It's not so straightforward though since you cannot mount something while building an image.

noobpwnftw:
Can declare a docker volume.

vdbergh:
As I understand it you can mount something while creating the container, but not while building the image.

vdbergh:
@noobpwnftw Actually now I understand what you are saying. Mounting an empty volume over a directory in the container causes the files in the directory to be copied to the volume. This then becomes permanent storage. Thanks!

vdbergh:
This is the command I use to run the image by @Dantist with permanent storage
```
sudo docker run -it -v worker-dir:/opt/fishtest/worker -e concurrency=MAX fishtest
```
The volume `worker-dir` has been created by
```
sudo docker volume create worker-dir
```
Pretty convenient actually. Given that this will work on all OS'es that docker supports.

vdbergh:
Interestingly the same volume can be mounted by two docker containers, so that effectively two workers share the same directory. The current lockfile implementation is defeated as the two workers have the same PID (in different containers). This can be solved by including the hostname in the lockfile (besides the PID).

vdbergh:
There is no a cutechess-cli binary that works in the docker image and that does not need the PR. See #1362 . This is a better solution. So I am closing this PR.

noobpwnftw:
Containerized worker is always nice to have. PID isolation can be disabled by running the container with `--pid=host`, although it is hardly necessary to run multiple containers in the first place.

vdbergh:
@noobpwnftw Thanks! The —pid option would indeed prevent two workers from running in the same directory.

PR #1364, opened 2022-06-20
Save the hostname in the lockfile. [WIP]

vdbergh:
This is to deal with the case that two workers on different hosts share a common file system. E.g. two docker instances mounting the same volume.

I haven't tested this on actual docker instances but on the common use case where all workers run on the same host the behavior is the same as before.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -11,6 +11,7 @@ import platform
 import random
 import re
 import signal
+import socket
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


vdbergh:
I made this WIP since running the docker image with `--pid=host` seems to be a better solution. However it currently does not work since busybox does not recognize the `-p` flag to the `ps` command. @ppigazzini Since you wrote that code: any suggestions?

vdbergh:
Closing this for now. See #1365

PR #1366, opened 2022-06-21
Fixes for busybox.

vdbergh:
See #1364 

Tested by replacing "ps" by "busybox", "ps" in the Popen statements and running the script
```sh
python worker.py &
python worker.py &
sleep 1
killall python g++ make
```

EDIT: It also works with the docker image
```
vdbergh@pl2:~/fishtest-docker$ sudo docker run -it  --pid=host -v worker-dir:/opt/fishtest/worker -e concurrency=1 fishtest

Worker started in /opt/fishtest/worker ... (PID=273258)
ps failed. It seems we are running busybox

*** Worker (PID=273258) stopped! ***
Another worker (PID=272865) is already running in this directory, using the lock file:
/opt/fishtest/worker/worker.lock
vdbergh@pl2:~/fishtest-docker$ 
```

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -944,7 +944,8 @@ def pid_valid(pid, name):
         else:
             p = stack.enter_context(
                 subprocess.Popen(
-                    ["ps", "-f", "-p", str(pid)],
+                    # for busybox these options are undocumented...
+                    ["ps", "-f", "-a"],
                     stdout=subprocess.PIPE,
                     stderr=subprocess.DEVNULL,
                     universal_newlines=True,
@@ -953,7 +954,7 @@ def pid_valid(pid, name):
                 )
             )
         for line in iter(p.stdout.readline, ""):
-            if name in line:
+            if name in line and str(pid) in line:
                 return True
     return False
 


vdbergh:
@vondele I guess your browser does not have the corresponding root certificate. My certificate is from belnet (which is an academic ISP in Belgium). I assumed that it used a root certificate recognized by the major browsers.... I never had a problem accessing my own website from abroad but I did not check recently. Anyway it is good that wget works.

vondele:
it is actually the other way around (wget fails). But yes, I think it is something minor at this point.

vdbergh:
I posted the wrong zip file. Sorry.

vdbergh:
I posted a better version which does not have hard baked user credentials and which passes on all arguments to the worker. After building the image you can now do things like
```
./worker.sh --concurrency MAX/2 --max_memory MAX/3
```
Or simply
```
./worker.sh --help
```

ppigazzini:
Using `alpine:edge` or `alpine:latests` is a container anti pattern, IMO better to use `alpine:3.16`
https://wiki.alpinelinux.org/wiki/Edge

vdbergh:
Well if you think that's better I can do it. But I had the impression that this is intentional. @Dantist wanted to have the bleeding edge gcc, python etc...

vondele:
i would agree that you'd like to pin the version of the environment as much as possible for a container. That's in part the purpose I would say.

vdbergh:
Ok the fact that each version of the container would be slightly different is a good argument. I pinned the alpine version to 3.16 as suggested.

worker/worker.py
```diff
@@ -946,14 +946,28 @@ def pid_valid(pid, name):
                 subprocess.Popen(
                     ["ps", "-f", "-p", str(pid)],
```
ppigazzini, line 947: If I'm not wrong changing only these 2 lines should work for either linux or busybox:
```python
                subprocess.Popen(
                    ["ps", "-a", "-f"],

...

            if name in line and str(pid) in line:
                return True
```

vdbergh: It seems you are right. I will change it. It is a bit strange though that `busybox --help` doesn't tell you this.
```
$ ps --help

BusyBox v1.35.0 (2022-05-09 17:27:12 UTC) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

	-o COL1,COL2=HEADER	Select columns for display
	-T			Show threads
```

ppigazzini: Neither here https://busybox.net/downloads/BusyBox.html

It seems that some common `ps` flags are accepted with null effect, perhaps for compatibility.

ppigazzini: By the way this could change in future...

vdbergh: I don't think they are going to make ps less compatible.

vdbergh: So I changed it.

vdbergh: This is from the busybox ps source.

	// POSIX:
	// -a  Write information for all processes associated with terminals
	//     Implementations may omit session leaders from this list
	// -A  Write information for all processes
	// -d  Write information for all processes, except session leaders
	// -e  Write information for all processes (equivalent to -A)
	// -f  Generate a full listing
	// -l  Generate a long listing
	// -o col1,col2,col3=header
	//     Select which columns to display
	/* We allow (and ignore) most of the above. FIXME.
	 * -T is picked for threads (POSIX hasn't standardized it).
	 * procps v3.2.7 supports -T and shows tids as SPID column,
	 * it also supports -L where it shows tids as LWP column.
	 */

So at least -f -a are documented there. I wonder what the FIXME means.

vdbergh: This comment was introduced in https://github.com/brgl/busybox/commit/9494919ea5116075e16daac1355265b023419cdc 16 years ago. I think the FIXME refers to the need for a more complete implementation (which was never done).

vdbergh: It is strange since ``-p is also part of the posix spec. https://www.open-std.org/jtc1/sc22/open/n4217.pdf page 2976.

ppigazzini: Perhaps as you have pointed out it simply a very old copy & paste, at example busybox allows and ignores `ps -f p <pid_number>`, while exits with a warning for the more standard `-p` and `--pid`. Better to stick with the somehow documented `-f -a`

https://man7.org/linux/man-pages/man1/ps.1.html

```
       p pidlist
              Select by process ID.  Identical to -p and --pid.

       -p pidlist
              Select by PID.  This selects the processes whose process
              ID numbers appear in pidlist.  Identical to p and --pid.

       --pid pidlist
              Select by process ID.  Identical to -p and p.
```

ppigazzini: Correction: busybox accepts this command `ps -f abcdef 123456`, so the `p 1234` is simply not interpreted as a flag.

https://github.com/mirror/busybox/blob/master/procps/ps.c#L598
Just below the comment that you have highlighted, if I'm not wrong, there is the list of the flags accepted:
```
		getopt32(argv, "Zo:*aAdefl"IF_FEATURE_SHOW_THREADS("T"), &opt_o);
```

ppigazzini:
66Triggered the workers update, tahnk you @vdbergh :)
The "books" repo has now the static `cutechess-cli` for linux.

vdbergh:
Thanks! I updated the docker image. It is now working correctly with the repo cutechess-cli.

PR #1367, opened 2022-06-23
Responsive website

dav1312:
- Update Bootstrap to 5.2.0-beta1 (responsive offcanvas)
- Add header for the hidden sidebar and other elements
- Make tables responsive (only when screen is small)
- Replace some inline styles with bootstrap classes
- Live elo now covers the whole screen
- Temporarily add a scrollbar if the screen is small to tests_run until it gets redesigned
- Other minor fixes


ppigazzini:
In my tests I found only this problem with a long machine list.

image

vondele:
is there a link to try this out?

ppigazzini:
@dav1312 bug fixed, thank you :)
@vondele PR applied to PROD

ppigazzini:
> Since it was using col-lg-9 I thought it was unintentional,

I'm a total noob for Bootstrap, I simply copied that tag form the test page.

dav1312:
Dw, either way, the raw stats page probably needs a refactor

dav1312:
That is weird... I didn't touch much on that template...

dav1312:
Reverted the changes on that page to see if it fixes it bcs I don't really know what happened

dav1312:
Yes... the issue is still there... hmm

ppigazzini:
> Yes... the issue is still there... hmm Edit: I found the issue, the bootstrap js that I used doesn't have popper.js 🤦‍♂️ I need to use the bundle one

PROD updated, please test.

ppigazzini:
I have `tmux` opened in `termux`, so no problem to update PROD from the mobile :)

ppigazzini:
@dav1312 PROD updated

ppigazzini:
Updated.

ppigazzini:
Bug found by noob in dark/light theme when duplicating a tab, steps to replicate:
- start with cleared cookies
- open the homepage (default light)
- switch to dark theme
- duplicate tab: this open a light homepage
After a page refresh the tab duplication seems to work fine.

dav1312:
That "bug" was not introduced by this PR tho

ppigazzini:
> Since it was using `col-lg-9` I thought it was unintentional, changing `container-fluid` to just `container` would probably be the recommended way to limit the width of the container

In my PRs I think to have monkey tried to set `container`, but in this way the text block for the "raw statistics" and "live Elo" pages is centered and not left aligned as for the "test" page (zoom 80% to exaggerate the effect). The only way I found was to copy the style (`col-lg-9`) from the "test" page.

image

ppigazzini:
Updated. Hyphenation seems to work only with the official repo name.

ppigazzini:
Updated, looks good to me.

ppigazzini:
On mobile (chrome app or the device toggle in chrome PC) for a SPSA test the option bar is moved at the left of the page, I need to zoom to show it.

Screenshot_20220701_145057_com android chrome

ppigazzini:
PROD updated

vondele:
Another minor suggestion move `4914 cores 0.88 MNps (4306.50 total MNps) 3268 games/minute 12.2 hours remaining` on a line before the button and `236 machines` so that the buttons are more nicely grouped on mobile.

ppigazzini:
> Another minor suggestion move `4914 cores 0.88 MNps (4306.50 total MNps) 3268 games/minute 12.2 hours remaining` on a line before the button and `236 machines` so that the buttons are more nicely grouped on mobile.

Check now if I understood correctly

```html
% if page_idx == 0:
    <h4>
      <div>
        ${cores} cores ${f"{nps / (cores * 1000000 + 1):.2f}"} MNps
        (${f"{nps / (1000000 + 1):.2f}"} total MNps)
        ${games_per_minute} games/minute
        ${pending_hours} hours remaining
      </div>
      <button id="machines-button" class="btn btn-sm btn-light border">
        ${'Hide' if machines_shown else 'Show'}
      </button>
      <span>
        ${len(machines)} machines
      </span>
    </h4>
```

vondele:
yes the result looks as intended and good IMO.

dav1312:
Yeah it looks nice
The only thing that comes to mind making the cores, nps, etc. a table? But that's all

vondele:
Another thing, but I think needs experimentation, is to have the boxes with the current unfinished results (those 4 LLR boxes in the screenshot) 'transparent' i.e. same color as the table rows

image

dav1312:
Not sure if it makes it easier to read on desktop

image

image

vondele:
can we still have the shape around it, just transparent ? Or filled with a color that is much more similar to the table row color?

vondele:
kind of like 
image

ppigazzini:
PROD updated, fixed on mobile the option window for the SPSA 👍

ppigazzini:
PROD updated.

ppigazzini (approved):

ppigazzini:
Very nice enhancement, thank you @dav1312 :)

Issue #1368, opened 2022-06-24
Only update cutechess-cli conditionally


vdbergh:
The updater does not update cutechess-cli on MacOS since we don’t have a MacOS binary.

This could be made more generic by testing if the downloaded cutechess-cli works. The code for this already exists in the worker.

Edit: I this way it would be less dangerous to try out new cutechess-cli binaries. We would not risk breaking existing setups.

vondele:
probably needs to be able to check the version and act somehow if the version doesn't match the requirements. Now we can update the version, even if the installed one 'works'.

vondele:
The cutechess version check is there, but it triggers and exception.

vdbergh:
It seems this has to be done in worker and not in the updater. The updater should save a backup copy of the existing cutechess-cli. The worker then should use that as a fallback if the downloaded cutechess-cli does not work.

vondele:
that's an idea.

ppigazzini:
> The updater does not update cutechess-cli on MacOS since we don’t have a MacOS binary.

We do have a cutechess-cli binary:
https://github.com/official-stockfish/books/blob/master/cutechess-cli-macos-64bit.zip

https://github.com/glinscott/fishtest/blob/5f9824621f4d587f0fac531686b0d9aa925a58e3/worker/games.py#L1170-L1180

Perhaps we could drop the `cutechess-cli` restore for MacOS:
https://github.com/glinscott/fishtest/blob/5f9824621f4d587f0fac531686b0d9aa925a58e3/worker/updater.py#L71-L74

> The updater should save a backup copy of the existing cutechess-cli.

The updater already backups `cutechess-cli` :)

https://github.com/glinscott/fishtest/blob/5f9824621f4d587f0fac531686b0d9aa925a58e3/worker/updater.py#L61-L63

vdbergh:
See #1369

PR #1369, opened 2022-06-25
Try a backup copy if the downloaded cutechess-cli is not working.

vdbergh:
Logic:

1) Check if cutechess-cli exists and is working.

2) If not: download it.

3) Check if downloaded cutechess-cli is working.

4) If not: try to restore a backup copy.

5) If successful check if the backup copy is working.

6) If not: bail out.

Also: do not treat MacOS specially in the updater.

Also: only setup/verify during worker startup.

Solves #1368 .

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -3,6 +3,7 @@ import atexit
 import base64
 import datetime
 import getpass
+import glob
 import hashlib
 import math
 import multiprocessing
@@ -11,17 +12,20 @@ import platform
 import random
 import re
 import signal
+import stat
 import subprocess
 import sys
 import threading
 import time
 import traceback
 import uuid
+import shutil
 import zlib
 from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
 from configparser import ConfigParser
 from contextlib import ExitStack
 from functools import partial
+from zipfile import ZipFile
 
 # Try to import the system wide package (eg requests).
 # Fall back to the local one if the global one does not exist.
@@ -35,10 +39,13 @@ sys.path.append(packages_dir)
 import packages.expression as expression
 import requests
 from games import (
+    EXE_SUFFIX,
     FatalException,
     RunException,
     WorkerException,
     backup_log,
+    download_from_github,
+    format_return_code,
     log,
     run_games,
     send_api_post_request,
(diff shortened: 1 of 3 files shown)


vdbergh:
@ppigazzini I wonder if you could have a look at this PR? I think it is a nice refactoring of the setup of cutechess-cli. Perhaps it might also deal correctly with this issue (if it is solvable at all):

https://tests.stockfishchess.org/actions?action=&user=technologov&before=1657958686.643&count=100

worker/worker.py
```diff
@@ -270,6 +277,130 @@ def get_credentials(config, options, args):
...
+    if not verify_required_cutechess(testing_dir, cutechess):
+        if len(EXE_SUFFIX) > 0:
+            zipball = "cutechess-cli-win.zip"
+        elif IS_MACOS:
+            zipball = "cutechess-cli-macos-64bit.zip"
+        else:
+            zipball = "cutechess-cli-linux-{}.zip".format(platform.architecture()[0])
+        try:
+            download_from_github(zipball, testing_dir)
+            zip_file = ZipFile(zipball)
```
ppigazzini, line 355: Context manager added in python 3.2, see https://docs.python.org/3.5/library/zipfile.html#zipfile.ZipFile

```python
            zip_file = ZipFile(zipball)
            zip_file.extractall()
            zip_file.close()
```

```python
            with ZipFile(zipball) as zip_file:
                zip_file.extractall()
```

vdbergh: Changed!

vdbergh:
> If I'm not wrong the error messages are now only worker side.

Yes that is true. 

But IMHO checking that cutechess is working is part of the start up of the worker (like selecting a compiler).

ppigazzini:
I meant that I don't view entries in Events log about cutechess-cli problems:
https://dfts-0.pigazzini.it/actions

vdbergh:
The messages in the event log come from failed tasks. In this PR the check for a working cutechess-cli is done before a task is started. So there are no messages in the event log. 

My plan is to create a log api to post messages to the event log not associated with a task.

ppigazzini:
A collateral problem found during the tests of this PR.
The worker update is checked/done in `fetch_and_handle_task()`, we have now some code than can prevent a fixing update.

If we add at example the support for ARM (eg the link to the `cutechess-cli-arm.zip`), an already installed worker will exit (because a not working cutechess) before checking the version.

vdbergh:
I agree! So the update test should also be done a startup….

vdbergh:
Latest push does an update test before checking cutechess.

ppigazzini (approved):

ppigazzini:
Nice improvement, thank you @vdbergh :)
Triggered the workers update.

PS: please try to update `black`. Mine is latest v 22.6.0 and writes `2**30`, yours still writes `2 ** 30`, see
https://github.com/psf/black/pull/2726

PR #1372, opened 2022-07-01
Use a short pagination

ppigazzini:


vondele:
still a tiny bit too wide on my phone (assuming fishtest is updated right now)

dav1312:
What about something like this? Having Prev, first, last and Next always and then 5 "cells" beween them

image

image

image

image

ppigazzini:
> What about something like this? Having Prev, first, last and Next always and then 5 "cells" beween them

Approved! I will update the code in a couple of hours.

ppigazzini:
PROD updated.

vondele:
that's nice.

PR #1374, opened 2022-07-03
Make sure to import the local expression package.

### LINKED ISSUE #1373: Worker bug with expression package
The error from the worker when I call `python3 fishtest/worker/worker.py <user> <pwd> --concurrency MAX`
```
Worker started in /home/<username>/fishtest/worker ... (PID=xxxxx)
Creating lock file /home/<username>/fishtest/worker/worker.lock
Found g++ version 10.3.0
Found clang++ version 10.0.0
module 'expression' has no attribute 'Expression_Parser'
Error parsing options. Config file not written.
Deleting lock file /home/<username>/fishtest/worker/worker.lock
```
Potential culprit:
https://github.com/glinscott/fishtest/blob/6c283d90f0d0ed21b86fd20f81db152b20fc8a9e/worker/worker.py#L157
A Python test:
```
> python3
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
>>> import expression
>>> e=expression.Expression_Parser()
AttributeError: module 'expression' has no attribute 'Expression_Parser'
```

vdbergh:
Fixes #1373

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -23,19 +23,20 @@ from configparser import ConfigParser
 from contextlib import ExitStack
 from functools import partial
 
-# Try to import an user installed package,
-# fall back to the local one in case of error.
 
 packages_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "packages")
-try:
-    import requests
-except ImportError:
-    sys.path.append(packages_dir)
-    import requests
+sys.path.append(packages_dir)
 
-if packages_dir not in sys.path:
-    sys.path.append(packages_dir)
-import expression
+# Try to import the system wide requests package.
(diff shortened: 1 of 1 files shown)


vdbergh:
Hmm this does not work correctly  if there is no global requests package (e.g. the docker image) since the local requests package does absolute imports (e.g. urllib3) which cannot be found using the new scheme. If there is no solution for this then we should go back to path manipulation.

The local expression package works fine since it doesn't do any further imports.

ppigazzini (approved):
Tested working.

ppigazzini:
Triggere the workers updated. I moved the comments to have a nice output from `isort`
Thank you @vdbergh :)

PR #1342, opened 2022-04-30
Refactoring of get_flag code

ppigazzini:
- little code simplification
- improve comments
- try except for a (very low probability) race condition:
  workers with the same ip, without country code, accessing
  simultaneously the global flag_cache

A flag_cache with 20k ip entries requires 590kB, is not necessary
to drop stale ip from the cache (see the PR for the code).

PR #1375, opened 2022-07-03
Hashed passwords in userdb

### LINKED ISSUE #1294: Passwords are stored in plaintext
https://github.com/glinscott/fishtest/blob/e16909c4994ac8b1533bda5bd41588424c6d2ad7/server/fishtest/userdb.py#L86

zungur:
Fixes  #1294
Adds argon2-cffi-hashed password storage to userdb
Hashes plaintext passwords during login as users login to website, or during authenticated API requests
Removes plaintext password logging to stderr
Changes authentication logic to filter out blocked users before password checks
Changes unit tests accordingly
Uses LRU cache for hash validation during authentication/API requests

### DIFF
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -4,6 +4,7 @@ import time
 from datetime import datetime
 
 from pymongo import ASCENDING
+from argon2 import PasswordHasher
 
 
 class UserDb:
@@ -33,16 +34,32 @@ class UserDb:
         with self.user_lock:
             self.cache.clear()
 
+    def hash_password(self, password):
+        return PasswordHasher().hash(password)
+
+    def check_password(self, hashed_password, password):
+        try:
+            return PasswordHasher().verify(hashed_password, password)
+        except:
+            return False
+
     def authenticate(self, username, password):
         user = self.find(username)
-        if not user or user["password"] != password:
-            sys.stderr.write("Invalid login: '{}' '{}'\n".format(username, password))
-            return {"error": "Invalid password for user: {}".format(username)}
-        if "blocked" in user and user["blocked"]:
-            sys.stderr.write("Blocked login: '{}' '{}'\n".format(username, password))
-            return {"error": "Account blocked for user: {}".format(username)}
-
-        return {"username": username, "authenticated": True}
+        if user:
+            if "blocked" in user and user["blocked"]:
+                sys.stderr.write("Blocked login attempt: '{}'\n".format(username))
+                return {"error": "Account blocked for user: {}".format(username)}
+            if self.check_password(user["password"], password):
+                return {"username": username, "authenticated": True}
+            # remove elif logic after all the passwords in userdb are hashed
+            elif user["password"] == password:
+                user["password"] = self.hash_password(user["password"])
+                self.save_user(user)
+                return {"username": username, "authenticated": True}
+            else:
+                return {"error": "Invalid credentials"}
+        else:
+            return {"error": "Invalid credentials"}
 
     def get_users(self):
         return self.users.find(sort=[("_id", ASCENDING)])
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -2,9 +2,11 @@ import sys
 import threading
 import time
 from datetime import datetime
+from functools import lru_cache
 
-from pymongo import ASCENDING
 from argon2 import PasswordHasher
+from argon2.exceptions import VerifyMismatchError, HashingError, InvalidHash, VerificationError
+from pymongo import ASCENDING
 
 
 class UserDb:
@@ -34,14 +36,26 @@ class UserDb:
         with self.user_lock:
             self.cache.clear()
 
-    def hash_password(self, password):
+    @lru_cache(maxsize=128)
+    def hash_password(self, password,
+        time_cost: int = 3, memory_cost: int = 65536, parallelism: int = 4,
+        hash_len: int = 32, salt_len: int = 16):
         return PasswordHasher().hash(password)
 
     def check_password(self, hashed_password, password):
         try:
             return PasswordHasher().verify(hashed_password, password)
-        except:
-            return False
+        except InvalidHash as e:
+            print("InvalidHash:", e, sep="\n")
+        except VerifyMismatchError as e:
+            print("VerifyMismatchError:", e, sep="\n")
+        except HashingError as e:
+            print("HashingError:", e, sep="\n")
+        except VerificationError as e:
+            print("VerificationError:", e, sep="\n")
+        except Exception as e:
+            print("Exception:", e, sep="\n")
+        return False
 
     def authenticate(self, username, password):
         user = self.find(username)
@@ -57,9 +71,9 @@ class UserDb:
                 self.save_user(user)
                 return {"username": username, "authenticated": True}
             else:
-                return {"error": "Invalid credentials"}
+                return {"error": "Invalid password"}
         else:
-            return {"error": "Invalid credentials"}
+            return {"error": "Invalid username"}
 
     def get_users(self):
         return self.users.find(sort=[("_id", ASCENDING)])
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -36,12 +36,12 @@ class UserDb:
         with self.user_lock:
             self.cache.clear()
 
-    @lru_cache(maxsize=128)
     def hash_password(self, password,
         time_cost: int = 3, memory_cost: int = 65536, parallelism: int = 4,
         hash_len: int = 32, salt_len: int = 16):
         return PasswordHasher().hash(password)
 
+    @lru_cache(maxsize=1024)
     def check_password(self, hashed_password, password):
         try:
             return PasswordHasher().verify(hashed_password, password)
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -5,7 +5,12 @@ from datetime import datetime
 from functools import lru_cache
 
 from argon2 import PasswordHasher
-from argon2.exceptions import VerifyMismatchError, HashingError, InvalidHash, VerificationError
+from argon2.exceptions import (
+    HashingError,
+    InvalidHash,
+    VerificationError,
+    VerifyMismatchError,
+)
 from pymongo import ASCENDING
 
 
@@ -36,8 +41,13 @@ class UserDb:
         with self.user_lock:
             self.cache.clear()
 
-    def hash_password(self, password,
-        time_cost: int = 3, memory_cost: int = 12288, parallelism: int = 1):
+    def hash_password(
+        self,
+        password,
+        time_cost: int = 3,
+        memory_cost: int = 12288,
+        parallelism: int = 1,
+    ):
         return PasswordHasher(time_cost, memory_cost, parallelism).hash(password)
 
     @lru_cache(maxsize=128)
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -37,9 +37,8 @@ class UserDb:
             self.cache.clear()
 
     def hash_password(self, password,
-        time_cost: int = 3, memory_cost: int = 12288, parallelism: int = 1,
-        hash_len: int = 32, salt_len: int = 16):
-        return PasswordHasher().hash(password)
+        time_cost: int = 3, memory_cost: int = 12288, parallelism: int = 1):
+        return PasswordHasher(time_cost, memory_cost, parallelism).hash(password)
 
     @lru_cache(maxsize=128)
     def check_password(self, hashed_password, password):
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -37,11 +37,11 @@ class UserDb:
             self.cache.clear()
 
     def hash_password(self, password,
-        time_cost: int = 3, memory_cost: int = 65536, parallelism: int = 4,
+        time_cost: int = 3, memory_cost: int = 12288, parallelism: int = 1,
         hash_len: int = 32, salt_len: int = 16):
         return PasswordHasher().hash(password)
 
-    @lru_cache(maxsize=1024)
+    @lru_cache(maxsize=128)
     def check_password(self, hashed_password, password):
         try:
             return PasswordHasher().verify(hashed_password, password)
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ name: CI lint
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
-  build:
+  check:
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/server.yaml
+++ b/.github/workflows/server.yaml
@@ -3,7 +3,7 @@ name: CI server
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/worker_msys2.yaml
+++ b/.github/workflows/worker_msys2.yaml
@@ -3,7 +3,7 @@ name: CI worker msys2
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
-  build:
+  test:
     runs-on: windows-latest
     strategy:
       matrix:
--- a/.github/workflows/worker_posix.yaml
+++ b/.github/workflows/worker_posix.yaml
@@ -3,7 +3,7 @@ name: CI worker posix
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
-  build:
+  test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 ### Overview
-[![CI server](https://github.com/glinscott/fishtest/actions/workflows/server.yaml/badge.svg)](https://github.com/glinscott/fishtest/actions/workflows/server.yaml) [![CI worker posix](https://github.com/glinscott/fishtest/actions/workflows/worker_posix.yaml/badge.svg)](https://github.com/glinscott/fishtest/actions/workflows/worker_posix.yaml) [![CI worker msys2](https://github.com/glinscott/fishtest/actions/workflows/worker_msys2.yaml/badge.svg)](https://github.com/glinscott/fishtest/actions/workflows/worker_msys2.yaml) [![CI lint](https://github.com/glinscott/fishtest/actions/workflows/lint.yaml/badge.svg)](https://github.com/glinscott/fishtest/actions/workflows/lint.yaml)
+[![CI server](https://github.com/official-stockfish/fishtest/actions/workflows/server.yaml/badge.svg)](https://github.com/official-stockfish/fishtest/actions/workflows/server.yaml) [![CI worker posix](https://github.com/official-stockfish/fishtest/actions/workflows/worker_posix.yaml/badge.svg)](https://github.com/official-stockfish/fishtest/actions/workflows/worker_posix.yaml) [![CI worker msys2](https://github.com/official-stockfish/fishtest/actions/workflows/worker_msys2.yaml/badge.svg)](https://github.com/official-stockfish/fishtest/actions/workflows/worker_msys2.yaml) [![CI lint](https://github.com/official-stockfish/fishtest/actions/workflows/lint.yaml/badge.svg)](https://github.com/official-stockfish/fishtest/actions/workflows/lint.yaml)
 
 Fishtest is a distributed task queue for testing chess engines. The main instance
 for testing the chess engine [Stockfish](https://github.com/official-stockfish/Stockfish) is at this web page https://tests.stockfishchess.org
@@ -17,4 +17,4 @@ The fishtest server:
 - updates and publishes the results of ongoing tests.
 - knows how to stop tests when they are statistically significant and publishes the final tests results.
 
-To get more information, such as the worker/server install and configuration instructions, visit the [Fishtest Wiki](https://github.com/glinscott/fishtest/wiki).
+To get more information, such as the worker/server install and configuration instructions, visit the [Fishtest Wiki](https://github.com/official-stockfish/fishtest/wiki).
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -118,6 +118,7 @@ def main(global_config, **settings):
     config.add_route("api_download_nn", "/api/nn/{id}")
     config.add_route("api_get_elo", "/api/get_elo/{id}")
     config.add_route("api_actions", "/api/actions")
+    config.add_route("api_calc_elo", "/api/calc_elo")
 
     config.scan()
     return config.make_wsgi_app()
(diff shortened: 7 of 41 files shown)


vdbergh:
Ideally there should be a salt in the db. Otherwise this scheme is still vulnerable to a rainbow table attack.

zungur:
argon2-cffi by default uses argon2id, which uses a random salt for hashing already

zungur:
There's a huge bug in this commit, users cannot be blocked, working on it

vdbergh:
I haven't really studied the PR but do you keep track which passwords are hashed and which aren't? This will necessary for future db conversion.

I am also thinking that once we convert the db, we will have lost the opportunity to add a salt, since the original passwords are lost.

dav1312:
> I haven't really studied the PR but do you keep track which passwords are hashed and which aren't? This will necessary for future db conversion.

Hashed passwords will have `$argon2i`

> I am also thinking that once we convert the db, we will have lost the opportunity to add a salt, since the original passwords are lost.

Passwords are already salted, are u thinking about a pepper?

vdbergh:
> > I haven't really studied the PR but do you keep track which passwords are hashed and which aren't? This will necessary for future db conversion.
> 
> Hashed passwords will have `$argon2i`

Ok good. I see that the hash has a lot of structure.
> 
> > I am also thinking that once we convert the db, we will have lost the opportunity to add a salt, since the original passwords are lost.
> 
> Passwords are already salted, are u thinking about a pepper?

I see that the salt is in fact part of the hash. So all is good! I do not know about "pepper".

ppigazzini:
Tested with `mongosh` and a script to reset the password to plain text.
Tests passed:
 - password hashed with web login
 - password hashed with worker login
 - password hashed with register
 - login blocked user
 - worker with blocked user
 - wrong password with web login
 - wrong username with web login
 - wrong password with worker login
 - wrong username with worker login
 - tested the fishtest server setup script
 
The error message in the PR is "Invalid credentials", perhaps a best practice to no leak information during a fraud attempt.
However we do write the usernames on the web so perhaps we could write a more detailed error to help the user.

```python
        if user:
            ...
            # remove elif logic after all the passwords in userdb are hashed
            elif user["password"] == password:
            ...
            else:
                return {"error": "Invalid password"}
        else:
            return {"error": "Invalid username"}
```


I'm not able to view a RAM usage difference with DEV and only few workers. 
However PROD configuration was slimmed down in the last months, so we have some free RAM as buffer.

```
~$ free -h
              total        used        free      shared  buff/cache   available
Mem:           5.0G        3.1G        395M        1.1M        1.5G        1.8G
Swap:            0B          0B          0B
```

```
top - 14:45:42 up 34 days,  5:09,  4 users,  load average: 0.08, 0.24, 0.33
Tasks:  34 total,   1 running,  33 sleeping,   0 stopped,   0 zombie
%Cpu0  :   1.0/0.0     1[|                                                    ]
%Cpu1  :   5.0/0.0     5[|||                                                  ]
%Cpu2  :   4.0/0.3     4[||                                                   ]
%Cpu3  :   2.0/0.3     2[|                                                    ]
KiB Mem : 62.9/5242880  [|||||||||||||||||||||||||||||||||                    ]
KiB Swap:  0.0/0        [                                                     ]
```

ppigazzini:
What about this code refactoring?

```diff
38,39c38,40
<     def hash_password(self, password):
<         return PasswordHasher().hash(password)
---
>     def hash_password(self, user):
>         user["password"] = PasswordHasher().hash(user["password"])
>         self.save_user(user)
57,58c58
<                 user["password"] = self.hash_password(user["password"])
<                 self.save_user(user)
---
>                 self.hash_password(user)
```

server/fishtest/userdb.py
```diff
@@ -4,6 +4,7 @@
 from datetime import datetime
 
 from pymongo import ASCENDING
+from argon2 import PasswordHasher
```
ppigazzini, line 7: Sort the imports:
```python
from argon2 import PasswordHasher
from pymongo import ASCENDING
```

zungur:
> What about this code refactoring?
> 
> ```diff
> 38,39c38,40
> <     def hash_password(self, password):
> <         return PasswordHasher().hash(password)
> ---
> >     def hash_password(self, user):
> >         user["password"] = PasswordHasher().hash(user["password"])
> >         self.save_user(user)
> 57,58c58
> <                 user["password"] = self.hash_password(user["password"])
> <                 self.save_user(user)
> ---
> >                 self.hash_password(user)
> ```

I would prefer to keep this function to a single functionality. One reason is here:
https://github.com/zungur/fishtest/blob/a83c3f8533755f676160ac97131856e70a947416/server/fishtest/views.py#L229
Signup function calls `create_user` with the hashed password. We cannot update a user who doesn't exist.

server/fishtest/userdb.py
zungur: will do

server/fishtest/userdb.py
```diff
@@ -34,16 +35,32 @@ def clear_cache(self):
         with self.user_lock:
             self.cache.clear()
 
+    def hash_password(self, password):
+        return PasswordHasher().hash(password)
+
+    def check_password(self, hashed_password, password):
+        try:
+            return PasswordHasher().verify(hashed_password, password)
+        except:
```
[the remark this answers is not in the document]
zungur: `except:` block captures all exceptions and hides them. Did you want to log what kind of mismatch we are getting? 
The `verify` return value upon mismatch has gotten more details over time:
https://argon2-cffi.readthedocs.io/en/stable/api.html#argon2.PasswordHasher.verify
Hypothetically, if the default parameters of `PasswordHasher()` change in the future, database errors can be easily understood.

zungur:
https://github.com/glinscott/fishtest/pull/1375#issuecomment-1180529195
Thank you for your detailed analysis :)
I will change consistent error feedback with your recommendation. I was far too focused on OWASP best practices.
I think 64 MB memory cost per hash is reasonable given the number of workers. If you wish, you can also check the timing performance of argon2-cffi for the default LOW_MEMORY profile on the server with  `python3 -m argon2 -t 3 -m 65536 -p 4 -l 32`

server/fishtest/userdb.py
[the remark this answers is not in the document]
zungur: To further future-proof against parameter changes, I should probably integrate `check_needs_rehash()` to the `check_password`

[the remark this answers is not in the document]
ppigazzini: > I would prefer to keep this function to a single functionality. One reason is here: https://github.com/zungur/fishtest/blob/a83c3f8533755f676160ac97131856e70a947416/server/fishtest/views.py#L229 Signup function calls `create_user` with the hashed password. We cannot update a user who doesn't exist.

In that case IMO is simpler to use directly the library function without adding a method.

I will prefer to have few high level methods as in this API documentation snippet (`db.get_password_hash_for_user()`, `db.set_password_hash_for_user()`) where to write the low level knowledge about `argon2` and the UserDB:

```python
import argon2

ph = argon2.PasswordHasher()

def login(db, user, password):
    hash = db.get_password_hash_for_user(user)

    # Verify password, raises exception if wrong.
    ph.verify(hash, password)

    # Now that we have the cleartext password,
    # check the hash's parameters and if outdated,
    # rehash the user's password in the database.
    if ph.check_needs_rehash(hash):
        db.set_password_hash_for_user(user, ph.hash(password))
```

ppigazzini:
We have a problem.
During that simple test on PROD the CPU usage was huge, I reverted the code otherwise the provider will shutdown the VPS in a couple of hours.
The web GUI was sluggish too.

image


```
top - 10:32:16 up 35 days, 55 min,  5 users,  load average: 5.67, 1.78, 0.78
Tasks:  35 total,   1 running,  34 sleeping,   0 stopped,   0 zombie
%Cpu0  :  65.4/16.6   82[||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||                  ]
%Cpu1  :  29.0/5.0    34[||||||||||||||||||||||||||||||||||                                                                  ]
%Cpu2  :  77.2/15.2   92[||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||        ]
%Cpu3  :  61.3/8.6    70[||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||                              ]
KiB Mem : 49.6/5242880  [||||||||||||||||||||||||||||||||||||||||||||||||||                                                  ]
KiB Swap:  0.0/0        [                                                                                                    ]

    1 root      20      224728   4724   2936 S        0.1   0:10.95 init -z
    2 root      20                           S                       `- [kthreadd/1988]
    3 root      20                           S              0:00.01      `- [khelper]
   69 root      20      218312  42836  33556 S        0.8   3:16.81  `- /lib/systemd/systemd-journald
  188 root      20       42096    796    584 S        0.0   0:02.91  `- /lib/systemd/systemd-udevd
  191 systemd+  20       71716    584        S        0.0   0:03.45  `- /lib/systemd/systemd-networkd
  199 root      20       30020    920    632 S        0.0   0:07.33  `- /usr/sbin/cron -f
  207 message+  20       47736   1208    640 S        0.0   0:00.42  `- /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
  223 root      20       61968   1492    920 S        0.0   0:03.71  `- /lib/systemd/systemd-logind
  224 syslog    20      189028    980        S   0.3  0.0   0:12.59  `- /usr/sbin/rsyslogd -n
  322 root      20      185932    972    660 S        0.0   0:00.11  `- /usr/bin/python3 /usr/share/unattended-upgrades/unattended-upgrade-shutdown --wait-for-signal
  329 root      20       14660    328    328 S        0.0            `- /sbin/agetty -o -p -- \u --noclear --keep-baud console 115200,38400,9600 linux
  335 root      20       13016    324    324 S        0.0            `- /sbin/agetty -o -p -- \u --noclear tty2 linux
  379 root      20       24180    764    516 S        0.0            `- /usr/sbin/xinetd -pidfile /run/xinetd.pid -stayalive -inetd_compat -inetd_ipv6
  386 root      20      100980   1404    552 S        0.0            `- /usr/sbin/saslauthd -a pam -c -m /var/run/saslauthd -n 2
  393 root      20      100980    856        S        0.0                `- /usr/sbin/saslauthd -a pam -c -m /var/run/saslauthd -n 2
  562 root      20       72292   1112    364 S        0.0   0:00.08  `- /usr/sbin/sshd -D
 9260 root      20       97184   6456   5516 S        0.1   0:00.02      `- sshd: fishtest [priv]
 9276 fishtest  20       97184   3316   2376 S        0.1   0:00.14          `- sshd: fishtest@pts/1
 9277 fishtest  20       21212   4776   3264 S        0.1   0:00.14              `- -bash
 2125 fishtest  20       38304   3364   2828 R        0.1   0:00.04                  `- top
  576 Debian-+  20       59456   1964   1244 S        0.0   0:06.76  `- /usr/sbin/exim4 -bd -q30m
  582 root      20      142544   3152   1700 S        0.1   0:00.02  `- nginx: master process /usr/sbin/nginx -g daemon on; master_process on;
 3747 www-data  20      147104   9800   3980 S   2.0  0.2  86:02.75      `- nginx: worker process
 3748 www-data  20      143752   4828   2640 S        0.1   0:06.85      `- nginx: worker process
 3749 www-data  20      144072   5372   2736 S   0.3  0.1   3:47.20      `- nginx: worker process
 3750 www-data  20      143156   4368   2508 S        0.1   0:00.56      `- nginx: worker process
27980 fishtest  20       29776   2568    716 S        0.0   0:09.12  `- tmux
27981 fishtest  20       21092    764    472 S        0.0   0:00.06      `- -bash
 4944 fishtest  20       21092   1888    472 S        0.0   0:00.06      `- -bash
 5942 fishtest  20       21224   1928    416 S        0.0   0:00.15      `- -bash
11365 fishtest  20       21224   2044    524 S        0.0   0:00.04      `- -bash
17584 mongodb   20     3414788 1.733g   9988 S   5.0 34.7 415:49.24  `- /usr/bin/mongod --config /etc/mongod.conf
 9727 fishtest  20     1171936 115872  31188 S  28.6  2.2   0:11.30  `- /home/fishtest/fishtest/server/env/bin/python3 /home/fishtest/fishtest/server/env/bin/pserve production.ini http_port=6544
 9745 fishtest  20     1748720 668360  31436 S 239.9 12.7   3:17.78  `- /home/fishtest/fishtest/server/env/bin/python3 /home/fishtest/fishtest/server/env/bin/pserve production.ini http_port=6543
```

vdbergh:
I read that it is recommended that hashing takes at least 0.5s CPU time to deter a brute force attack. That  is huge of course if it has to happen with every api call.

dav1312:
> I read that it is recommended that hashing takes at least 0.5s CPU time to deter a brute force attack. That is huge of course if it has to happen with every api call.

0.125-0.25s should be more than enough

ppigazzini:
Better CPU usage with
```python
#ph = PasswordHasher(time_cost: int = 3, memory_cost: int = 65536, parallelism: int = 4, hash_len: int = 32, salt_len: int = 16)
ph = PasswordHasher(time_cost=1, memory_cost=65536, parallelism=1, hash_len=16, salt_len=8)

    def authenticate(self, username, password):
        pwd = "correct horse battery staple"
        #hash = "$argon2id$v=19$m=65536,t=3,p=4$MIIRqgvgQbgj220jfp0MPA$YfwJSVjtjSU0zzV/P3S9nnQ/USre2wvJMjfCIjrTQbg"
        hash = "$argon2id$v=19$m=65536,t=1,p=1$YV0m6YaZWcs$hxN4QG/0ktUF+CDaQYAOkA"

        try:
            ph.verify(hash, pwd)
        except Exception as e:
            print("Exception verifying with argon2:", e, sep="\n")
```

image

vdbergh:
Are there serious issues with caching the hash?

ppigazzini:
PROD is running with this configuration, see the live monitoring https://cloud.mongodb.com/freemonitoring/cluster/V726YST63VGLDMUUVB4GT34WZHIY6L4D

```pithon
ph = PasswordHasher(time_cost=1, memory_cost=16384, parallelism=1, hash_len=16, salt_len=16)

    def authenticate(self, username, password):
        pwd = "correct horse battery staple"
        hash = "$argon2id$v=19$m=16384,t=1,p=1$EFK4hPnTi90P1FB4uyUU9w$UBvF8QWlmENG7T2aOxeACg"
        try:
            ph.verify(hash, pwd)
        except Exception as e:
            print("Exception verifying with argon2:", e, sep="\n")
```

ppigazzini:
> Are there serious issues with caching the hash?

Hash cached worker side? We already write the password in the config.

ppigazzini:
> No. Just maintaining a dictionary on the server with { password: hash} . Then the hash for every password has to be computed only once (until the server reboots).

The users can use the same password and at the moment I fail to find an use for the hash info without doing a costly verification. So the cache for the password should be a set `{(user, password)}` or, if the hash info should be useful, a dict `{(user, password): hash}` or `{hash: (user, password)}`. In any case we need to update the cache after a password update.

vdbergh:
> > No. Just maintaining a dictionary on the server with { password: hash} . Then the hash for every password has to be computed only once (until the server reboots).
> 
> The users can use the same password and at the moment I fail to find an use for the hash info without doing a costly verification. So the cache for the password should be a set `{(user, password)}` or, if the hash info should be useful, a dict `{(user, password): hash}` or `{hash: (user, password)}`. In any case we need to update the cache after a password update.
> 
> EDIT_000: With a cache if the user forget the password I need to write to the DB a hash from a random password and restart fishtest to clear the cache. Reading the user, password from the cache is not an option :)

Different schemes would be possible but I was just suggesting the conceptually simplest one. Computing the hash of a password is expensive. So saving the result of the computation and reusing that ("memoizing") would be a standard thing to do. No need to update the dictionary if the user password changes (although one can of course).

ppigazzini:
> Different schemes would be possible but I was just suggesting the conceptually simplest one. Computing the hash of a password is expensive. So saving the result of the computation and reusing that ("memoizing") would be a standard thing to do. No need to update the dictionary if the user password changes (although one can of course).

If I'm not wrong to retrieve the hash we need to save in any case the tuple (user, password) in the cache, so we can test directly the plain text password, the hash is checked only at the first authentication to populate the cache. In this way we skip the DB search too.

By the way this is how works the cache for the flag.

vdbergh:
Yes I know of course. I just suggested a standard design pattern (https://en.wikipedia.org/wiki/Memoization) to make it simple (no code changes would be necessary).

ppigazzini:
This extreme parameters configuration has only 4% lower CPU usage than the previous configuration:
```python
ph = PasswordHasher(time_cost=1, memory_cost=8192, parallelism=1, hash_len=16, salt_len=8)

    def authenticate(self, username, password):
        pwd = "correct horse battery staple"
        hash = "$argon2id$v=19$m=8129,t=1,p=1$hWmMFlsmi5s$xh8nykmTYumJhU1efu2Vog"

        try:
            ph.verify(hash, pwd)
        except Exception as e:
            print("Exception verifying with argon2:", e, sep="\n")
```
image

zungur:
I agree that a caching/memoization would be helpful. I'm not well versed in cryptography to evaluate the safety of parameter choices, unfortunately.

I can implement a caching structure such that we create {user:[hash,NONE]} mappings in memory, and populate the NONEs with the plaintext password at the time of the user's first authentication. Consecutive authentications would check against the plaintext, and hash checks would only happen once per user per server reboot.

As a next step, I can add a timeout to the structure so that the existing hash->plaintext mappings are invalidated after a while. A password reset would require access to in-memory structure, but still be possible.

For the in-memory structure, I'm considering this:
```
import sqlite3
connection = sqlite3.connect(":memory:")
...
```
Sqlite is good for concurrent read accesses, but doesn't do well for concurrent write operations. As this structure is read-heavy, I'm assuming it would be fine.

What are your thoughts?

ppigazzini:
Check the code for the flag cache, I personally don't like to have different implementations of the same pattern:
https://github.com/glinscott/fishtest/blob/da4aa2a46d33b1c95925d7ec4af603da5011a218/server/fishtest/api.py#L249-L304

Here the code to drop a flag entry after a time out, not implemented because the memory usage of the flag cache is low.
https://github.com/glinscott/fishtest/pull/1342

python has several ready made decorators for memoization https://docs.python.org/3/library/functools.html , but this is just a note.

zungur:
I didn’t know about the functools cache/lru_cache decorators. No need to reinvent the wheel of course. I can use the built in functions. 

Another solution to repetitive hash checks is server-assigned access tokens, but this would require a major overhaul of authentication and change workers to handle such tokens. This solution would also be out of scope.

noobpwnftw:
Why is it not possible to store auth status per (username,ip) pair?

vdbergh:
IMHO memoizing the hash function is harmless. It is just an implementation detail.

peregrineshahin:
Well, I'm still stuck on the phase that this hashing shouldn't result in this CPU usage reported at the first place

After upgrading mongodb and python version and getting extra space, Can we get another report if this is still an issue?

vdbergh:
The has function is supposed to take 0.5s of CPU time. Of course this has a substantial effect on the overall CPU time use.

noobpwnftw:
What justifies a 20 QPS(according to its docs, with parallelism enabled on recent-ish hardware takes 40~50ms per verification) hash function given our limited CPU time? How is for example a salted SHA256 not secure?

noobpwnftw:
In some sense, the current situation is almost a joke by itself, as if we used some RSA-1048576 encryption and now trying to salvage the slowness by caching modexp, while do not want to shorten the modulus length due to 1048576 > 2048 being more secure.

vdbergh:
The salt protects against the use of rainbow tables. But I guess a salted hash that is quick to compute is still not secure against a brute force search when the user chooses a weak password.

Ofc it is still infinitely more secure than plain text passwords :)

server/fishtest/userdb.py
```diff
@@ -35,14 +37,26 @@ def clear_cache(self):
         with self.user_lock:
             self.cache.clear()
 
-    def hash_password(self, password):
+    @lru_cache(maxsize=128)
+    def hash_password(self, password,
+        time_cost: int = 3, memory_cost: int = 65536, parallelism: int = 4,
```
vdbergh, line 42: I am confused. The parameters to hash_pashword are not passes on it appears.

noobpwnftw:
If it is a weak password that is brute-forcible, then no amount of slowness will really save it from getting broken. I assume we are not implementing MFA or rate limits on such attempts anytime soon, so if that happens it will inflict us with "DDOS" due to wrong passwords can't be cached and will go through our expensive hashing every time.

vdbergh:
The aim of hashing is to make it expensive to recover passwords from a stolen db (so e.g. rate limits are not relevant to the discussion).

vdbergh:
But I agree that expensive hashing makes Fishtest more DOSable.

noobpwnftw:
The more likely event of such a brute force compared to a stolen DB is to test over fishtest API, which makes the slowness and/or rate limits relevant. Even if we use a 20-seconds-per-attempt hashing scheme, passwords like 123456 are still not safe either way. With salted hashes, in case of a stolen DB, passwords are not immediately vulnerable(assuming not 123456-alike) therefore it served the purpose and we'll just notify people to change passwords ASAP.

maximmasiutin:
Please consider calling [`verify()`](https://argon2-cffi.readthedocs.io/en/stable/api.html) of the library to check whether the password matches the hash, rather then using the '==' operator. This potentially helps avoiding [strict comparison](https://stackoverflow.com/questions/28033852/strict-comparison) issues. There are known strict comparison issues in the other languages, such as PHP, see PHP magic hashes, there were practical vulnerabilities based on non-strict comparison, where `"gtahleueylc" == "0e028942374804414919719779763451845794416362003862731183"` and `sha224("gtahleueylc") == "gtahleueylc"`, for example, see the [other examples I discovered](https://github.com/maximmasiutin/PHP-magic-hashes-Open-MPI). A [practical attack is explained in an OWASP Day 2015](https://owasp.org/www-pdf-archive/PHPMagicTricks-TypeJuggling.pdf), but I can send you a PHP script to illustrate the vulnerability. Python is not as vulnerable to type juggling, but anyway consider using the `verify()` function since the library implements it.

dav1312:
> Please consider calling [`verify()`](https://argon2-cffi.readthedocs.io/en/stable/api.html) 

But... verify() is already used?

maximmasiutin:
@dav1312, yes, you are right, the only place where the `==` is used is to compare existing password before hashing it (user["password"] == password). Maybe add type comparison there additionally also to ensure that both variables are of string type, and, when accepting new passwords, also verify that they are strings rather than numbers?

zungur:
You are correct that the comparison with == is only there for the cases when the already existing passwords which haven't been hashed. After user's first authentication, plaintext will be automatically replaced. In time, when all active users' passwords are hashed (definition of active is pending) this part of the logic will be removed. New user's passwords will be hashed immediately before db insert as well, so won't be subject to this conversion.

Is there a way to pass an input to the password field in a Mako template file, which we retrieve with urllib, such that type of input isn't string but int/float/bool/none? In that case, the right way to do this is to fetch the existing plaintext password, check type (we don't know if what we have in the db is a string either), and compare this to the form field submission's type too.

Since we aren't using PHP, I'd like to focus on potential vulnerabilities from Python and Mako format in this case.

server/fishtest/userdb.py
zungur: I'm using the current default parameters explicitly to ensure future consistency in case the defaults are changed by the package maintainers

PR #1376, opened 2022-07-12
Cards for cores, nps...

dav1312:
The idea was popular in the Discord server so I created it
Cards can be removed, modified or reordered if necessary

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -84,6 +84,13 @@
   .offcanvas-body::-webkit-scrollbar {
     display: none
   }
+
+  .card-lg-sm {
+    --bs-card-spacer-y: 0.5rem;
+    --bs-card-spacer-x: 0.5rem;
+    --bs-card-cap-padding-y: 0.25rem;
+    --bs-card-cap-padding-x: 0.5rem;
+  }
 }
 
 .main {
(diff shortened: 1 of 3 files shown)


vondele:
I would drop `Avg. nodes / sec / core' can be computed easily enough.

ppigazzini:
DEV updated https://github.com/glinscott/fishtest/pull/1376

vondele:
on desktop, it feels like the info is a bit too dominant. Possibly because the numbers/data is bold and a large font. Maybe also because the cards span the whole width?

dav1312:
> on desktop, it feels like the info is a bit too dominant. Possibly because the numbers/data is bold and a large font.

It uses the same font size as before, `h4`

> Maybe also because the cards span the whole width?

I can try to limit the width

vondele:
yes, the font is the same, but since it is in the cards it becomes more prominent. I think it could probably be the same font size as what is in the LLR boxes.

dav1312:
> yes, the font is the same, but since it is in the cards it becomes more prominent. I think it could probably be the same font size as what is in the LLR boxes.

That looks a little bit ridiculous...

image

ppigazzini:
DEV updated.

ppigazzini:
Update to PROD by hand to collect a clear feedback.

ppigazzini:

PR #1377, opened 2022-07-13
Limit SPSA Parameter dropdown

Disservin:
This changes the amount of parameteres that are shown for the SPSA dropdown before the user has to scroll down.

With this change the SPSA dropdown should look like this:
image

Before it was:
image

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -121,6 +121,11 @@
   border: none;
 }
 
+.dropdown-menu {
+  max-height: 10em;
+  overflow: auto;
+}
+
 a {
   color: #008002;
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated, this SPSA has a long list of parameters:
https://dfts-0.pigazzini.it/tests/view/6164493983dd501a05b0b1fe

ppigazzini:
@Disservin please revert the commit from #1308, with your PR the overlap problem cannot happen.

Disservin:
Should be removed with the latest commit.

ppigazzini (approved):
Looks good to me on PC and mobile.

ppigazzini:
PROD updated, thank you @Disservin :)

I opened #1378 writing "Max A. (Disservin)" as in SF AUTHORS file.
Let me know if you want a change.

ppigazzini:
Updated with the full name.

PR #1379, opened 2022-07-13
Open graph meta tags

dav1312:
This PR aims to add open graph meta tags for a better integration with external websites such as Discord
image

I also added a condition in `tests_finished.mak` in case someone tries to search for green and yellow tests at the same time

I'm not entirely familiar with mako templates so if there is a cleaner way of doing things please tell me

### DIFF
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -262,7 +262,7 @@ td[style*="background-color:#FF6A6A"] {
 }
 
 /* Darken white box */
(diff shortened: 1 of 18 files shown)


dav1312:
Looks like Discord doesn't like `yellow` as a color... *sigh*

ppigazzini:
DEV updated after https://github.com/glinscott/fishtest/pull/1379/commits/552517d661d4541b3ad3f4e7084b665916c9f4f1

dav1312:
I guess I did something wrong...

PR #1381, opened 2022-07-16
show pending, active and finished by default

Disservin:
As mentioned by vondele on discord <https://discord.com/channels/435943710472011776/733545871911813221/997505676182040657>
`probably we should set defaults for 'Pending approval', 'Active' and 'Finished' to show... btw`
So that a user who visits fishtest for the first time immediately sees pending, active and finished tests.

This pr does just that.

### DIFF
--- a/server/fishtest/templates/run_table.mak
+++ b/server/fishtest/templates/run_table.mak
@@ -49,7 +49,11 @@
 <h4>
 % if toggle:
     <button id="${toggle}-button" class="btn btn-sm btn-light border" onclick="toggle_${toggle}()">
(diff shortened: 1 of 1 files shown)


vdbergh:
Hmm. I spent quite a bit of time making "runtable.mak" generic and this PR breaks this badly.

The idiomatic way to implement this PR is to add an extra argument "default_state" to "runtable.mak" (line 1) and then invoke "runtable.mak" with the appropriate arguments in "runtables.mak".

Disservin:
Ah okay, I didn’t know that. This felt more of a cheat anyway. Will have to think about your suggestion thanks

PR #1382, opened 2022-07-17
Update forms

dav1312:
Update events log form to bootstrap 5
Display the selected values after submitting in events log
Make SPRT forms mobile friendly
Update action buttons and forms in test view
Update CSS forms for dark theme

### DIFF
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -36,10 +36,6 @@ a {
   color: #75bb76;
 }
 
-.form-control {
-  background-color: var(--color-dark-1);
-}
-
 pre .hljs {
   color: #c9d1d9;
   background: var(--bg-dark-2);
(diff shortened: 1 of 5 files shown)


ppigazzini:
All my tests on DEV look good. Applied by hand to PROD.
"Events log" has a little change that personally I like: after filtering for a class of event the dropdown menu doesn't fall back to the default "All" allowing to refine the filtering with a username.

ppigazzini:

vdbergh:
An issue with this PR is that it triggers a login dialog for the Event log on an ipad (browser=safari). Presumably because of the hint "username" in the "From user" field. 

In some sense the hint is superfluous since "From user" already adequately describes the field.

vdbergh:
A related cosmetic issue is that on Firefox the "From user" field is yellow and contains the username. The fact that it is yellow indicates that it has been auto-filled (again because of the hint "username"). I am not sure we want this.

ppigazzini:
> A related cosmetic issue is that on Firefox the "From user" field is yellow and contains the username. The fact that it is yellow indicates that it has been auto-filled (again because of the hint "username"). I am not sure we want this.

I replicated the Firefox problem on Windows, it happens after answering "yes" to "save the password" after a login.
I always decline the request of saving the credentials so I missed the problem during the tests.

PR #1388, opened 2022-07-27
Draft for replacing jQuery to JavaScript

peregrineshahin:
Now that we don't need it in Bootstrap 5


ppigazzini:
DEV updated.

peregrineshahin:
Refactored all jQuery from the templates, and fixed bootstrap collapse instance,
I think this is ready to be tested again

ppigazzini:
DEV updated.
```
$ git log --oneline -n 15
fc8bcb6 (HEAD -> master) Merge commit 'refs/pull/1388/head' of https://github.com/glinscott/fishtest
3646fd1 Merge branch 'task_duration_180' of https://github.com/ppigazzini/fishtest
1ea774e missing parentheses
2a32868 fix typo in document.head
b4831c7 Update tests_view.mak
3f8f595 Bootstrap 5 collapse instance
28217ad Update tests_run.mak
7d780a8 Update base.mak
b66085f Update actions.mak
0317aaa convert jQ to JS
917a17c Replace legacy `var` with ES6 `let` and `const`
d54aa7a Convert jQuery to JavaScript
423f4c1 (origin/master, origin/HEAD) Update forms Update events log form to bootstrap 5 Display the selected values after submitting in events log Make SPRT forms mobile friendly Update action buttons and forms in test view Update CSS forms for dark theme
7303459 Raise worker version to 175 (also server side)
e198edb Use ZipFile context manager
```

Not working:
- all show/hide buttons (it works only for workers but after a browser refresh, not with a clean start)
- diff missing in test page
- theme switch
- login "check_csrf_token(): Invalid token"

Solved:
- SPSA graph

Working:
- column content sorting
- SPRT calc
- liveElo graph (not tested yet the live updated)
- show/hide default nets

peregrineshahin:
seems like one problem in the bootstrap collapse instance is making the script stop resulting is other problems,
and seems like I'm assigning a bootstrap instance to the `div`s target before they are loaded;

will have a look on how I can solve this

ppigazzini:
Take your time, I know very little javascript so my help is limited to testing the PR.

ppigazzini:
DEV updated.

ppigazzini:
This throws error when not in homepage:
```js
  const machinesButton = document.querySelector("#machines-button");
  machinesButton?.addEventListener("click", (e) => {
```
This throws error when in homepage:
```js
  const tasksButton = document.querySelector("#tasks-button");
  tasksButton?.addEventListener("click", (e) => {
```

Commenting out the second one, and loading the homepage, the theme switch works fine.

ppigazzini:
DEV updated, the `collapse,js` error in console seems gone in every page.

In a test page (https://dfts-0.pigazzini.it/tests/view/62e0ba678e4fa6ae4726947e) there are 2 SyntaxErrors:

image

ppigazzini:
Login works now (use "user01"/"user01" to login as normal user and "user00"/"user00" as admin)!
Error in the new test page.

image

ppigazzini:
DEV updated, I was able to submit a new SPRT test.
However, I was not able to change the type of test (eg NumGames, SPSA) or TC.

ppigazzini:
DEV updated. In homepage the show/hide buttons are changing status, to load the table require a F5, though.

peregrineshahin:
I will try to fix that lastly as I'm thinking about using CSS classes and not using this bootstrap instance instead, will see,

there are still some errors I still see jQuery even in the last commit and there are a couple of syntax errors,

ppigazzini:
I'm thinking about this PR as a marathon, not a 100m dash. There are so much things to be well tested before merge. Take your time.

server/fishtest/templates/tests_view.mak
```diff
@@ -421,46 +421,46 @@ if 'spsa' in run['args']:
...
+          copyDiffBtn && copyDiffBtn.style.display = "none";
+          toggleBtn.innerText = "Show";
         }
         document.querySelector("#diff-section").style.display = "block";
         hljs.highlightElement($diffText[0]);
 
 
         // Show # of comments for this diff on Github
-        fetch(diffApiUrl).then((response) => {
+        fetch(diffApiUrl).then(response => response.json()).then(text => {
```
peregrineshahin, line 461: todo: this text variable should actually be renamed json in both places here

ppigazzini:
DEV updated.
2 Uncaught SyntaxError in console for a test page.

ppigazzini:
DEV updated.

ppigazzini:
DEV updated

ppigazzini:
Updated

ppigazzini:
Updated and added by hand in `tests.mak`, `tests_view.mak`:
```python
<%
from fishtest.util import get_cookie
%>
```

All show/hide buttons seems to work fine in a fast test.

ppigazzini:
If I recall correctly the first code for show/hide buttons was all in `application.js`.

Please mind that the function `get_cookie()` was added as workaround for a bug with chrome, see
https://github.com/glinscott/fishtest/commit/fdee74e066152811d3da23af017e471f9aee5fb6

ppigazzini:
Updated.
In `tests_view.mak` the import of `get_cookie` can be simplified in this way:
```
<%
from fishtest.util import get_cookie, worker_name

if 'spsa' in run['args']:
  import json
  spsa_data = json.dumps(run["args"]["spsa"])
%>
```

ppigazzini:
Pyramid works in this way.
In `views.py` we set the mako template to be used to render the page and the variables passed to that template, eg

https://github.com/glinscott/fishtest/blob/7a732ef837612cca4113540e4a4dfcdd8d2ff434/server/fishtest/views.py#L1294-L1325

Please mind the last lines, where we read a cookie to set a variable used in the mako template.

Here is some documentation about pyramid request, views etc.

https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#request-module
https://docs.pylonsproject.org/projects/pyramid/en/latest/glossary.html#term-request
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html

Here is some doc about mako template (default template in Pyramid):
https://www.makotemplates.org/
https://docs.makotemplates.org/en/latest/

ppigazzini:
DEV updated.

server/fishtest/templates/tests_run.mak
```diff
@@ -409,157 +410,165 @@ else:
...
-            $('#base-branch').removeAttr('readonly').val(initial_base_branch);
-            $('#base-signature').removeAttr('readonly').val(initial_base_signature);
-            $('#test-branch').off('input');
-            $('#test-signature').off('input');
+            document.querySelector('#base-branch').removeAttribute('readonly');
+            document.querySelector('#base-branch').value = initial_base_branch;
+            document.querySelector('#base-signature').removeAttribute('readonly');
+            document.querySelector('#base-signature').value = initial_base_signature;
+            document.querySelector('#test-branch').removeEventListener('input', testBranchHandler);
+            document.querySelector('#test-signature').removeEventListener('input', testSignatureHandler);
```
dav1312, line 497: Instead of using `querySelector()` for IDs why not just use `getElementById()`

dav1312:
Instead of using `querySelector()` for IDs why not just use `getElementById()` ?

peregrineshahin:
> Instead of using `querySelector()` for IDs why not just use `getElementById()` ?

I think it's becoming the standard way of performing selections on elements, since it is now supported by all major browsers.

For consistency, it is probably best to choose one and stick with it throughout your scripts. It just looks less messy.

ppigazzini:
Updated

ppigazzini:
DEV updated. Git diff is finally there but with wrong format.

peregrineshahin:
the last patch escaping HTML and using `textContent` to display code is the equivalent for jQuery.text() which turns out to escape HTML by default and attach text node to the `<code>`, this also seems necessary for highlight.js as code should not be rendered as is, or else considered security attach/injection

ppigazzini:
Updated. Diff text shown correctly. Highlight works fine for dark theme, not for the light theme.
The button label is split on several rows, see PROD vs DEV:
https://tests.stockfishchess.org/tests/view/62e0c5fe8e4fa6ae4726964c
https://dfts-0.pigazzini.it/tests/view/62e0c5fe8e4fa6ae4726964c

ppigazzini:
Updated. Diff windows looks good now with either themes. The only minor difference with PROD is that the "Copied!" message appears on the following row. I played unsuccessfully with the html tags trying to fix.

dav1312:
Hide and Show buttons are reversed in some places

image

image

peregrineshahin:
there is still design issue when showing diff the page expands width
I'm looking into that rn,
in regards to text reversed I think the lame solution would be to reverse the condition
but I'm looking forward to understanding the reason why in view.py it says
```"machines_shown": request.cookies.get("machines_state") == "Hide",```

**Comment by ppigazzini** [comment]
Updated.

**Comment by ppigazzini** [comment]
Updated.
Diff "Copied!" fixed.
Show/Hide button seem working fine.
IMO the default should be "table hidden".
Please mind that tasks table (rarely read by the user) for some test can be huge, taking high memory and time to load in the browser, see https://github.com/glinscott/fishtest/pull/1316

**Comment by ppigazzini** [comment]
Updated, show/hide looks good to me!

**Comment by dav1312** [comment]
Machines and tasks are now always hidden by default, regardless of cookies

**Comment by peregrineshahin** [comment]
Yeah that's because of non-consistent naming between states in PROD,

other buttons have the format like "cookiename-state" while machines and tasks now that I see it have tasks_state and machines_state

will fix this now

**Comment by dav1312** [comment]
> Yeah that's because of non-consistent naming between states in PROD,

Consistency is nice... would be nice to have them all use the same one

**Comment by ppigazzini** [comment]
Updated.

**Comment by ppigazzini** [comment]
Updated. The In my tests the browser starts with the right default, the browser load the right info from the pages cookie when restarting.

**Comment by ppigazzini** [comment]
I think to have hit a bug that IMO already affects master.
I created several tests on DEV (SPRT, NumGames, SPSA), testing the different options and triggering the GitHub API rate limit.
The Diff block shows:
```
{"message":"API rate limit exceeded for 93.56.160.56. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```
The consoles has these errors (the first two are the error messages show above):
image

Here is the code where happens the third error:
image

**Comment by ppigazzini** [comment]
Live update for "Live Elo" page works.

File: server/fishtest/static/js/application.js [hunk 3]
```diff
@@ -1,102 +1,116 @@
...
+        .sort(
+          comparer(
+            Array.from(th.parentNode.children).indexOf(th),
+            (this.asc = !this.asc)
+          )
+        )
+        .forEach((tr) => body.appendChild(tr));
+    }
+  });
+  const nonDefaultButton = document.querySelector("#non_default-button");
```
ppigazzini, line 52: I'm wondering if it's possible to move the `#non_default-button` code in `nns.mak` like `"#machines-button"` (code moved in `tests.mak`) and `"#tasks-button"` (code moved in `tests_view.mak`)

peregrineshahin:
Another good solution to not getting request limit rate issue in diff would be to cache the response code (because the diff code doesn't change over time), if such a solution were to take place, I'll be more than happy to help in front-end aspect of it

server/fishtest/templates/tests_run.mak
```diff
@@ -392,13 +394,14 @@ else:
...
     // If pressing the 'back' button to get back to this page, make sure
     // the submit test button is enabled again.
-    $('#submit-test').removeAttr('disabled').text('Submit test');
+    document.querySelector('#submit-test').removeAttribute('disabled');
+    document.querySelector('#submit-test').innerText = 'Submit test';
     // Also make sure that the odds TC fields have the right visibility.
-    update_odds($('#checkbox-time-odds')[0]);
-  });
+    update_odds(document.querySelector('#checkbox-time-odds'));
+  };
```
peregrineshahin, line 404: Issue: 
The scenario is if SPSA was clicked then Base branch and Base signature empty out content which is correct but when reaching the run_tests page with (Back/Forward) the stop rule enabled is SPRT then Base branch and Base signature are not refilled at all

Todo: Make sure Base branch and Base Signature are refilled when window.onpageshow

ppigazzini:
Updated:
 - non default nns cookie works fine
 - diff in test page is now broken `Uncaught SyntaxError: Unexpected token '}' (at 62e1b66c8e4fa6ae4726bbfe:2247:3)`
 
image

I don't recall someone reporting the rate limit message in the diff window (hard to be ignored because the long one line message pushes the "Actions", "Stats", "Time" fields in the far right). I triggered the rate limit because I created a very high number of new runs to test the vanilla js code.
If a proper solution is hard to be coded or is against the best practice, for me me is fine to revert https://github.com/glinscott/fishtest/pull/1388/commits/517999b03dd154242f344b5bd5b9dafe6845a9b0

server/fishtest/templates/tests_run.mak
ppigazzini: I suggests to keep in this PR only the conversion from jQuery to vanilla js (this is valid also for the github rate limit).
When frozen, we can test the PR on PROD.

dav1312:
What about checking if the user has reached its rate limit before making a request? https://api.github.com/rate_limit

ppigazzini:
> What about checking if the user has reached its rate limit before making a request? https://api.github.com/rate_limit

Nice idea but IMO better to no add new features to this PR (BTW I started requesting new features so I'm to blame here :) )

peregrineshahin:
Please notice that the jQuery Ajax solution(master) also has the same problem of 403 (the rate limit problem)with diff box, I have been experiencing it since a very long time, when you refresh enough times on a particular test diff doesn't show anymore,
if you tried the same test on the master you will also get the same error.

ppigazzini:
> I think to have hit a bug that IMO already affects master.

I do know, I wrote it in my old post.

peregrineshahin:
I think this should make it equivalent to master
didn't notice at first that there are nested fetching in the code

ppigazzini:
DEV Updated. Looks good to me!

PROD updated by hand to test the PR with the sharks :)

ppigazzini:
I will run PROD with this PR for few days to collects feedbacks from the users.

@peregrineshahin please at your convenient time squash the commits or in one commit or in few logical commits, writing a/some final commit message/s (I use `git rebase -i master` to do the work, ymmv)

vdbergh:
Very substantial PR!

ppigazzini:
DEV updated, I submitted 11 tests in a row using the back button, it seems to work (with the previous code I got a random true/false value for `form_submitted` when pushing the back button).
PROD updated by hand to have a bigger number of testers.

peregrineshahin:
Just for the record, I don't think you get random as garbage memory random before the fix, I think the randomity comes from if the back button triggered a full reload or rendered the cached page and that depends on how fast you click it - it seems (bc I'm still not able to reproduce it)-

please @ppigazzini feel free to format the PR with Prettier when merging, and I think I can squash the commits after learning how git commands, tonight or tomorrow

ppigazzini:
@peregrineshahin 
```
git checkout jQ-to-JS
git checkout -b jQ-to-JS-backup
git checkout master
git fetch -p upstream
git reset --hard upstream master
git checkout jQ-to-JS
git rebase -i master
# change all the "pick" to "squash" but the commits that you want to keep, save and exit
# write the final commit messages for the commits, save and exit
# if a commit doesn't depend on previous ones, you can change the order moving the line
git push -f origin jQ-to-JS
```

dav1312 (approved):

server/fishtest/templates/run_table.mak
```diff
@@ -34,13 +34,13 @@
...
 % if toggle:
 <script>
-  function toggle_${toggle}() {
-  const button = document.querySelector("#${toggle}-button");
-  const div = document.querySelector("#${toggle}");
-  const active = button.innerText.trim() === "Hide";
-  button.innerText = active ? "Show" : "Hide";
-  document.cookie =
-    "${cookie_name}" + "=" + button.innerText.trim() + ";max-age=315360000";
+  function toggle_k() {
```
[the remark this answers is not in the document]
peregrineshahin: Changing the name was to force Prettier accept the compilation of formatting as mako templates are not supported
so I  changed the same back in the last force pushes

server/fishtest/templates/tests_run.mak
```diff
@@ -432,118 +436,160 @@ else:
...
+              document.querySelector("#base-branch").value);
+          document
+            .querySelectorAll(".stop_rule")
+            .forEach((el) => (el.style.display = "none"));
+          document.querySelector("#stop_rule_field").value = stopRule;
+          document
+            .querySelectorAll("." + stopRule)
+            .forEach((el) => (el.style.display = ""));
+          % if not is_rerun:
+          if (stopRule === "spsa") {
```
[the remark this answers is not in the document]
peregrineshahin: this solved in the last commit, please confirm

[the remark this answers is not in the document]
peregrineshahin: this is the commit to review 
https://github.com/glinscott/fishtest/commit/7c356e29272e614f252a32035863ea12bbbf2b5d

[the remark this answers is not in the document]
dav1312: Yes, I marked the review as approved
I opened the review a few days ago and approved it now bcs I saw that u fixed them
I still don't fully understand how reviews work tho

dav1312 (changes requested):

ppigazzini:
PROD and DEV updated.

ppigazzini (approved):

ppigazzini:
Merged, PROD updated, thank you @peregrineshahin :)

PR #1389, opened 2022-07-28
Fix bug in help message.

vdbergh:


ppigazzini:
Good catch, triggered the workers update.

PR #1392, opened 2022-07-30
New SPRT bounds

vondele:
introduce new bounds, requiring more games to pass, but makes passing more feasible.


ppigazzini:
I'm wondering if need the update the custom Elo1 bound too:
https://github.com/glinscott/fishtest/blob/7a732ef837612cca4113540e4a4dfcdd8d2ff434/server/fishtest/templates/tests_run.mak#L244-L247

vondele:
I don't think we need to update the custom bound.. it is custom after all. It is also almost never used.

At first sight dev looks good.

ppigazzini:

vondele:
Thanks!

dav1312:
> I don't think we need to update the custom bound.. it is custom after all. It is also almost never used.

They need to be updated or the script needs to be changed because the values in the **custom bounds are the default bounds**

That's why a lot of users are currently having issues not being able to submit STC tests with the correct bounds

image

ppigazzini:
Custom elo1 fixed by hand on PROD, tonight I will double check the fix.

Issue #1393, opened 2022-07-30
split worker for reduced concurrency


vondele:
it seems like the time loss problem is getting more acute with multiple large core workers. 
See e.g. https://tests.stockfishchess.org/tests/view/62e523e2b383a712b1386193 
We know this is probably due to cutechess not being able to deal with a large concurrency, and probably our best workaround is to split the worker internally (so not visible from the user side), to have multiple cutechess processes each with reduced concurrency.

dubslow:
No matter what the prevention solution is -- fixing cutechess or making workarounds inside the worker -- fishtest should be able to manage tasks with timelosses separately from high residual tasks (e.g. rejecting them or purging them etc)

vdbergh:
technologov-28cores-r345 has lots of issues. See e.g. #1360.

vondele:
I've pinged him on discord. But there are at least two other workers with a lot of losses.

ppigazzini:
Script for linux:
- creates the "fishtest" user
- creates 5 copies of the "worker" directory to be able to run 5 workers
- creates an unit systemd file (name "fishtest") that takes the worker ID as parameter
- start, stop, check the status for any single worker with `sudo systemctl start fishtest@3`
- start, stop all the workers with `sudo systemctl start fishtest@{0..4}`
- set the auto start for the workers with `sudo systemctl enable fishtest-worker@{0..4}.service`

To revert:
- `sudo systemctl disable fishtest-worker@{0..4}.service`
- `sudo rm /etc/systemd/system/fishtest-worker@.service`
- `sudo deluser --remove-home fishtest` (this delete the fishtest user and all his files/folders) 

```bash
#!/bin/bash
# setup_worker.sh
# to setup a fishtest worker on Ubuntu 20.04, simply run: 
# sudo bash setup_workers.sh 2>&1 | tee setup_workers.sh.log

# print CPU information
cpu_model=$(grep "^model name" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
n_cpus=$( grep "^physical id" /proc/cpuinfo | sort | uniq | wc -l)
online_cores=$(grep "^bogo" /proc/cpuinfo | wc -l)
n_siblings=$(grep "^siblings" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
n_cpu_cores=$(grep "^cpu cores" /proc/cpuinfo | sort | uniq | cut -d ':' -f 2)
[...]

PR #1394, opened 2022-08-02
Untested try at rejecting workers submitting bad data

dubslow:
Marked as draft because it needs review/testing but in principle, it should be merged as soon as possible.

This addresses server-side filtering of problematic tasks caused by #1393. Obviously we should also prevent this worker-side problem in the first place, but in the short run this should at least prevent fishtests from being polluted by garbage as they currently are

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -912,6 +912,13 @@ class RunDb:
                 error = "Update_task: the number of games received for task {}/{} is incompatible with the SPRT batch size".format(
                     run_id, task_id
                 )
+        elif    ("stats" in task
+            and ((task["stats"]["time_losses"] + task["stats"]["crashes"])*100 > num_games)
+            # Allow workers at most 1% bad data, however patches which break TM need special handling (TODO?)
+            and error == ""):
+            error = "Update task: task {}/{} has too many crashes+timelosses ({}+{})".format(
+                run_id, task_id, task["stats"]["crashes"], task["stats"]["time_losses"]
+            )
 
         if error != "":
             print(error, flush=True)
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -912,13 +912,13 @@ class RunDb:
                 error = "Update_task: the number of games received for task {}/{} is incompatible with the SPRT batch size".format(
                     run_id, task_id
                 )
-        elif    ("stats" in task
-            and ((task["stats"]["time_losses"] + task["stats"]["crashes"])*100 > num_games)
+        elif "stats" in task and error == "":
+            crashes, time_losses = task["stats"].get("crashes", 0), task["stats"].get("time_losses", 0)
             # Allow workers at most 1% bad data, however patches which break TM need special handling (TODO?)
-            and error == ""):
-            error = "Update task: task {}/{} has too many crashes+timelosses ({}+{})".format(
-                run_id, task_id, task["stats"]["crashes"], task["stats"]["time_losses"]
-            )
+            if (crashes + time_losses) * 100 > num_games:
+                error = "Update task: task {}/{} has too many crashes+timelosses ({}+{})".format(
+                    run_id, task_id, crashes, time_losses
+                )
 
         if error != "":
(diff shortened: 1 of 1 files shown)


dubslow:
Well at least it passes CI now, still can't say what it might do in production, I hope someone more knowledgeable can work on this and merge it

vdbergh:
On issue with this PR is that it run after every update. So if if a task start with a time loss the task will be cancelled, even if the the total number of time losses turns out to be less than 1%. 

As an alternative to this PR we could also simply turn on auto purge again...

vdbergh:
An advantage of the current PR would be that the offending workers would be flagged in the event log. The owner could then split the worker into several lower core ones. 

Of course doing this internally in fishtest would be preferable but it is a substantial task.

dubslow:
Ah, I see. Thanks for the review. Yes I concur that rejecting a 1/1 timeloss isn't ideal, altho it would have been an improvement when many workers were doing too many of these.

So `purge_run`, https://github.com/glinscott/fishtest/blob/d5ea015623a94f95295fa9b71443705a507ba75e/server/fishtest/rundb.py#L1062, when does this function get called? How does a user trigger a purge on fishtest? Is that admin only?

vondele:
both purging and deleting these batches is a hack... this needs to be solved at the root IMO. Cutechess simply can't deal with the concurrency. Even if there are no time losses, this could severely distort time management nevertheless. The only real solution IMO is to internally split the worker.

We know the purging doesn't work well with NNUE, as machines with different hardware architecture are different.

dubslow:
> both purging and deleting these batches is a hack... this needs to be solved at the root IMO.

Installing bad-data filters on fishtest isn't a hack, it's simply good engineering.

That said, any sources of bad data should also be solved at the root, prevented from occurring in the first place. By no means are the two approaches mutually exclusive. Both strategies should be pursued to completion

vdbergh:
@vondele 

> We know the purging doesn't work well with NNUE, as machines with different hardware architecture are different.

I think this problem is gone now. At least the chi^2 test does not indicate systemic deviating workers (other than time losses).

Perhaps this has to do with the switch to the UHO book. The noob book with its extremely high draw ratio was perhaps more sensitive to small hardware differences.

vondele:
I think this would still show for tests of new NNUE architectures. 

However, if we don't fix the problem with concurrency, we'll just remove all contributions from linrock with his 95 core servers, and sebastronomy with 48 cores.

dubslow:
> However, if we don't fix the problem with concurrency, we'll just remove all contributions from linrock with his 95 core servers, and sebastronomy with 48 cores.

We should of course fix the cause of the symptom, but nevertheless bad data is bad data and should be removed from test statistics.

That said, I now believe this PR isn't the best way to go about filtering the bad data. I think being able to separately purge crashes/timelosses from "standard" residuals is the way to go. (I'm fine with keeping autopurging off, it's not a very future-proof tool.) The main problem is I don't know where the purge button is for existing tests on the site.

vondele:
image

The red 'purge' button

dubslow:
Ahhhhhh. I think, back when I was first poking around, I assumed that meant "purge from the DB", not anything relating to individual tasks... thereafter it was forever immediately filtered from my vision whenever I load a page...

vdbergh:
> I think this would still show for tests of new NNUE architectures.
> 
> However, if we don't fix the problem with concurrency, we'll just remove all contributions from linrock with his 95 core servers, and sebastronomy with 48 cores.

@vondele We would encourage them to to split up their workers into several lower core workers. This is a completely trivial operation (just copy the worker directory and restart the workers with a lower number of cores).

I agree this is not ideal but doing the splitting internally in the worker is not a trivial operation and a volunteer needs to step up to do it.

Note: while the linrock workers seem to genuinely suffer from the high core cutechess issue, I am not so sure about sebastronomy. Technologov has several 56 core workers and these seem to run without issues.

dav1312:
> I assumed that meant "purge from the DB"

I thought so too... maybe the name should be changed to something like "Clear tasks"?

PR #1395, opened 2022-08-04
Minor style changes

dav1312:
Dark theme checkbox
Darkt theme hover file selector
Autofocus username in login and signup
Pointer style cursor for checkbox and checkbox label
Bootstrap alert for live elo warning

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -541,3 +541,8 @@ option.grayedoutoption {
 .monospace {
   font-family: var(--bs-font-monospace);
 }
+
+input[type="checkbox"],
+.form-check-label {
+  cursor: pointer;
+}
--- a/server/fishtest/static/css/live_elo.css
+++ b/server/fishtest/static/css/live_elo.css
@@ -1,7 +1,5 @@
 div.error {
   display: none;
-  color: black;
-  text-align: center;
 }
 
 span.pointer {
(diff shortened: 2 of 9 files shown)


ppigazzini:
If I'm not wrong the alert for Live Elo is unreachable because `retry == true`:

https://github.com/glinscott/fishtest/blob/0150bbf3ce467b5135704ac35190aba19ee07e16/server/fishtest/static/js/live_elo.js#L18-L19
https://github.com/glinscott/fishtest/blob/0150bbf3ce467b5135704ac35190aba19ee07e16/server/fishtest/static/js/live_elo.js#L224-L225
https://github.com/glinscott/fishtest/blob/0150bbf3ce467b5135704ac35190aba19ee07e16/server/fishtest/static/js/live_elo.js#L247-L256

So this is debug code (aka dead code):
https://github.com/glinscott/fishtest/blob/0150bbf3ce467b5135704ac35190aba19ee07e16/server/fishtest/static/js/live_elo.js#L210-L221

`retry == true also before the last major refactoring https://github.com/glinscott/fishtest/pull/1334 , but in that case `alert_()` was used in other places:
https://github.com/glinscott/fishtest/blob/4709fa65250248d27d9c6587286437c92ee6acab/server/fishtest/static/js/live_elo.js

ppigazzini (approved):
Looks good to me

ppigazzini:

PR #1396, opened 2022-08-04
Save diffs in sessionStorage

dav1312:
Draft in case someone wants to make suggestions and because it has debug code

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -453,58 +453,82 @@ if 'spsa' in run['args']:
       "//api.github.com/repos/"
     );
 
-    fetch(diffApiUrl, {
-      headers: {
-        Accept: "application/vnd.github.v3.diff",
-      },
-    })
-      .then((response) => {
-        if (response.ok) return response.text();
+    // Check if the diff is already in sessionStorage and use it if it is
+    const diffId = sessionStorage.getItem("${run['_id']}");
+    if (!diffId) {
+      fetchDiff();
+    } else {
+      showDiff(diffId);
+    }
+
+    function fetchDiff() {
+      fetch(diffApiUrl, {
+        headers: {
+          Accept: "application/vnd.github.v3.diff",
+        },
       })
-      .then((text) => {
-        if (!text) return;
-        const numLines = text.split("\n").length;
-        const toggleBtn = document.querySelector("#diff-toggle");
-        const diffContents = document.querySelector("#diff-contents");
-        const diffText = diffContents.querySelector("code");
-        diffText.textContent = text;
-        toggleBtn.addEventListener("click", function () {
-          diffContents.style.display =
-            diffContents.style.display === "none" ? "" : "none";
-          if (copyDiffBtn)
-            copyDiffBtn.style.display =
-              copyDiffBtn.style.display === "none" ? "" : "none";
-          if (toggleBtn.innerText === "Hide") toggleBtn.innerText = "Show";
-          else toggleBtn.innerText = "Hide";
+        .then((response) => {
+          if (response.ok) return response.text();
+        })
+        .then((text) => {
+          if (!text) return;
+          showDiff(text);
         });
+    }
 
-        // Hide large diffs by default
-        if (numLines < 50) {
-          diffContents.style.display = "";
-          if (copyDiffBtn) copyDiffBtn.style.display = "";
-          toggleBtn.innerText = "Hide";
-        } else {
-          diffContents.style.display = "none";
-          if (copyDiffBtn) copyDiffBtn.style.display = "none";
-          toggleBtn.innerText = "Show";
-        }
-        document.querySelector("#diff-section").style.display = "";
-        hljs.highlightElement(diffText);
+    function showDiff(text) {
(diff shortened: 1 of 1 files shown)


peregrineshahin:
I don't think saving the diff in local storage is going to be a good practice, as some may find saving a whole block of code inside their browser storage does feel a bit malicious

dav1312:
> saving a whole block of code inside their browser storage does feel a bit malicious

I think that if the user is knowledgeable enough to even check its localStorage it shouldn't have a hard time understanding what the contents are or just checking the source code

ppigazzini:
DEV updated.

peregrineshahin:
If you insist on the idea, I suggest that you minimize the save to only the session storage and not the local storage, because in real situation, refreshes are made during the same session, and deleting those when user exist browser would make more sense

ppigazzini:
Some thoughts:
- it's more dangerous running the code on the worker than saving a code diff in the browser cache. The browser load already that code diff
- a caching of the diff should not done on the server: the diff could be big (16 MB max size for a MongoDB document), delay downloading the data from PROD, PROD could hit the GH API rate etc

peregrineshahin:
Thoughts:

I picked a random diff size calculating the difference by this script one before saving any diff and another time after saving one:
using this script:
```
 let limit = 1024 * 1024 * 5; // 5 MB
 let remSpace = limit - unescape(encodeURIComponent(JSON.stringify(sessionStorage))).length;
 remSpace ``` in chrome 
 
 The Session size limit per domain in chrome is 5 MB = 5242880
 one random diff size = 5242880 - 5241359 = 1521 B
 How many diffs should be stored before the browser stops saving and hits and error (no overwriting) =
           5242880 / 1521 ≈ 3447 diffs
  
 therefore this amount of diffs is practically impossible to fetch by a sane user, so I think it is a safe solution minding that the code saved is publicly available and is not sensitive and open source,
 performance wise it's surprisingly blazingly fast, LGTM

**Comment by dav1312** [comment]
> 5242880 / 1521 ≈ 3447 diffs

Actually, if the user is _really_ stubborn is not hard to ran out of sessionStorage space
Regression tests have pretty big diffs (200-300kB), it would fill with less than 20 tests

**Comment by ppigazzini** [comment]
```
ls -lah big_diff_00*
-rw-r--r-- 1 usr00 usr00 1.4M Aug  5 13:15 big_diff_000.txt
-rw-r--r-- 1 usr00 usr00 4.1M Aug  5 13:14 big_diff_001.txt
-rw-r--r-- 1 usr00 usr00 7.5M Aug  5 13:20 big_diff_002.txt
```
https://dfts-0.pigazzini.it/tests/view/62ecf2b5921046e23f7fc1ac
https://dfts-0.pigazzini.it/tests/view/62ecf9f7921046e23f7fc1ae
https://dfts-0.pigazzini.it/tests/view/62ecfc44921046e23f7fc1b0

Master has no problem loading the diff for any tests, the PR only if there is space left in SessionStorage, otherwise we get this error:
```
62ecfc44921046e23f7fc1b0:631 Fetching diff
62ecfc44921046e23f7fc1b0:661 Uncaught (in promise) DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of '62ecfc44921046e23f7fc1b0' exceeded the quota.
    at showDiff (https://dfts-0.pigazzini.it/tests/view/62ecfc44921046e23f7fc1b0:661:22)
    at https://dfts-0.pigazzini.it/tests/view/62ecfc44921046e23f7fc1b0:649:11
showDiff @ 62ecfc44921046e23f7fc1b0:661
(anonymous) @ 62ecfc44921046e23f7fc1b0:649
Promise.then (async)
fetchDiff @ 62ecfc44921046e23f7fc1b0:647
(anonymous) @ 62ecfc44921046e23f7fc1b0:632
```

I'm wondering if we can deal with this exception (I don't know JS).

BTW there is a library to compress text in storage with a nice demo page to test size and time for LZ-String and LZMA-JS (but the original text size seems to be doubled...):
https://pieroxy.net/blog/pages/lz-string/index.html
https://pieroxy.net/blog/pages/lz-string/demo.html

- big_diff_000.txt:

image

- big_diff_001.txt:

image

- big_diff_002.txt:

image

ppigazzini:
DEV updated, LGTM :)

ppigazzini:
DEV updated, I will review after a fast lunch.

ppigazzini (approved):
Looks good to me, tested on Firefox, Chrome, Edge, Opera, Chrome mobile, Edge mobile.

Checked github api rate usage with https://api.github.com/rate_limit

ppigazzini:

ppigazzini:
BTW with previous master any F5 consumed 2 github api calls, so this PR should be a life changer for some users.

peregrineshahin:
So I tried using the library @ppigazzini referenced above and it does a great job compressing the diff text, but I don't know about how malicious the text looks in session storage though

should do we it?
Capture

ppigazzini:
I do like to post in the PR/Issue some links than can be useful right now or for future improvements.
IMO at the moment the sessionStorage is fine as in master, that library could be useful if we start saving other info in sessionStorage.

PR #1397, opened 2022-08-05
Drop debug code from live_elo.js

ppigazzini:


vdbergh:
I think you can probably also drop live_elo.css.

These were things from the time when live_elo.html wasn’t integrated in Fishtest.

PR #1398, opened 2022-08-06
Clean up more code for live_elo page

ppigazzini:
Companion PR for https://github.com/glinscott/fishtest/pull/1397

Thanks to @vdbergh for the idea and @dav1312 for the review.


ppigazzini:
DEV updated, LGTM, please @dav1312 have a look at your convenient time.

server/fishtest/static/css/live_elo.css
```diff
@@ -1,30 +0,0 @@
...
-  color: #5f9960;
-}
-
-a:focus {
-  color: #497449;
-}
-
-td:last-child {
-  word-break: break-word;
-}
```
dav1312, line 30: Looks like removing this is adding a very small horizontal scroll on my phone

server/fishtest/static/css/live_elo.css
```diff
@@ -1,30 +0,0 @@
-div.error {
-  display: none;
-}
```
dav1312, line 3: We could also remove all `error` related code right?

https://github.com/glinscott/fishtest/blob/2dc3970bf45ad5c3415e2e658cffed9e1d7b4d3c/server/fishtest/templates/tests_live_elo.mak#L25

https://github.com/glinscott/fishtest/blob/2dc3970bf45ad5c3415e2e658cffed9e1d7b4d3c/server/fishtest/static/js/live_elo.js#L146

vdbergh: I think so.

server/fishtest/static/css/live_elo.css
vdbergh: I find horizontal scrolling rather annoying. Perhaps we could move this to application.css?

server/fishtest/static/css/live_elo.css
ppigazzini: Correct, the alert_() function was dropped in #1397
DEV updated.

server/fishtest/static/css/live_elo.css
ppigazzini: Confirmed. Embarrassedly I haven't idea how to fix...

ppigazzini: I added a style tag in the mako template to break word for the last child, it works.

PR #1399, opened 2022-08-06
Refill base branch/signature when page is shown

peregrineshahin:
When the run test page is reached with 'back'/'forward' button and the test is not a reschedule, Make sure that the base branch and base signature are refilled if the last stop rule has set their value to an empty string (e.g. last stop rule chosen is SPSA)

I'm expecting this to be tested first in DEV

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -395,11 +395,17 @@ else:
 
 <script>
   let form_submitted = false;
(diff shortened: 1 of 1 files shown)


peregrineshahin:
I will close this as there seem to be a conflict in the behavior of the back button between browsers and devices

I'm now able to reproduce the old issues only in mobile, and I see now this makes no sence as in my phone the back button doesn't set the info button back to SPRT

PR #1400, opened 2022-08-07
try: except: block to catch recent "FileNotFoundError".

vdbergh:


vdbergh:
Made this WIP since there is an issue with error reporting.

vdbergh:
Fixed.

vdbergh:
Most likely the FileNotFoundError is a red herring. Probably the cutechess command line for the tuning test was just too long (I think the FileNotFoundError occurs only on windows).

vdbergh:
@ppigazzini Is anything stopping committing this trivial PR? Each time a FileNotFoundError occurs (cfr the event log) the affected worker will quit (an unclassified exception is a fatal error). It will encourage users to run their workers in a loop (which is not recommended in general).

peregrineshahin:
I believe this error occurs back before this PR predicts, it is caused by a large JSON parsing overflowing the Memory because of bad python's implementation of `response.json()` which is at `send_api_post_request` call in games.py
the bad implementation is the whole JSON file gets loaded wholly before it gets parsed no chunking or iterating is used which consumes too much RAM

I believe following down the road to cutechess there should be a safe way of handling this exception,

but the root cause should be solved also
I talked about this issue, the suggested solution, and the resources of problems with JSON with python  in the discord channel
after it got reported with more info in hand by Bryan_Stockfish
please check and following messages:
https://discord.com/channels/435943710472011776/900771437177094156/1012926646468227123

vdbergh:
For me the logical explanation for this error is a cutechess-cli command line that is too long for Windows (note that the error only happens on Windows).

ppigazzini (approved):
This should be used only to debug the problem, the 3 nested try are a bit ugly.
Not related to this PR, we should change the bare except after the wait with a TimeoutExpired exception:
```python
                    p.wait(timeout=CUTECHESS_KILL_TIMEOUT)
                except TimeoutExpired:
                    print("timeout", flush=True)
```
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait
https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired

vdbergh:
I don't think it is ugly. It is a standard pattern used in many places in the worker: Popen combined with catching OSError and SubprocessError.

ppigazzini:
The try added in this PR is fine, this 3 level try is ugly.
https://github.com/glinscott/fishtest/blob/599898efbd0488741eb67eb413d7daff64f32185/worker/games.py#L1025-L1043

IMO this should work, dropping one nested try and dropping a bare except.
```python
            finally:
                # We nicely ask cutechess-cli to stop.
                try:
                    send_sigint(p)
                except Exception as e:
                    print(
                        "\nException in send_sigint:\n", e, sep="", file=sys.stderr
                    )
                # now wait...
                print(
                    "\nWaiting for cutechess-cli to finish ... ", end="", flush=True
                )
                try:
                    p.wait(timeout=CUTECHESS_KILL_TIMEOUT)
                except TimeoutExpired::
                    print("timeout", flush=True)
                    kill_process(p)
                else:
                    print("done", flush=True)
```

ppigazzini:

ppigazzini:
> Most likely the FileNotFoundError is a red herring. Probably the cutechess command line for the tuning test was just too long (I think the FileNotFoundError occurs only on windows).

You should be right, see https://github.com/glinscott/fishtest/issues/1360#issuecomment-1248344165

PR #1401, opened 2022-08-08
Tests run redesign

dav1312:
Makes the page responsive
Tests in dev are encouraged, it has lots of changes

---

#### Default
image

#### Definitions on hover
image

image

#### SPSA Selected
image

#### SPSA Autoselect information modal
image

#### Mobile
image

image

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -1,3 +1,9 @@
+:root {
+  --green-1: #008002;
+  --green-2: #5f9960;
+  --green-3: #497449;
+}
+
 @media (max-width: 575.98px) {
   body {
     font-size: 14px;
@@ -144,16 +150,16 @@
 }
 
 a {
-  color: #008002;
+  color: var(--green-1);
   text-decoration: none;
 }
 
 a:hover {
-  color: #5f9960;
+  color: var(--green-2);
 }
 
 a:focus {
-  color: #497449;
+  color: var(--green-3);
 }
 
 .main .default-net {
@@ -542,7 +548,47 @@ option.grayedoutoption {
   font-family: var(--bs-font-monospace);
 }
 
+.form-control[readonly] {
+  background: var(--bs-gray-300);
+}
+
 input[type="checkbox"],
 .form-check-label {
   cursor: pointer;
+}
+
+.no-arrows::-webkit-inner-spin-button,
+.no-arrows::-webkit-outer-spin-button {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+  margin: 0;
+}
+
+.no-arrows {
+  -moz-appearance: textfield;
+}
+
+.list-group-checkable .list-group-item {
+  cursor: pointer;
+}
+
+.list-group-item-check {
+  position: absolute;
+  clip: rect(0, 0, 0, 0);
+}
+
+.list-group-item-check:hover + .list-group-item {
+  background-color: var(--bs-list-group-border-color);
+}
+
+.list-group-item-check:checked + .list-group-item {
+  color: #fff;
+  background-color: var(--green-3);
+}
+
+.list-group-item-check[disabled] + .list-group-item,
+.list-group-item-check:disabled + .list-group-item {
+  pointer-events: none;
+  opacity: 0.5;
 }
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -8,11 +8,10 @@
   --bg-dark-5: #555;
   --color-dark-1: #b1ada7;
   --color-dark-2: #9b9b9b;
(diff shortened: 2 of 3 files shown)


ppigazzini:
DEV updated.

dav1312:
> DEV updated.

I don't see any changes

ppigazzini:
Typo in PR number, DEV updated with PR1402 too.

ppigazzini:
@dav1312 rn the only problem found is with SPSA, "Missing required option: base_signature"
To queue a SPSA test you need to write the test and base signature as SPRT or Games test, then select SPSA tests.

ppigazzini:
DEV running without this PR to tests other GUI PRs.

ppigazzini:
DEV updated. With https://github.com/glinscott/fishtest/pull/1401/commits/c04edca407d82774e4a29ea60826d558fb7baf59 "SPRT" "Games" "SPSA" buttons stopped working.

dav1312:
U merged my branch with the book fix PR and there is a conflict and u pushed that to DEV
image

ppigazzini:
Thank you. DEV update only with this PR.

ppigazzini:
SPSA submitted!
https://dfts-0.pigazzini.it/tests/view/63127081f532204cdb46af02

vondele:
two comments:
1)
image
the little up and down buttons are not needed, the field is just used for copy and paste.
2)
After rescheduling a test, certain fields (like branch name) can't be modified. I believe that these immutable fields where greyed out in the past.

dav1312:
> I believe that these immutable fields where greyed out in the past.

The style in that regard is the same as PROD. With light theme you can't realize that they are readonly, with dark theme u can. I can make it so it's easier to realize with the light theme too.
image

image

image

image

ppigazzini:
Book depth works fine.

ppigazzini:
PR applied by hand to PROD to collect feedbacks.

snicolet:
Nice work, thanks! Three small comments for feedback:

0) same as Joost, in the Test signature boxes, the arrow will bring more problems than good. They should be removed (even if that means that the boxes become pure text rather than numerics).

1)  the "Info" text box should be bigger (4 lines)

2) green buttons styling: center text in the button, and resize the buttons to make them smaller => same size of button and font as the blue button.

vondele:
BTW, one thing to consider (maybe not easy and maybe not part of the PR). 

It would be helpful if we could have buttons for the progression test (LTC, SMP), that automatically sets the right master (SF_15 right now), book, and things. It is not very critical, but right now it is some manual work to do that.

dav1312:
> 1. They should be removed (even if that means that the boxes become pure text rather than numerics).

It can be done without changing them back to text, dw.
I'll remove them from the test signature but keep them in other places.

> the "Info" text box should be bigger (4 lines)

Okay. Keep in mind that you can resize it if need be.

> center text in the button, and resize the buttons to make them smaller

I personally don't like it over the current one

image

> buttons for the progression test

I already explored that option a few weeks ago but not setting the right master, just test type, time control, number of threads, book etc.

https://user-images.githubusercontent.com/63931154/188280350-ffc00028-ba1a-4afb-9ebe-69a24d7d59d0.mp4

vondele:
>  I already explored that option a few weeks ago but not setting the right master, just test type, time control, number of threads, book etc.
looks good to me, hardcoding master and its bench would be helpful (that's what I have to look up in git log), I'm not sure it is worth the trouble to do it automatic.

vondele:
> center text in the button, and resize the buttons to make them smaller

I would agree with making the font smaller, left-aligned or centered would both work for me, probably left-aligned is a bit more logical for that part of the page.

ppigazzini:
LGTM on DEV, PROD updated by hand.

vondele:
maybe something that can be addressed (or noted down). If one submits a test without bench, there will be an error message (` This commit has no signature: please supply it manually.`) and all input will be forgotten. Would be nice if that error message would come, but all content would be retained.

snicolet:
@vondele I usually used "back" to get the inputs in this situation, maybe it still works (can't test it anymore on my Mac as fishtest is broken at the moment, see https://github.com/glinscott/fishtest/issues/1412)

peregrineshahin:
Good job!
One thing I would add is it might look better if the font inside boxes is responsive to the screen size
Because some of the text is overhidden and large in the box for smaller mobile screens,

Also the blue color of submit button is a bit out of context as a fitting color for the whole design feel.

Cool design

dav1312:
> font inside boxes is responsive to the screen size

Very hard to do properly and the text would need to be _too_ small in some cases.

ppigazzini:
@dav1312 not complains from PROD. I will merge as soon as you will squash the commits.

ppigazzini (approved):

ppigazzini:

PR #1404, opened 2022-08-12
Fix book depth input not showing

peregrineshahin:
.book-depth has more than one element to be shown/hidden,
this is basically a miss on my recent PR of Replacing jQuery with JavaScript

as jQuery selector $('.<class-name>') operates on all elements with that selector,
which was simplified when needed for selectors with only one element
not to have unnecessary loops over a length of 1
and this was carried out by mistake with correct simplifications of other cases

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -429,10 +429,11 @@ else:
   }
 
   function update_book_depth_visibility(book) {
-    if (book.match('\.pgn$')) {
-      document.querySelector('.book-depth').style.display = "";
+    const bookDepth = document.querySelectorAll(".book-depth");
+    if (book.match("\\.pgn$")) {
+      bookDepth.forEach((el) => (el.style.display = ""));
     } else {
-      document.querySelector('.book-depth').style.display = "none";
+      bookDepth.forEach((el) => (el.style.display = "none"));
     }
   }
 


dav1312:
https://github.com/glinscott/fishtest/pull/1401 already solves this issue
Also, why did you remove the backslash from the regex?

peregrineshahin:
Yeah I thought it would be just a temporary fix if your redesign took a longer time to get reviewed/merged,

regarding the backslash I assume you're talking about this `document.head.querySelector('link[href*="styles/github.min.css"]')?.remove()`

yeah I removed the backslash when I faced the crossorigin blocking that it should be camelCase as a JavaScript property when set to the link element, before realizing that and fixing it (now both ways work) browser didn't allow me in DEV to remove the link or add one to the document so removing the backslash worked like a work-around for just deleting iirc but still couldn't add links to the document because of cross origin block, so I kept it as is

but generally I don't care if it kept this way or back with the backslash, both ways have good points tho apart from the weird errors I was getting

server/fishtest/templates/tests_run.mak
```diff
@@ -429,10 +429,11 @@ else:
   }
 
   function update_book_depth_visibility(book) {
-    if (book.match('\.pgn$')) {
-      document.querySelector('.book-depth').style.display = "";
+    const bookDepth = document.querySelectorAll(".book-depth");
+    if (book.match(".pgn$")) {
```
dav1312, line 433: No, I'm talking about this

peregrineshahin:
omg just realized that prettier did it and not me
Capture

dav1312:
I checked, it looks like it's doing it for a good reason... because it doesn't matter 😂
Looks like the regex was wrong in the first place. It needs 2 backslashes, not just 1!

PR #1406, opened 2022-08-14
Make sure Back button reloads the tests page

### LINKED ISSUE #1405: "back" does not restore the button state
Steps to trigger

- Make sure that the default state of the "Finished" button is "Show" (i.e. the content of the corresponding div is hidden).

- Press the "Finished" button. The state now becomes "Hide" and the content of the div becomes visible.

- Click on one of the tests in the div.

- Press the browser back button. The state of the "Finished" button is again "Show" (the content of the div is hidden).

EDIT:

- If we then reload the page, the state becomes "Hide" again. So the cookie was set correctly.

peregrineshahin:
More details in the issue comments
Fixes #1405

### DIFF
--- a/server/fishtest/templates/base.mak
+++ b/server/fishtest/templates/base.mak
@@ -9,6 +9,20 @@ monitoring = request.rundb.conn["admin"].command("getFreeMonitoringStatus")
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
     <script>
+      window.addEventListener("pageshow", (e) => {
+        if (
+          location.pathname !== "/tests" &&
+          !location.pathname.includes("tests/user")
+        )
+          return;
+        const perfEntries = performance.getEntriesByType("navigation");
+        if (e.persisted || perfEntries[0].type === "back_forward") {
+          location.reload();
+        }
+      });
+    </script>
+
+    <script>
       darkThemeHash = "${cache_busters['css/theme.dark.css']}";
     </script>
 
@@ -249,25 +263,5 @@ monitoring = request.rundb.conn["admin"].command("getFreeMonitoringStatus")
         </div>
       </main>
     </div>
-
   </body>
-
-  <script>
-    (function (i, s, o, g, r, a, m) {
-      i["GoogleAnalyticsObject"] = r;
-      (i[r] =
-        i[r] ||
-        function () {
-          (i[r].q = i[r].q || []).push(arguments);
-        }),
-        (i[r].l = 1 * new Date());
-      (a = s.createElement(o)), (m = s.getElementsByTagName(o)[0]);
-      a.async = 1;
-      a.src = g;
-      m.parentNode.insertBefore(a, m);
-    })(window, document, "script", "//www.google-analytics.com/analytics.js", "ga");
-
-    ga("create", "UA-41961447-1", "stockfishchess.org");
-    ga("send", "pageview");
-  </script>
 </html>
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated.

ppigazzini:
On Windows only Edge shown the problem (Chrome, Opera, Firefox were fine).
With the PR on DEV the problem is fixed on Edge, but:
 - Edge, Chrome, Opera: the back returns with the right button state and after 1 second refresh
 - Firefox: the back returns with the right button state and there is a very fast refresh

peregrineshahin:
I think the right button state does not return correctly before the refresh if the issue reproduces on edge for you on master

Maybe you mistyped that the button returns the correct state

If master produce wrong state on a certain browser this addition should make it only return the correct state after the refresh/reload

Regarding the speed I think a more speedup can be obtained by moving the script tag even up a bit above of the already rendered html elements (I will force push that tonight)

Note that no refreshing (if the correct state took place) has been tried by changing the state to the correct one whitout a reload and failed because we don't get the correct existing state for buttons most of the times when asking for the state of the buttons due to cache conflict with the renderer I think

Also refreshing even if you have the right state is not 100% pointless as the values of the test page when pressing back will be up-to-date

ppigazzini:
> I think the right button state does not return correctly before the refresh if the issue reproduces on edge for you on master
> 
> Maybe you mistyped that the button returns the correct state

No. On my PC and on my Windows Sandbox, Edge returns the correct state (even with PROD) after running at least once DEV with this PR.

ppigazzini:
DEV updated.

ppigazzini:
DEV updated.

ppigazzini:
IMO this is faster. All the "back"s but the first one didn't need the refresh.

ppigazzini:
DEV updated.

A little off topic.
Google analytics was added in https://github.com/glinscott/fishtest/commit/10a5200f4e5fd8314f5149bdb208c2cee269f5c5
I don't know if someone is still checking the analytics and if make sense to still have the script.

There is a new faster script for modern browser:
https://developers.google.com/analytics/devguides/collection/analyticsjs/

Edge with default setting stops the tracking:
image

ppigazzini:
PROD updated by hand with this PR to collect feedbacks.

I agree to drop google analytics.

vondele:
I agree with dropping google analytics, I'm not aware of this being used.

ppigazzini:
DEV updated. This "back" refresh doesn't work for the user page, eg
https://dfts-0.pigazzini.it/tests/user/fauzi2

ppigazzini:
@peregrineshahin why do you have closed the PR?

peregrineshahin:
because I'm having too much trouble wrapping my head to squash in one commit
I pushed from backButton-my to backButton branch and boom the PR is closed automatically
At least now I can squash in one commit

peregrineshahin:
Ready for merge/review

ppigazzini:
DEV updated.
The refresh now work on the user page.
Unfortunately it works on the new test page too, so after a "back" the refresh cleans the written fields.

ppigazzini:
LGTM on Edge, Chrome, Opera, Firefox (all Windows desktop).
PROD updated by hand to collect feedbacks from the users (@vdbergh)

vdbergh:
Yes it works on firefox on Linux. There is indeed a small delay before the button shows the correct state after "back" but it is not really a problem.

ppigazzini (approved):

ppigazzini:

PR #1410, opened 2022-09-01
Catch exception when the call to os.path.getmtime fails.

vdbergh:
See https://tests.stockfishchess.org/actions?action=&user=ChessDBCN&before=1662019685.675&count=100

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -955,7 +955,12 @@ def launch_cutechess(
             print("Server told us task is no longer needed")
             return False
 
-        result["spsa"] = {"num_games": games_to_play, "wins": 0, "losses": 0, "draws": 0}
+        result["spsa"] = {
+            "num_games": games_to_play,
+            "wins": 0,
+            "losses": 0,
+            "draws": 0,
+        }
 
         w_params = req["w_params"]
         b_params = req["b_params"]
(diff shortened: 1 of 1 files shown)


worker/games.py
```diff
@@ -1126,18 +1131,27 @@ def parse_options(s):
...
-                os.remove(old_engine)
-            except Exception as e:
-                print(
-                    "Failed to remove an old engine binary {}:\n".format(old_engine),
-                    e,
-                    sep="",
-                    file=sys.stderr,
-                )
-
+        try:
```
ppigazzini, line 1134: What about this?
```python
    if len(engines) > num_bkps:
        try:
            engines.sort(key=os.path.getmtime)
        except Exception as e:
            print(
                "Failed to obtain modification time of old engine binary:\n",
                e,
                sep="",
                file=sys.stderr,
            )
        else:
            for old_engine in engines[:-num_bkps]:
                try:
                    os.remove(old_engine)
                except Exception as e:
                    print(
                        "Failed to remove an old engine binary {}:\n".format(
                            old_engine
                        ),
                        e,
                        sep="",
                        file=sys.stderr,
                    )
```

ppigazzini:

I will clean up all the nested try in a separate commit.

PR #1415, opened 2022-09-06
Progression tests presets vs SF15

dav1312:
Automatically set base branch, base signature, base options, stop rule, amount of games, threads, time control and book

image


ppigazzini:
DEV updated, LGTM.
I suggest to rename the commit and the PR title "Progression tests presets vs SF15" to make clear that there are some hardcoded data (the commit hash and the bench) to be changed with a new official release.

vondele:
nice, thanks!

ppigazzini (approved):

ppigazzini:

PR #1416, opened 2022-09-07
Progression tests v2

dav1312:
Refactor for easier future modifications and also set test branch and signature

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -6,10 +6,16 @@
   fb = lambda e0,e1:format_bounds(elo_model, e0, e1)
 
   base_branch = args.get('base_tag', 'master')
-  base_signature = args.get('base_signature', bench)
+  latest_bench = args.get('base_signature', bench)
+
+  pt_branch = "e6e324eb28fd49c1fc44b3b65784f85a773ec61c"
+  pt_signature = 8129754
 
   tc = args.get('tc', '10+0.1')
   new_tc = args.get('new_tc', tc)
+
+  test_book = "UHO_XXL_+0.90_+1.19.epd"
+  pt_book = "8moves_v3.pgn"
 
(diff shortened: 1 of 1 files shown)


vondele:
I might tag SF 15.1 but we still keep testing against SF 15, so fully automatic based on the latest tag might not work.

Bonus points for an automatic PT test Info message that picks the commit header, commit date, and formulates something like `PT of "Remove an unneeded randomization of evals." of Aug 12th vs SF 15.` not necessarily for this PR.

server/fishtest/templates/tests_run.mak
```diff
@@ -517,7 +527,7 @@
       if (btn.dataset.options) testOptions = btn.dataset.options;
 
       if (testOptions) {
-        const { tc, new_tc, threads, options, book, stop_rule, bounds, games, base_branch, base_signature } = JSON.parse(testOptions);
+        const { tc, new_tc, threads, options, book, stop_rule, bounds, games, test_branch, base_branch, test_signature, base_signature } = JSON.parse(testOptions);
```
ppigazzini, line 530: @dav1312 what are doing these new parameters?

dav1312: Set the test branch and the test signature...?

ppigazzini: I'm testing the PR on DEV but I don't know how to check these new parameters vs master. I need some help to understand the high level function they are used for (eg new test, reschedule, back button etc.) :)

dav1312: It's just for this
image

And if it is a reschedule pressing the PT presets won't change those values

ppigazzini: Thank you :)

ppigazzini (approved):

ppigazzini:

PR #1420, opened 2022-09-09
Redirect to home in /login and /signup if already logged in

peregrineshahin:
a try with Mako template's conditional rendering


dav1312:
I would do this check in views.py
```py
def login(request):
    if userid:
        return home(request)
```
https://github.com/glinscott/fishtest/blob/f52c8e683e71335d38924904172e999de2680b55/server/fishtest/views.py#L55

ppigazzini (approved):
LGTM on DEV

PR #1421, opened 2022-09-10
Make sure "back" reloads tests and tests/user V2

peregrineshahin:
- This solves the issue on all browsers without an additional refresh and without flickering thus improving the speed and ensuring the comfort of the user's eyes (especially people with epilepsy)

- Revert the previous front-end solution


ppigazzini (approved):
Looks good with all browsers on Windows.

ppigazzini:
Merged via https://github.com/glinscott/fishtest/commit/38e18e25344dd10dafad1d9f1b5a3c7a4a93fe55 after formatting with `black`

PR #1422, opened 2022-09-10
Progression tests presets v3

dav1312:
Set info (https://github.com/glinscott/fishtest/pull/1416#issuecomment-1239326536)

Notes:
- Only makes the API call when the PT or PT SMP buttons are pressed
- The message is the first line of the commit message

image

image

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -8,6 +8,7 @@
   base_branch = args.get('base_tag', 'master')
   latest_bench = args.get('base_signature', bench)
 
+  pt_version = "SF 15"
   pt_branch = "e6e324eb28fd49c1fc44b3b65784f85a773ec61c"
   pt_signature = 8129754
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, LGTM, waiting @vondele for approval/suggestion.

vondele:
excellent. I'll leave it up to David to consider/decide using 'Progression test' instead of 'PT' in the Info box, might be useful in the future searching for it.

dav1312:
Whatever you prefer. Most use "Progression test" but some use "PT" and most other tests use the short version (e.g "LTC" or "VLTC")

vondele:
let's use 'Progression test' in this case. The other abbreviations are more common.

dav1312:
Do you also want me to replace the " with ' for the message?

vondele:
whatever is the proper quoting to use in text like that :-), both are fine with me.

ppigazzini (approved):

ppigazzini:

PR #1423, opened 2022-09-11
Adapt to new MDN standard with explicit SameSite

peregrineshahin:
Explicitly set SameSite to Lax to treat FireFox warning (in the browser console) of new Cookie standard when clicking on the "finished button"

- Lax is the default we already have but setting it explicitly to make cookies standardized for most browsers and treat the warning
- Lax option allows HTTP cookie transmission which is what we have for buttons state and is convenient for Dev purposes in testing on other devices in the same network
- Note that any sensitive cookies introduced in the future should be flagged as secured and not Lax but these button-states are not sensitive data and we don't provide them to third parties after dropping google analytics 

For any new use of cookies in sensitive data,
see the options available at: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

Warning Screenshot:
Capture

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -75,7 +75,7 @@ document.addEventListener("DOMContentLoaded", () => {
         ?.remove();
     }
     // Remember the theme for 30 days
-    document.cookie = `theme=${theme};path=/;max-age=${30 * 24 * 60 * 60}`;
+    document.cookie = `theme=${theme};path=/;max-age=${30 * 24 * 60 * 60};SameSite=Lax;`;
   };
 
   const getPreferredTheme = () => {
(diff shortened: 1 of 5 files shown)


ppigazzini:
The cookie "theme" shows the same warning.

image

This seems to be a proper fix:
```diff
diff --git a/server/fishtest/static/js/application.js b/server/fishtest/static/js/application.js
index 2c54e55..900ebe7 100644
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -75,7 +75,7 @@ document.addEventListener("DOMContentLoaded", () => {
         ?.remove();
     }
     // Remember the theme for 30 days
-    document.cookie = `theme=${theme};path=/;max-age=${30 * 24 * 60 * 60}`;
+    document.cookie = `theme=${theme};path=/;max-age=${30 * 24 * 60 * 60};SameSite=Lax;`;
   };

   const getPreferredTheme = () => {
```

peregrineshahin:
Yes I was sure I missed another place also

ppigazzini (approved):
LGTM on the 4 major browsers.

ppigazzini:

PR #1424, opened 2022-09-12
Progression tests presets v3.1

dav1312:
Use the preexistent API call done in the backend to retrieve the bench to also get the message and date

Opened it as draft to gather suggestions about the python code


ppigazzini:
DEV updated with the PR.

ppigazzini (approved):
Progression tests on DEV LGTM, logs clean, console clean.

ppigazzini:

PR #1425, opened 2022-09-14
Add a basic search to finished tests and rename URL parameters

dav1312:


ppigazzini:
DEV updated.

vondele:
In the longer term just differentiating STC / LTC by the time will be a bit difficult over the full period, i.e. if we change the definitions of STC and LTC.

ppigazzini:
DEV updated.

dav1312:
Can we speed up pagination making something like this?
If the page_size is 25, search page_size + 1 elements (skipping if necessary) to see if there is a next page
This would not show the amount of available pages or return the total amount of elements but it would still allow pagination right?

ppigazzini:
> Can we speed up pagination making something like this? If the page_size is 25, search page_size + 1 elements (skipping if necessary) to see if there is a next page This would not show the amount of available pages or return the total amount of elements but it would still allow pagination right?

All the pages with pagination open very fast for me and in fishtest development we always check if the code is still working for the very old pages. Clicking 19270 times to get the last Events page is a no go for me :)

With master the navigation to a very old page is slow, but this is due to the `skip()` function.
https://www.mongodb.com/docs/manual/reference/method/cursor.skip/
https://www.codementor.io/@arpitbhayani/why-mongodb-s-cursor-skip-is-slow-90j5f3fst

BTW it seems that there are several approach to try to get a fast and consistent pagination, we can experiment with some of these.

dav1312:
> All the pages with pagination open very fast for me and in fishtest development we always check if the code is still working for the very old pages. Clicking 19270 times to get the last Events page is a no go for me :)

Ah sorry I meant speed up the search. Instead of trying to find all documents that match a info search, it would only need to find 26 right?

> Clicking 19270 times to get the last Events page is a no go for me :)

Yeah It would suck but idk I'm not sure if adding even more indexes is fine...

PR #1383, opened 2022-07-18
Administrators group

dav1312:
Create a new group called administrators that has the same permissions as approvers but that can also change the group of other users

image

Users in this group can't be blocked and their group can't be changed

image

When the group of a user is changed, an action will be added to events

image

PR #1430, opened 2022-09-26
Use a select with options for the books in tests_run

dav1312:
Easier for the users to select the book that they want without having to manually check the repo and type it correctly

image

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -490,6 +490,18 @@ def get_master_bench():
     return None
 
 
+def get_valid_books():
+    response = requests.get(
+        "https://api.github.com/repos/official-stockfish/books/contents"
+    ).json()
+    books_list = (
+        b["path"].replace(".zip", "")
+        for b in response
+        if b["path"].endswith((".epd.zip", ".pgn.zip"))
+    )
+    return books_list
+
+
 def get_sha(branch, repo_url):
     """Resolves the git branch to sha commit"""
     api_url = repo_url.replace("https://github.com", "https://api.github.com/repos")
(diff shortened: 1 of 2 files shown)


vondele:
hmm, since most users should not be changing the book, I'm not sure that listing the full choice is a good idea, it will lead to more random book testing. While it is possible to test different books, we don't want to encourage it, I think. So, not 100% sure.

dav1312:
> it will lead to more random book testing
I think most of the fishtest contributors are "responsible" in that regard.
Tests still need to pass through approvers and they can raise their concerns and delete tests if a user starts requesting all sorts of random books for no real reason.

Currently there is not really a way to know which books are available for testing. You have to figure out that the books are the ones in the stockfish org by luck or by looking at the source code.

peregrineshahin:
you can just make it available for admin or approver group with your pending PR https://github.com/glinscott/fishtest/pull/1383

vdbergh:
_if_ it has to be restricted then I would favor an "Advanced" tick box enabling it (like the tick box for time odds matches).

server/fishtest/views.py
```diff
@@ -490,6 +490,19 @@ def get_master_bench():
     return None
 
 
+def get_valid_books():
+    response = requests.get(
+        "https://api.github.com/repos/official-stockfish/books/contents"
+    ).json()
+    return list(
```
ppigazzini, line 497: In modern python an iterable comprehension (or in this case a memory saving generator expression) is replacing the use of functional programming elements such as `map`, `filter`, see
https://stackabuse.com/functional-programming-in-python/

`endswith()` accepts a tuple of strings to be searched for
https://docs.python.org/3/library/stdtypes.html?highlight=endswith#str.endswith

```python
    books_list = (
        b["path"].replace(".zip", "")
        for b in response
        if b["path"].endswith((".epd.zip", ".pgn.zip"))
    )
    return books_list
```

DEV running with this change.

ppigazzini (approved):

ppigazzini:

vondele:
I like that the drop down is behind a tick box.. nice.

vdbergh:
Something that might be considered is to shorten the list by giving only one name, without extension, per book. Whether a book is epd or pgn is an implementation detail that should not concern the user.

dav1312:
> Something that might be considered is to shorten the list by giving only one name, without extension, per book. Whether a book is epd or pgn is an implementation detail that should not concern the user.

Well, the thing is that pgn books can have custom depths that the user can customize that their epd counterpart doesn't have

vdbergh:
> > Something that might be considered is to shorten the list by giving only one name, without extension, per book. Whether a book is epd or pgn is an implementation detail that should not concern the user.
> 
> Well, the thing is that pgn books can have custom depths that the user can customize that their epd counterpart doesn't have

Ok I hadn’t thought of this. So dropping the extension is not a good idea. Still it seems .pgn is strictly better than .epd. So one might say that there is no point showing the .epd file if a .pgn file exists.

PR #1431, opened 2022-09-30
Events pagination

### LINKED ISSUE #1284: Allow more than 100 results in "Events" log

dav1312:
Add pagination to actions/events

image

Closes #1284

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -8,7 +8,15 @@ class ActionDb:
         self.db = db
         self.actions = self.db["actions"]
 
-    def get_actions(self, max_num, action=None, username=None, before=None):
+    def get_actions(
+        self,
+        username = None,
+        action = None,
+        limit = 0,
+        skip = 0,
+        utc_before = None,
+        max_actions = None,
+    ):
         q = {}
         if action:
             q["action"] = action
@@ -16,9 +24,17 @@ class ActionDb:
             q["action"] = {"$nin": ["update_stats", "dead_task"]}
         if username:
             q["username"] = username
-        if before:
-            q["time"] = {"$lte": before}
-        return self.actions.find(q, sort=[("_id", DESCENDING)], limit=max_num)
+        if utc_before:
+            q["time"] = {"$lte": utc_before}
+
+        count = self.actions.count_documents(q)
+        if max_actions:
+            limit = min(limit, max_actions - skip)
+            count = min(count, max_actions)
+
+        actions_list = self.actions.find(q, limit=limit, skip=skip, sort=[("_id", DESCENDING)])
+
+        return actions_list, count
 
     def update_stats(self):
         self._new_action("fishtest.system", "update_stats", "")
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -328,23 +328,42 @@ def sprt_calc(request):
 @view_config(route_name="actions", renderer="actions.mak")
 def actions(request):
     search_action = request.params.get("action", "")
-    search_user = request.params.get("user", "")
-    search_before = request.params.get("before", None)
-    try:
-        search_before = datetime.datetime.utcfromtimestamp(float(search_before))
-    except:
-        search_before = datetime.datetime.utcnow()
-    count = request.params.get("count", "100")
-    try:
-        count = int(count)
-    except:
-        count = 100
-    count = min(count, 100)
+    username = request.params.get("user", "")
+    before = request.params.get("before", None)
+    max_actions = request.params.get("max_actions", None)
+
+    utc_before = before
+    if before:
+        utc_before = datetime.datetime.utcfromtimestamp(float(before))
+    if max_actions:
+        max_actions = int(max_actions)
+
+    page_idx = max(0, int(request.params.get("page", 1)) - 1)
+    page_size = 25
+
+    actions, num_actions = request.actiondb.get_actions(
+        username=username,
+        action=search_action,
+        skip=page_idx * page_size,
+        limit=page_size,
+        utc_before=utc_before,
+        max_actions=max_actions,
+    )
+
+    pages = pagination(page_idx, num_actions, page_size)
+
+    for page in pages:
+        if username:
+            page["url"] += "&user={}".format(username)
+        if search_action:
+            page["url"] += "&action={}".format(search_action)
+        if max_actions:
+            page["url"] += "&max_actions={}".format(max_actions)
+        if before:
+            page["url"] += "&before={}".format(before)
 
     actions_list = []
-    for action in request.actiondb.get_actions(
-        count, search_action, search_user, search_before
-    ):
+    for action in actions:
         item = {
             "action": action["action"],
             "time": action["time"],
@@ -366,33 +385,19 @@ def actions(request):
             item["_id"] = action["data"]["before"]["_id"]
             item["description"] = []
 
-            before = action["data"]["before"]["args"]["priority"]
-            after = action["data"]["after"]["args"]["priority"]
-            if before != after:
-                item["description"].append(
-                    "priority changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["num_games"]
-            after = action["data"]["after"]["args"]["num_games"]
-            if before != after:
-                item["description"].append(
-                    "games changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["throughput"]
-            after = action["data"]["after"]["args"]["throughput"]
-            if before != after:
-                item["description"].append(
-                    "throughput changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["auto_purge"]
-            after = action["data"]["after"]["args"]["auto_purge"]
-            if before != after:
-                item["description"].append(
-                    "auto-purge changed from {} to {}".format(before, after)
-                )
+            for k in ("priority", "num_games", "throughput", "auto_purge"):
+                try:
+                    before = action["data"]["before"]["args"][k]
+                    after = action["data"]["after"]["args"][k]
+                except KeyError:
+                    pass
+                else:
+                    if before != after:
+                        item["description"].append(
+                            "{} changed from {} to {}".format(
+                                k.replace("_", "-"), before, after
+                            )
+                        )
 
             item["description"] = "modify: " + ", ".join(item["description"])
         else:
@@ -412,7 +417,13 @@ def actions(request):
 
         actions_list.append(item)
 
-    return {"actions": actions_list, "approver": request.has_permission("approve_run")}
+    return {
+        "actions": actions_list,
+        "approver": request.has_permission("approve_run"),
+        "pages": pages,
+        "num_actions": num_actions,
+        "page_idx": page_idx,
+    }
 
 
 def get_idle_users(request):
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -1,18 +1,16 @@
 <%inherit file="base.mak"/>
+
 <%
   import datetime
 %>
+
 <h2>Events Log</h2>
+
 <script>
   document.title = 'Events Log | Stockfish Testing';
-
-  function timestamp(){
-    document.querySelector('#before').value = Date.now() / 1000;
-    return true;
-  }
 </script>
 
-<form onsubmit="timestamp();" class="row mb-3">
+<form class="row mb-3">
   <div class="col-12 col-md-auto mb-3">
     <label for="restrict" class="form-label">Show only</label>
     <select id="restrict" class="form-select" name="action">
@@ -35,12 +33,12 @@
     <input id="user" type="text" name="user" class="form-control" placeholder="username" value="${request.GET.get('user') if request.GET.get('user') != None else ''}">
   </div>
 
-  <input type="hidden" id="before" name="before" value=-1>
-  <input type="hidden" id="count" name="count" value=100>
   <div class="col-12 col-md-auto mb-3 d-flex align-items-end">
     <button type="submit" class="btn btn-success w-100">Select</button>
   </div>
 </form>
+
+<%include file="pagination.mak" args="pages=pages"/>
 
 <div class="table-responsive-lg">
   <table class="table table-striped table-sm">
@@ -56,7 +54,7 @@
       % for action in actions:
           <tr>
             ## Dates in mongodb have millisecond precision. So they fit comfortably in a float without precision loss.
-            <td><a href=/actions?count=1&before=${action['time'].replace(tzinfo=datetime.timezone.utc).timestamp()}>
+            <td><a href=/actions?max_actions=1&before=${action['time'].replace(tzinfo=datetime.timezone.utc).timestamp()}>
              ${action['time'].strftime(r"%y&#8209;%m&#8209;%d %H:%M:%S")|n}</a></td>
             % if approver and 'fishtest.' not in action['username']:
                 <td><a href="/user/${action['username']}">${action['username']}</a></td>
@@ -76,6 +74,8 @@
     </tbody>
   </table>
 </div>
+
+<%include file="pagination.mak" args="pages=pages"/>
 
 <script>
   document.querySelector('#restrict').value = ('${request.GET.get("action") if request.GET.get("action") != None else ''}');


server/fishtest/views.py
```diff
@@ -328,23 +328,35 @@ def sprt_calc(request):
...
-        search_before = datetime.datetime.utcfromtimestamp(float(search_before))
-    except:
-        search_before = datetime.datetime.utcnow()
-    count = request.params.get("count", "100")
+    username = request.params.get("user", "")
+    timestamp = request.params.get("timestamp", None)
+
     try:
-        count = int(count)
+        timestamp = datetime.datetime.utcfromtimestamp(float(timestamp))
```
ppigazzini, line 335: DEV updated.
Nearly all our codebase is using the LBYL approach, so better to use the simple: 
```python
    if timestamp is not None:
        timestamp = datetime.datetime.utcfromtimestamp(float(timestamp))
```
In any case for the EAFP approach it's better to not use a bare exception:
```python
    try:
        timestamp = datetime.datetime.utcfromtimestamp(float(timestamp))
    except TypeError:
        pass
```
In some cases python requires EAFP, see:
https://realpython.com/python-lbyl-vs-eafp/

dav1312:
Going past page 16206 crashes and I'm not sure why... 🤔

vdbergh:
I am trying to read you code. Did you drop the count argument? I think count=1 is useful if you want to link to a single event.

dav1312:
> I am trying to read you code. Did you drop the count argument? I think count=1 is useful if you want to link to a single event.

I removed that argument and kept the link functionality at the same time
https://dfts-0.pigazzini.it/actions?timestamp=1664625873.969

vdbergh:
I see. I have also used the count argument for referring to the last x events of a user (in case a worker started recently developing a particular problem). This is not ideal though. It would be handled better by a keyword search on the comment field (so that we could search for a particular type of event).

vdbergh:
What I want to say is that it now seems to be no longer possible to link to the event log at a particular point in time (one can only link to a single event).  I consider this a loss in functionality (functionality that has been used).

ppigazzini:
> Going past page 16206 crashes and I'm not sure why... 🤔 Do old actions have something that we don't use currently?

```
Oct 02 12:24:44 dfts-0 pserve[21241]:   File "/home/usr00/fishtest/server/fishtest/views.py", line 393, in actions
Oct 02 12:24:44 dfts-0 pserve[21241]:     before = action["data"]["before"]["args"]["throughput"]
Oct 02 12:24:44 dfts-0 pserve[21241]: KeyError: 'throughput'
```

ppigazzini:
Fixed with this (in this case it's shorter the EAFP way, the LBYL way requires 2 controls):
```diff
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 6ae4470..8e97888 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -376,33 +376,16 @@ def actions(request):
             item["_id"] = action["data"]["before"]["_id"]
             item["description"] = []

-            before = action["data"]["before"]["args"]["priority"]
-            after = action["data"]["after"]["args"]["priority"]
-            if before != after:
-                item["description"].append(
-                    "priority changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["num_games"]
-            after = action["data"]["after"]["args"]["num_games"]
-            if before != after:
-                item["description"].append(
-                    "games changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["throughput"]
-            after = action["data"]["after"]["args"]["throughput"]
-            if before != after:
-                item["description"].append(
-                    "throughput changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["auto_purge"]
-            after = action["data"]["after"]["args"]["auto_purge"]
-            if before != after:
-                item["description"].append(
-                    "auto-purge changed from {} to {}".format(before, after)
-                )
+            for k in ("priority", "num_games", "throughput", "auto_purge"):
+                try:
+                    before = action["data"]["before"]["args"][k]
+                    after = action["data"]["after"]["args"][k]
+                    if before != after:
+                        item["description"].append(
+                            "{} changed from {} to {}".format(k, before, after)
+                        )
+                except KeyError:
+                    pass

             item["description"] = "modify: " + ", ".join(item["description"])
         else:
```

ppigazzini:
Are you talking about this search type, right?
https://tests.stockfishchess.org/actions?=failed_task&user=ChessDBCN&count=4&before=1664312574.521
on DEV has stopped working.
https://dfts-0.pigazzini.it/actions?action=failed_task&user=ChessDBCN&count=3&before=1664312574.521

ppigazzini:
> What I want to say is that it now seems to be no longer possible to link to the event log at a particular point in time (one can only link to a single event). I think this a loss of a of functionality (functionality that has been used).

Are you talking about this search type, right?
https://tests.stockfishchess.org/actions?=failed_task&user=ChessDBCN&count=4&before=1664312574.521
on DEV has stopped working.
https://dfts-0.pigazzini.it/actions?action=failed_task&user=ChessDBCN&count=3&before=1664312574.521

vdbergh:
Your examples show that "before" still works. My bad. I somehow thought that that was removed. I do miss the count argument though (as long that there is not a more convenient way to restrict the output of a query).

vdbergh:
Deleted my last post since it seems that "before" was indeed removed. It seems it is replaced by "timestamp" which shows only one event.

ppigazzini:
I re added on the fly those API (to be checked), see
https://dfts-0.pigazzini.it/actions?action=failed_task&user=ChessDBCN&before=1664312574.521&num_actions=4

```diff
diff --git a/server/fishtest/actiondb.py b/server/fishtest/actiondb.py
index a189c2f..2c044d6 100644
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -15,6 +15,8 @@ class ActionDb:
         limit = 0,
         skip = 0,
         timestamp = None,
+        before = None,
+        num_actions = None,
     ):
         q = {}
         if action:
@@ -25,6 +27,10 @@ class ActionDb:
             q["username"] = username
         if timestamp:
             q["time"] = timestamp
+        if before:
+            q["time"] = {"$lte": before}
+        if num_actions is not None and num_actions < limit:
+            limit = num_actions

         count = self.actions.count_documents(q)
         actions_list = self.actions.find(q, limit=limit, skip=skip, sort=[("_id", DESCENDING)])
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 6ae4470..ea447bc 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -330,9 +330,13 @@ def actions(request):
     search_action = request.params.get("action", "")
     username = request.params.get("user", "")
     timestamp = request.params.get("timestamp", None)
+    before = request.params.get("before", None)
+    num_actions = request.params.get("num_actions", None)

     if timestamp is not None:
         timestamp = datetime.datetime.utcfromtimestamp(float(timestamp))
+    if before is not None:
+        before = datetime.datetime.utcfromtimestamp(float(before))

     page_idx = max(0, int(request.params.get("page", 1)) - 1)
     page_size = 25
@@ -342,7 +346,9 @@ def actions(request):
         action=search_action,
         skip=page_idx * page_size,
         limit=page_size,
-        timestamp=timestamp
+        timestamp=timestamp,
+        before=before,
+        num_actions=int(num_actions),
     )
```

vdbergh:
It seems to work but now "timestamp" has become redundant....

vdbergh:
Hmm there seems to be a bit of a conflict between num_actions and pagination.

ppigazzini:
Dropped `timestamp`
- single event form link:
https://dfts-0.pigazzini.it/actions?num_actions=1&before=1664421199.44
- events group:
https://dfts-0.pigazzini.it/actions?action=failed_task&user=ChessDBCN&before=1664312574.521&num_actions=4

```diff
diff --git a/server/fishtest/actiondb.py b/server/fishtest/actiondb.py
index a189c2f..81e2162 100644
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -14,7 +14,8 @@ class ActionDb:
         action = None,
         limit = 0,
         skip = 0,
-        timestamp = None,
+        before = None,
+        num_actions = None,
     ):
         q = {}
         if action:
@@ -23,8 +24,10 @@ class ActionDb:
             q["action"] = {"$nin": ["update_stats", "dead_task"]}
         if username:
             q["username"] = username
-        if timestamp:
-            q["time"] = timestamp
+        if before:
+            q["time"] = {"$lte": before}
+        if num_actions is not None and num_actions < limit:
+            limit = num_actions

         count = self.actions.count_documents(q)
         actions_list = self.actions.find(q, limit=limit, skip=skip, sort=[("_id", DESCENDING)])
diff --git a/server/fishtest/templates/actions.mak b/server/fishtest/templates/actions.mak
index c838b3b..384c6d7 100644
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -54,7 +54,7 @@
       % for action in actions:
           <tr>
             ## Dates in mongodb have millisecond precision. So they fit comfortably in a float without precision loss.
-            <td><a href=/actions?timestamp=${action['time'].replace(tzinfo=datetime.timezone.utc).timestamp()}>
+            <td><a href=/actions?num_actions=1&before=${action['time'].replace(tzinfo=datetime.timezone.utc).timestamp()}>
              ${action['time'].strftime(r"%y&#8209;%m&#8209;%d %H:%M:%S")|n}</a></td>
             % if approver and 'fishtest.' not in action['username']:
                 <td><a href="/user/${action['username']}">${action['username']}</a></td>
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 6ae4470..617f4da 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -329,10 +329,13 @@ def sprt_calc(request):
 def actions(request):
     search_action = request.params.get("action", "")
     username = request.params.get("user", "")
-    timestamp = request.params.get("timestamp", None)
+    before = request.params.get("before", None)
+    num_actions = request.params.get("num_actions", None)

-    if timestamp is not None:
-        timestamp = datetime.datetime.utcfromtimestamp(float(timestamp))
+    if before is not None:
+        before = datetime.datetime.utcfromtimestamp(float(before))
+    if num_actions is not None:
+        num_actions = int(num_actions)

     page_idx = max(0, int(request.params.get("page", 1)) - 1)
     page_size = 25
@@ -342,7 +345,8 @@ def actions(request):
         action=search_action,
         skip=page_idx * page_size,
         limit=page_size,
-        timestamp=timestamp
+        before=before,
+        num_actions=num_actions,
     )

     pages = pagination(page_idx, num_actions, page_size)
@@ -375,35 +379,19 @@ def actions(request):
             item["run"] = action["data"]["before"]["args"]["new_tag"]
             item["_id"] = action["data"]["before"]["_id"]
             item["description"] = []
-
-            before = action["data"]["before"]["args"]["priority"]
-            after = action["data"]["after"]["args"]["priority"]
-            if before != after:
-                item["description"].append(
-                    "priority changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["num_games"]
-            after = action["data"]["after"]["args"]["num_games"]
-            if before != after:
-                item["description"].append(
-                    "games changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["throughput"]
-            after = action["data"]["after"]["args"]["throughput"]
-            if before != after:
-                item["description"].append(
-                    "throughput changed from {} to {}".format(before, after)
-                )
-
-            before = action["data"]["before"]["args"]["auto_purge"]
-            after = action["data"]["after"]["args"]["auto_purge"]
-            if before != after:
-                item["description"].append(
-                    "auto-purge changed from {} to {}".format(before, after)
-                )
-
+            for k in ("priority", "num_games", "throughput", "auto_purge"):
+                try:
+                    before = action["data"]["before"]["args"][k]
+                    after = action["data"]["after"]["args"][k]
+                except KeyError:
+                    pass
+                else:
+                    if before != after:
+                        item["description"].append(
+                            "{} changed from {} to {}".format(
+                                k.replace("_", "-"), before, after
+                            )
+                        )
             item["description"] = "modify: " + ", ".join(item["description"])
         else:
             item["run"] = action["data"]["args"]["new_tag"]
```

vdbergh:
This seems ok. The superfluous pagination remains (when num_actions exists) but that's not a big deal.

ppigazzini:
> This seems ok. The superfluous pagination remains (when num_actions exists) but that's not a big deal.

It should be fixed, please check the changes in my previous post.

vdbergh:
> > This seems ok. The superfluous pagination remains (when num_actions exists) but that's not a big deal.
> 
> It should be fixed, please check the changes in my previous post.

Yes it seems to be fixed! Thanks.

ppigazzini:
DEV updated from mobile.

server/fishtest/actiondb.py
```diff
@@ -8,17 +8,33 @@ def __init__(self, db):
...
             q["username"] = username
-        if before:
-            q["time"] = {"$lte": before}
-        return self.actions.find(q, sort=[("_id", DESCENDING)], limit=max_num)
+        if utc_before:
+            q["time"] = {"$lte": utc_before}
+
+        count = self.actions.count_documents(q)
+        if max_actions:
+            limit = min(limit, max_actions)
```
ppigazzini, line 32:
```python
            limit = min(limit, max_actions - skip)
```
In this way we have the right last page when `max_actions` is bigger than and not multiple of `page_size`:
https://dfts-0.pigazzini.it/actions?user=ChessDBCN&action=failed_task&max_actions=53&before=1664312574.521

ppigazzini (approved):

ppigazzini:
PROD updated, great job, thank you @dav1312 :)

ppigazzini:
> The query seems to be very slow when `max_actions` is present (even with `max_actions=1`). I wonder if this was also the case before. I never noticed it.

DEV running on previous master, the query is very fast.
https://dfts-0.pigazzini.it/actions?count=1&before=1664394722.655&count=7

dav1312:
Could it be bcs of that loop hack that u did on views.py to avoid the crashes? The rest of the code is pretty similar

ppigazzini:
> Could it be bcs of that loop hack that u did on views.py to avoid the crashes? The rest of the code is pretty similar Or maybe that now it has to count the 500k documents for the pagination

DEV on master, I dropped the try/except (the loop is only a code simplification)

```diff
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 4dcf09d..86ad609 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -386,18 +386,26 @@ def actions(request):
             item["description"] = []

             for k in ("priority", "num_games", "throughput", "auto_purge"):
-                try:
-                    before = action["data"]["before"]["args"][k]
-                    after = action["data"]["after"]["args"][k]
-                except KeyError:
-                    pass
-                else:
-                    if before != after:
-                        item["description"].append(
-                            "{} changed from {} to {}".format(
-                                k.replace("_", "-"), before, after
-                            )
+                before = action["data"]["before"]["args"][k]
+                after = action["data"]["after"]["args"][k]
+                if before != after:
+                    item["description"].append(
+                        "{} changed from {} to {}".format(
+                            k.replace("_", "-"), before, after
                         )
+                    )

             item["description"] = "modify: " + ", ".join(item["description"])
         else:
```

ppigazzini:
This count could be costly `count = self.actions.count_documents(q)`

vdbergh:
I suspect the count as well. But why is it not an issue if we don’t use max_actions?

ppigazzini:
> I suspect the count as well. But why is it not an issue if we don’t use max_actions?

We have indexes for "actions" and "username" keys to speedup the search. Try to use `max_actions` without `before` (I'm commuting now)

ppigazzini:
DEV on master is very fast with:
```
        #count = self.actions.count_documents(q)
        count = 1000
```
Try opening:
https://dfts-0.pigazzini.it/actions?user=vdv&action=approve_run&max_actions=10&before=1643129087.036

ppigazzini:
I try to add an index on the "time" key,

ppigazzini:
It seems fast with the index

```
Current indexes on actions:
{ '_id_': {'key': [('_id', 1)], 'v': 2},
  'action_1__id_-1': {'key': [('action', 1), ('_id', -1)], 'v': 2},
  'time_1__id_-1': {'key': [('time', 1), ('_id', -1)], 'v': 2},
  'username_1__id_-1': {'key': [('username', 1), ('_id', -1)], 'v': 2}}
```

ppigazzini:
This seems to have fixed the problem without the index, PROD and DEV updated by hand:
```diff
diff --git a/server/fishtest/actiondb.py b/server/fishtest/actiondb.py
index b4f692c..c408a41 100644
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -27,10 +27,11 @@ class ActionDb:
         if utc_before:
             q["time"] = {"$lte": utc_before}

-        count = self.actions.count_documents(q)
         if max_actions:
+            count = self.actions.count_documents(q, limit=max_actions)
             limit = min(limit, max_actions - skip)
-            count = min(count, max_actions)
+        else:
+            count = self.actions.count_documents(q)

         actions_list = self.actions.find(q, limit=limit, skip=skip, sort=[("_id", DESCENDING)])
```

dav1312:
Something like that is what I wanted to use before u sent ur code yesterday 😂 Glad I wasn't that far from the right path in the end

ppigazzini:
@dav1312 DEV seems running fine without these 2 variables (if I'm not wrong they are not used in `actions.mak`)

```diff
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 1ee5ebe..17c532d 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -421,8 +421,8 @@ def actions(request):
         "actions": actions_list,
         "approver": request.has_permission("approve_run"),
         "pages": pages,
-        "num_actions": num_actions,
-        "page_idx": page_idx,
+#        "num_actions": num_actions,
+#        "page_idx": page_idx,
     }
```

ppigazzini:
Fixed with https://github.com/glinscott/fishtest/commit/a1f80eb2c251639fca92e21163cb94446a807336

PR #1432, opened 2022-10-03
Fix events pagination

dav1312:
Limit the count if max_actions is used (credits to @ppigazzini)
Correct value in url in case user uses a max_actions value larger than count


ppigazzini (approved):

ppigazzini:
PROD updated, thank you @dav1312 for the fix, thank you @vdbergh for the report.

PR #1433, opened 2022-10-04
Minimal solution for cached requests with wrong states

peregrineshahin:
this should in principle increase the performance of the load on the test pages, this minimal solution is dropping out cache control support of old legacy browsers while still unifying the back button mismatched behavior in different modern browsers


ppigazzini:
DEV updated.

ppigazzini (approved):
I didn't find problems running on DEV, it's surely not a slowdown checking the page load time in Inspect.

ppigazzini:

Issue #1412, opened 2022-09-03
Impossible to create/modify/stop tests on Safari (Mac and IPhone)


snicolet:
Hi,

Since a couple of days, I encounter a problem when trying to either create/modify/stop any test on Safari (Mac or IPhone), or even trying to login in fishtest. I suspect that it may be related to password checking: has anything changed in the recent days?

Steps to reproduce:

1) launch Safari
2) try to create/modify/stop any of my tests

==>  I get a blank page with an `check_csrf_token(): Invalid token` error message

Disservin:
I cannot recreate this, I just successfully modified a test of yours and everything worked fine on safari. I have iOS 15.6.1

snicolet:
I can make a video stream of my screen if you wish to see the bug in action: https://discord.gg/4r4C8x65

snicolet:
I managed to change my password, but now even the logout link is broken:

```
404 Not Found
The resource could not be found.
predicate mismatch for view logout (request_method = POST)
```

peregrineshahin:
What OS version do you have?
This is because of the use of using JavaScript ES6 (2015) syntax
The arrow functions, let, const instead of using normal functions and var
Your browser seems not to support these new syntactic sugar from ES6 because it's maybe old

peregrineshahin:
Well I even think it's because of the optional chaining introduced in (ES2020) specifically
Not the ES6 (2015) features

I think the script running on your device is stopping way before password checking
Not allowing the script to continue and check login

Can you confirm where the script is failing by checking your browser console?

PR #1434, opened 2022-10-05
Catch error in case of no browser support

dav1312:
Use a try catch because some old safari versions might not support this

Closes https://github.com/glinscott/fishtest/issues/1412

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -88,9 +88,13 @@ document.addEventListener("DOMContentLoaded", () => {
     setTheme(getPreferredTheme());
   }
 
-  window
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated with #1434, #1433, #1425

ppigazzini:
DEV updated (I don't know if there was a change with the last push).
Let me know when this is ready to be merged.

ppigazzini:

dav1312:
> DEV updated (I don't know if there was a change with the last push). Let me know when this is ready to be merged.

Didn't read this on time but no, I just changed the commit message

ppigazzini:
I checked the previous version on DEV :)

snicolet:
thanks everybody :-)

PR #1435, opened 2022-10-10
Adjust adjudication params by +50% in wake of new net arch with equivalent eval inflation

dubslow:
Data:
https://tests.stockfishchess.org/tests/view/6343f4b04bc7650f075401c7 with old adjudication params https://tests.stockfishchess.org/tests/view/6343f4ba4bc7650f075401ca with no adjudication

If these tests come out with "noticeably different" results then we should consider this or similar adjustments. If those tests come back with "substantially identical" results than this should be rejected.

(Terms in quotes are left for future definition)

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1363,7 +1363,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
                 [
                     "-resign",
                     "movecount=3",
-                    "score=400",
+                    "score=600",
                     "-draw",
                     "movenumber=34",
                     "movecount=8",


ppigazzini:
I wait @vondele opinion before merging this.

dubslow:
Certainly, this is indeed marked as draft, significant consensus should be achieved before merger. I just want to start the discussion and provide some starting point data

dubslow:
I must say I didn't expect one to pass and one to fail, and even if they did, I would have thought that w/o adj passes while w/ adj fails. Nevertheless, here we are. For comparison's sake, I've rescheduled both for LTC

peregrineshahin:
@dubslow I think you've rescheduled both STC again?

dubslow:
oops. fixed

vondele:
Adjudication is something I don't really like, but certainly when tested in the past, it turned out to be really accurate. It tends to speedup testing by 20% (again, based on older data). I have been considering to remove it completely, so increasing the win adjudication is a step in the right direction. Increasing the draw one, I think, is not a good idea.

dubslow:
Your thoughts exactly align with my own. I also prefer adjusting the win, and not the draw, but was trying to be minimally controversial. And I would be happy to delete them altogether too, altho there's probably good reasons to not do that

snicolet:
+1 for increasing the win adjudication level and keeping the current draw adjudication level.

vondele:
so this PR is fine with me.

ppigazzini (approved):

ppigazzini:
Started the workers update, thank you @dubslow :)
Let me know if you want your real name too in the AUTHORS file.

PR #1436, opened 2022-10-18
clickable username in liveElo page

snicolet:
So that we can jump directly from the LiveElo pages to the page with all the tests of the user.


snicolet:
Thanks to the Discord chat for help :-)

snicolet:
@ppigazzini

Note that this PR becomes irrelevant if we decide to merge https://github.com/glinscott/fishtest/pull/1437, as https://github.com/glinscott/fishtest/pull/1437 also contains the clickable username feature (with another implementation).

PR #1437, opened 2022-10-19
Template literals, avoid innerHTML and user link

dav1312:
Use template literals
Avoid innerHTML as much as possible, using textContent instead
Add a link to the user tests
Reorder based on table order

### DIFF
--- a/server/fishtest/static/js/live_elo.js
+++ b/server/fishtest/static/js/live_elo.js
@@ -124,84 +124,48 @@ function clear_gauges() {
   set_gauges(0, -2.94, 2.94, 0.5, 0, 0, 0);
 }
 
-const entityMap = {
-  "&": "&amp;",
-  "<": "&lt;",
-  ">": "&gt;",
-  '"': "&quot;",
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated, LGTM. If I'm not wrong this PR makes #1436 obsolete.

dav1312:
Yes, this PR also adds a link to the user tests

ppigazzini:
PROD updated, thank you @dav1312 (and @snicolet) :)

snicolet:
Slight surprise with the username link : it should not open in a new tab. We should remove the « target = blank »

dav1312:
> it should not open in a new tab

Why not

snicolet:
because if we do a sequence of 
LiveElo -> username -> LiveElo -> username -> LiveElo -> username, 
we end up with an infinite number of tabs.

All other links pointing to https://tests.stockfishchess.org remain in the same page, giving the global impression that https://tests.stockfishchess.org is a single app (there must be a technical term for that). I think it's a good feature.

PR #1438, opened 2022-10-19
Remove target=blank attribute for username link

snicolet:
If we do a sequence of
LiveElo -> username -> LiveElo -> username -> LiveElo -> username,
we end up with an infinite number of tabs.

All other links in fishtest pointing to https://tests.stockfishchess.org/ remain in the same page, giving the global impression that https://tests.stockfishchess.org/ is a single window app (there must be a technical term for that). This patch is coherent with this.


Disservin:
I agree, all links should be in the same tab, unless the link links to some page from where you cant get back to fishtest. (like the diff link to goes to github).

server/fishtest/templates/tests_live_elo.mak
```diff
@@ -6,7 +6,12 @@
 </script>
 
 <div class="container">
-  <h2>Live Elo for SPRT test <a href="/tests/view/${str(run['_id'])}">${str(run['_id'])}</a></h2>
+  <h2>
+    <center>
```
peregrineshahin, line 10: Hello Stéphane, The `<center>` tag has been deprecated in HTML 4 and been removed from HTML5 in favor of the CSS text-align property, which can be applied to the <div> element, but by doing so the <a> tag (username) is alone centered and other fields (i.e Commit, Info, Submitter etc.) are not

peregrineshahin, line 10: Hello Stéphane, The <center> tag has been deprecated in HTML 4 and removed from HTML5 in favor of the CSS text-align property, which can be applied to the <h2> element.

snicolet:
Done

ppigazzini:
DEV updated.
This PR restores one consistency but breaks another one: all the h2/h3 titles in this and other pages are left aligned.

snicolet:
thanks @ppigazzini :-)

I'm sure the centered title will be very nice above the dialers

ppigazzini:
I dislike by heart the centered text, it keeps moving when resizing a windows from the right side disrupting the reading.
When I fixed the pages formatting I simply failed to get the three gauges left aligned due to my incompetence with the CSS/HTML/bootstrap.

snicolet:
Do whatever you want with the centering, the core of the PR is the target for the link :)

PR #1439, opened 2022-10-25
Limit the scope of no-storing the response to run_table instead of t…

peregrineshahin:
…he whole home page

this limits the scope of demanding a no-store response only to be applied on the run tables instead of the tests page,
should result in some page loading speedup of `tests` and `tests\user` pages while still solving the "back" button issue of mismatched behavior between browsers.

DEV testing and PROD testing are needed to confirm if any speed-up which loading time can be measured by dev-tools network tab


ppigazzini:
DEV updated.

peregrineshahin:
Closing for no performance gain.

PR #1440, opened 2022-10-30
Use github actions to add md5sums to branch

vdbergh:
For every push or pull request or push this PR adds a commit (via github actions) with the md5sums of the worker files. By downloading this file from the master branch at github the worker can check if it has been changed locally. If the worker has been changed it can communicate this information to the server which can suitably annotate event log messages.

This is to aid in debugging. See e.g. https://github.com/glinscott/fishtest/issues/1360#issuecomment-1294506544 .

The worker/server changes are not included yet since I first would like to know if this PR is acceptable.

### DIFF
--- a/.github/workflows/compute_md5sums.py
+++ b/.github/workflows/compute_md5sums.py
@@ -1,0 +1,15 @@
+import hashlib
+import json
+from pathlib import Path
+
+worker_dir = Path(__file__).parent.parent.parent / "worker"
+
+FILE_LIST = ["updater.py", "worker.py", "games.py"]
+md5dict = {}
+for file in FILE_LIST:
+    item = worker_dir / file
+    bytes = item.read_bytes()
+    md5 = hashlib.md5(bytes).hexdigest()
+    md5dict[file] = md5
+
+(worker_dir / "md5sums").write_text(json.dumps(md5dict) + "\n")
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,6 +22,8 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       # Use MongoDb
       - name: Start MongoDB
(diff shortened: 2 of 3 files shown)


vdbergh:
Sadly it does not appear to work correctly for PRs only for pushes. I will leave this open for a bit in case people want to comment on what I am doing wrong.

.github/workflows/compute_md5sums.py
```diff
@@ -0,0 +1,18 @@
+import hashlib
+import json
+from pathlib import Path
+worker_dir = Path(__file__ ).parent.parent.parent / "worker"
+
+FILE_LIST = ["updater.py", "worker.py", "games.py"]
+md5dict = {}
+for file in FILE_LIST:
+    item = worker_dir / file
+    with open(item, "rb") as f:
```
ppigazzini, line 10: We could use the standard SRI hash already used for all our resources:
https://github.com/glinscott/fishtest/blob/c74e8ed329efc01a689a4daede7240d8456507a8/server/fishtest/__init__.py#L29-L32

ppigazzini:
> EDIT: Another not nice thing is that I am forced to create a new commit with the added md5sum file. Changing the top commit with `git commit --amend --no-edit` appears not to work in github actions. It yields a history consisting of a single commit.

If I'm not wrong `fixup` and `autosquash` could help (I dont' know if they work in gh actions)
https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

.github/workflows/compute_md5sums.py
vdbergh: Thanks. I have no strong feeling about this. The reason I used md5sum is that the result is easy to verify using the shell command `md5sum`.

vdbergh:
> > EDIT: Another not nice thing is that I am forced to create a new commit with the added md5sum file. Changing the top commit with `git commit --amend --no-edit` appears not to work in github actions. It yields a history consisting of a single commit.
> 
> If I'm not wrong `fixup` and `autosquash` could help (I dont' know if they work in gh actions) https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

I experimented a bit on the command line (not on github) but I do not know how to do it non-interactively (interactively it works).

vdbergh:
I think I discovered why `git commit --amend --no-edit` did not work. One has to supply the argument `fetch-depth:0` to `checkout@v2`.

.github/workflows/compute_md5sums.py
vdbergh: I used read_bytes and write_text now. Thanks! If you don't mind then I will md5sum, just because the md5sum is easy to generate on the command line.

vdbergh:
It seems to be working well now. I implemented the corresponding worker changes (in a different branch). Of course they depend on the file md5sums being present in the master branch (which is not yet the case).

vdbergh:
After thinking it through, I don’t really like GitHub actions messing with the repository. So I think it is better to leave the responsibility of the generation (before a push) of the md5sums file to the developer (e.g. with an argument to worker.py). However CI can _check_ if the md5sums file is correct.

vdbergh:
Superceeded by #1442 .

PR #1441, opened 2022-10-30
Handle make with invalid permissions gracefully.

vdbergh:
See

https://tests.stockfishchess.org/actions?action=&user=markkulix&max_actions=2


ppigazzini (approved):

ppigazzini:
Workers updated, thank you @vdbergh :)

PR #1442, opened 2022-10-31
Compute sri hashes of worker files and compare with master

vdbergh:
The sri hashes of the current worker files are stored in a file `sri.txt`. This file is updated each time the worker runs, and it can also be updated manually with
```    
    python worker.py --only_config
```    
CI is used to enforce that the file `sri.txt` matches the worker files during pushes or pull requests.
    
The worker will also download the master `sri.txt` (currently not existing) 

https://raw.githubusercontent.com/glinscott/fishtest/master/worker/sri.txt
    
and check if the hashes in that file coincide with the sri hashes of the current worker. If not then a message is printed on the terminal. A tainted worker is not stopped but the flag `worker_info['modified']` is set to `True`. In a subsequent PR the server will use this information in the event log messages (for debugging purposes).
    
In order to fix the race condition between the update of the master worker on GitHub and the update of the Fishtest server (which pulls its code from GitHub), the file `sri.txt` contains the version number of the worker. We ignore the master `sri.txt` if its version is higher than the version of the running worker.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -5,6 +5,7 @@ import datetime
 import getpass
 import glob
 import hashlib
+import json
 import math
 import multiprocessing
 import os
@@ -47,6 +48,7 @@ from games import (
     download_from_github,
     format_return_code,
     log,
+    requests_get,
     run_games,
     send_api_post_request,
     str_signal,
@@ -54,6 +56,8 @@ from games import (
 from updater import update
 
 WORKER_VERSION = 181
+FILE_LIST = ["updater.py", "worker.py", "games.py"]
+SRI_URL = "https://raw.githubusercontent.com/glinscott/fishtest/master/worker/sri.txt"
 HTTP_TIMEOUT = 30.0
 INITIAL_RETRY_TIME = 15.0
 THREAD_JOIN_TIMEOUT = 15.0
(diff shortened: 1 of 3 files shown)


worker/worker.py
```diff
@@ -215,6 +219,84 @@ def safe_sleep(f):
...
+def verify_sri(install_dir):
+    sri = generate_sri(install_dir)
+    sri_file = os.path.join(install_dir, "sri.txt")
+    if not os.path.exists(sri_file):
+        print("{} does not exist".format(sri_file))
+        return False
+    try:
+        with open(sri_file, "r") as f:
+            sri_ = json.load(f)
+    except:
```
ppigazzini, line 255: Better to write the expected exceptions or to print them if unknown:
```python
except Exceptions as e:
    print("Exception reading {}:\n".format(sri_file), e, sep="", file=sys.stderr)
```

vdbergh: Changed it!

worker/worker.py
```diff
@@ -215,6 +219,84 @@ def safe_sleep(f):
...
+    # Returns:
+    # None in case of network error
+    # True if verification succeeded
+    # False if verification failed
+    sri = generate_sri(install_dir)
+    sri_ = download_sri()
+    if sri_ is None:
+        return None
+    tainted = False
+    for k, v in sri_.items():
```
ppigazzini, line 288: Just for fun :)
```python
if sri_ != sri:
    tainted = True
    for k, v in (set(sri_.items()) - set(sri.items())):
        print("{} has been modified!".format(k))
```

vdbergh: Not sure about this one. Personally I find the original code more straightforward and hence more readable...

ppigazzini:
If I'm not wrong we could have some unavoidable warnings during a worker update.
The cleanest process should be:
  - push the commit that raises the worker version (worker and server side) and computes the sri.txt
  - update the server to trigger the workers update (there are some seconds of delay, causing the warnings)

The warnings on DEV during a test of a worker PR can be ignored.

vdbergh:
Yes this is correct! Workers that are restarted between the push and the server update will consider themselves tainted (before they are upgraded). I don't think this is a serious problem.

One way to circumvent is, is to include the worker version in the file `sri.txt` (perhaps with key `__version`) and to ignore the master `sri.txt` if its version does not match the version of the worker that is running. It requires a bit of thinking to get this correct so I would prefer to defer this to another PR.

vdbergh:
Latest push was a rebase.

worker/worker.py
ppigazzini: Just as not, the pythonic version is faster in checking the equality but slower in finding the differences:
```python
import timeit
a = {str(i): i for i in range(100)}
#b = {str(i): i+1 for i in range(10)}
b = {str(i): i for i in range(100)}
b["0"]=1

def v():
    for k, v in a.items():
        if k not in b or v != b[k]:
            tainted = True
            #print("{} has been modified!".format(k))

def p():
    if a != b:
        tainted = True
        for k in dict(set(a.items()) ^ set(b.items())):
            pass
            #print("{} has been modified!".format(k)

print("v():", timeit.timeit('[v()]', globals=globals()))
print("p():", timeit.timeit('[p()]', globals=globals()))
```

vdbergh: Thanks. Speed is not really a priority for the worker though...

It stands to reason that `dict == dict` is faster than manually going through the key value pairs. In the former case we are executing C code and in the latter case we are executing python code.

ppigazzini: Of course, I wasted some time to learn&refresh a bit of python. As previously stated, just for fun :)

vdbergh:
In the latest commit I put the worker version in the sri.txt file.

vdbergh:
I guess it is much better to move the CI verification that srt.txt is up to date to test_worker.py, certainly in view of #1443  . Next commit…

ppigazzini:
Testing the workers on DEV (fixed SRI_URL merging a branch of mine).

vdbergh:
There is still a small race condition between the merging of a PR and the commit that updates the worker version.

I was thinking that, to completely avoid race conditions, the server could perhaps get the master `sri.txt` from GitHub at start up and then send it to the worker as part of the reply to the version api call. Then the worker would not need to download the master `sri.txt`.

ppigazzini:
> There is still a small race condition between the merging of a PR and the commit that updates the worker version.
> 
> I was thinking that, to completely avoid race conditions, the server could perhaps get the master `sri.txt` from GitHub at start up and then send it to the worker as part of the reply to the version api call. Then the worker would not need to download the master `sri.txt`. Next PR?

Ok, I'm merging this one.

ppigazzini (approved):

ppigazzini:
Triggered the workers update with https://github.com/glinscott/fishtest/commit/37df400a0f258e27cadff78952cc20216250ad0a

sri.txt generated with:
```
python3 worker.py aaa 123 --only_config --no_validation
git clean -fd
```

Thank you @vdbergh :)

vdbergh:
Maybe we should make this last step more straightforward. Something like
```
python worker.py --generate_sri
```

ppigazzini:
Well, it's mostly a maintainer task, not a problem for me. cc to @zungur

vdbergh:
It seems race conditions are unavoidable even if the server would send the sri file (read at start up). Assume that in that case the master branch is updated  but not yet the server. If somebody installs the worker (downloading it from GitHub) at that point, it would be considered tainted.

PR #1443, opened 2022-10-31
Update github actions

ppigazzini:
Split in server and worker workflows.
Set up the server workflow with the SW versions used on PROD.
Set up the worker workflows to test on Linux with few python versions and on Windows msys2.

See:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
https://github.com/marketplace/actions/setup-msys2


vdbergh:
Would it be possible to test the worker also on MacOSX? There is some Mac specific code in the worker and it would be nice to be able to test it.

ppigazzini:
That was my intention but I got an error when adding MacOSX in the workflow matrix.

ppigazzini:
@vdbergh now it works :) https://github.com/ppigazzini/fishtest/actions/runs/3445013827/jobs/5748241705

Disservin:
What error ? We have macos in the stockfish CI if you are interested take a look there only the avx2 compile doesn’t work for macOS ci runner

ppigazzini:
My first implementation had python 3.5 too in the matrix (now dropped because I would like to switch to `pathlib`), the macos action works now for python 3.6-3.11
https://github.com/ppigazzini/fishtest/actions/runs/3445037852

ppigazzini:
I got the error when using only one workflow for linux and macos, perhaps we are hitting some limits.

https://github.com/ppigazzini/fishtest/actions/runs/3445092768/jobs/5748408359

```
test_worker_script_with_no_args (test_worker.workerTest.test_worker_script_with_no_args) ... ok

ERROR: test_item_download (test_worker.workerTest.test_item_download)
  File "/Users/runner/work/fishtest/fishtest/worker/games.py", line 146, in requests_get
    result.raise_for_status()  # also catch return codes >= 400
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/fishtest/fishtest/worker/.eggs/requests-2.28.1-py3.11.egg/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/official-stockfish/books/git/trees/master

During handling of the above exception, another exception occurred:

  File "/Users/runner/work/fishtest/fishtest/worker/test_worker.py", line 22, in test_item_download
    games.download_from_github("README.md", ".")
  File "/Users/runner/work/fishtest/fishtest/worker/games.py", line 396, in download_from_github
    tree = requests_get(
           ^^^^^^^^^^^^^
  File "/Users/runner/work/fishtest/fishtest/worker/games.py", line 154, in requests_get
    raise WorkerException("Get request to {} failed".format(remote), e=e)
games.WorkerException: Get request to https://api.github.com/repos/official-stockfish/books/git/trees/master failed
```

Ran 6 tests in 20.487s

ppigazzini:
I don't understand why only the macos workflow is hitting that rate limit...

vdbergh:
Perhaps we can fake the rate limit request during testing?

ppigazzini:
The macos workflow still fails when using the github token, even with only one python version.
https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp

```yaml
      - name: Run worker tests
        run: |
          ./test_worker.sh
        env:
          GITHUB_TOKEN: ${{ github.token }}
```

vdbergh:
@ppigazzini Do you have the link of that action?

ppigazzini:
https://github.com/ppigazzini/fishtest/actions/

I wonder why we are hitting the link https://api.github.com/repos/official-stockfish/books/git/trees/master with this code
```python
    def test_item_download(self):
        try:
            games.download_from_github("README.md", ".")
            self.assertTrue(os.path.exists(os.path.join(".", "README.md")))
        except KeyError:
            pass
```

vdbergh:
It is strange that the github token does not work since it is officially supported.

ppigazzini:
Doubts:
 - we are hitting an external repo
 - we are using `run:`

ppigazzini:
This failed randomly too:
```yaml
      - name: Run worker tests
        run: |
          cd worker && python setup.py test
        env:
          #GITHUB_TOKEN: ${{ github.token }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

vdbergh:
Maybe we can download things using raw.githubusercontent.com ?   I think this is not affected by api credits.

vdbergh:
I made a PR. Have a look!

PR #1444, opened 2022-11-01
Retire uuid.txt.

vdbergh:
The first (constant) part of the UUID is now written as a comment in the config file. This is sufficient to identify the worker.


ppigazzini (approved):

ppigazzini:
Raised the workers update, thank you @vdbergh :)

PR #1445, opened 2022-11-01
Give some indications that a worker is modified

vdbergh:
This is the server counter part to #1442 .

Add (M) to workername of modified worker.
Add + to version of modified worker in machine table.

### DIFF
--- a/server/fishtest/templates/machines_table.mak
+++ b/server/fishtest/templates/machines_table.mak
@@ -20,7 +20,7 @@
           gcc_version = ".".join([str(m) for m in machine['gcc_version']])
           compiler = machine.get('compiler', 'g++')
           python_version = ".".join([str(m) for m in machine['python_version']])
-          version = machine['version']
+          version = str(machine['version']) + "*" * machine['modified']
         %>
         <tr>
           <td>${machine['username']}</td>
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -51,6 +51,7 @@ def validate_request(request):
             "gcc_version": [int, int, int],
             "compiler": union("g++", "clang++"),
             "unique_key": str,
+            "modified": bool,
             "rate": {"limit": int, "remaining": int},
             "ARCH": str,
             "nps": float,
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -16,6 +16,7 @@ def worker_name(worker_info):
     username = worker_info["username"]
     cores = str(worker_info["concurrency"])
     uuid = worker_info["unique_key"]
+    modified = worker_info.get("modified", False)
     name = "{}-{}cores".format(username, cores)
     if len(uuid) != 0:
         uuid_split = uuid.split("-")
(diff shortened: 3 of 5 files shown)


server/fishtest/templates/machines_table.mak
```diff
@@ -20,7 +20,7 @@
           gcc_version = ".".join([str(m) for m in machine['gcc_version']])
           compiler = machine.get('compiler', 'g++')
           python_version = ".".join([str(m) for m in machine['python_version']])
-          version = machine['version']
+          version = str(machine['version'])+("+" if machine['modified'] else "")
```
ppigazzini, line 23: Black adds spaces around the plus sign:
```python
version = str(machine['version']) + ("+" if machine['modified'] else "")
```
Other options are still ugly for the flood of single/double quotes, parentheses/brackets/braces, the last one is funny :)
```python
version = "".join([str(machine['version']), "+" if machine['modified'] else ""])
version = f"{machine['version']}{'+' if machine['modified'] else ''}"
version = "{}{}".format(machine['version'], "+" if machine['modified'] else "")
version = str(machine['version']) + '+' * machine['modified']
```

vdbergh: The last one is short but it is a bit of a hack (it depends on the fact that Python interprets True as 1 and False a 0). But I guess it is clear what it is meant. I will take it.

vdbergh: Done (not tested since this laptop doesn't have fishtest).

ppigazzini:
DEV updated, LGTM.
"183+" seems positive ("I run the plus version of the worker"), perhaps it's better to use another symbol: "183<", "183#", "183*"

server/fishtest/templates/machines_table.mak
ppigazzini: DEV updated, it works.

dav1312:
> DEV updated, LGTM. "183+" seems positive ("I run the plus version of the worker"), perhaps it's better to use another symbol: "183<", "183#", "183*"
> 
> EDIT: or letter, "183M" or "183m" to be consistent with the renaming of the worker.

I think the best symbol is `*`, the other ones seem confusing.

vdbergh:
Maybe also for the worker name (instead om (M))?

vdbergh:
I changed both of them to *. AFK for an hour or so.

ppigazzini (approved):
DEV updated, LGTM, logs clean.

ppigazzini:

ppigazzini:
Modders caught! :)
image

vdbergh:
That's a bit weird though. I would have thought this would only happen for developers (why would one otherwise modify the worker). Also technologov seems to have some modded workers, but not all of them are modded.

ppigazzini:
Asked to technologov and sebastronomy on Discord.

ppigazzini:
BabyShark on Discord pointed out that the raw data on Github can have a 5 minutes delay.
https://stackoverflow.com/questions/62785962/get-raw-file-from-github-without-waiting-for-5-minute-cache-update
https://stackoverflow.com/questions/46551413/github-not-update-raw-after-commit
> GitHub serves "raw" pages with Cache-Control: max-age=300. That's specified in seconds, meaning the pages are intended to be cached for 5 minutes.
You can see this if you open the Developer Tools in your web browser of choice before clicking the "Raw" button on GitHub.

PR #1446, opened 2022-11-06
Ignore the master sri if it has a different version number

vdbergh:
This is to handle another race condition. Namely the fact  that it takes some time for https://raw.githubusercontent.com to update after a push. See
    
https://github.com/glinscott/fishtest/pull/1445#issuecomment-1304848578


ppigazzini:
Merged by hand via https://github.com/glinscott/fishtest/commit/5aa4450be0f31eba71881346707bd9e83ccfaad0

I updated PROD just after the merge to test the change.

PR #1447, opened 2022-11-07
No placeholder for the username field in the event log

vdbergh:
The placeholder causes annoying behaviour in some browsers that think that this is a login field: e.g. Safari on iPad.

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -30,7 +30,7 @@
   </div>
   <div class="col-12 col-md-auto mb-3">
     <label for="user" class="form-label">From user</label>
-    <input id="user" type="text" name="user" class="form-control" placeholder="username" value="${request.GET.get('user') if request.GET.get('user') != None else ''}">
+    <input id="user" type="text" name="user" class="form-control" value="${request.GET.get('user') if request.GET.get('user') != None else ''}">
   </div>
 
   <div class="col-12 col-md-auto mb-3 d-flex align-items-end">


vdbergh:
26121522-19F6-413B-9412-684CB1C9005A

peregrineshahin:
`autocomplete="off"` will not work on most browsers (gets ignored), you may try to change the name of the input to something else other than `name="user"` (e.g. name="from-user") to not lose autocomplete which works on other browsers and will most probably work in Safari too.

If that didn't work it might be that Safari is smart enough to read the placeholder then it can be changed to another word or the same word can have invisible Unicode or extra HTML spaces or a Cyrillic e -> "е" if necessary.

vdbergh:
@peregrineshahin Can you explain why autocomplete is important for this field? IMHO it serves no purpose.

vdbergh:
On the other hand, while `autocomplete="off"` is generally supported on mobile browsers, it is typically ignored on desktop browser according to this site https://caniuse.com/input-autocomplete-onoff . So it is perhaps better to not give the `username` suggestion (it is obvious from the label that this is what this field should contain).

peregrineshahin:
Well, it may serve little benefit here but not much yes,
I don't fully understand what is meant by this
> So it is perhaps better to not give the username suggestion (it is obvious from the label that this is what this field should contain).

but yes, changing the `for="user"` in label and `name="user"` to another name in input will stop giving suggestions of previous usernames and give suggestions on the history of the new name given which I'm not sure what the behavior would be on iPad TBH,

but you really may have a shot with `autocomplete="off"` as I see now that this surprisingly works in my latest chrome Desktop!

vdbergh:
> 
> but you really may have a shot with `autocomplete="off"` as I see now that this surprisingly works in my latest chrome Desktop!

Well according to this site I linked to it seems to depend on the browser configuration for desktop browsers. The main issue is password saving. Should a website be allowed to disable password saving, even if the user wants it?

peregrineshahin:
Well, my idea was that based on the `name="user"` in input Safari thinks that this input is a login field thus I think changing the name `for="user"` and `name="user"` will consider input as a non-login and this annoying `use "vdberg"` should not appear and have a normal auto-complete.

If that doesn't work, the placeholder should also be removed or changed, but removing the placeholder still gives suggestions on my desktop.
better to test that first

and yeah it is allowed to force disable password saving providing one is successful really doing it but may be bad as a quality metric for SEO reasons.

vdbergh:
Let me add that the current behaviour is extremely annoying on Safari (on iPad). If you page through the event log then the login dialog pops up at _every_ new page.

vdbergh:
Closing because of #1463 .

PR #1448, opened 2022-11-08
Correct singular (1 minute ago instead of 1 minutes ago)

vdbergh:


server/fishtest/util.py
```diff
@@ -375,18 +375,22 @@ def diff_date(date):
...
-        delta = "{:d} days ago".format(diff.days)
-    elif diff.seconds / 3600 > 1:
-        delta = "{:d} hours ago".format(diff.seconds // 3600)
-    elif diff.seconds / 60 > 1:
-        delta = "{:d} minutes ago".format(diff.seconds // 60)
-    else:
-        delta = "seconds ago"
-    return delta
+        return "Never"
+    s = [
```
ppigazzini, line 385: Alternatives (just for fun) using `tv` for time_values and `td` for time_descriptors:
```python
    tv = diff.days, diff.seconds // 3600, diff.seconds // 60
    td = "day", "hour", "minute"
    for v, d in zip(tv, td):
        if v >= 1:
            return "{:d} {} ago".format(v, plural(v, d))
    return "seconds ago"
```
The hard pythonic way :)
https://stackoverflow.com/questions/19502378/find-the-first-instance-of-a-nonzero-number-in-a-list-in-python
https://stackoverflow.com/questions/2361426/get-the-first-item-from-an-iterable-that-matches-a-condition

```python
    tv = diff.days, diff.seconds // 3600, diff.seconds // 60
    td = "day", "hour", "minute"
    idx = next((i for i, x in enumerate(tv) if x), None)
    return "seconds ago" if idx is None else "{:d} {} ago".format(tv[idx], plural(tv[idx], td[idx]))
```

vdbergh: As usual I do not like the "pythonic" way since it makes something simple complicated. Using zip makes sense since it saves a few lines.

vdbergh: Changed!

ppigazzini: > As usual I do not like the "pythonic" way since it makes something simple complicated.

What about purescript?
easy way (stackoverflow prone):
```haskell
fibs :: Int -> Int
fibs 0 = 1
fibs 1 = 1
fibs n = fibs (n-1) + fibs (n-2)
```
hard way:
```haskell
fibs :: Int -> Int
fibs n = f n 0 1
  where
    f 0 _ b = b
    f i a b = f (i-1) b (a+b)
```

ppigazzini (approved):

ppigazzini:

vdbergh:
Strange. The worker CI failed on the master commit, despite the worker not having changed…

ppigazzini:
Restarted the failed CI jobs.

PR #1449, opened 2022-11-09
Verify the sri hashes of a worker update

vdbergh:


ppigazzini (approved):
The update tests seems fine

PR #1450, opened 2022-11-11
Use raw.githubusercontent.com to download long living stuff from github

vdbergh:
See the discussion  for the PR #1443 .

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -70,7 +70,7 @@ HTTP_TIMEOUT = 30.0
 CUTECHESS_KILL_TIMEOUT = 15.0
 UPDATE_RETRY_TIME = 15.0
 
-REPO_URL = "https://github.com/official-stockfish/books"
+RAWCONTENT_URL = "https://raw.githubusercontent.com/official-stockfish/books/master/"
 EXE_SUFFIX = ".exe" if IS_WINDOWS else ""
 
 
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):
It worked in my tests on my repo with the macos workflow.

ppigazzini:
Thank you @vdbergh :)

A couple of notes.
I can either add a separate workflow for macOS or change the matrix of the linux workflow: do you have a preference?
On my repo the linux workflow fails because `ubuntu-latest` (alias for v 20.04 right now) starts unexpectedly v 22.04 that has dropped the python 3.6 support, see
https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
noob's fleet uses python 3.6, so perhaps better to force `ubuntu-20.04` in the workflow.

vondele:
This leads to problems with noob's workers. Maybe a proxy issue or some thing?

vondele:
blocked in China:
https://github.com/openvinotoolkit/openvino/issues/8492

vdbergh:
Bah. I guess it is best to revert for now. The solution is to first use the api and if that fails, as a fallback raw.gitusercontent.com. Or perhaps the other way round.

vdbergh:
Hmm sri.txt is also downloaded from raw.gituserconttent.com. So it seems it is not always blocked. Otherwise the worker would not start.

vdbergh:
Maybe we should wait a bit. It is not the first time that the noob workers have transient problems with github.

vdbergh:
The noob workers appear to be working normally again.

vdbergh:
E.g. this worker https://tests.stockfishchess.org/actions?max_actions=1&before=1668548747.379 now appears to be producing games.

vondele:
probably because it doesn't need the different book, i.e. that book is just available ?

vdbergh:
Well it was trying to download the book, which failed at first, but now apparently it has succeeded.

vondele:
some more discussion in discord on this https://discord.com/channels/435943710472011776/900771437177094156/1042243969867849839

PR #1455, opened 2022-11-16
Revert 731369e

ppigazzini:
Workers from China are unable to download the items.
Drop macOS from GitHub Actions.


ppigazzini:
I tried to use the GITHUB_TOKEN downloading items from the fishtest repo during the test but the macOS workflow keeps failing at random for the GitHub API rate limit, so I dropped macOS from the workflow matrix.

image

PR #1458, opened 2022-11-21
Switch to pathlib

ppigazzini:


vdbergh:
Nice cleanups!

vdbergh:
@ppigazzini That's a nuisance. No opinion really... It would be dangerous to break the updater. Perhaps in the updater you can use `str(...)` with a comment?

ppigazzini:
I replaced `shutil.move` with pathlib functions and used string for `shutil.rmtree`

```diff
-        bkp_testing_dir = os.path.join(worker_dir, "_testing_" + time_stamp)
-        shutil.move(testing_dir, bkp_testing_dir)
-        os.makedirs(testing_dir)
+        bkp_testing_dir = worker_dir / ("_testing_" + time_stamp)
+        testing_dir.replace(bkp_testing_dir)
+        testing_dir.mkdir()
```

`shutil.rmtree` is the only function used now, it could be replaced with pathlib functions, see https://stackoverflow.com/questions/50186904/pathlib-recursively-remove-directory

```python
def rm_tree(pth: Path):
    for child in pth.iterdir():
        if child.is_file():
            child.unlink()
        else:
            rm_tree(child)
    pth.rmdir()
```

The `shutil.rmtree` implementation is very sophisticated, though.
https://github.com/python/cpython/blob/0ac5372bf6b937ed44a8f9c4e402d024fcd80870/Lib/shutil.py#L683

ppigazzini:
It seems that `shutil` accepts path-like parameters from version 3.6, but the documentation was updated only for version 3.8. This is consistent with my tests.
https://stackoverflow.com/questions/33625931/copy-file-with-pathlib-in-python

> This generalized a lot of stdlib functionality, and documentation was not consistently updated as a result, including most of shutil which as of 3.7 [only documents support in a single function](https://docs.python.org/3.7/library/shutil.html#shutil.unpack_archive).

vdbergh:
I browsed a bit https://github.com/python/cpython/blob/v3.6.8/Lib/shutil.py but it is unclear to me how the pathlib support is implemented in shutil.py (the module does not itself import pathlib). I wonder if perhaps some of the os.path functions accept path-like objects themselves...

ppigazzini:
`os` functions started accepting path-like parameters from 3.6

vdbergh:
Ah ok  that explains things. `shutil` is implemented via `os.path` and the latter accepts path-like objects since 3.6. So it seems we are pretty safe.

ppigazzini:
@noobpwnftw if you can run this snippet (at your convenient time) on the VPS that runs your workers we can made a safe choice. 

```bash
#!/bin/bash

mkdir -p d0/d1
touch d0/f0_{0..9}
touch d0/d1/f1_{0..9}
ls d0/**

python3 << EOF
import shutil
from pathlib import Path

shutil.rmtree(Path("d0"))
EOF

ls d0/**
```

PR #1462, opened 2022-11-24
Enable the standard exceptions for make.

vdbergh:
Workers come up with ever more obscure exceptions, so it seems it difficult to catch them all. See

https://tests.stockfishchess.org/actions?max_actions=1&before=1669310498.047


ppigazzini:
Triggered the workers update, finger crossed.

PR #1463, opened 2022-11-26
Give users list suggestion instead of history suggestions when autoco…

peregrineshahin:
…mplete in events

and explicitly set which fields to fill in for the username and password autocomplete.

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -28,13 +28,28 @@
       <option value="update_stats" class="grayedoutoption">System Events</option>
     </select>
   </div>
+
   <div class="col-12 col-md-auto mb-3">
     <label for="user" class="form-label">From user</label>
-    <input id="user" type="text" name="user" class="form-control" placeholder="username" value="${request.GET.get('user') if request.GET.get('user') != None else ''}">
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated, ready for tests by users.

vdbergh:
This works well for me on iPad. Thanks! I’ll close my PR.

server/fishtest/templates/actions.mak
```diff
@@ -28,13 +28,28 @@
...
-    <input id="user" type="text" name="user" class="form-control" placeholder="username" value="${request.GET.get('user') if request.GET.get('user') != None else ''}">
+    <input
+      id="user"
+      class="form-control"
+      autocomplete="off"
+      placeholder="Search by username"
+      type="text"
+      name="user"
+      list="users-list"
+      value="${request.GET.get('user') if request.GET.get('user') != None else ''}"
```
ppigazzini, line 42: Better to use `is not None` here:
```
value="${request.GET.get('user') if request.GET.get('user') is not None else ''}"
```

ppigazzini (approved):

ppigazzini:

PR #1464, opened 2022-11-28
Detect the presence of make during start up

vdbergh:
A missing make is a common toolchain issue.

https://tests.stockfishchess.org/actions?max_actions=1&before=1669622636.797

 Also: verify during CI if the tests for make and cutechess work.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -437,6 +437,8 @@ def setup_cutechess(worker_dir):
     testing_dir = worker_dir / "testing"
     testing_dir.mkdir(exist_ok=True)
 
+    curr_dir = Path.cwd()
+
     try:
         os.chdir(testing_dir)
     except Exception as e:
@@ -444,6 +446,8 @@ def setup_cutechess(worker_dir):
         return False
 
     cutechess = "cutechess-cli" + EXE_SUFFIX
+
+    ret = True
 
     if not verify_required_cutechess(testing_dir, cutechess):
         if len(EXE_SUFFIX) > 0:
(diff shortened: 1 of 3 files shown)


worker/worker.py
```diff
@@ -1083,6 +1083,24 @@ def detect_compilers():
     return ret
 
 
+def detect_make():
+    try:
+        with subprocess.Popen(
+            ["make"],
```
vondele, line 1089: may be this could better check for the output of `make -v`, since just invoking make without arguments is an error (i.e. non-zero exit code).

vdbergh: @vondele I agree to some extent but I was afraid that different OSes may have different make options. I wonder what @ppigazzini thinks.

vdbergh: Actually `make` without arguments is dangerous since there could be a spurious makefile in the worker directory.

vdbergh: These are the arguments for posix make https://pubs.opengroup.org/onlinepubs/009695299/utilities/make.html . Unfortunately `-v` is not among them. This does not mean it is not a good idea to use it. But it requires some testing to see if the option is supported on the platform the worker supports. In principle this could be done in CI.

vdbergh: @vondele I now use 'make -v' and check in CI that the detection works (the current CI checks that the worker works correctly in quite a few environments, including the troublesome MacOS).

ppigazzini (approved):
Tested working on linux and msys2

ppigazzini:

PR #1468, opened 2022-12-05
Let the worker decide if it is close to the github rate limit.

vdbergh:
Also: more feedback if api.github.com/rate_limit fails.

See #1466.

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -961,18 +961,19 @@ def get_uuid(options):
     return uuid_prefix[:8] + str(uuid.uuid4())[8:]
 
 
-def get_rate():
+def get_remaining_github_api_calls():
     try:
-        rate = requests.get(
-            "https://api.github.com/rate_limit", timeout=HTTP_TIMEOUT
-        ).json()["resources"]["core"]
+        rate = requests.get("https://api.github.com/rate_limit", timeout=HTTP_TIMEOUT)
+        rate.raise_for_status()
+        return rate.json()["resources"]["core"]["remaining"]
     except Exception as e:
-        print("Exception fetching rate_limit:\n", e, sep="", file=sys.stderr)
-        rate = {"remaining": 0, "limit": 5000}
-        return rate, False
-    remaining = rate["remaining"]
-    print("API call rate limits:", rate)
-    return rate, remaining < math.sqrt(rate["limit"])
+        print(
+            "Exception fetching rate_limit (invalid ~/.netrc?):\n",
+            e,
+            sep="",
+            file=sys.stderr,
+        )
+        return 0
 
 
 def gcc_version():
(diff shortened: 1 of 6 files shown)


vdbergh:
I am making this WIP since the rate limit thing should be rethought. As far as I can tell the only thing that matters is that `remaining` should be above a certain small bound (to make sure that the worker can download the engine binaries). So I do not understand the purpose of the current complicated code.

vdbergh:
Note that this contains both worker and (small) server changes.

ppigazzini:
PROD updated, started a couple of workers,

vdbergh:
If the server is restarted after the update the running workers will get an error message from the server but this is harmless. The workers will restart and update.

ppigazzini:
Tested without `.netrc`, few workers, exausting the api calls (changed the raw address to speedup the process):

- Already running worker:
```
Task exited
Remaining number of github api calls = 10

  We have almost exhausted our github api calls.
  The server will only give us tasks for tests we have seen before.

Verify worker version...
Post request https://dfts-0.pigazzini.it:443/api/request_version handled in 118.55ms (server: 1.39ms)
Current time is 2022-12-06 14:15:42.772423 UTC (offset: +01:00)
Fetching task...
Post request https://dfts-0.pigazzini.it:443/api/request_task handled in 134.56ms (server: 7.94ms)
Working on task 46 from https://dfts-0.pigazzini.it:443/tests/view/638f398e72008c09724c7ade
Running master vs master
Variable task sizes used. Opening offset = 5696
Obtaining EvalFile of stockfish_758f9c9350abee36a5865ec701560db8ea62004d ...
Obtaining EvalFile of stockfish_758f9c9350abee36a5865ec701560db8ea62004d ...
Verifying signature of stockfish_758f9c9350abee36a5865ec701560db8ea62004d ...
Verifying signature of stockfish_758f9c9350abee36a5865ec701560db8ea62004d ...
```

New worker:
```
Verify worker version...
Post request https://dfts-0.pigazzini.it:443/api/request_version handled in 164.42ms (server: 0.11ms)
Current time is 2022-12-06 14:15:17.302798 UTC (offset: +01:00)
Fetching task...
Post request https://dfts-0.pigazzini.it:443/api/request_task handled in 183.09ms (server: 22.34ms)
No tasks available at this time, waiting...
Waiting 60.0 seconds before retrying
  Send heartbeat for 3b2eceb9-b2df-4c71-8e7e-b44bb9d9addf ... Skipping heartbeat ...
Remaining number of github api calls = 10

  We have almost exhausted our github api calls.
  The server will only give us tasks for tests we have seen before.

Verify worker version...
Post request https://dfts-0.pigazzini.it:443/api/request_version handled in 169.46ms (server: 1.69ms)
Current time is 2022-12-06 14:16:17.933045 UTC (offset: +01:00)
Fetching task...
Post request https://dfts-0.pigazzini.it:443/api/request_task handled in 181.93ms (server: 21.78ms)
No tasks available at this time, waiting...
Waiting 120.0 seconds before retrying
```

worker/worker.py
```diff
@@ -1303,12 +1304,18 @@ def fetch_and_handle_task(worker_info, password, remote, lock_file, current_stat
         return False
 
     try:
```
ppigazzini, line 1306: I wonder if we can shrink the code in this `try`.
If I'm not wrong `get_remaining_github_api_calls()` already deals with its exception.

The `try` here is dealing only with the `send_api_post_request()` exceptions, one directly in `fetch_and_handle_task()` and one nested in `verify_worker_version()`

vdbergh: Ok I changed it. Have a look.

ppigazzini: Reading and testing.

ppigazzini (approved):

ppigazzini:

PR #1471, opened 2022-12-06
Also break the word if no pre-wrap white space can be done

peregrineshahin:
Reference:
1- [tests.stockfishchess.org/tests/view/638ec34268532fcbf79df94f](https://tests.stockfishchess.org/tests/view/638ec34268532fcbf79df94f)
2- [tests.stockfishchess.org/tests/view/638e49d468532fcbf79de4fd](https://tests.stockfishchess.org/tests/view/638e49d468532fcbf79de4fd)
3- [tests.stockfishchess.org/tests/view/638e497b68532fcbf79de4e6](https://tests.stockfishchess.org/tests/view/638e497b68532fcbf79de4e6)

Before:
capture02

After:
capture01

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -139,6 +139,7 @@
 
 code {
   white-space: pre-wrap; 
+  word-break: break-word !important;
 }
 
 /* Confirm deny button */


ppigazzini:
DEV updated.
1 - https://dfts-0.pigazzini.it/tests/view/638ec34268532fcbf79df94f
2 - https://dfts-0.pigazzini.it//tests/view/638e49d468532fcbf79de4fd
3 - https://dfts-0.pigazzini.it//tests/view/638e497b68532fcbf79de4e6

If I'm not wrong this PR breaks the line trying to keep a whole command on the same line.

peregrineshahin:
Idea is that actual code lines can be easily noticed as an actual code line has either `-` or `+` before it.

ppigazzini:
To be honest I don't have the know how to verify if the PR is working as intended, so let me know if I can merge :)

peregrineshahin:
It works as intended as the x-scroll bar does not appear anymore and the layout has not been ruined this one is a follow-up on the previous PR but because of the full command, we had to break word also.

also, we use `!important` to prevent GitHub CSS override our line as the code element is styled as `word-break: normal` in the GitHub CSS imported after this file.

We basically do the same stuff `word-break: break-word` for the info message for tests on the main page and user page.

DEV lgtm

ppigazzini (approved):

ppigazzini:

PR #1472, opened 2022-12-08
Make the URLs in the event log more specific.

vdbergh:
The current scheme does work correctly for URLs corresponding to a dead_task event.


ppigazzini:
DEV updated.

ppigazzini (approved):

ppigazzini:

PR #1473, opened 2022-12-09
Change the handling and presentation of events (PR1)

vdbergh:
Currently the handling of events is a mess, so we want to fix this. The practical aim however is to provide more options for searching in the future (on the worker name and the comment field).

This project requires a fair number of changes. So the plan is to break it up into several consecutive mergeable PR's.

*Strategy*

- Extract data from the current description field (i.e. `task_id`, `worker name`, `message`) and use it in the new event log format (PR1, i.e. this one).
- Put the extracted data back into the action collection and use the action collection directly to show the event log. Also convert the old action records to the new format (PR2).
- Change the rest of fishtest to put the data directly into the action collection in the new format. No more parsing of the description field (PR3).
- Allow searching on the worker name and the message field. This requires the creation of text indices (PR4).
- Possibly delete the superfluous never shown data in the action collection. This would free a lot of space, but should be discussed (PR5).

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -60,9 +60,10 @@
     <thead class="sticky-top">
       <tr>
         <th>Time</th>
-        <th>Username</th>
-        <th>Run/User</th>
         <th>Event</th>
+        <th>Source</th>
+        <th>Target</th>
+        <th>Comment</th>
       </tr>
     </thead>
     <tbody>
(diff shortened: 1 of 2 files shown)


vdbergh:
It seems CI failed but I think this is just a fluke.

ppigazzini:
DEV updated.

vdbergh:
I can't see anything wrong with the new format. The parsing of the previous description field appears to be ok.

vdbergh:
It would be nice is a link like `dblExtDLow/712` would effectively take you to task `712`. Something for a future PR!

vdbergh:
In the latest push I made some cosmetic changes in the headers of the event table: `User/Worker->Source` and `Message->Comment`.

ppigazzini:
DEV updated. LGTM. Let me know if I can merge :)

ppigazzini (approved):

vdbergh:
I think is can be merged!

ppigazzini:
Done!

PR #1477, opened 2022-12-11
PR2: Extract the event fields directly from the data base. 

vdbergh:
To make this work the action database should be converted using the script `utils/convert_actions.py`.

Note: this script does not delete any data. Still would be good to temporarily keep a backup of the current action database, just in case something goes wrong.

This PR has no user visible changes.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -2,6 +2,77 @@ from datetime import datetime
 
 from pymongo import DESCENDING
 
+"""
+schema = {
+"action" : str,
+"username" : str,
+"time" : datetime.datetime,
+optional_key("message") : str,
+optional_key("user") : str,
+optional_key("run_id") : ObjectId,
+optional_key("run") : str,
+optional_key("worker") : str,
+optional_key("nn") : str,
+}
+"""
+
+
+def expand_action(action):
+    action["message"] = ""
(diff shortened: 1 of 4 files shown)


vdbergh:
Made it WIP since I need to think about a possible issue.

vdbergh:
Unwipped!

ppigazzini:
DEV updated.

```
usr00@dfts-0:~$ sudo systemctl stop cron
usr00@dfts-0:~$ sudo systemctl stop fishtest@{6543..6544}
usr00@dfts-0:~$ ${VENV}/bin/python3 fishtest/server/utils/convert_actions.py | tee convert_actions.py.log
Starting conversion...
Actions converted: 681041.
Conversion finished in 1597.95 seconds. Time per run: 2.35ms.
usr00@dfts-0:~$ sudo systemctl start fishtest@{6543..6544}
usr00@dfts-0:~$ sudo systemctl start cron
```

vdbergh:
Everything seems to be ok with the converted db.

ppigazzini (approved):

ppigazzini:
PROD updated, DB backupped, starting collection conversion.

ppigazzini:
@vdbergh collection converted :)

vdbergh:
Thanks! Next step!

PR #1478, opened 2022-12-11
PR3: Directly write to the actiondb using the new schema.

vdbergh:
The actiondb now has a well defined schema which is checked before insertions. This should facilitate debugging.
    
The script: utils/convert_actions.py deletes superfluous data from the actiondb. This should save a lot of space.
    
Note that the database still contains enough information to reconstruct the deleted data.
    
 No user visible changes.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -1,77 +1,63 @@
 from datetime import datetime
 
+from bson.objectid import ObjectId
+from fishtest.util import optional_key, union, validate, worker_name
 from pymongo import DESCENDING
 
-"""
-schema = {
-"action" : str,
-"username" : str,
-"time" : datetime.datetime,
-optional_key("message") : str,
-optional_key("user") : str,
-optional_key("run_id") : ObjectId,
-optional_key("run") : str,
-optional_key("worker") : str,
-optional_key("nn") : str,
-}
-"""
-
-
-def expand_action(action):
-    action["message"] = ""
-    if action["action"] == "update_stats":
-        action["message"] = "Update user statistics"
-    elif action["action"] == "upload_nn":
-        action["nn"] = action["data"]
-    elif action["action"] == "block_user":
-        action["message"] = "blocked" if action["data"]["blocked"] else "unblocked"
-        action["user"] = action["data"]["user"]
-    elif action["action"] == "modify_run":
-        action["run"] = action["data"]["before"]["args"]["new_tag"]
-        action["run_id"] = action["data"]["before"]["_id"]
-        message = []
-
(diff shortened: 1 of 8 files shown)


vdbergh:
I made this WIP since I think it needs to be refactored a bit more.

ppigazzini:
> I made this WIP since I think it needs to be refactored a bit more.

GitHub has introduced the "draft" status. In that way nobody can erroneously merge the PR until the status is switched back to "ready for review", see:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

vdbergh:
Thanks!

vdbergh:
Unwipped and undrafted!

ppigazzini:
DEV updated!

ppigazzini:
Events seems registered correctly, logs clean.

vdbergh:
I retested all events on DEV except `upload_nn` since I cannot generate a new nn.  But I did test this locally.

ppigazzini:
`upload_nn` works, see https://dfts-0.pigazzini.it/actions?max_actions=1&before=1670926794.616

for future reference:
```bash
#!/bin/bash
dd if=/dev/urandom of=nn.nnue bs=1MiB count=16
nn_sha=$(sha256sum nn.nnue)
mv nn.nnue nn-${nn_sha:0:12}.nnue
```

vdbergh:
> `upload_nn` works, see https://dfts-0.pigazzini.it/actions?max_actions=1&before=1670926794.616
> 
> for future reference:
> 
> ```shell
> #!/bin/bash
> dd if=/dev/urandom of=nn.nnue bs=1MiB count=16
> nn_sha=$(sha256sum nn.nnue)
> mv nn.nnue nn-${nn_sha:0:12}.nnue
> ```

Yes I know but I was a afraid that the fake nn would find its way to the cdn. I guess this is not the case then.

ppigazzini:
The nn file is uploaded on DEV, courtesy of the server setup script.

```
# set secrets
sudo -i -u ${user_name} << EOF
echo '' > fishtest.secret
echo '' > fishtest.captcha.secret
echo 'http://127.0.0.1/upload_net' > fishtest.upload
```

The download of the nn is from the cdn as default, but there is a script to switch back and forth with DEV.

vdbergh:
Ok. Thanks anyway for testing!

ppigazzini (approved):
`purge_run` finally tested :)
https://dfts-0.pigazzini.it/actions?max_actions=1&before=1670931593.383

ppigazzini:

PR #1479, opened 2022-12-13
PR4: Add a free text search field to the event log page.

vdbergh:
For this to work one needs to run

utils/create_indexes.py actions

to create a text index on the string fields of an event:

action, username, worker, message, run, nn.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -69,6 +69,7 @@ class ActionDb:
         self,
         username=None,
         action=None,
+        text=None,
         limit=0,
         skip=0,
         utc_before=None,
@@ -81,6 +82,8 @@ class ActionDb:
             q["action"] = {"$nin": ["update_stats", "dead_task"]}
         if username:
             q["username"] = username
+        if text:
+            q["$text"] = {"$search": text}
         if utc_before:
             q["time"] = {"$lte": utc_before}
 
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -2,6 +2,7 @@
 
 <%
   import datetime
+  import urllib
 %>
 
 <h2>Events Log</h2>
@@ -39,7 +40,7 @@
       type="text"
       name="user"
       list="users-list"
-      value="${request.GET.get('user') if request.GET.get('user') is not None else ''}"
+      value='${username_param|n}'
     />
     <datalist id="users-list">
       % for user in request.userdb.get_users():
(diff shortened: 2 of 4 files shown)


ppigazzini:
DEV updated.

ppigazzini:
DEV updated on the fly.

vdbergh:
> DEV updated. Creating indexes... done. LGTM, but it seems that are required at least 4 letters to start a search,

MongoDb text indices are a bit limited.  You can only search for full words (e.g. when searching for `user01-2cores-9e48562f-f444*` you can search for `9e48562f` but not for `9e4856`). An alternative for text search is regex search but this cannot use an index.

ppigazzini:
The search with the incomplete word "fail" works, though.
https://dfts-0.pigazzini.it/actions?action=&user=ChessDBCN&text=fail

vdbergh:
> The search with the incomplete word "fail" works, though. https://dfts-0.pigazzini.it/actions?action=&user=ChessDBCN&text=fail

Yes there seems to be an undocumented exception that you can do partial word matches on the start of a field...

vdbergh:
On the internet I saw some recommendations to redo a text search using regex search in case the text search returns nothing. However if the regex search also doesn't match anything then this means scanning the whole collection (even with a paginated search) which will be very slow.

vdbergh:
There are still a few bugs with escaping/unescaping the test field. I will look at it tonight. For now I am marking this as draft.

vdbergh:
> BTW I was asking only to understand how to use the search field.

Yes I agree it is not very discoverable (although it is similar to Google search). I think it will be best to add a dialog box with some explanations. A good model is the dialog box for the autoselect SPSA parameters.

vdbergh:
Undrafted. I think I fixed the quoting issues and I also added a message box with some information about the free text search.

vdbergh:
Hmm still not fully fixed.

vdbergh:
Hopefully it is fixed now.

ppigazzini:
DEV update on the fly before running to take the train :)

vdbergh:
> DEV update on the fly before running to take the train :)

Thanks!

vdbergh:
I changed the index to have no default language (so it should be regenerated with the script). Otherwise the text search does some funky stuff with stemming which may be useful for large text databases but which in our context is only confusing.

ppigazzini:
DEV updated, actions indexes rebuilt.

vdbergh:
> DEV updated, actions indexes rebuilt.
> 
> Now the search returns only for a whole word of any length: https://dfts-0.pigazzini.it/actions?action=&user=&text=to
> 
> The info box is a classy touch :)
> 
> > When passed a search string that only contains negated words, text search will not match any documents
> 
> :(

Yeah I got to dislike a mongodb a bit for all these arbitrary restrictions. But we have to live with it!

Anyway I think the PR is fine now.

ppigazzini (approved):

ppigazzini:
PROD and indexes updated, thank you @vdbergh :)

PR #1480, opened 2022-12-15
Fix quoting the proper way

vdbergh:
It seems that using `${string|u}` in a mako template implies that `string` is first encoded as a URL and then again as html. This seems wrong to me if the URL is used as a href and gives strange results if the parameters are then used in a text field.
    
The workaround seems to be to use `${string|u,n}`. This is what we do in this PR.

Although it seems to be no longer required in html5 we are encoding the `&` separating the parameters by `&amp;`.

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -40,7 +40,7 @@
       type="text"
       name="user"
       list="users-list"
-      value='${username_param|n}'
+      value="${username_param}"
     />
     <datalist id="users-list">
       % for user in request.userdb.get_users():
@@ -60,11 +60,11 @@
             <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
           </div>
           <div class="modal-body text-break">
-            This will perform a case insensitive free text search on the fields event name, user name, worker name,
+            This will perform a case insensitive text search on the fields event name, user name, worker name,
             target user, network name, branch name and comment.
-            In most cases one can only search for a list of full words, i.e. strings bounded by delimiters.
-            For example the worker name <i>chessuser-128cores-abcdefgh-uvwx</i> matches <i>abcdefh</i> but not <i>abcde</i>.
-	    To force a word or a combination of words to be
+            One can only search for a list of full words, i.e. strings bounded by delimiters.
+            For example the worker name <i>fishtestfan-128cores-abcdefgh-uvwx</i> matches <i>abcdefgh</i> but not <i>abcde</i>.
+            To force a word or a combination of words to be
             included in the result, use quotes like in <i>&quot;dog cat&quot;</i>. To exclude a word, precede
             it with a minus sign like in <i>dog &minus;cat</i>. For more information see
             <a https://www.mongodb.com/docs/manual/reference/operator/query/text/#mongodb-query-op.-text>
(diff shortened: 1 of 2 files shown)


vdbergh:
I put some weird stuff in the text input field and ran the resulting html through the html validator at https://validator.w3.org/ . The good news is that the URLs generated by this PR are valid html.

The bad news that there are quite a few html errors at other places. These should be fixed (preferably by the people which are responsible for the current html code) but this is outside the scope of this PR.

server/fishtest/templates/actions.mak
```diff
@@ -60,11 +60,11 @@
             <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
           </div>
           <div class="modal-body text-break">
-            This will perform a case insensitive free text search on the fields event name, user name, worker name,
+            This will perform a case insensitive text search on the fields event name, user name, worker name,
             target user, network name, branch name and comment.
-            In most cases one can only search for a list of full words, i.e. strings bounded by delimiters.
+            One can only search for a list of full words, i.e. strings bounded by delimiters.
```
ppigazzini, line 65: > For example the worker name chessuser-128cores-abcdefgh-uvwx matches **abcdefh** but not abcde.

If I'm not wrong should be "abcdefgh", reported by salvo_sf-fan on discord

ppigazzini:
DEV updated, LGTM.

server/fishtest/templates/actions.mak
vdbergh: Fixed!

vdbergh:
LGTM too...

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @vdbergh and thank you at "salvo_sf_fan" on discord!

ppigazzini:
> I put some weird stuff in the text input field and ran the resulting html through the html validator at https://validator.w3.org/ . The good news is that the URLs generated by this PR are valid html.
> 
> The bad news that there are quite a few html errors at other places. These should be fixed (preferably by the people which are responsible for the current html code) but this is outside the scope of this PR.

Could you open an issue making some examples?

I copied the html for the text search "too -slow" and I view few errors (if I understood correctly your suggestion)
https://tests.stockfishchess.org/actions?action=failed_task&user=ChessDBCN&text=too+-slow

image

ppigazzini:
Anyway I thin to have understood. We need to write our URIs in the validator and try to fix all the errors. I just caught a not closed `<a>` tag in `users.mak`

ppigazzini:
@vdbergh 
I'm not able to fix the url in href for an username with a space.
I write the right url in mako with python (str.replace or urllib.parse.quote) 
https://dfts-0.pigazzini.it/tests/user/Andrew%20Grant
The page ie rendered with link:
https://dfts-0.pigazzini.it/tests/user/Andrew Grant

I tried unsuccessfully some mako filtering (or disabling the filtering). Suggestions?

peregrineshahin:
Prettier forces the `\` on void elements.

ppigazzini:
Yes, I read, but we can fix the mako templates that are not formatted with `prettier`

https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#configuring-tools-to-not-output-trailing-slashes-for-void-elements

PR #1483, opened 2022-12-16
Make the task target links in the event log point to tasks.

vdbergh:
Also: include the bytes 2-4 of the run_id in the run name (there may be many tests with the same branch, currently these are indistinguishable).

One cannot yet search for the extra data. This will require again a database conversion.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -500,6 +500,11 @@ td {
   background-color: #cff4fc;
 }
 
+/* Background colors for a highlighted task */
+.table tbody tr.highlight > td {
+  background-color: #fff0f5;
+}
+
 /* Pagination links */
 .pagination {
   --bs-pagination-color: var(--bs-body-color);
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -236,6 +236,12 @@ div#tasks thead {
   background-color: #0f3373;
 }
 
+/* Background colors for a highlighted task */
+.table tbody tr.highlight > td {
+  background-color: #800080;
+}
+
+
 /* Override - test tasks table - purged tasks red background */
 td[style*="background-color:#ffebeb"] {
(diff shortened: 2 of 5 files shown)


vdbergh:
There is currently a bug in fishtest that after purging `task_id`'s change (a task does not know its own id) so links to specific tasks in the event log may become incorrect  and more seriously, theoretically workers can become confused as well when suddenly the `task_id` they are using points to a different task. 

I will fix this in another PR but it cannot be done retroactively (the required information is gone).

ppigazzini:
DEV updated, LGTM.

https://dfts-0.pigazzini.it/tests/view/639011178536e3e199afe346?show_task=299

vdbergh:
Do you have the RGB values for your suggestion?

There is a slight issue with the diff which sometimes pushes the selected task down. I'll have a look at this.

ppigazzini:
> Do you have the RGB values for your suggestion?
> 
> There is a slight issue with the diff which sometimes pushes the selected task down. I'll have a look at this.

I'm experimenting, I noted that we have already this color for the purged tasks, searching for the RGB

image

ppigazzini:
What about the light purple `#ffbbff`?

image

image

ppigazzini:
@dav1312 suggestion for the colors?

ppigazzini (approved):
I will fix the color for the light theme during the merge,

ppigazzini:

vdbergh:
I am still struggling with the diff issue (this part of the page is entirely done in javascript). On my local installation the diff seems to be always hidden so I cannot replicate the issue. Perhaps someone who understands the javascript code can help?

ppigazzini:
@peregrineshahin could help with javascript code

vdbergh:
Ok whether a diff is shown or hidden appears to depend on the diff itself (perhaps its size?).

vdbergh:
Ok found it in the javascript code (hide large diffs by default).

PR #1484, opened 2022-12-17
Hide diff in case we are scrolling to a specific task.

vdbergh:


ppigazzini:
DEV updated, let me known if it's working :)

ppigazzini (approved):
Tested working

ppigazzini:

PR #1485, opened 2022-12-17
fixe access to the response variable

vondele:
fixes:
```
Dec 17 13:17:10 tests.stockfishchess.org pserve[22144]:   File "/home/fishtest/fishtest/server/fishtest/views.py", line 194, in sync_upload
Dec 17 13:17:10 tests.stockfishchess.org pserve[22144]:     "Network upload failed: " + str(response.status_code), "error"
Dec 17 13:17:10 tests.stockfishchess.org pserve[22144]: UnboundLocalError: local variable 'response' referenced before assignment
```

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -191,7 +191,7 @@ def sync_upload(request):
     except Exception as e:
         print("Network upload failed: " + str(e))
         request.session.flash(
-            "Network upload failed: " + str(response.status_code), "error"
+            "Post request for the network upload failed", "error"
         )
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:
Applied as fast bug fix, but that function should be not called without clicking on the button, so something is still broken IMO.

vondele:
I think that uncaught exception results in somehow a missing return, and the reset of the logic being influenced by it... just wishful thinking probably.

PR #1486, opened 2022-12-17
Append a hash of the run_id to the run name (=branch).

vdbergh:
The run_id itself is not very suitable since it is not a hash but a combinations of a timestamp, a random number, computed once per process, and a counter initialized to a random value. It is not clear which part one should take.

The sha of the commit is also not suitable since there may be multiple runs with the same commit (e.g. different TC).

If the scheme in this PR is acceptable then I will make a PR to regenerate the actiondb so that the hash becomes searchable.

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -2,6 +2,7 @@
 
 <%!
   import datetime
+  from fishtest.util import hex_print
 %>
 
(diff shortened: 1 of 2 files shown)


vdbergh:
I am closing this since without db conversion it is a bit pointless. I am going to submit a PR with db conversion.

PR #1487, opened 2022-12-17
Append a hash of the run_id to the run name (=branch).

vdbergh:
The run_id itself is not very suitable since it is not a hash but a combinations of a timestamp, a random number, computed once per process, and a counter initialized to a random value. It is not clear which part one should take.

The sha of the commit is also not suitable since there may be multiple runs with the same commit (e.g. different TC).

Before this works utils/convert_actions.py should be run.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 from bson.objectid import ObjectId
-from fishtest.util import optional_key, union, validate, worker_name
+from fishtest.util import hex_print, optional_key, union, validate, worker_name
 from pymongo import DESCENDING
 
 schema = union(
@@ -60,6 +60,12 @@ schema = union(
 )
 
 
+def run_name(run):
+    run_id = run["_id"]
+    run = run["args"]["new_tag"]
+    return run[:23] + "-" + hex_print(run_id)[0:7]
+
+
 class ActionDb:
     def __init__(self, db):
         self.db = db
@@ -79,7 +85,7 @@ class ActionDb:
         if action:
             # update_stats is no longer used, but included for backward compatibility
             if action == "system_event":
-                q["action"] = {"$in" : ["system_event", "update_stats"]}
+                q["action"] = {"$in": ["system_event", "update_stats"]}
             else:
                 q["action"] = action
         else:
@@ -110,7 +116,7 @@ class ActionDb:
             username=username,
             worker=worker_name(task["worker_info"]),
             run_id=run["_id"],
-            run=run["args"]["new_tag"],
+            run=run_name(run),
             task_id=task_id,
             message=message[:1024],
         )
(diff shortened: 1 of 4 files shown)


vdbergh:
@ppigazzini  I wonder if you have some time to look into this PR? The conversion of the actiondb is referersible (no information is lost).

For the event log it is really important to disambiguate the runs since typically several runs are created from the same branch. Currently the disambiguation is done visually in `actions.mak` but one cannot search for the disambiguating hex string since it is not contained in the actiondb. 

After this PR I would like to add events for the creation of a run, and for a run that finishes normally. In this way one can search for the full life cycle of a run in the event log.

ppigazzini:
In the next days I should have some spare time to check the complex PRs waiting tests/reviews.

ppigazzini:
Still converting the actions collection...
While waiting I configured a maintenance page.

vdbergh:
Hmm. I guess it is much slower since I am taking every run from the rundb. I wanted to make sure that the conversion is idempotent.  If we drop this requirement then it can be done much faster.

vdbergh:
The page is rather nice! But the down time seems unacceptably large.

vdbergh:
I added a commit with a version which is still idempotent but which should be faster during the first conversion.

ppigazzini:
Finished now the first version with an error:
```
time sudo bash apply_pr1487.sh
Starting conversion...
  File "/home/usr00/fishtest/server/utils/convert_actions.py", line 21, in <module>
    action["run"] = run_name(run)
  File "/home/usr00/fishtest/server/fishtest/actiondb.py", line 64, in run_name
    run_id = run["_id"]
TypeError: 'NoneType' object is not subscriptable

real    195m12.281s
user    161m30.516s
sys     8m48.127s
```

vdbergh:
Yek. This means an inconsistency in the actiondb (reference to a run not existing in the rundb). I had not anticipated this. I will make it skip such records.

vdbergh:
New version. If you still have access to the non-converted version, then it better to use that one since it will be faster.

ppigazzini:
Reloading the DB and starting the conversion.
Added some snow to the maintenance page :)

ppigazzini:
Started now.

ppigazzini:
Successful and fast, DEV online now :)
```
$ time sudo bash x/apply_pr1487.sh
Starting conversion...
Actions converted: 708062.
Conversion finished in 1721.66 seconds. Time per run: 2.43ms.

real    29m25.315s
user    15m18.934s
sys     3m16.027s
```

vdbergh:
It seems to work. The hash of the run is now searchable. I started a worker and then killed it and the run got the correct hash in the event log. I will squash the PR (the squashing only affects the conversion program).

ppigazzini (approved):
LGTM

ppigazzini:

PR #1491, opened 2022-12-19
Better handling of scroll to task.

vdbergh:
The case of the asynchronously appearing SPSA graph should now also be fixed, although I did not test this.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -10,12 +10,20 @@
 
 % if show_task >= 0:
 <script>
-  window.addEventListener("load", (event) => {
-    document.getElementById("task${show_task}")
+  function scroll_to(task_id) {
+    document.getElementById("task" + task_id)
     .scrollIntoView();
+  }
+  document.addEventListener("DOMContentLoaded", (event) => {
+    const resize_observer = new ResizeObserver((entries) => {
+      scroll_to(${show_task});
+    });
+    resize_observer.observe(document.body);
+    scroll_to(${show_task});
   });
 </script>
 % endif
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, here is a worker with a failed SPSA task, let me know if it's working as intended:

https://dfts-0.pigazzini.it/actions?max_actions=1&action=failed_task&user=mgrabiak&text=mgrabiak-3cores-0a8fbe43&before=1670038940.979
https://dfts-0.pigazzini.it/tests/view/638a9b91d2b9c924c4c636d8?show_task=35

vdbergh:
It works even better than I expected :) Without this PR the purple row would scroll away once the the graph became visible.

vdbergh:
I did some more testing and it really works well. No more need to hide the diff.

ppigazzini (approved):
LGTMVDB!

ppigazzini:
Ops, one error with the W3C validator
https://validator.w3.org/nu/?doc=https%3A%2F%2Fdfts-0.pigazzini.it%2Ftests%2Fview%2F638a9b91d2b9c924c4c636d8%3Fshow_task%3D35
```html
Error: Attribute value missing.
At line 616, column 18
  <thead class=>↩      <tr>↩  
```

ppigazzini (approved):
The validation problem is already in master:
https://validator.w3.org/nu/?showsource=yes&doc=https%3A%2F%2Ftests.stockfishchess.org%2Ftests%2Fview%2F638a9b91d2b9c924c4c636d8%3Fshow_task%3D35#cl608c18

ppigazzini:

PR #1492, opened 2022-12-19
Fix some missed W3C Validation errors.

peregrineshahin:
Basically, the first one regarding the button is not catchable because it only "triggers" while submitting the test, and the other one was overlooked because of rebasing.

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -859,7 +859,7 @@
       }
       form_submitted = true;
       const submitButton = document.getElementById("submit-test");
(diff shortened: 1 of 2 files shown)


ppigazzini:
Fixed!
https://validator.w3.org/nu/?doc=https%3A%2F%2Fdfts-0.pigazzini.it%2Ftests%2Fview%2F638a9b91d2b9c924c4c636d8%3Fshow_task%3D35

ppigazzini (approved):

ppigazzini:
PRDO updated, thank you @peregrineshahin :)

PR #1494, opened 2022-12-21
Provide task links in the machine table.

vdbergh:


vdbergh:
There are currently many machines working on runs all called `master` even though these represent different runs. One can disambiguate them by adding a hash of the run id to their name (as will be done in the event log) but I think in the machine table this will look rather ugly and it seems not necessary as in practice several different runs running at the same time using the same branch name will be rather rare. The only reasonable case would be regression tests.

ppigazzini (approved):
LGTM

vdbergh:
Yes it seems ok (it is only a small patch).

ppigazzini:
> There are currently many machines working on runs all called `master` even though these represent different runs.

The run disambiguation is only a click away, so IMO it's not necessary to add other information.

ppigazzini:

ppigazzini:
> Yes it seems ok (it is only a small patch).

Small patch but very helpful, it avoids to scroll to the end of the tasks list.

PR #1496, opened 2022-12-23
Faster rendering and no more artifacts when processing ?show_task=xxx.

vdbergh:
We set `scroll-behavior` to `auto` (instantaneous scrolling) and we only make the page visible when the scrolling is complete.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -10,6 +10,7 @@
 
 % if show_task >= 0:
 <script>
+  document.documentElement.style="scroll-behavior:auto !important; visibility:hidden !important;";
   function scroll_to(task_id) {
     document.getElementById("task" + task_id)
       .scrollIntoView();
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
DEV updated, more faster when opening in a new window, I never viewed artifacts so I'm not able to verify the claim :)

vdbergh:
I am not sure artifacts is the right terminology. Before, the top of the page would be rendered first (not interesting in this case) and then there would be a _slow_ scrolling to the requested task. This PR fixes both issues.

ppigazzini:
Ok, that was "the faster opening" for me.

ppigazzini:
Mmmm, there is a little problem.
When clicking on a task now the left bar disappear for some time.
DEV running on previous master to compare.

vdbergh:
Your observation is correct. It did not bother me (I thought the fact that there is no visible scrolling anymore if more important). I think the issue of the flashing left pane can be fixed by only setting the visibility of the containing diff on the right to hidden.

ppigazzini:
Fixed by https://github.com/glinscott/fishtest/pull/1497

PR #1497, opened 2022-12-23
Make sure that the left pane does not disappear during scrolling.

vdbergh:


ppigazzini:
DEV updated, LGTM on Chrome and Edge.

vdbergh:
I bug: a string, intended as a mako comment appears on the page. Need to check.

ppigazzini:
https://docs.makotemplates.org/en/latest/syntax.html#comments

> The single line comment uses ## as the first non-space characters on a line

vdbergh:
Fixed. I think.

ppigazzini:
DEV updated, comment fixed, firefox OK.

ppigazzini (approved):

ppigazzini:

PR #1499, opened 2022-12-24
Keep table heading visible after scrolling.

vdbergh:
Also: cleanups of latest scroll patch.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -10,10 +10,13 @@
 
 % if show_task >= 0:
 <script>
-  document.documentElement.style="scroll-behavior:auto !important; overflow:hidden !important";
+  document.documentElement.style="scroll-behavior:auto; overflow:hidden;";
   function scroll_to(task_id) {
-    document.getElementById("task" + task_id)
-      .scrollIntoView();
+    const task_offset = document.getElementById("task" + task_id).offsetTop;
+    const tasks_head_height = document.getElementById("tasks-head").offsetHeight;
+    const tasks_div = document.getElementById("tasks");
+    tasks_div.scrollIntoView();
+    tasks_div.scrollTop = task_offset - tasks_head_height;
   }
 </script>
 % endif
@@ -38,7 +41,7 @@
     </script>
 % endif
 
-<div id="enclosure" ${"style=\"visibility:hidden !important;\"" if show_task >= 0 else "" |n}>
+<div id="enclosure"${' style="visibility:hidden;"' if show_task >= 0 else "" |n}>
 
 <h2>
   <span>${page_title}</span>
@@ -341,7 +344,7 @@
 <div id="tasks"
      class="overflow-auto ${'collapse show' if tasks_shown else 'collapse'}">
   <table class='table table-striped table-sm'>
-    <thead ${'class="sticky-top"' if show_task == -1 else "" | n}>
+    <thead id="tasks-head" class="sticky-top">
       <tr>
         <th>Idx</th>
         <th>Worker</th>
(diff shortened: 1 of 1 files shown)


server/fishtest/templates/tests_view.mak
```diff
@@ -10,10 +10,13 @@
 
 % if show_task >= 0:
 <script>
-  document.documentElement.style="scroll-behavior:auto !important; overflow:hidden !important";
+  document.documentElement.style="scroll-behavior:auto; overflow:hidden;";
   function scroll_to(task_id) {
-    document.getElementById("task" + task_id)
-      .scrollIntoView();
+    task_offset = document.getElementById("task" + task_id).offsetTop;
```
peregrineshahin, line 15: suggestions on little cleanup comments on the places that needs some cleanups as I didn't test the effects yet:
1- Variables should be declared as const or let.
2- We use camelCase for JavaScript and snake_case for python.

server/fishtest/templates/tests_view.mak
```diff
@@ -341,7 +344,7 @@
 <div id="tasks"
      class="overflow-auto ${'collapse show' if tasks_shown else 'collapse'}">
   <table class='table table-striped table-sm'>
-    <thead ${'class="sticky-top"' if show_task == -1 else "" | n}>
+    <thead id="tasks_head" class="sticky-top">
```
peregrineshahin, line 347: Better use kebab-case for id's

server/fishtest/templates/tests_view.mak
```diff
@@ -38,7 +41,7 @@
     </script>
 % endif
 
-<div id="enclosure" ${"style=\"visibility:hidden !important;\"" if show_task >= 0 else "" |n}>
+<div id="enclosure" ${"style=\"visibility:hidden;\"" if show_task >= 0 else "" |n}>
```
peregrineshahin, line 44: For consistency, I think it is better to use a combination of single quotes and double quotes,
something like:
`<div id="enclosure" ${'style="visibility:hidden;"' if show_task >= 0 else "" |n}>`
would be better read no?

peregrineshahin (commented):
Mind these small suggestions.

vdbergh:
> Mind these small suggestions.

I fixed all your comments, except for the use of camel case in Javascript. All my code in fishtest is snake_case, so I better keep using that for consistency (changing all the code would be too much work now). Besides I think snake_case is much easier on the eye.

ppigazzini:
Mmm, the server workflow is a couple of weeks that fails randomly with:
```
ERROR: test_10_get_bench (test_run.CreateRunTest)
  File "/home/runner/work/fishtest/fishtest/server/tests/test_run.py", line 9, in test_10_get_bench
    self.assertTrue(re.match("[0-9]{7}|None", str(get_master_info()["bench"])))
TypeError: 'NoneType' object is not subscriptable

```

vdbergh:
> Mmm, the server workflow is a couple of weeks that fails randomly with:
> 
> ```
> ======================================================================
> ERROR: test_10_get_bench (test_run.CreateRunTest)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/work/fishtest/fishtest/server/tests/test_run.py", line 9, in test_10_get_bench
>     self.assertTrue(re.match("[0-9]{7}|None", str(get_master_info()["bench"])))
> TypeError: 'NoneType' object is not subscriptable
> 
> ----------------------------------------------------------------------
> ```

I noticed that too. We should debug this.

ppigazzini:
DEV updated with this and the #1500

ppigazzini:
LGTM

ppigazzini (approved):

ppigazzini:

PR #1500, opened 2022-12-24
Convert some promiscuous code to async/await.

vdbergh:
The async/await framework allows one to write asynchronous code as if it were synchronous. It uses promises under the hood.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -8,6 +8,14 @@
     spsa_data = json.dumps(run["args"]["spsa"])
 %>
 
+<script>
+function event_listener(event) {
+  return new Promise((resolve,reject) => {
+    window.addEventListener(event, resolve);
+  });
+}
+</script>
+
 % if show_task >= 0:
 <script>
   document.documentElement.style="scroll-behavior:auto !important; overflow:hidden !important";
@@ -30,7 +38,7 @@
             integrity="sha384-${cache_busters['js/spsa.js']}"
             crossorigin="anonymous"></script>
     <script>
-      const spsa_promise = spsa();
+      const spsa_promise = handle_spsa();
     </script>
 % else:
     <script>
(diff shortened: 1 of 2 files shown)


server/fishtest/templates/tests_view.mak
```diff
@@ -559,22 +566,23 @@
...
-        .then((response) => {
-          if (response.ok) return response.text();
-        })
-        .then((text) => {
-          if (!text) return Promise.resolve();
-          return showDiff(text);
-        });
+      });
+      if(response.ok) {
+        var text = await response.text();
```
peregrineshahin, line 576: The deceleration of this variable here will most certainly make it undefined if the response was not ok, resulting in a fatal runtime error, the same goes for the json variable.

better is to declare the variable with a null value outside before the if checking.

vdbergh: This is just a literal translation of the current code (which has no error handling) to async/await. I would like to keep this as a separate PR. Later we can clean up the code and add error handling, which is easy in the async/await framework. The reason for not doing it now is that it is laborious to test the error handling code paths, and moreover error handling may paper over genuine bugs.

ppigazzini:
LGTM. The code is very clean now, it was a nightmare when I wrote some PRs :)

peregrineshahin:
I notice that the SPSA pre-loader (image/giff) has stopped showing even before this PR, but I'm not sure it is needed from a feature POV.

ppigazzini:
AFTK for a couple of hours, I kept open the PRs for review and test.

vdbergh:
> I notice that the SPSA pre-loader (image/giff) has stopped showing even before this PR not sure when, but I'm not sure it is needed from a feature POV.

I noticed that too. I’ll have a look at (probably not today…, family here).

ppigazzini:
I wonder what is the last warning, it's in all our pages, I don't find that header in the source.

image

peregrineshahin:
It is related to chrome extensions, not Fishtest, I disabled all my chrome extensions and the error disappeared, not sure which one is the culprit in my case though.

vdbergh:
In the latest push there are a few tweaks to make the code slightly more iconic.

vondele:
iconic :-)

ppigazzini (approved):

ppigazzini:

PR #1501, opened 2022-12-25
Record server shutdown and restart as system events in the event log.

vdbergh:
Also: retire the update_stats event and replace it with the more generic system_event.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -22,7 +22,7 @@ schema = union(
         "run": str,
         "task_id": int,
     },
-    {"action": "update_stats", "username": "fishtest.system", "message": str},
+    {"action": "system_event", "username": "fishtest.system", "message": str},
     {"action": "new_run", "username": str, "run_id": ObjectId, "run": str},
     {"action": "upload_nn", "username": str, "nn": str},
     {
@@ -77,9 +77,13 @@ class ActionDb:
     ):
         q = {}
         if action:
-            q["action"] = action
+            # update_stats is no longer used, but included for backward compatibility
+            if action == "system_event":
+                q["action"] = {"$in" : ["system_event", "update_stats"]}
+            else:
+                q["action"] = action
         else:
-            q["action"] = {"$nin": ["update_stats", "dead_task"]}
+            q["action"] = {"$nin": ["system_event", "update_stats", "dead_task"]}
         if username:
             q["username"] = username
         if text:
@@ -143,9 +147,9 @@ class ActionDb:
             task_id=task_id,
         )
 
-    def update_stats(self, message=None):
+    def system_event(self, message=None):
         self.insert_action(
-            action="update_stats",
+            action="system_event",
             username="fishtest.system",
             message=message,
         )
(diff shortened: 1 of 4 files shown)


vdbergh:
CI fails. This is because the bench test does a github api call and it likely hits the ratelimit.

ppigazzini:
> CI fails. This is because the bench test does a github api call and it likely hits the ratelimit.

Testing the github token
https://github.com/glinscott/fishtest/commit/1cfb4317b37dc22411e368b9d6177e5a0f89ec9b

vdbergh:
The documentation says that the GITHUB_TOKEN is only valid for the workflow repo.

vdbergh:
Unless I misunderstand and you were using a token of your own somehow (I don't really understand how secrets work).

ppigazzini:
I simply forgot that that "should" work (with the macos runner didn't worked) only for the workflow repo.

ppigazzini:
DEV updated, LGTM.
Tested with:
```
sudo reboot
sudo systemctl restart fishtest@{6543..6544}
sudo systemctl restart fishtest@6543
sudo systemctl restart fishtest@6544
sudo systemctl stop fishtest@{6543..6544}
sudo systemctl start fishtest@{6543..6544}
sudo systemctl stop fishtest@6543
sudo systemctl start fishtest@6543
${VENV}/bin/python3 fishtest/server/utils/delta_update_users.py
```

vdbergh:
Looking at the messages, it seems it would be useful to include the port number in the message.

vdbergh:
Hmm this seems to be not so obvious. The pyramid application does not seem to know the port it is listening on. The only way I see is to parse the command line via `sys.argv`.

vdbergh:
Ok I did it via parsing of the command line (until a better method is found to get the port number).

ppigazzini:
DEV updated, LGTM, I'm away from the keyboard right now, not able to search for a better way for the port. I wonder why 6542 as default.

vdbergh:
6542 is what is contained in development.ini. I think a more robust way is to parse the config file using the command line arguments (assuming the config file is a regular ini file).

vdbergh:
Ok I am now parsing the config file. This should be robust.

vdbergh:
Still a small bug. If we instantiate `Rundb` directly (e.g. in `delta_update_users()` or in CI ) there should be no event log messages.

vdbergh:
Fixed.

ppigazzini:
DEV updated. It look very good to me.
Nit-picking. What about using the same command name from `systemctl` (ie `stop` and `start`)?

vdbergh:
> DEV updated. It look very good to me. Nit-picking. What about using the same command name from `systemctl` (ie `stop` and `start`)?

Fine for me! I will change it. I think the order of the arguments is also the opposite (i.e. start fishtest:6542).

ppigazzini:
Yes `<sytemd_unit_file>@<parameter>`
```
sudo systemctl start fishtest@6543
sudo systemctl start fishtest@{6543..6544}
# this is the development/debug systemd unit file
sudo systemctl start fishtest_dbg
```

vdbergh:
Done. Now the systemd syntax is followed.

ppigazzini (approved):

ppigazzini:

PR #1502, opened 2022-12-27
(cleanup) Support the `listen host:port` directive in the ini file.

vdbergh:
This directive is legal. It is currently not used but someone might use it in the future and then things would fail silently.

Also better error handling if we are not able to obtain the port number.


ppigazzini (approved):
DEV updated. Tests LGTM.

ppigazzini:

PR #1503, opened 2022-12-27
Rewrite live_elo.js using async/await.

vdbergh:


ppigazzini:
DEV update, started some workers.

In this test the Elo Gauge the red/green zones are missing...
https://dfts-0.pigazzini.it/tests/live_elo/63ab646a77ef25e68d3be6b0

image

vdbergh:
That's also in master. Basically the confidence interval (which would be green) is "too far out" to be visible.

ppigazzini:
LGTM

vdbergh:
I wonder if the notify stuff still works. I do not know how to test it.

ppigazzini:
> I wonder if the notify stuff still works. I do not know how to test it as I don't know what it is supposed to do. Also my local fishtest is on http and Notify does not work over http.

:)

image

vdbergh:
In that case I think things work as they should!

ppigazzini (approved):

ppigazzini:

PR #1507, opened 2022-12-29
Simplified main loop in live_elo.js.

vdbergh:
It should also be robust now. All error paths were tested.

To simplify the logic we implement `raise_for_status()` as in the python requests package.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -142,6 +142,20 @@ function supportsNotifications() {
   return false;
 }
 
+function raise_for_status(response) {
+  // A useful function from the python requests package
+  if (response.ok) {
+    return;
+  }
+  throw `request failed with status code ${response.status}`;
+}
+
+async function fetch_json(url) {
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated, tests scheduled, workers joined. I will be away form the keyboard for some hours.

vdbergh:
I created a generic async function `fetch_json()` and moved it to `application.js`. It will be useful to simplify `fetch_diff()`  (and make it more robust) in `tests_view.py`.

ppigazzini:
DEV updated.

ppigazzini:
I wonder if these PRs are a workout before https://docs.python.org/3.6/library/asyncio.html

vdbergh:
You are right! This is a useful learning exercise to understand the async/await paradigm!

ppigazzini (approved):

ppigazzini:
Notifications still working fine, PROD updated, thank you @vdbergh :)

PR #1508, opened 2022-12-29
Refactor and flatten the function `handle_diff()` in tests_view.mak.

vdbergh:
Also: delete v3 in the media type for a diff. It is no longer required. See:
    
https://docs.github.com/en/rest/overview/media-types?apiVersion=2022-11-28


vdbergh:
This is a followup on the last PR.

ppigazzini:
DEV updated. LGTM.

ppigazzini:

PR #1510, opened 2022-12-30
Replace quotation marks by standard ones.

### LINKED ISSUE #1509: Free text search in the event log appears to be broken on safari/ios.
Searching for "unknown argument" on safari/ios returns something different from other browsers (the first hit only contains the word "argument" and not the word "unknown"). Tested on opera/konqueror/chrome/firefox on linux.

The URL shown on the top of the screen appears to be the same as what is shown on other browsers, but one cannot be sure that it is really the same, as the browsers partially remove the encoding in the displayed URLs. Of course no other debugging is possible on ios.

Can people try on safari on macos?

vdbergh:
This fixes #1509.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -325,11 +325,39 @@ def sprt_calc(request):
     return {}
 
 
+# Different LOCALES may have different quotation marks.
+# See https://op.europa.eu/en/web/eu-vocabularies/formex/physical-specifications/character-encoding/quotation-marks
+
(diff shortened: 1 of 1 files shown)


vdbergh:
I checked on my ipad that the bug is fixed on DEV. The ipad now returns the same result as the browsers on my Linux system.

ppigazzini (approved):

ppigazzini:

PR #1511, opened 2022-12-30
Give some information about new runs in the event log.

vdbergh:
Examples
```
NumGames(40000)(PT) TC:30+0.3 Book:8moves_v3.pgn Threads:8(SMP)
SPRT<0.00,2.00> TC:10+0.1 Book:UHO_XXL_+0.90_+1.19.epd Threads:1
```
The format is terse, but the idea is that it should be easily searchable.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -23,7 +23,7 @@ schema = union(
         "task_id": int,
     },
     {"action": "system_event", "username": "fishtest.system", "message": str},
-    {"action": "new_run", "username": str, "run_id": ObjectId, "run": str},
+    {"action": "new_run", "username": str, "run_id": ObjectId, "run": str, "message": str},
     {"action": "upload_nn", "username": str, "nn": str},
     {
         "action": "modify_run",
@@ -160,12 +160,13 @@ class ActionDb:
             message=message,
         )
 
-    def new_run(self, username=None, run=None):
+    def new_run(self, username=None, run=None, message=None):
         self.insert_action(
             action="new_run",
             username=username,
             run_id=run["_id"],
             run=run_name(run),
+            message=message,
         )
 
     def upload_nn(self, username=None, nn=None):
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -91,6 +91,8 @@ class RunDb:
         self.task_runs = []
 
         self.task_duration = 900  # 15 minutes
+        self.ltc_lower_bound =40 # Beware: this is used as a filter in an index!
+        self.pt_info = {'pt_version' : 'SF_15', 'pt_branch' : "e6e324eb28fd49c1fc44b3b65784f85a773ec61c", 'pt_bench' : 8129754}
 
         global last_rundb
         last_rundb = self
@@ -538,7 +540,7 @@ class RunDb:
         if username:
             q["args.username"] = username
         if ltc_only:
-            q["tc_base"] = {"$gte": 40}
+            q["tc_base"] = {"$gte": self.ltc_lower_bound}
         if success_only:
             q["is_green"] = True
         if yellow_only:
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -8,9 +8,9 @@
   base_branch = args.get('base_tag', 'master')
   latest_bench = args.get('base_signature', master_info["bench"])
 
-  pt_version = "SF 15"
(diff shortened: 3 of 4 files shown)


vdbergh:
BTW. I feel that "If you need, you can always contact us Discord " is not grammatically correct and also a bit heavy. To me "If needed you may contact us on Discord" sounds better. Perhaps native English speakers can comment?

vdbergh:
Needless to say that I like the new page. It's a great improvement over the lamentations by waitress.

ppigazzini:
What about keeping simple with "You may contact us on Discord" ?

ppigazzini:
DEV updated.

vdbergh:
> What about keeping it simple with "You may contact us on Discord" ?

That's also good. But then the "otherwise" that follows is a bit strange me.  What about 

We'll be back online soon but in the meantime you may contact us on Discord ...

ppigazzini:
> We'll be back online soon but in the meantime you may contact us on Discord ...

Approved!

We could commit the page in fishtest repo. The trick is done by the nginx maintenance config (btw I never searched for a better solution).

```
server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2 ipv6only=on;
    ssl_certificate /etc/letsencrypt/live/dfts-0.pigazzini.it/fullchain.pem;
    ssl_certificate_key /etc/letsencrypt/live/dfts-0.pigazzini.it/privkey.pem;
    include /etc/letsencrypt/options-ssl-nginx.conf;
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;

    add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;

    server_name dfts-0.pigazzini.it;

    location ~ ^/(img|favicon.ico|robots.txt) {
        root /home/usr00/fishtest/server/fishtest/static;
        expires     1y;
        add_header  Cache-Control public;
        access_log  off;
    }

    location / {
        root /var/www/dfts-0.pigazzini.it/public_html;
        add_header  Cache-Control public;
        access_log  off;
        try_files /maintenance.html =404;
    }
}

server {
    listen 80;
    listen [::]:80;

    server_name dfts-0.pigazzini.it;

    location / {
        return 301 https://$host$request_uri;
    }
}
```

vdbergh:
> 
> We could commit the page in fishtest repo. The trick is done by the nginx maintenance config (btw I never searched for a better solution).
>

Yes why not. I guess that page is served by nginx itself. I would expect a webserver to have custom error pages. But perhaps not...

vdbergh:
Here is something about custom error pages in nginx:

https://www.tecmint.com/create-custom-nginx-error-page/

vdbergh:
I think the count down is visually nice but a bit misleading. IMHO it should start counting from a realistic ETA value, and if there isn't any then the count should go up. I do not know if there is a way to template or otherwise parametrize the page.

vdbergh:
BTW I tested the current PR, so if you want to experiment again with the error page, then please go ahead!

ppigazzini:
I added too early the confetti but people thought that they were for the passed new net, so now they expect fireworks for tomorrow...

ppigazzini:
DEV updated.

ppigazzini:
> `nginx` would have to react to `502 Bad Gateway` I would think.

IMO the maintenance page shouldn't be displayed for every `502 Bad Gateway`, otherwise if fishtest should die for whatever reason, people will not report the problem on Discord.

dav1312:
Here are some custom error pages for nginx https://github.com/alexphelps/server-error-pages/tree/gh-pages/_site
* [403](https://alexphelps.github.io/server-error-pages/403.html)
* [404](https://alexphelps.github.io/server-error-pages/404.html)
* [500](https://alexphelps.github.io/server-error-pages/500.html)
* [502](https://alexphelps.github.io/server-error-pages/502.html)
* [504](https://alexphelps.github.io/server-error-pages/504.html)
* [Maintenance](https://alexphelps.github.io/server-error-pages/maintenance.html)

ppigazzini:
https://serverfault.com/questions/310819/maintenance-page-on-nginx-best-practices

In effect the best practices for the maintenance require always an action (eg create a file/link for the maintenance page, reload a nginx config, etc).

ppigazzini:
DEV updated.

vdbergh:
I am happy with the latest push https://dfts-0.pigazzini.it/actions?max_actions=4&action=&user=&text=&before=1672424226.7763801857 .

ppigazzini (approved):

ppigazzini:

PR #1515, opened 2022-12-31
Index on "finished" + "username".

vdbergh:


ppigazzini (approved):
LGTM :)

https://dfts-0.pigazzini.it/tests/user/snicolet?page=7
https://dfts-0.pigazzini.it/tests/user/Vizvezdenec?page=5

vdbergh:
Indexes are magic!

vdbergh:
https://dfts-0.pigazzini.it/tests/user/Vizvezdenec?page=773 is still slow...

vdbergh:
I guess the issue is that skip/limit is just slow on mongod. Even if the cursor comes from an index (covered query).

ppigazzini:
LGTM :)

https://dfts-0.pigazzini.it/tests/user/snicolet?page=7
https://dfts-0.pigazzini.it/tests/user/Vizvezdenec?page=5


> I guess the issue is that skip/limit is just slow on mongod. Even if the cursor comes from an index (covered query).

https://medium.com/swlh/mongodb-pagination-fast-consistent-ece2a97070f3

vdbergh:
The issue is not so much the paging (which can be sped up when paging sequentially) but the fact that we want random access (`?page=xxx`). 

Still it is a bit of a mystery why the paging of the main page is fast (even going to the last page). Perhaps this is because the cursor is the full index. It is conceivable that mongo has a special case solution for this.

vdbergh:
If there is indeed such a special case then a dirty solution would be to use extra indexes for the few prolific users (there aren't that many).

ppigazzini:

(not easy to follow the discussion when playing with the PROD db)

peregrineshahin:
Very substantial PR. Thank you vdbergh!

PR #1516, opened 2023-01-01
Adapt links on the events log to the run_id parameter.

vdbergh:
The pagination was not tested since on my local fishtest I don't have events associated with runs spanning multiple pages.

### DIFF
--- a/server/fishtest/templates/actions.mak
+++ b/server/fishtest/templates/actions.mak
@@ -110,7 +110,7 @@
       % for action in actions:
           <tr>
             ## Dates in mongodb have millisecond precision. So they fit comfortably in a float without precision loss.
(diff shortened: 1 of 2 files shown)


vdbergh:
This one has two pages and seems to work correctly

https://dfts-0.pigazzini.it/actions?page=1&run_id=63abfb7f39af998100ce2ab1

server/fishtest/templates/actions.mak
```diff
@@ -1,4 +1,4 @@
-<%inherit file="base.mak"/>
+-<%inherit file="base.mak"/>
```
ppigazzini, line 1: I wonder about this change.

ppigazzini: Confirmed typo

image

vdbergh: Fixed!

ppigazzini (approved):

ppigazzini:

PR #1518, opened 2023-01-02
Create an api for querying the event log.

vdbergh:
A valid request consists of a POST request with a body that is a stringified version of the json query.

The response is limited to 200 events.

To test the api there is an example document

<server>/html/finished_runs.html

This will be removed later.

### DIFF
--- a/server/fishtest/static/html/finished_runs.html
+++ b/server/fishtest/static/html/finished_runs.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head> </head>
+  <body>
+    <script>
+      // Example script to demonstrate the api.
+      // Not used by Fishtest.
+
+      URL = "/api/actions";
+
+      payload = { action: "finished_run" };
+
+      function async_sleep(ms) {
+        return new Promise((resolve, reject) => {
+          setTimeout(resolve, ms);
+        });
+      }
+
+      async function doit() {
+        while (true) {
+          try {
+            response = await fetch(URL, {
+              method: "POST",
+              mode: "cors",
+              cache: "no-cache",
+              headers: {
+                "Content-Type": "application/json",
+              },
+              body: JSON.stringify(payload),
+            });
+          } catch (e) {
+            console.log(e);
+            return;
+          }
+          if (response.ok) {
+            json = await response.json();
+          } else {
+            console.log("response.status=" + response.status);
+            return;
+          }
+          document.body.innerText = JSON.stringify(json);
+          await async_sleep(10000);
+        }
+      }
+
+      doit();
+    </script>
+  </body>
+</html>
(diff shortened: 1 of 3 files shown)


ppigazzini:
Updated.
https://dfts-0.pigazzini.it/html/finished_runs.html

server/fishtest/static/html/finished_runs.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html>
```
ppigazzini, line 2: <html lang="en"> :D

vdbergh:
It seems the database on DEV is not recent enough. There aren't any finished runs... We need to create some.

ppigazzini:
I'm loading the latest DB backup.

server/fishtest/static/html/finished_runs.html
vdbergh: Fixed.

ppigazzini:
Done!

vdbergh:
It works now... Perhaps we can leave the page html/finished_runs.html for a while as an example for people that want to play with the api? What I want to do is to generate notifications for finished runs.

ppigazzini:
It's fine for me, perhaps better to add a comment in the html page.
Not easy after some time to clean up the old scripts, old dbs, old collections,

vdbergh:
I made the test page auto refreshing.

vdbergh:
Comment added in test file.

server/fishtest/static/html/finished_runs.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head> </head>
+  <body>
+    <script>
+      // Example script to demonstrate the api.
```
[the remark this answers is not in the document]
ppigazzini: No, sorry, it's in the script tag

ppigazzini:
DEV updated.

ppigazzini:
The 1000 Elo is an upper bound, I suppose.

ppigazzini:
Auto refresh works.

vdbergh:
> The 1000 Elo is an upper bound, I suppose.

Yes. The SPRT formulas don't work for extremely big Elo differences.

ppigazzini (approved):

ppigazzini:
PROD updated, posted a message on Discord, thank you @vdbergh :)

vdbergh:
Bummer. In retrospect it was not a good idea to store binary data in the actiondb (e.g. the time field). A generic remote query on the value of the time field is now impossible since json does not support the date format. 

I guess another db conversion would be best sometime.

ppigazzini:
We have a beautiful maintenance page with a timer :)

dav1312:
> We have a beautiful maintenance page with a timer :)

We can now also see it here: https://tests.stockfishchess.org/html/maintenance.html

ppigazzini:
I added the html location to the nginx config for the example page of this PR. Before it was not reachable.

PR #1519, opened 2023-01-02
Dbconversion to remove the opaque date and run_id types in actiondb.

vdbergh:
The opaque types are a nuissance since they cannot be encoded as json. So it becomes difficult to do remote queries.

As usual the conversion program is

utils/convert_actions.py

PR #1520, opened 2023-01-02
Notifications (WIP)

vdbergh:
This is a proof of concept implementation of a notification system, that generalizes the one that is currently
implemented in fishtest.

If the user agrees to receive notifications then they will be informed about every finished test.

The notifications are not really tested since one cannot enable notifications over a http protocol (my local fishtest). But the PR prints something reasonable on the console.

Note that this PR includes and depends on #1519 .


ppigazzini:
DEV updated with this one and the required PR.

vdbergh:
I saw some notifications appearing on Chrome. Just having the main page open.

ppigazzini:
Working with Chrome, Edge, Opera and Firefox.

vdbergh:
Yes it seems to work.

ppigazzini:
I suspect that duplicating few times the home page triggers past notification.

vdbergh:
It should not normally.

vdbergh:
The PR keeps track of the notifications it has shown in localStorage. Of course javascript has no locking so there is a possibility for a race condition, but I think the probability is low.

vdbergh:
This PR has a flaw that events that happen when the Fishtest app is not active (e.g. when a laptop is closed) are forgotten. It is possible to fix this but then the user would be bombarded with notifications when the Fishtest app becomes active.

I think it might be better to have an opt in system where the user clicks on a button on the live_elo page to register for notifications.

vdbergh:
I  made this draft. However the parent PR #1519 should be fine.

PR #1521, opened 2023-01-03
Opt-in system for notifications.

vdbergh:
The live_elo page now has a button where one can subscribe to notifications.

Some information is saved in local storage. Since Fishtest may have been opened in several tabs, this may lead to race conditions. We try to mitigate this by avoiding slow operations in critical sections.

Note that the current version prints a few messages on the console for debugging.

As usual the notifications themselves are not tested.

Note: it is currently not possible subscribe to notifications for fixed length tests. To make this possible one should extend the live elo page to handle NumGames tests. Or put the subscription button somewhere else (e.g. on the main page).


ppigazzini:
Problem on Edge:

image

ppigazzini:
Opera:

image

ppigazzini:
Firefox

image

ppigazzini:
Chrome:

image

vdbergh:
I saw a notification on chrome.

However while trying to disable notifications in the settings (to test the dialog for allowing notifications) I saw an error appear in the console. Not sure where it came from.

vdbergh:
@dav1312 

> It doesn't make sense that u get different results on opera, edge and chrome since they are all basically the same

I am not sure I understand. What is the difference you are referring to?

ppigazzini:
Notifications work on Chrome, Opera, Firefox, tested with live_elo opened and with home page opened.

ppigazzini:
master shows those warning on Edge as well

image

dav1312:
> @dav1312
> 
> > It doesn't make sense that u get different results on opera, edge and chrome since they are all basically the same
> 
> I am not sure I understand. What is the difference you are referring to?

Idk he seemed to get an error on edge but not in the others, maybe I just misread

ppigazzini:
I started testing Edge on a VM, perhaps is ublocker that breaks the script.

vdbergh:
> master shows those warning on Edge as well
> 
> image

Are you in a private browser window perhaps?

vdbergh:
image

This is the error I sometimes get when disabling notifications on chrome.

vdbergh:
Ok I see. The previous dialog no longer exists since one now has to push a button anyway to opt in.

ppigazzini:
DEV updated.

vdbergh:
> master with Edge (not InPrivate) in a clean VM
> 
> image

It seems these errors are benign since the gauges are moving. It seems google charts tries to access some kind of storage on the browser and Edge is disallowing this (because of CORS maybe?).

ppigazzini:
I started the 4 browser clearing the permission for the notification, then I clicked "Follow". Only Firefox asked for the permission to send the notification. Waiting the finish of the test.

vdbergh:
In my case Chrome usually asks for permission to send notifications. But clicking allow does not always seem to have an immediate effect.

ppigazzini:
> In my case Chrome usually asks for permission to send notifications. But clicking allow does not always seem to have an immediate effect.

After a change of the permission the browser ask for a refresh..

ppigazzini:
!!!

image

vdbergh:
This is the code executed as a response to an click event (borrowed from @dav1312 )
```js
      if (supportsNotifications() && Notification.permission === "default") {
        Notification.requestPermission();
      }
```
This is the code for `supportsNotifications()`: (also from dav123)
```js
function supportsNotifications() {
  if (
    // Notifications only work over a secure connection                                                                                                                                
    window.location.protocol == "https:" &&
    // Safari on iOS doesn't support them                                                                                                                                              
    "Notification" in window &&
    // Chrome and Opera on Android don't support them                                                                                                                                  
    !(
      navigator.userAgent.match(/Android/i) &&
      navigator.userAgent.match(/Chrome/i)
    )
  ) {
    return true;
  }
  return false;
}
```
It's not clear to me if there could be a problem.

ppigazzini:
If I understood correctly if the browser has the system notification blocked it shows a browser notification.

vdbergh:
@ppigazzini  Yes if you block notifications then the browser will not ask. If you still insist on following the test the PR will use a bootstrap alert to display notifications.

ppigazzini:
Resetting to "Ask" all browser asked for the permission.

ppigazzini:
Passed! The firefox notification seems disappear after few seconds.

image

dav1312:
> Passed! The firefox notification seems disappear after few seconds.

Firefox doesn't support `requireInteraction`

ppigazzini:
Tested working with Edge InPrivate with uBlock.

vdbergh:
I fixed a bug (the LRU object losing its capacity field). This seems to have had no consequences, but it is not nice.

ppigazzini:
DEV updated, new round of tests.

vdbergh:
Things work on ipad. The notification is displayed in a bootstrap alert.

ppigazzini:
It seems that I don't have the "Follow" button on Chrome...

vdbergh:
@dav1312 Do you know what would be a good way to make sure that the bootstrap alert div does not scroll under the main (it should be floating).

vdbergh:
> It seems that I don't have the "Follow" button on Chrome...

I see them. Does the button appear if you reload the page?

vdbergh:
If there is a network error during fetching the finished runs, then the follow button will not show.

ppigazzini:
> > It seems that I don't have the "Follow" button on Chrome...
> 
> I see them. Does the button appear if you reload the page?

Yes, in another browser the button appeared after a reload.

ppigazzini:
> Things work on ipad. The notification is displayed in a bootstrap alert.

Bootstrap alert in Android/Chrome as weel.

dav1312:
> @dav1312 Do you know what would be a good way to make sure that the bootstrap alert div does not scroll under the main header (it should be floating).

A [modal](https://getbootstrap.com/docs/5.2/components/modal/#live-demo)?

vdbergh:
> > @dav1312 Do you know what would be a good way to make sure that the bootstrap alert div does not scroll under the main header (it should be floating).
> 
> A [modal](https://getbootstrap.com/docs/5.2/components/modal/#live-demo)?

Hmm I don't think notifications should be modal.... They should be as non-intrusive as possible. 

I think I will leave this to another PR.

ppigazzini (approved):
Let's merge!

vdbergh:
> > Things work on ipad. The notification is displayed in a bootstrap alert.
> 
> Bootstrap alert in Android/Chrome as well. On Android/Edge the alert appeared after a reload of the home page.

It may have been hidden under the main header. Ideally the alert should stay in place like the main header stays in place when you scroll. But I do not know how to do this.

vdbergh:
> Let's merge!

I will first remove the instrumentation.

vdbergh:
Oops its already merged. I'll make a separate commit.

ppigazzini:
PROD update, sorry @vdbergh :D

vdbergh:
Done.

dav1312:
> > > @dav1312 Do you know what would be a good way to make sure that the bootstrap alert div does not scroll under the main header (it should be floating).
> > 
> > 
> > A [modal](https://getbootstrap.com/docs/5.2/components/modal/#live-demo)?
> 
> Hmm I don't think notifications should be modal.... They should be as non-intrusive as possible.
> 
> I think I will leave this to another PR.

You can also use `position-absolute` but be careful because darkmode modals have slighly transparent backgrounds

vdbergh:
> > > > @dav1312 Do you know what would be a good way to make sure that the bootstrap alert div does not scroll under the main header (it should be floating).
> > > 
> > > 
> > > A [modal](https://getbootstrap.com/docs/5.2/components/modal/#live-demo)?
> > 
> > 
> > Hmm I don't think notifications should be modal.... They should be as non-intrusive as possible.
> > I think I will leave this to another PR.
> 
> You can also use `position-absolute` but be careful because darkmode alerts have slighly transparent backgrounds

Thanks! I'll have a look.

vdbergh:
Actually the instrumentation isn't so bad. It is useful for debugging.

PR #1522, opened 2023-01-04
Make the bootstrap notification alert fixed-top.

vdbergh:
In that way it will always be visible.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -612,3 +612,7 @@ input[type="checkbox"],
   pointer-events: none;
   opacity: 0.5;
 }
+
+/* remove transparency of background */
+.alert-success-non-transparent {
+}
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -452,3 +452,8 @@ pre[style*="border-color:Pink"] {
   --bs-modal-footer-border-color: var(--bs-modal-border-color);
   --bs-modal-header-border-color: var(--bs-modal-border-color);
 }
+
+/* remove transparency of background */
+.alert-success-non-transparent {
+    --bs-alert-bg: rgba(32,51,32,1);
+}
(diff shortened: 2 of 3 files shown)


vdbergh:
As @dav1312 pointed out this does not work in dark mode as alerts are semi-transparent. I have not figured out yet what bootstrap incantations are necessary to fix this so making this PR draft for now.

vdbergh:
Fixed I think.

vdbergh:
> I'm currently testing on DEV this https://getbootstrap.com/docs/5.2/utilities/opacity/

I think this does not work because opacity is cumulative. But let's see.

ppigazzini:
Opacity didn't worked, updating DEV with your last commit.

ppigazzini (approved):

server/fishtest/static/css/application.css
```diff
@@ -612,3 +612,7 @@ input[type="checkbox"],
   pointer-events: none;
   opacity: 0.5;
 }
+
+/* remove transparency of background */
+.alert-success-non-transparent {
+}
```
dav1312, line 618: Empty rule?

ppigazzini:
PROD update, sorry @dav1312 :(

dav1312:
It's fine, it doesn't change anything so

server/fishtest/static/css/application.css
ppigazzini: If I understood correctly we can drop this rule. Let me know that I will push directly a commit.

ppigazzini: @dav1312 BTW fell free to open a PR to clean up the rule yourself :)

vdbergh:
@dav1312  I wasn't sure if it is legal to refer to a class that is not defined.

ppigazzini:
BTW I'm testing what happens with multiple bootstrap notifications.

vdbergh:
Currently they will overwrite each other. I will make a PR where they will append instead.

vdbergh:
It seems it is legal to refer to a class that has no associated style. That would mean the empty rule can go.

PR #1523, opened 2023-01-04
Omit the follow button in mako template for finished tests.

vdbergh:
Otherwise stopped tests will have a follow button which makes no sense.


ppigazzini (approved):
LGTM

ppigazzini:

PR #1524, opened 2023-01-04
Multiple fallback notifications no longer overwrite.

vdbergh:
To test this one should block notifications, follow some tests, and wait for them to finish. They will all appear in the same alert dialog.


ppigazzini:
DEV updated, pc and android browsers following the tests after blocking notifications.

ppigazzini:
LGTM, worked fine an Android (Edge, Chrome) as well.

image

ppigazzini (approved):

ppigazzini:

PR #1525, opened 2023-01-04
Move the follow button to the main test page.

vdbergh:
In that way it applies to all tests.


ppigazzini (approved):
SPRT and NumGames

image

PR #1527, opened 2023-01-05
Make it possible to (un)follow a run on the main page via an icon.

vdbergh:


vdbergh:
Perhaps it should be an extra column in the table. Now it is looks rather zigzaggy.

ppigazzini:
Yes, and I think that the icon should be centered to have a static eye when clicking.
Perhaps @dav1312 can help with the icon :)

vdbergh:
Tried to make it a separate column but the browser insists on putting a huge amount of space on the right.

ppigazzini:
The Delete button has a proper space, though.

image

vdbergh:
Fixed in the latest commit.

ppigazzini:
The layout is ok, the question now is "it really works"?

vdbergh:
Got the puzzle of the day wrong btw. That cost me 40 points in my puzzle rating (it's an 1800 puzzle).

vdbergh:
> The layout is ok, the question now is "it really works"?

It should... The main logic wasn't touched.

vdbergh:
The basic architecture is simple. The browser keeps a list or run_ids it should follow (it is stored in local storage so it is persistent). There is a main loop which checks every 20 seconds if one of the run_ids that are being followed has had a finished_run event. In that case a notification is shown.

The main functions are
```js
follow_run(run_id);
unfollow_run(run_id);
following_run(run_id);
```

Currently at most 15 runs can be followed (if you select more then some runs will be quietly unfollowed).

vdbergh:
I wonder if the logic should be inverted (like the sun and the moon for the theme selection). Clicking on the eye should then follow a run.

ppigazzini:
It disrupt the portrait layout on my phone.

mobile_portrait

ppigazzini:
> I wonder if the logic should be inverted (like the sun and the moon for the theme selection). Clicking on the eye should then follow a run.

I switched sun/moon in https://github.com/glinscott/fishtest/commit/172f6b36c0419ef11617ea9807de22e0930f9dfe after reading these:

> The current status is obvious for light/dark modes.
> Use the sun/moon icons for the action (switch to the alternative mode).
> https://ux.stackexchange.com/questions/39418/night-day-mode-toggle
> https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan?rq=1

vdbergh:
I am looking at the iphone. It seems that the browser want to put the eye in a separate column. I may need some help on this @dav1312

ppigazzini:
Excerpt from "About Face: The Essentials of Interaction Design":

> Flip-flop button controls are very efficient. They save space by controlling two mutually exclusive options with a single control. The problem with flip-flop controls is that they fail to fulfill the second duty of every control - to inform the user of their current state. If the button says ON when the state is off, it is unclear what the setting is. If it is OFF when the state is off, however, where is the ON button? Don't use them. Not on buttons and no on menus!

Here the browser (on Discord are suggesting to use a Bell)

image

ppigazzini:
BTW the logic is still working fine.

image

vdbergh:
I added a few css properties that are also enjoyed by the other elements of the table. Perhaps this will fix the iphone issue.

ppigazzini:
> I added a few css properties that are also enjoyed by the other elements of the table. Perhaps this will fix the iphone issue.

Fixed!

mobile_portrait_fixed

ppigazzini:
> I guess I will leave the flip-flop redesign for another PR but I will switch the logic to be consistent with the sun/moon. I will also add a tooltip which will hopefully make it clearer what the effect is of switching the state.

Ok, let me know when this PR is ready to be merged.

ppigazzini:
I replaced everywhere "eye" with "bell" on DEV.

vdbergh:
The bell looks nice. I will change it but having dinner now. The latest push switches the logic and also adds tool tips.

vdbergh:
Converted to bell. I am happy with the PR.

vdbergh:
I wonder if fontawesome has the little switches that Google uses. Somehow these seem unambiguous (don't know exactly why).

ppigazzini:
> Converted to bell. I am happy with the PR.

What about replacing "eye" with "notification" or "notif" ? In that way we need only to replace the fontawesome name when changing icon.

> I wonder if fontawesome has the little switches that Google uses. Somehow these seem unambiguous (don't know exactly why).

Searching...

https://fontawesome.com/search

vdbergh:
Plenty of switches https://fontawesome.com/search?q=switches&o=r .

ppigazzini:
And toggles

https://fontawesome.com/search?q=toggle&o=r

ppigazzini:
> Plenty of switches https://fontawesome.com/search?q=switches&o=r .

There is a trap, push "Free" as well :)

https://fontawesome.com/search?q=switch&o=r&m=free
https://fontawesome.com/search?q=toggle&o=r&m=free
https://fontawesome.com/search?q=bell&o=r&m=free

vdbergh:
I added a commit where we use a bell plus a toggle.

vdbergh:
It's possible to make to column less wide.

vdbergh:
Maybe a bell with the original logic is the best. It is not consistent with the sun/moon but the tooltip should make clear what we mean. What do you think?

ppigazzini:
I vote for bell and tooltip.

vdbergh:
Ok. Then have a look at the latest push which a less wide column (and perhaps check on iphone).

ppigazzini:
> Perhaps https://fontawesome.com/icons/toggle-on?s=solid&f=classic and https://fontawesome.com/icons/toggle-off?s=solid&f=classic

We should be sure to use the free version, hoping than @dav1312 can help :)

ppigazzini:
DEV updated.

vdbergh:
Squashed.

vdbergh:
> > Converted to bell. I am happy with the PR.
> 
> What about replacing "eye" with "notification" or "notif" ? In that way we need only to replace the fontawesome name when changing icon.
> 

I did not see this. It's now a bit late for that since it will introduce new errors.

ppigazzini:
On DEV I switched the positions of bell and toggle hoping to fix the position of toggle, nope.

vdbergh:
> On DEV I switched the positions of bell and toggle hoping to fix the position of toggle, nope.

In LaTeX one would create a phantom box with the dimensions of the crossed out bell and put the usual bell in it. Not sure if this can be done in html.

vdbergh:
I guess one could add a bit of padding to the bell to make it the same size of the crossed out bell.

dav1312:
I thought you guys would opt to use something like what I mentioned in Discord a couple of days ago. Basically the same as the delete button but with the bell.
image

ppigazzini:
Sorry, I was very busy with the ~lichess puzzles~ PRs tests in the last days, I missed your suggestion on Discord.

ppigazzini:
> I thought you guys would opt to use something like what I mentioned in Discord a couple of days ago. Basically the same as the delete button but with the bell. image

How to insert that button for "Paused Tests"?

image

vdbergh:
I did not succeed in applying padding directly to an `<i></i>` element.

dav1312:
> Where to insert that button for "Paused Tests"?
> 
> image

But the button is already there right? The div just needs to have a `btn btn-sm btn-secondary` class(?)

dav1312:
> But the button is already there right? The div just needs to have a `btn btn-sm btn-secondary` class(?)

So just move the button to after the delete button
Change the icon to `fa-solid fa-bell` and `fa-solid fa-bell-slash`
Add `btn btn-sm btn-secondary` to the div
And everything should work?

image

vdbergh:
@dav1312 Yes the icon can be moved. But @ppigazzini pointed out that it is not clear if the bell represents the current state (following) or the action (hence currently not following). That's why we were experimenting with an icon pair (following google).

vdbergh:
@ppigazzini Can you update again?

ppigazzini:
> @ppigazzini Can you update again?

Ops, I missed the new push, updating...

dav1312:
> it is not clear if the bell represents the current state

Oh okay, that's what I was missing... The "title" should already clear it up in case someone doesn't know it but

What about "regular" and "solid" bells instead of the switch?
image

vdbergh:
@dav1312 Do you think this solves the issue? For some reason a switch seems to solve the issue although I do not really understand why.

vdbergh:
@ppigazzini I think the situation with the double icon is improved although still not perfect (and a rendering bug at large magnification was fixed). I am going the squash the PR again.

dav1312:
> @dav1312 Do you think this solves the issue? For some reason a switch seems to solve the issue although I do not really understand why.

I personally don't like the double icon thing. Specially since it doesn't have a background right now, it makes it look like 2 separate things.
For example, youtube has this:

image

ppigazzini:
> @dav1312 Do you think this solves the issue? For some reason a switch seems to solve the issue although I do not really understand why.

On the browser you can click only on the toggle, so the toggle makes the action and the bell reports the status. In our case we can click on either icons, so that not really solve the indetermination.

vdbergh:
> > @dav1312 Do you think this solves the issue? For some reason a switch seems to solve the issue although I do not really understand why.
> 
> On the browser you can click only on the toggle, so the toggle makes the action and the bell reports the status. In our case we can click on either icons, so that not really solve the indetermination.

To me it is clear that the on/off switch represents the turning on and off of the feature. The fact that you can also click on a somewhat large area to get the same effect seems irrelevant to me... 

Of course it is possible to make it so that the bell itself is irresponsive but it would mean changing the javascript code. Since there is currently no real agreement what the best solution is I propose to leave it so.

vdbergh:
> > @dav1312 Do you think this solves the issue? For some reason a switch seems to solve the issue although I do not really understand why.
> 
> I personally don't like the double icon thing. Specially since it doesn't have a background right now, it makes it look like 2 separate things. For example, youtube has this:
> 
> image

The key thing is that there are words written next to the icons. That removes any ambiguity.

dav1312:
Having the `disabled` class too to make it dimmer

image

vdbergh:
@dav1312 From your picture it is still not clear to me if the icons represent the state or the action. Of course as you point out the tool tip will clear this up. So I really don't feel strongly about this icon issue.

Since there seems to be no real agreement on what the best solution is, my proposal is to keep the current PR as it is to get some feedback from users.

vdbergh:
> I will merge the PR to introduce the new feature. We can improve the layout with new PRs.
> 
> @vdbergh
> 
> > > What about replacing "eye" with "notification" or "notif" ? In that way we need only to replace the fontawesome name when changing icon.
> > 
> > 
> > I did not see this. It's now a bit late for that since it will introduce new errors.s.
> 
> I can fix this in a new PR, I have only to replace all the "eye" occurrences, right? I already done a couple of hours ago on the server and I didn't viewed any error.

Yes. Actually this is good since "eye" is rather a short word. So there would be more risk with collisions in the future.

ppigazzini (approved):

ppigazzini:

PR #1530, opened 2023-01-06
Nonmoving double icon even at large magnification.

vdbergh:


ppigazzini (approved):
LGTM: DEV tested with the 4 browsers from 25% to 500% scaling

ppigazzini:

PR #1531, opened 2023-01-06
Some enhancements

vdbergh:
The PR has two independent commits

- Display the bootstrap notifications in all fishtest browser windows (currently a bootstrap notification is displayed only once, like a system notifications).

- Display the number of received (and non closed) notifications in the title of the tab.

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -47,26 +47,49 @@ LRU.load = function (name) {
 };
 
 const fishtest_notifications_key = "fishtest_notifications";
+const fishtest_broadcast_key = "fishtest_broadcast";
 
-let current_nofitication = "";
+function notify_fishtest_(message) {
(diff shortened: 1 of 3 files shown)


vdbergh:
It seems to work. I noticed as small bug in the string handling which has not consequence for the titles used in Fishtest. I will fix it in a separate PR.

ppigazzini:
I think to have found a strange behavior.
If the home page is not reloaded after the tests are finished, it seems that you can trigger a new notification for the already finished test clicking on the icon.

vdbergh:
> I think to have found a strange behavior. If the home page is not reloaded after the tests are finished, it seems that you can trigger a new notification for the already finished test clicking on the icon.

Yes that is correct. Good catch! Since it is an unrelated problem I will fix it in a separate PR.

ppigazzini (approved):

ppigazzini:

PR #1532, opened 2023-01-06
Make the notification toggle disappear after a finished test.

vdbergh:
Also fix a small bug in the last PR.

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -55,7 +55,9 @@ async function process_title(count) {
   title = document.title;
   // check if there is already a number
   if (/\([\d]+\)/.test(title)) {
-    title = title.split(") ")[1]; // if so, remove it
+    // if so, remove it
+    let idx = title.indexOf(" ");
+    title = title.slice(idx + 1);
   }
   // add it again if necessary
   if (count > 0) {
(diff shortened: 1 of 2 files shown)


vdbergh:
The toggle will disappear immediately if the user was following the test (and hence was notified). If the user is not following then after enabling the toggle he will get a final notification and then the toggle will disappear as well.

vdbergh:
Yes I see how this can happen. It is due to the interaction between the browser tabs (they realize that a run is finished at different times). I need to think about a better solution.

vdbergh:
I hope it is fixed.

ppigazzini:
Tests started.

ppigazzini:
Clicking very fast on the unfollowed test icon I was able to get 2 notifications on Firefox before the icon grayed out. The other 3 browser were more smart, though.
On Android I got 2 notification by the SPSA, but the browser goes in sleep mode on mobile.
I opened a big number of tabs in any browser.

LGTM :)

image

ppigazzini (approved):

vdbergh:
Yes there are some race conditions. With fast clicking or a lot of tabs open they are more likely to be triggered I guess. An occasional double notification under stringent conditions is not the end of the world.

To get rid of the race conditions the communication between the tabs should be sanitized. But not now....

vdbergh:
If a tab is sleeping  while receiving a message then I guess strange things might be happening. I did not consider this possibility.

ppigazzini:
I will merge the PR as soon as in "Ready for review" mode.

vdbergh:
> I will merge the PR as soon as in "Ready for review" mode.

Ah forgotten that I had put it into draft mode. Undrafted.

ppigazzini:

PR #1533, opened 2023-01-07
Cleanup and refactoring of cross tab communication.

vdbergh:
No functional change.

With all the bug fixing the code had grown a lot of hair. Now it is much cleaner I think. 

Some generic stuff (handling the document title and broadcasts) has been taken out of `notifications.js` and put into `application.js` .

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -195,6 +195,24 @@ async function fetch_text(url, options) {
   return response.text();
 }
 
+async function process_title(count) {
+  // async because the document title is often set in javascript
+  await DOM_loaded();
+  let title = document.title;
+  // check if there is already a number
+  if (/\([\d]+\)/.test(title)) {
+    // if so, remove it
+    let idx = title.indexOf(" ");
+    title = title.slice(idx + 1);
+  }
+  // add it again if necessary
+  if (count > 0) {
+    document.title = `(${count}) ${title}`;
+  } else {
+    document.title = title;
+  }
+}
+
 function DOM_loaded() {
   // Use as
   // await DOM_loaded();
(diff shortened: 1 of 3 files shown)


server/fishtest/static/js/application.js
```diff
@@ -195,6 +195,24 @@ async function fetch_text(url, options) {
   return response.text();
 }
 
+async function process_title(count) {
+  // async because the document title is often set in javascript
+  await DOM_loaded();
+  title = document.title;
```
ppigazzini, line 201:
```js
let title = document.title;
```

server/fishtest/static/js/application.js
```diff
@@ -213,3 +231,25 @@ function async_sleep(ms) {
     setTimeout(resolve, ms);
   });
 }
+
+const fishtest_broadcast_key = "fishtest_broadcast";
+let broadcast_dispatch = {};
+
+window.addEventListener("storage", (event) => {
+  if (event.key == fishtest_broadcast_key) {
+    cmd = JSON.parse(event.newValue);
```
ppigazzini, line 240:
```js
const cmd = JSON.parse(event.newValue);
```

server/fishtest/static/js/application.js
```diff
@@ -213,3 +231,25 @@ function async_sleep(ms) {
...
+    cmd_function = broadcast_dispatch[cmd["cmd"]];
+    if (cmd_function) {
+      cmd_function(cmd["arg"]);
+    }
+  }
+});
+
+function broadcast(cmd, arg) {
+  cmd_function = broadcast_dispatch[cmd];
```
ppigazzini, line 249:
```js
const cmd_function = broadcast_dispatch[cmd];
```

server/fishtest/static/js/application.js
```diff
@@ -213,3 +231,25 @@ function async_sleep(ms) {
...
   });
 }
+
+const fishtest_broadcast_key = "fishtest_broadcast";
+let broadcast_dispatch = {};
+
+window.addEventListener("storage", (event) => {
+  if (event.key == fishtest_broadcast_key) {
+    cmd = JSON.parse(event.newValue);
+    cmd_function = broadcast_dispatch[cmd["cmd"]];
```
ppigazzini, line 241:
```js
const cmd_function = broadcast_dispatch[cmd["cmd"]];
```

ppigazzini (commented):
I'm a noob with js, so take my review with a grain of salt. Not tested.

ppigazzini:
OK, I'm testing the PR right now.

server/fishtest/static/js/application.js
vdbergh: All fixed. BTW it might been better to set the title in the html header
```
<title>
Title here ...
</title>
```
Then this function `process_title` wouldn't have to be async.

ppigazzini:
I already tested the fixes by hand.

image

ppigazzini (approved):

ppigazzini:

ppigazzini:
@vdbergh I just learned on Discord that the original notification PR was able to put the right test page on focus clicking on the notification, courtesy of this code:
https://github.com/glinscott/fishtest/blob/8c6d3d4efe0613edf831112d062392bf39b0c58c/server/fishtest/static/js/application.js#L155-L157

With the change of the framework that code should be dead now, the feature could be interesting if easy to implement (it should open a new tab or put in focus the tab if already opened).

vdbergh:
I see! I will post a quick fix.

ppigazzini:
Playing a bit with fastapi, impressive!
https://fastapi.tiangolo.com/tutorial/

Finally I took some time to understand the difference between asyncio and threading ;)

PR #1534, opened 2023-01-09
Make notifications clickable.

vdbergh:
Not tested with native notifications.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -145,7 +145,7 @@ function supportsNotifications() {
   return false;
 }
 
-function notify(title, body, fallback) {
+function notify(title, body, link, fallback) {
   if (supportsNotifications() && Notification.permission === "granted") {
     const notification = new Notification(title, {
       body: body,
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):

ppigazzini:
PROD updated thank you @vdbergh :)

vdbergh:
This PR is not quite right. We still need the focus call at onclick before loading the test page. And I guess it is best to load the test page in a new tab (`taget=_blank`), rather than overwriting a random page.

dav1312:
> @dav1312 Would that also give focus to the browser?

I tested it pasting this in the devtools inside Fishtest and it did so I guess it works
```js
function notify(title = "title", body = "body", link = "/tests/view/63bc1f35cd3db0c8d39a3016") {
  if (Notification.permission === "granted") {
    const notification = new Notification(title, {
      body: body,
      requireInteraction: true,
      icon: "https://tests.stockfishchess.org/img/stockfish.png",
    });
    notification.onclick = () => {
      window.open(link, "_blank");
    };
  }
}

setTimeout(notify, 5000);
```

vdbergh:
> > @dav1312 Would that also give focus to the browser?
> 
> I tested it pasting this in the devtools inside Fishtest and it did so I guess it works
> 
> ```js
> function notify(title = "title", body = "body", link = "/tests/view/63bc1f35cd3db0c8d39a3016") {
>   if (Notification.permission === "granted") {
>     const notification = new Notification(title, {
>       body: body,
>       requireInteraction: true,
>       icon: "https://tests.stockfishchess.org/img/stockfish.png",
>     });
>     notification.onclick = () => {
>       window.open(link, "_blank");
>     };
>   }
> }
> 
> setTimeout(notify, 5000);
> ```

That sounds like next level hacking to me :smile: But thanks for checking!

PR #1535, opened 2023-01-09
More cleanups and refactoring.

vdbergh:
Handle the follow/unfollow button in the same way as the clock+toggle icon.

Do the right thing when the user wants to follow > 15 runs.

Fix some typos.


ppigazzini:
Tested ok on DEV.

ppigazzini (approved):

PR #1537, opened 2023-01-09
Make a network error while verifying the worker version non fatal.

vdbergh:


vdbergh:
It's not really about the worker update but about the fact that currently the worker quits if the server goes down for too long while the worker is running (with this PR it will start the usual sleep/retry cycle).

vdbergh:
Concerning your first comment. I think we want the worker to quit if the user has been blocked after the worker started but not if there is a network error (in that case we want to sleep and retry). Your solution (which I assume refers to master) does not distinguish between these two cases.

Concerning your second comment: I think it is currently difficult to avoid that the worker quits at start up if the server is down since the main loop hasn't started yet. It would be a serious architectural change.

ppigazzini:
I had a little problem testing the PR because I missed the first call of `verify_worker_version()`, so the worker keeps exiting at the start skipping the instrumentation code.
I just wonder if we really want to exit at the first call of `verify_worker_version()` if the return value is `None`.

ppigazzini:
> Concerning your first comment. I think we want the worker to quit if the user has been blocked after the worker started but not if there is a network error (in that case we want to sleep and retry). Your solution (which I assume refers to master) does not distinguish between these two cases.

I didn't propose an alternative solution to this PR but a complement to this PR to treat the return value of the two calls of `verify_worker_version()` in the same way.

vdbergh:
I think that at the first call of `verify_worker_version` we could indeed allow the worker to proceed if the result is `None`.  But I think this would make little difference, since just before this we have checked the credentials of the worker. So if the result is None (of verify_worker_version) then the network was likely down during the checking of the credentials, in which case the worker would have quit already.

vdbergh:
But of course we can treat the two cases the same. It doesn't hurt. I will change it.

ppigazzini:
Thank you. I was surprised why the worker kept exiting when forcing the return `None`, and after having found the first call I wondered why the calls were treated in different ways.

ppigazzini (approved):

ppigazzini:

PR #1538, opened 2023-01-10
Introduce a lockless critical section.

vdbergh:
When the browser comes out of sleep then there is a race between the tabs which may lead to duplicate notifications. For some reason this only seems to happen on mobile devices.

This issue could be trivially fixed by not displaying duplicate notifications but this seems like a hack.

Here we implement a lockless critical section following an algorithm by Michael Fischer.

It seems impossible to test this without just trying....

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -253,3 +253,44 @@ function broadcast(cmd, arg) {
     JSON.stringify({ cmd: cmd, arg: arg, rnd: Math.random() })
   );
 }
+
+const tab_id = String(Math.random()).slice(2) + String(Math.random()).slice(2);
+
+async function critical_section(lock, critical_section_) {
+  // This is the simplest mutual exclusion algorithm not depending on
+  // locking primitives. It is due to Michael Fischer and is mentioned in
+  // Leslie Lamport's seminal paper "A Fast Mutual Exclusion Algorithm".
+  // Lamport regards the algorithm as inadequate since we cannot
+  // rigorously bound the delay necessary to execute (1)(2)(3) because
+  // there is no bound on memory contention (here corresponding to the
+  // setItem instructions). However for us this is not an issue -
+  // 30ms should be sufficient for plenty of tabs.
+  // Another issue is that taking the lock always takes at least 30ms
+  // (with this implementation) even if there is no contention. This is also
+  // not a problem since our application is not time critical.
+  while (true) {
+    let X;
+    // We wait for the lock to clear. After 1.5s we assume that the current
+    // holder of the lock has disappeared.
+    for (let i = 0; i <= 50; i++) {
+      X = localStorage.getItem(lock);
+      // (1)
+      if (X == null || X == "free") {
+        break; // (2)
+      }
+      async_sleep(30);
+    }
+    // Ok the lock is clear. Let's try to grab it. Of course competing tabs will now do the same.
+    localStorage.setItem(lock, tab_id); // (3)
+    async_sleep(30); // Should be long enough to do (1)(2)(3) (by other tabs)
+    X = localStorage.getItem(lock);
+    if (X == tab_id) {
+      // Hurray we won the race with the other tabs! Go to the critical section.
+      break;
+    }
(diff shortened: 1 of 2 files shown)


noobpwnftw:
Can it be solved by having unique notification IDs so that the OS/browser notifier does the duplication?

vdbergh:
Weblocks is marked as experimental technology only supported in the latest browsers. So while being very useful it is perhaps a bit early to use them. I think for the application here a simple lockless scheme is sufficient.

dav1312:
> Weblocks is marked as experimental technology only supported in the latest browsers.

I think that "experimental" notice is outdated, we probably use features with less support than this.
Supported in Chrome since 2018 and Firefox since 2021

https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API#browser_compatibility
https://caniuse.com/mdn-api_lock

vdbergh:
> > Weblocks is marked as experimental technology only supported in the latest browsers.
> 
> I think that "experimental" notice is outdated, we probably use features with less support than this. Supported in Chrome since 2018 and Firefox since 2021
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API#browser_compatibility https://caniuse.com/mdn-api_lock
> 
> The only problem that I see is probably Safari that only supports it since last year and we have some users with old versions

Safari would be an issue. There are plenty of people with old iphones.

vdbergh:
I added some comments explaining how the algorithm works.

server/fishtest/static/js/application.js
```diff
@@ -253,3 +253,43 @@ function broadcast(cmd, arg) {
...
+  // rigorously bound the delay necessary to execute (1)(2)(3) because
+  // there is no bound on memory contention (here corresponding to the
+  // setItem instructions). However for us this is not an issue -
+  // 30ms should be sufficient for plenty of tabs.
+  // Another issue is that taking the lock always takes at least 30ms
+  // (with this implementation) even if there is no contention. This is also
+  // not a problem since our application is not time critical.
+  while (true) {
+    let X;
+    // We wait for the lock to clear. Afer 1.5s we assume that the current
```
dav1312, line 273: After*

vdbergh: Fixed! Thanks.

vdbergh:
This is an example with triplicate notifications:

IMG-0238

vdbergh:
I think I can test this PR by using an instrumented version fishtest which periodically answers the `/api/actions` api-call
with some fake finished runs.

vdbergh:
I tried this and for now the verdict seems to be that it does not work at all, even extending the 30ms delay to 750ms (tested on safari an ipad). I don't understand this but it seems impossible to debug this. 

For now the best solution seems to be to simply suppress duplicate notifications. Someday I might do a test with weblocks. If it still does not work then there would be an issue with the code.

vdbergh:
https://www.javaer101.com/en/article/35579652.html

It seems that tabs do not see local storage updates at the same time (like memory behind a cache). This invalidates the algorithm. 

It appears on old browsers one can implement real locks via indexeddb (which has transaction support).  https://www.npmjs.com/package/idb-mutex .

vdbergh:
The only sensible solution seems to be to use weblocks when they are supported and ignore the issue otherwise (it only happens in rather extreme cases with many tabs open).

vdbergh:
So it seems weblocks also don’t solve the issue on Safari on iPad (weblocks are supported). I don’t understand what’s wrong.

vdbergh:
Discovered a stupid bug. async_sleep is an async function so one needs to write await async_sleep.

vdbergh:
But it still doesn't work. I conjecture that changes in local storage take some time to propagate to other tabs, especially if the OS is on the borderline of sleeping (this is on an ipad). So in addition to locks one would need memory barriers for local storage.

vdbergh:
I guess the correct way to handle this is via a service worker.

vdbergh:
But it seems they became only very recently supported by Chrome on Android whereas Safari on iOS had them for a long time:  https://caniuse.com/serviceworkers Strange. Usually Chrome is ahead of the pack.

PR #1539, opened 2023-01-11
Open test window and give focus when clicking a native notification.

vdbergh:
Suggested by dav1312.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -153,7 +153,7 @@ function notify(title, body, link, fallback) {
       icon: "https://tests.stockfishchess.org/img/stockfish.png",
     });
     notification.onclick = () => {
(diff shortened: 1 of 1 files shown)


vdbergh:
I managed to get fishtest working over ssl. So I could actually test this myself...

ppigazzini:
A click on system notification opens the test in a new tab of the correct browser.
A click on a bootstrap notification opens the test in the tab where we clicked.

vdbergh:
I tried to open the link in a bootstrap notification in a different tab but that did not work very well as this tab now also started showing notifications.

It might work if we could close individual notifications (and they would then close in all tabs). But we aren't there yet....

ppigazzini:
PROD updated with both PRs, thank you @vdbergh :)

PR #1540, opened 2023-01-13
Asynchronous network upload

ppigazzini:
To avoid duplicated netowrk entries in the DB was forced a synchronous upload, see: https://github.com/glinscott/fishtest/pull/757

An easy way to enable to avoid duplicate entries is to look up the network in the DB just before adding the entry.

This PR speed up the upload process and reduces possible race conditions checking for duplicate files on the file server as well, raising the appropriate HTTP errors if necessary.

Other changes:
 - use a fixed path for the POST (usually the {id} is used in PUT and DELETE)
 - use a multipart form POST to upload the network. In this way the file server can use an UploadFile object that uses in a smart way the RAM for big data

Requirements:
 - update of the file server
 - adding a "/" in the fishtest.upload

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -124,23 +124,8 @@ def login(request):
     return {}
 
 
-# Guard against upload timeouts/retries
-uploading = threading.Semaphore()
-
-
 @view_config(route_name="nn_upload", renderer="nn_upload.mak", require_csrf=True)
 def upload(request):
-    if not uploading.acquire(False):
-        request.session.flash(
-            "An other upload is in progress, please try again later", "error"
-        )
-        return {}
-    result = sync_upload(request)
-    uploading.release()
-    return result
-
-
-def sync_upload(request):
     userid = request.authenticated_userid
     if not userid:
         request.session.flash("Please login")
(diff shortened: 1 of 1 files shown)


ppigazzini:
Companion server
```
env/bin/python3 -m pip install fastapi uvcorn[standard] gunicorn python-multipart
env/bin/gunicorn net_server:app --workers 4 --worker-class uvicorn.workers.UvicornWorker
```

```python
import gzip
from pathlib import Path

from fastapi import FastAPI, HTTPException, UploadFile

app = FastAPI()


@app.post("/upload_net/", status_code=201)
async def create_upload_net(upload: UploadFile) -> None:
    net_file = upload.filename
    net_file_gz = Path.cwd() / "nn" / (net_file + ".gz")
    try:
        with gzip.open(net_file_gz, "xb") as f:
            f.write(await upload.read())
    except FileExistsError as e:
        print(e)
        raise HTTPException(
            status_code=409,
            detail=f"File {net_file} already uploaded",
        )
    except Exception as e:
        print(e)
        raise HTTPException(
            status_code=500,
            detail=f"Failed to write file {net_file}",
        )
```

ppigazzini:
Tested on DEV:
 - 8 concurrent uploads (different networks)
 - 4 concurrent uploads of the same networks, triggering or the 409 HTTP status on the file server or the duplicated name in the DB lookup

ppigazzini:
PROD, file server and DEV updated with the asynchronous net upload server.

PR #1544, opened 2023-01-16
Give more information in notifications.

vdbergh:
Also: include hash in new_run event message.


ppigazzini:
LGTM

image

vdbergh:
Yes I think it is ok. I tried inserting an explicit line break but that looked ugly if the screen width does not match the length of the lines. So now it is one long line...

ppigazzini (approved):

PR #1545, opened 2023-01-16
Use better estimation for remaining hours calculation.

MinetaS:
This patch rewrites estimation algorithm to calculate remaining hours of SPRT type tests.

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -1,4 +1,5 @@
 import hashlib
+import math
 import re
 import smtplib
 from datetime import datetime, timedelta
@@ -345,10 +346,35 @@ def is_active_sprt_ltc(run):
 def remaining_hours(run):
     r = run["results"]
     if "sprt" in run["args"]:
-        # current average number of games. Regularly update / have server guess?
-        expected_games = 58000
-        # checking randomly, half the expected games needs still to be done
-        remaining_games = expected_games / 2
+        # Current average number of games. The number should be regularly updated.
+        average_total_games = 95000
+
+        # SPRT tests always have pentanomial stats.
+        p = r["pentanomial"]
+        N = sum(p)
+
+        sprt = run["args"]["sprt"]
+        llr, alpha, beta = sprt["llr"], sprt["alpha"], sprt["beta"]
+        o0, o1 = 0, 0
+        if "overshoot" in sprt:
+            o = sprt["overshoot"]
+            o0 = -o["sq0"] / o["m0"] / 2 if o["m0"] != 0 else 0
+            o1 = o["sq1"] / o["m1"] / 2 if o["m1"] != 0 else 0
+        llr_bound = (math.log((1 - beta) / alpha) - o1 if llr > 0.0 else
+                     math.log(beta / (1 - alpha)) + o0)
+
+        # Use 0.1 as a safeguard if LLR is too small.
+        llr, llr_bound = max(0.1, abs(llr)), abs(llr_bound)
+        if llr >= llr_bound:
+            return 0
+
+        # Assume all tests use default book (UHO_XXL_+0.90_+1.19).
+        book_positions = 223070
+        t = scipy.stats.beta(1, 15).cdf(min(N / book_positions, 1.0))
+        expected_games_llr = int(2 * N * llr_bound / llr)
+        expected_games = min(run["args"]["num_games"],
+                             int(expected_games_llr * t + average_total_games * (1-t)))
+        remaining_games = max(0, expected_games - 2 * N)
     else:
         expected_games = run["args"]["num_games"]
         remaining_games = max(0, expected_games - r["wins"] - r["losses"] - r["draws"])


MinetaS:
Please check if the code is fine.
Also note that I'm totally unfamiliar with fishtest code structures so there might be some bugs...

ppigazzini:
DEV updated (https://dfts-0.pigazzini.it)
Something seems wrong (master top, PR bottom):

image

ppigazzini:
I agree in principle that we should not compare the already ran games with the expected one: those numbers should be comparable only with a big number of tests.

server/fishtest/util.py
```diff
@@ -329,10 +330,33 @@ def estimate_game_duration(tc):
...
+            o1 = o["sq1"] / o["m1"] / 2 if o["m1"] != 0 else 0
+        if abs(llr) < 1e-4:
+            llr = math.copysign(1e-4, llr)
+        llr_bound = (math.log(beta / (1 - alpha)) - o1 if llr > 0 else
+                     math.log((1 - beta) / alpha) + o0)
+
+        # Assume all tests use default book (UHO_XXL_+0.90_+1.19).
+        book_positions = 223070
+        t = scipy.stats.beta(1, 15).cdf(min(N / book_positions, 1.0))
+        expected_games_llr = int(2 * N * llr_bound / llr)
```
ppigazzini, line 357: IMO this should be:
```python
expected_games_llr = int(abs(2 * N * llr_bound / llr))
```

server/fishtest/util.py
```diff
@@ -329,10 +330,33 @@ def estimate_game_duration(tc):
...
     if "sprt" in run["args"]:
-        # current average number of games. Regularly update / have server guess?
-        expected_games = 58000
-        # checking randomly, half the expected games needs still to be done
-        remaining_games = expected_games / 2
+        # Current average number of games. The number should be regularly updated.
+        average_total_games = 95000
+
+        # SPRT tests always have pentanomial stats.
+        assert "pentanomial" in r.keys()
```
ppigazzini, line 337: Not a fan of assert, however:
```python
assert "pentanomial" in r
```

vondele:
For the purpose of estimating the remaining number of games left, probably a very simple prior (like a Gaussian centered at 0, with sigma 1), would be an improvement over what we have now.

ppigazzini:
To visualize this function:
```python
t = scipy.stats.beta(1, 15).cdf(min(N / book_positions, 1.0))
```
download

vdbergh:
> For the purpose of estimating the remaining number of games left, probably a very simple prior (like a Gaussian centered at 0, with sigma 1), would be an improvement over what we have now.

The last Elo prior was centered at -1 with sigma 1. This was before NNUE. I have not recomputed it recently.

vdbergh:
Note that you cannot just take the observed Elo's. You have to take the observed outcomes (# games + LLR) and find the Elo distribution that best matches these outcomes (e.g. by minimizing the `chi^2` value of the computed outcome distribution vs the observed outcome distribution).

vdbergh:
Note that you cannot just take the observed Elo's. You have to take the observed outcomes (# games + LLR) and find the Elo distribution that best matches these outcomes (e.g. by minimizing the `chi^2` value of the computed outcome distribution vs the observed outcome distribution).

vdbergh:
A complication is given by people stopping tests. Such tests should be discarded but they introduce bias in the non stopped tests.

server/fishtest/util.py
MinetaS: I forgot to fix the order between two bounds after changing the sign.
However, I'll add the abs in case something going wrong.

vdbergh:
I reran my elo prior calculation program (with very old data).  You can see that the prior predicted the observations very accurately (the prediction is the unlabeled orange line).

Screenshot from 2023-01-17 20-19-00

ppigazzini:
Some noob's notes:
- the stopped tests can be or Elo losing tests or wrong tests (ops, I inserted the SF7 branch instead than master!). the second ones could be a problem 
- if people keep stopping losing Elo tests with the same random process, to compute the remaining time perhaps the prior from the completed tests is not so bad
- I wonder how many tests are stopped on the grand total

vondele:
> EDIT: I am thinking of the following procedure. Once we have an Elo prior (like in the above post) we can compute the Elo posterior for each stopped test and turn it into a fuzzy finished test using the resulting outcome distribution. We can then recompute the prior and iterate. Presumably this will converge.

practically speaking, I think doing a single iteration will be relatively accurate and stable by design. For the current use case, having the perfect estimate is not particularly important.

vdbergh:
@ppigazzini 

> I wonder how many tests are stopped on the grand total

These are the statistics for the last ~ 5000 tests.

Passed (not stopped) stc tests:  median elo= -0.3256 count= 4517
Stopped stc tests:  median elo= -0.5047 count= 984

The Elo is the logistic elo as computed by sprt.py.

So the number of stopped tests is substantial and there should be some effect on the prior.

dubslow:
"Perfect" is the enemy of "better than the status quo"

ppigazzini:
This is the difference between master (top) and the PR (bottom) using the latest DB backup.
I don't have the experience to judge if the new estimate is a better fit, I need help.

image

MinetaS:
DEV looks better because there's no way 43 tests can be completed within 24 hours.
However I'm not sure about guessing remaining time based on LLR.
The metric is totally random on most of tests except decisive Elo gaining/losing patches.

Rebased + set minimum bound to 0.1 (1e-4 is to small).

ppigazzini:
DEV updated with last commit.

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @MinetaS :)

ppigazzini:
@MinetaS for AUTHORS file please confirm that your name is Syine Mineta.

MinetaS:
Yes, it's good.

Issue #1547, opened 2023-01-20
duplicate entries in event log


dubslow:
23‑01‑20 02:18:15    finished_run    Vizvezdenec    mcpTtcaptL2-ba0e5fd    Result:red Elo:-0.76[-2.15,0.65] LOS:14% Games:25416 Ptnml:[18,5172,15167,5043,16]
    23‑01‑20 02:13:38    finished_run    Vizvezdenec    mcpTtcaptL2-ba0e5fd    Result:red Elo:-0.75[-2.14,0.65] LOS:15% Games:25412 Ptnml:[18,5170,15165,5043,16]

I think it's reporting the game count incorrectly, perhaps as pairs, not games, compare the actual stats from the test page:

    Total: 50832 W: 13375 L: 13508 D: 23949
    Ptnml(0-2): 18, 5172, 15167, 5043, 16```

https://tests.stockfishchess.org/tests/view/63c7059e18c20f4929c63833

vdbergh:
😦 … I’ll fix it.

vdbergh:
Actually I am noticing you are pointing out two problems. The issue of duplicate entries in the event log is much more serious. It points to a race condition in the server.

The `finished_run` event is created by `update_task` which acquires a lock... The `finished_run` event is only generated after checking that the run is really finished.

The two `finished_run` events were five minutes apart.

Nothing happened between the two `finished_run` events:

https://tests.stockfishchess.org/actions?run_id=63c7059e18c20f4929c63833

The issue must be that `update_task` does not check if a run is already finished. However if does check that the current task is active. And if a run is stopped all tasks are set to non-active.

But conceivably a worker may request a new task while a run is being stopped in `update_task`. If I am reading the code correctly then `request_task` does not acquire the same lock as `update_task` (presumably because normally they don't interfere).

dubslow:
My basic assumption would be that some task which started before the run finished then later the task after the run ended. If I understand you right, your second-to-last paragraph says this couldn't be the cause, since "late" task results are ignored, but your last paragraph says that somehow one last task was assigned right as the run ended the first time, in an unlikely-but-plausible race condition? That would be funny lol. But as you say, if this is so, then the fix should be as simple as making `request_task` get the same lock.

vdbergh:
>  But as you say, if this is so, then the fix should be as simple as making `request_task` get the same lock.

Yeah basically, but currently this will deadlock, so it requires a bit of thought.

dubslow:
[...]

PR #1549, opened 2023-01-20
Should fix the other issue mentioned in #1547

vdbergh:
A race condition between request_task and update_task may cause a test to be stopped multiple times.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -962,7 +962,7 @@ class RunDb:
         pentanomial = results["pentanomial"]
         total = 2 * sum(pentanomial)
         ret += f" Games:{total} Ptnml:{str(pentanomial).replace(' ','')}"
-        return ret;
+        return ret
 
     def update_task(self, worker_info, run_id, task_id, stats, spsa):
         lock = self.active_run_lock(str(run_id))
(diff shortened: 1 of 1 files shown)


server/fishtest/rundb.py
```diff
@@ -962,7 +962,7 @@ def finished_run_message(self, run):
         pentanomial = results["pentanomial"]
         total = 2 * sum(pentanomial)
         ret += f" Games:{total} Ptnml:{str(pentanomial).replace(' ','')}"
-        return ret;
+        return ret
```
ppigazzini, line 965: I didn't know that semicolon are legal in python.

vdbergh:
In this case it was a typo but they can be used as a statement separator. Useful in lambda functions which must be one line.

ppigazzini (approved):
I ran some tests on DEV, it seem good.

ppigazzini:

ppigazzini:
> In this case it was a typo but they can be used as a statement separator. Useful in lambda functions which must be one line.

That seems wrong.
Lambda in python3 have only one expression and zero statements.
https://stackoverflow.com/questions/49942050/define-a-lambda-function-in-single-line-using-semicolons-in-python-3

vdbergh:
I guess I was wrong. The restrictions on lambda expressions is something that I don't like in Python. Compare this to the flexibility of using anonymous functions in javascript ... Of course the restrictions in Python are caused by the lack of proper block delimiters in the language.

vdbergh:
That pattern is called currying. In a language where functions are first class object you need only functions with one argument.

ppigazzini:
Even possible in python :)

```python
add_three = lambda a: lambda b: lambda c: a + b + c
add_three(1)(2)(3)
```

ppigazzini:
I started learning with this free book "Category Theory for Programmers"
https://github.com/hmemcpy/milewski-ctfp-pdf/

PR #1552, opened 2023-01-23
Parameters Tweak New Bounds (0 , 2)

FauziAkram:
I suggest changing the LTC bounds for the patches that only change the values of parameters, without touching anything else, from <0.5, 2.5> to <0, 2> As there is no complexity addition in these types of patches. And if it proves to pass STC and LTC both of them at <0, 2> bounds, this should confirm that the patch is good enough to be merged. This should also help in getting out local max/min.

For a very long time, in Fishtest history, we had easier bounds for parameters-tweak patches, so is not something against the logic of Stockfish maintainers.
The "paramater-tweak-only patches" does not increase the complexity at all, so the code would remain basically the same.
And having more parameters-tweak patches being merged would help in improving the engine overall by getting out local max/min, and by improving the strength of the engine even if by only 0.75 Elo each time.

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -334,7 +334,7 @@
                   <select class="form-select" id="bounds" name="bounds">
                     <option value="standard STC">Standard STC ${fb(0.0, 2.0)}</option>
                     <option value="standard LTC">Standard LTC ${fb(0.5, 2.5)}</option>
-                    <option value="Parameter's Tweak LTC">Parameter's Tweak LTC ${fb(0, 2)}</option>
+                    <option value="Parameters Tweak LTC">Parameters Tweak LTC ${fb(0, 2)}</option>
                     <option value="regression STC">Non-regression STC ${fb(-1.75, 0.25)}</option>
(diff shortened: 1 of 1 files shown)


dav1312:
"value" is not what the user will see, you need to switch those

vdbergh:
I guess it is "parameters tweak" (as it is not a tweak owned by a parameter...). Personally I would even write "parameter tweak" even if more than one parameter is being tweaked, by analogy of "parameter tweaking". But I am not sure what a native English speaker would say.

vondele:
Personally, I think the suggested benefit is small, the work associated in testing, merging etc. is the same for these patches. Also parameter tweaks risk to regress, so reducing the lower bound is not risk free. I'm thus not really in favour of merging this.

Vizvezdenec:
Yeah, lowering lower bound is bad. Sf is so strong that it may result in average regression.
And narrowing bounds isn't a good thing either (if you suggest it instead of just shifting bounds) because fishtest is busy as it is.

ppigazzini:
> Personally, I think the suggested benefit is small, the work associated in testing, merging etc. is the same for these patches. Also parameter tweaks risk to regress, so reducing the lower bound is not risk free. I'm thus not really in favour of merging this.

Closing

PR #1553, opened 2023-01-26
Fix blue showing as red

peregrineshahin:
make it consistent with the db state variable.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -921,14 +921,10 @@ class RunDb:
     def finished_run_message(self, run):
         if "spsa" in run["args"]:
             return "SPSA tune finished"
-        result = "red"
-        if run["is_yellow"]:
-            result = "yellow"
-        elif run["is_green"]:
-            result = "green"
-        ret = f"Result:{result}"
+        result = "unknown"
         if "sprt" in run["args"]:
             sprt = run["args"]["sprt"]
+            result = sprt.get("state", "")
             elo_model = sprt["elo_model"]
             alpha = sprt["alpha"]
             beta = sprt["beta"]
@@ -943,6 +939,7 @@ class RunDb:
                 elo1=elo1,
                 elo_model=elo_model,
             )
+            ret = f"Result:{result}"
             ret += f" Elo:{a['elo']:.2f}[{a['ci'][0]:.2f},{a['ci'][1]:.2f}]"
             ret += f" LOS:{a['LOS']:.0%}"
         else:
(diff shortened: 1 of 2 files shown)


vdbergh:
I don't think we should say in the event log that the result of a test is blue. It is a passed test so the result is green. In the gui it is colored blue to draw attention to the fact that it is a simplification. But this is not necessary for the event log. 

Perhaps it would be better to just say passed/failed?

peregrineshahin:
I'm not sure about calling it a green, since in that case we would need to flag blue ones as `isGreen`, pretty sure passed/failed are good enough but since we can argue that yellow is also a failed test but we mention it and flag it in the db then I'm not sure why not blue..

vdbergh:
> I'm not sure about calling it a green, since in that case we would need to flag blue ones as `isGreen`, pretty sure passed/failed are good enough but since we can argue that yellow is also a failed test but we mention it and flag it in the db then I'm not sure why not blue..

Yes I think mentioning yellow in the event log was also a big mistake. I only did it because one can currently search for yellow tests in the gui, but in fact yellow has little statistical meaning.

Therefore I am proposing the labels `passed` and `failed` in the event log, instead of the color labels.

peregrineshahin:
I'm a bit in favor of `accepted` and `rejected` since `failed` can mislead the reader because we can have failed tests for the wrong benches or if the code has errors, probably more consistent with the state variable also, do you think that will do?

probably with one-liner:
```result = state``` ?

peregrineshahin:
Let me know please if fixed games please can either be green or red because I'm not sure, so that we don't calculate based on los and have two sources of truth (if the bounds changed requires changing here) or if we have any simpler way to do it.

vdbergh:
> Let me know please if fixed games can either be green or red because I'm not sure, so that we don't calculate based on los and have two sources of truth (if the bounds changed requires changing here) or if we have any simpler way to do it.

I would say that fixed is accepted if it currently labeled green and rejected otherwise. For SPRT we can indeed just take the state!

vdbergh:
I am happy with the new logic!

vdbergh:
Actually I see that also the title of the notification has to be changed. From `XXX by XXX finished <color>` to `XXX by XXX was <state>`.

ppigazzini:
DEV updated, running some tests with expected outcome.

ppigazzini:
Duplicated notification for the first test, Events log is fine, though.

image

image

vdbergh:
The duplicate notifications are the result of a benign race condition. This can be partially fixed with weblocks but there is still the issue that the communication between tabs using local storage is not instantaneous. A more serious issue is that I think that when IOS goes to sleep, writes to local storage may simply be discarded and so they are never seen by another tab. This seems unfixable.

In short I have not succeeding in eliminating the problem on IOS. I will have to settle for mitigation.

ppigazzini:
If the text of the notifications meets the expectation , please squash the commits.

ppigazzini:

PR #1555, opened 2023-01-30
Fix spsa tunes show as `was undefined` in notifications and some clea…

peregrineshahin:
…nups


peregrineshahin:
just missed up appending html elements instead of creating elements them, draft for the moment

ppigazzini:
Started on DEV some tests (SPSA, NumGames, SPRT) with expected results.

ppigazzini:
LGTM
image

ppigazzini:
The Non-regression tests are fine as well.

image

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @peregrineshahin , nice clean-up :)

PR #1557, opened 2023-02-01
Reduce latency of LTC relative to STC

dubslow:
Adjust the TC power law exponent to make base LTC TP of 4x STC TP, with the goal of reducing LTC latency to only 1.5x STC latency.

With the recent of busyness of fishtest (a good thing IMO), the current LTC latency of sqrt(6) ~ 2.5x is making for painfully long feedback time, potentially causing authors to submit more later-unnecessary STC tests rather than wait for LTC results. The latency is actually more than 2.5x on average since LTC elo bounds are higher as well, requiring more games at the same elo performance (on average).

This adjustment reduces LTC base latency from 2.5x to 1.5x, which should hopefully reduce the number of red-herring STC submissions as the test result feedback loop tightens up.

Edit: And, quite aside from red-herring STCs and test result feedback loops, the simple fact of semantically overlapping tests taking a while even while master changes "underneath them" could lead to conflicts of results and ideas, leading to more wasted tests and/or untried ideas. In other words, there are multiple ways in which this patch should benefit stockfish.

I also tweaked the LLR bonus TP, reducing it slightly on both ends, but I'm not nearly as attached to this as to the exponent change. (1.37/0.63 new max/min vs old max/min 1.59/0.41)

Also, don't boost shorter-than-STC latency.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -599,10 +599,7 @@ class RunDb:
 
     def calc_itp(self, run):
         itp = run["args"]["throughput"]
-        if itp < 1:
-            itp = 1
-        elif itp > 500:
-            itp = 500
+        itp = max(min(itp, 500), 1)
 
         # Base TP derived from power law of TC relative to STC
         tc_ratio = run["args"]["threads"] * estimate_game_duration(run["args"]["tc"]) \
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -603,15 +603,22 @@ class RunDb:
             itp = 1
         elif itp > 500:
             itp = 500
-        itp *= math.sqrt(
-            estimate_game_duration(run["args"]["tc"]) / estimate_game_duration("10+0.1")
-        )
-        itp *= math.sqrt(run["args"]["threads"])
-        if "sprt" not in run["args"]:
-            itp *= 0.5
-        else:
+
+        # Base TP derived from power law of TC relative to STC
+        tc_ratio = run["args"]["threads"] * estimate_game_duration(run["args"]["tc"]) \
+                                          / estimate_game_duration("10+0.1")
+        # Discount longer test TP, but don't boost shorter tests
+        if tc_ratio > 1:
+            # LTC/STC tc_ratio = 6, target latency ratio = 3/2,
+            # => LTC base tp = 4 => log(4)/log(6) ~ 0.774
+            itp *= tc_ratio**0.774
+
+        # TP bonus derived from LLR (or min-bonus for nonSPRTs)
+        llr = -2.94
+        if "sprt" in run["args"]:
             llr = run["args"]["sprt"].get("llr", 0)
-            itp *= (5 + llr) / 5
+        itp *= (llr + 8) / 8 # max/min bonus 1.37/0.63
+
         run["args"]["itp"] = itp
 
     def sum_cores(self, run):
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -613,11 +613,17 @@ class RunDb:
             # => LTC base tp = 4 => log(4)/log(6) ~ 0.774
             itp *= tc_ratio**0.774
 
-        # TP bonus derived from LLR (or min-bonus for nonSPRTs)
+        # Gentle bonus/malus for positive/negative LLR (non-SPRTs are like worst LLR)
         llr = -2.94
         if "sprt" in run["args"]:
-            llr = run["args"]["sprt"].get("llr", 0)
+            # Sanity check that we're no lower than standard lower bound
+            llr = max(llr, run["args"]["sprt"].get("llr", 0))
         itp *= (llr + 8) / 8 # max/min bonus 1.37/0.63
+
+        # Extra bonus for most promising tests -- LTC at strong-gainer bounds with good LLR
+        if tc_ratio >= 3.0 and llr > 1.5 \
+           and run["args"].get("sprt", {}).get("elo0", 0) > 0:
+            itp *= 1.2 # LLR 1.5 bonus from 1.19 to 1.42, LLR 2.9 bonus from 1.36 to 1.64
 
         run["args"]["itp"] = itp
 


dubslow:
I believe that these are tests in the last few days which have been submitted because a then-pending LTC test had taken several days to look like it was going to pass. (One of those LTCs has passed, two more might yet, it remains to be seen if the corresponding followup STCs are "red herrings" or not.)

https://tests.stockfishchess.org/tests/view/63d944c7bbadd17b3787b96f
https://tests.stockfishchess.org/tests/view/63da214dbbadd17b3787dd7f

https://tests.stockfishchess.org/tests/view/63d7bfe7a67dd929a5564fcd
https://tests.stockfishchess.org/tests/view/63d7ebc5a67dd929a5565751

https://tests.stockfishchess.org/tests/view/63d9a01ebbadd17b3787c811

server/fishtest/rundb.py
```diff
@@ -603,15 +603,22 @@ def calc_itp(self, run):
...
+        # Discount longer test TP, but don't boost shorter tests
+        if tc_ratio > 1:
+            # LTC/STC tc_ratio = 6, target latency ratio = 3/2,
+            # => LTC base tp = 4 => log(4)/log(6) ~ 0.774
+            itp *= tc_ratio**0.774
+
+        # TP bonus derived from LLR (or min-bonus for nonSPRTs)
+        llr = -2.94
+        if "sprt" in run["args"]:
             llr = run["args"]["sprt"].get("llr", 0)
```
vondele, line 619: while we're at it, I would use, so we have a guaranteed lower bound.
`llr = max(llr, run["args"]["sprt"].get("llr", 0))

server/fishtest/rundb.py
```diff
@@ -603,15 +603,22 @@ def calc_itp(self, run):
...
-            itp *= (5 + llr) / 5
+        itp *= (llr + 8) / 8 # max/min bonus 1.37/0.63
```
vondele, line 620: I would like to go in the different direction, but even more cores on the tests with high llr. That will improve time to answer for strong patches. I suggest `itp *= (llr + 4) / 4`
`

dubslow: I have mixed feelings about this.

1) I have indeed been thinking for some weeks about implementing some extra bonus for the most promising tests, as candirufish and I had been doing manually;

2) More recently, I tried a couple weeks ago to set my most garbage tests to low TP to let them finish while letting more resources go to other, less-garbage tests; however, the end result turned out to be that I had the same dozen tests languishing at the bottom of my page for multiple weeks while other users only ever looked at my testcount, not overall TP. In other words, I've reached the conclusion that too low TP for bad tests only increases the clutter in the test list, so I'm against increasing the malus for bad tests, as implied by `+4/4`. A minimum TP of 0.63 is already low enough (and higher than the current 0.41, and I certainly think going below 0.41 would be even worse).

3) Regardless of 1) and 2), the considerable change in LTC thruput suggested by just tweaking the power law will go a long way to getting the LTC passers going quicker, without needing to give them any extra bonus at all.

I think we should be conservative and see how the power law tweak feels, and then later we can add an extra bonus for the best tests if we still want to. That said, I can include such extra bonus code in this PR here and now for consideration. But again, just the power law tweak will be a considerable shift in LTC latency.

ppigazzini: I agree with the proposal of do incremental changes, the code can be tested on DEV only for bugs, the effect on the queue must be tested on PROD.

server/fishtest/rundb.py
```diff
@@ -603,15 +603,22 @@ def calc_itp(self, run):
...
-        itp *= math.sqrt(
-            estimate_game_duration(run["args"]["tc"]) / estimate_game_duration("10+0.1")
-        )
-        itp *= math.sqrt(run["args"]["threads"])
-        if "sprt" not in run["args"]:
-            itp *= 0.5
-        else:
+
+        # Base TP derived from power law of TC relative to STC
+        tc_ratio = run["args"]["threads"] * estimate_game_duration(run["args"]["tc"]) \
```
ppigazzini, line 608: We use black to format the code form time to time, feel free to keep long lines.

dubslow: I've always thought that autoformatters tend to produce suboptimal results. If it's going to be spread over two lines, I may as well make that look the best I can, and line up the two function calls

dubslow:
Per my review comment, I've added code for an extra bonus for LTCs at strong-gainer bounds, however I believe it's premature to use this -- we should feel out how the power law change will work before worrying about even more bonuses. (And again, I'm not in favor of worsening the bad LLR malus.)

server/fishtest/rundb.py
ppigazzini: I routinely run black, isort, prettier on the code base, so the format by hand could be changed.

dubslow:
There's also the fact that LLR is of course, by definition, a statistically noisy number, and several recent LTC passers, not to mention all kinds of STC passers, sometimes languish at negative LLR for tens of thousands of games before bouncing back to passing. Reducing those too hard up front only means we're assigning meaning to noise, which is quite opposite to the purpose of fishtest. Overdoing the negative LLR malus is simply harmful.

ppigazzini:
The queue on DEV looks fine, the LLR bonus not kicked-in yet.

image

ppigazzini (approved):

dubslow:
@ppigazzini i suggest commenting out the extra bonus before pushing to PROD (and then we can revisit it in a month or so when we're more familiar with the base change)

ppigazzini:
Ops, PROD already updated. I will push a commit to comment out that code.
I exposed the itp by hand in the run table to help monitoring the new algorithm. Let me know if this should be a permanent info.

dubslow:
itp can already be found in each individual test page, in the long run i think it's fine, maybe even better, to keep it there, not on the run page, but im open to other opinions

dubslow:
actually, allow me to strengthen my opinion a bit: newcomers or otherwise casual viewers, which is the large majority of viewers (hundreds or thousands of people take interest in what stockfish is doing and trying), probably won't have any reference for what "itp" means, and how that does or does not relate to "games per second", or for that matter how games/second relates to time control. it's called **i**tp for a reason, and now having thought for a few minutes i think it should definitely be kept off the main run page, and left where it is in the test pages.

altho admittedly for the next day or two it's nice to have for the dozen or two people who know what it actually means in wake of the change. so i propose leave it for now, revert it in 24-48 hours.

ppigazzini:
@dubslow To avoid a mistake, I must comment out this code, right?

```python
        # Extra bonus for most promising tests -- LTC at strong-gainer bounds with good LLR
        if tc_ratio >= 3.0 and llr > 1.5 \
           and run["args"].get("sprt", {}).get("elo0", 0) > 0:
            itp *= 1.2 # LLR 1.5 bonus from 1.19 to 1.42, LLR 2.9 bonus from 1.36 to 1.64
```

dubslow:
Yes that's the chunk I was referring to, altho it was enlarged by the autoformatter lol

```        # Extra bonus for most promising tests -- LTC at strong-gainer bounds with good LLR
        if (
            tc_ratio >= 3.0
            and llr > 1.5
            and run["args"].get("sprt", {}).get("elo0", 0) > 0
        ):
            itp *= (
                1.2  # LLR 1.5 bonus from 1.19 to 1.42, LLR 2.9 bonus from 1.36 to 1.64
            )```

vondele:
I think the itp on the main page is not really needed, it is pretty redundant with the number of cores assigned.

ppigazzini:
Temporary added based on
> altho admittedly for the next day or two it's nice to have for the dozen or two people who know what it actually means in wake of the change. so i propose leave it for now, revert it in 24-48 hours.

image

dubslow:
> I think the itp on the main page is not really needed, it is pretty redundant with the number of cores assigned.

only for sanity and verification purposes, it will indeed be removed in the next day or so.

PR #1558, opened 2023-02-02
Try pre-including worker cores in cores/itp priority estimate

dubslow:
At present, STC tasks appear to systematically have higher cores/itp than LTC tasks. If this is due to low workers-per-test granularity issues, this is the fix.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -714,12 +714,21 @@ class RunDb:
 
         last_run_id = self.worker_runs.get(unique_key, {}).get("last_run", None)
 
+        # Collect some data about the worker that will be used below.
+        max_threads = int(worker_info["concurrency"])
+        min_threads = int(worker_info.get("min_threads", 1))
+        max_memory  = int(worker_info.get("max_memory", 0))
+
+        near_github_api_limit = worker_info["near_github_api_limit"]
+
         def priority(run):  # lower is better
             return (
                 -run["args"]["priority"],
-                # Try to find a new run for this worker.
+                # Try to avoid working on the same run consecutively
                 run["_id"] == last_run_id,
-                run["cores"] / run["args"]["itp"] * 100.0,
+                # Judge itp by including this potential worker, to solve
+                # low workers-per-run granularity issues
+                (run["cores"] + max_threads) / run["args"]["itp"],
                 -run["args"]["itp"],
                 run["_id"],
             )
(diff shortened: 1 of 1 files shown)


dubslow:
Altho perhaps as is, this would only result in the reverse problem, with STCs getting shortchanged. I guess I'd like to try it as is, and if that's a problem, we can also try `+ max_threads//2`

dubslow:
Upon further reflection, I have indeed switched it to `max_threads//2`:

> Tests with low itp/workers-per-test can cause granularity issues. If
we simply use oldcores/itp, then low-core tests will be overweighted
when they're assigned large workers. If we simply use newcores/itp,
then low-core tests will be underweighted when they're *not* assigned
large workers. Split the difference by splitting the difference, and
also ensuring at least one worker at all times.

dubslow:
Recently, I just sampled the main page as itp 7756, cores 5079, global avg c/i 0.655, and a trend line fit of `c(i) = 0.560i + 15.3` with `R^2 = 0.964`. I hope that after merging this the trend slope will be closer to global average (altho a perfect fit is impossible and perhaps undesirable).

vdbergh:
The idea behind this PR seems plausible to me. Note however that noob's workers cannot take LTC tasks. This might also influence the ratio `cores/itp`.

dubslow:
> Note however that noob's workers cannot take LTC tasks.

Fascinating, do you know why? Usually the limitation is the opposite.

At any rate, yes the averaging and the `> 0` check should both work to keep it all in balance, even with a no-LTC restriction on part of the fleet (and since a significant other part of the fleet has the opposite restriction).

dubslow:
I suppose I'll mention that I reviewed the actual worker-run-assignment code as well, and that looks fairly robust. Any worker can take any run at any time which is suitable for that worker, it's just that the code tests "suitable" in order of cores/itp. So an adjustment like this will in no way prevent noob from getting STCs -- it just might take a few more iters of "is_run_suitable()". (Meanwhile, workers will the opposite restriction will take less such iters, so overall it's still a wash for the server.)

ppigazzini (approved):
Looks good on DEV

ppigazzini:

dubslow:
A recent sample shows itp 8060 (more tests but less itp/test than before), cores 3770, `c/i = 0.468`, and a trend line fit of `c(i) = 0.506i - 5.42` with `R^2 = 0.970`.

Tis surprising to me that the slope is higher than the global average, but I assume a fluctuation and not a big deal. It's at least in part driven by two outlier LTCs with the second and third largest `c/i` of the lot (first is a test with 8 itp and 16 cores for `c/i = 2` lol). Checking those two outliers just now, they are returned much closer to the global norm, so perhaps it was some assignment flux as one larger was finishing while another was just starting.

Overall I think this patch has met its goal, and produced its intended effect.

I do have some small concerns that smaller workers will be biased towards lower itp/TC tests, and I assume that smaller workers are correlated with more unreliability/variability. Previously lots of large workers got assigned low itp tests regularly, but now less regularly, so some small-worker-effects may now be "louder" in STC results. But I assume that this concern is nearly negligible.

dubslow:
itp 7750, cores 3783, global `c/i = 0.488`, trend line `c(i) = 0.600i - 12.4` and `R^2 = 0.964`. That's quite the gap between the slope and global average, but there's a lot less LTCs than a few days ago.

At any rate, I'm satisfied, and will stop monitoring.

PR #1559, opened 2023-02-03
 try at highlighting LTC tests

dubslow:
untested. i cannot into mako or html, pls critique

### DIFF
--- a/server/fishtest/templates/elo_results.mak
+++ b/server/fishtest/templates/elo_results.mak
@@ -10,7 +10,7 @@
       ret = f'style="background-color: {style};"'
 
     classes = 'rounded elo-results results-pre'
-    if is_active_sprt_ltc(run)
+    if is_active_sprt_ltc(run):
       classes += ' ltc-highlight'
     tc = run ['args']['tc']
     new_tc = run['args'].get('new_tc',tc)
@@ -81,7 +81,7 @@
         <div style="margin-left:90px;">
     % endif
 % endif
-<pre ${results_pre_attrs(run)}>
+<pre ${results_pre_attrs(run)|n}>
   ${list_info(run)}
 </pre>
 % if show_gauge:
--- a/server/fishtest/templates/elo_results.mak
+++ b/server/fishtest/templates/elo_results.mak
@@ -1,15 +1,23 @@
 <%page args="run, show_gauge=False"/>
 
 <%!
-  def get_run_style(run):
-    ret = 'white-space:nowrap; border: 1px solid rgba(0,0,0,0.15);'
-    style = run['results_info'].get('style','')
-    if style != '':
-      ret += 'background-color:' + style+';'
+  from fishtest.util import is_active_sprt_ltc
+
+  def results_pre_attrs(run):
+    ret = ''
+    style = run['results_info'].get('style', '')
+    if style:
+      ret = f'style="background-color: {style};"'
+
+    classes = 'rounded elo-results results-pre'
+    if is_active_sprt_ltc(run)
+      classes += ' ltc-highlight'
     tc = run ['args']['tc']
     new_tc = run['args'].get('new_tc',tc)
     if tc != new_tc:
-       ret += 'border-style:solid;border-color:Pink;border-width:medium;'
+      classes += ' time-odds'
+    ret += f' class="{classes}"'
+
     return ret
 %>
 
@@ -73,7 +81,7 @@
         <div style="margin-left:90px;">
     % endif
 % endif
-<pre style="${get_run_style(run)}" class="rounded elo-results">
+<pre ${results_pre_attrs(run)}>
   ${list_info(run)}
 </pre>
 % if show_gauge:
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -20,6 +20,7 @@ from fishtest.userdb import UserDb
 from fishtest.util import (
     crash_or_time,
     estimate_game_duration,
+    get_tc_ratio_stc,
     format_bounds,
     format_results,
     get_bad_workers,
@@ -602,11 +603,7 @@ class RunDb:
         itp = max(min(itp, 500), 1)
 
         # Base TP derived from power law of TC relative to STC
-        tc_ratio = (
-            run["args"]["threads"]
-            * estimate_game_duration(run["args"]["tc"])
-            / estimate_game_duration("10+0.1")
-        )
+        tc_ratio = get_tc_ratio_stc(run["args"]["tc"], run["args"]["threads"])
         # Discount longer test TP, but don't boost shorter tests
         if tc_ratio > 1:
             # LTC/STC tc_ratio = 6, target latency ratio = 3/2,
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  background: rgba(255, 150, 0, 0.08);
+  border: 2px solid rgba(100, 200, 50, 0.4);
 }
 
 .time-odds {
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -206,6 +206,19 @@ legend {
   line-height: normal;
 }
 
+.results-pre {
+  white-space: nowrap;
+  border: 1px solid rgba(  0,   0,  0, 0.15);
+}
+
+.ltc-highlight {
+  background: rgba(255, 150, 0, 0.08);
+}
+
+.time-odds {
+  border: medium solid Pink;
+}
+
 .checkbox.inline {
   margin-left: 20px;
   margin-right: 0;
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -267,37 +267,47 @@ td[style*="background-color:#FF6A6A"] {
 }
 
 /* Darken white box */
-pre[style*="white-space:nowrap"] {
+pre[class*="results-pre"] {
   background: var(--bg-dark-3) !important;
   color: var(--color-dark-1) !important;
 }
 
+pre[class*="ltc-highlight"] {
+  background: rgb(40, 40, 15) !important;
+  color: var(--color-dark-1) !important;
+}
+
+span[class*="ltc-highlight"] {
+  background: rgb(40, 40, 15) !important;
+  color: var(--color-dark-1) !important;
+}
+
 /* Darker yellow */
-pre[style*="background-color:yellow"] {
+pre[style*="background-color: yellow"] {
   background: #efef22cc !important;
   color: var(--bg-dark-3) !important;
 }
 
 /* Darker red */
-pre[style*="background-color:#FF6A6A"] {
+pre[style*="background-color: #FF6A6A"] {
   background: #ed6464cc !important;
   color: var(--bg-dark-3) !important;
 }
 
 /* Darker green */
-pre[style*="background-color:#44EB44"] {
+pre[style*="background-color: #44EB44"] {
   background: #75bb76 !important;
   color: var(--bg-dark-3) !important;
 }
 
 /* Darker blue */
-pre[style*="background-color:#66CCFF"] {
+pre[style*="background-color: #66CCFF"] {
   background: #88d7ffcc !important;
   color: var(--bg-dark-3) !important;
 }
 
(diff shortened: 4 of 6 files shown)


ppigazzini:
```diff
diff --git a/server/fishtest/templates/run_table.mak b/server/fishtest/templates/run_table.mak
index eff1dc8..dd2cba7 100644
--- a/server/fishtest/templates/run_table.mak
+++ b/server/fishtest/templates/run_table.mak
@@ -9,7 +9,7 @@
 % endif

 <%
-  from fishtest.util import get_cookie
+  from fishtest.util import get_cookie, estimate_game_duration
   if toggle:
     cookie_name = toggle+"_state"
 %>
@@ -114,7 +114,15 @@
                 % else:
                   ${run['args']['num_games']}
                 % endif
-                @ ${run['args']['tc']} th ${str(run['args'].get('threads',1))}
+                <%
+                   tc_ratio = estimate_game_duration(run['args']['tc']) / estimate_game_duration("10+0.1")
+                   ## SMP-STC = 5+0.5 * 8 th --> tc_ratio == 4
+                %>
+                % if tc_ratio > 4 and 'sprt' in run['args']:
+                   @ <mark> ${run['args']['tc']} th ${str(run['args'].get('threads',1))} </mark>
+                % else:
+                   @ ${run['args']['tc']} th ${str(run['args'].get('threads',1))}
+                % endif
                 <br>
                 ${f"itp: {round(run['args']['itp'])} " if not run['finished'] else ''}
                 ${f"cores: {run['cores']}" if not run['finished'] and 'cores' in run else ''}
```


image

dubslow:
that.... looks basically what i thought it should look like? pog

what do you think then?

ppigazzini:
To be honest an eye killer in dark theme.
Hoping @dav1312 can suggest a proper fix.

dubslow:
well ive never been known for my design skills but at least the functionality is there! will gladly merge changes, especially design fixes

dubslow:
ah i see now that my code did not work, your diff is different from mine

ppigazzini:
If I'm not wrong the conditional should be done on:
```html
              <td style="width: 11%;" class="run-live">
```

Torom:
I personally liked the way tests with distinct time controls were highlighted. But since this is very rarely used, wouldn't a decent frame around the "LLR field" be a good way to highlight LTC tests?

dubslow:
> If I'm not wrong the conditional should be done on:

Well the conditional can be put anywhere of course. Anyone can tell me what the special html would be!

> I personally liked the way tests with distinct time controls were highlighted. But since this is very rarely used, wouldn't a decent frame around the "LLR field" be a good way to highlight LTC tests?

If you can write HTML corresponding to your choice, post it here or PR it!

ppigazzini:
> Well the conditional can be put anywhere of course

Sorry, I wrote in hurry. I meant that the conditional should be done on the CSS styling.

dubslow:
By "HTML" I meant "HTML/CSS" lol, the whole web thing is really not my forte

dubslow:
I've put the first version here https://github.com/glinscott/fishtest/compare/master...dubslow:fishtest:highlightLTCold and force-push replaced it with a new version, which attempts (and probably fails) to add a border around the LLR box and around the tc description, please let me know how this works ppigazzini

dubslow:
I played around in my browser some, confirmed that what I've tried at least works in theory, and dare I say even looks good to my unqualified eye. I even factored out some styling from python+mako into CSS!

Please review and test!

peregrineshahin:
Haven't checked how it looks design wise yet, but code wise you're doing well factoring the styling to the CSS, the remaining part towards the best practice is that all style should be refactored there also, changing the backend function not to return styles so instead it should return your desired labels and then you conditionally switch over the returned string to apply the predefined css class you desire.

peregrineshahin:
Ah I see the backend style function already exists in master! Probably refactoring it for another day.

server/fishtest/templates/elo_results.mak
```diff
@@ -2,14 +2,14 @@
 
 <%!
```
ppigazzini, line 3: Missing import:
```python
  from fishtest.util import is_active_sprt_ltc
```

ppigazzini:
DEV https://dfts-0.pigazzini.it/tests after fixing the bug by hand.

server/fishtest/templates/elo_results.mak
dubslow: Honestly it should just be a method of `run` lol

dubslow:
> DEV https://dfts-0.pigazzini.it/tests

Hm, it does seem perhaps understated, I wonder if some background shading might be better. But then, definitely way better to be understated than overstated, and I think it's at least an improvement over nothing for now

dav1312:
> To be honest an eye killer in dark theme. Hoping @dav1312 can suggest a proper fix.

Just add another style in the dark theme css to override it :)

dubslow:
What about something like this, replacing the border with a mostly-transparent orange background?

image

ppigazzini:
Added a LTC with time odds

image

image

dubslow:
Maybe make both the ltc and odds highlighting as backgrounds? (If both, odds pink highlighting gets priority in the results box, while the ltc orange highlighting remains on the base tc)

dubslow:
image

I like this enough that I will push this to this PR.

dubslow:
Pushed my most recent screenshot. It also includes considerable factoring out of styles into the CSS, making future style tweaks much easier (for example, switching from the orange background to the green border would be just changing one line, unlike before).

vondele:
I guess I would prefer something less strong as a highlight, not necessarily adding an additional color to the interface. 
We already have a dedicated page for LTC tests, what about having that show the active and pending tests as well? I think having some toggles on the main page (e.g. for LTC, 'my tests',  passed etc) could be at the same time visually simpler, and more powerful..

dubslow:
I've squashed and force-pushed since this version is quite a bit better than master, even if one removes the LTC highlighting entirely.

Switching back to the green border is as easy as this:

```
diff --git a/server/fishtest/static/css/application.css b/server/fishtest/static/css/application.css
index 2e467ae..a1b4f14 100644
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  background: rgba(255, 150, 0, 0.08);
+  border: 2px solid rgba(100, 200, 50, 0.5);
 }
 
 .time-odds {
```

dubslow:
> We already have a dedicated page for LTC tests, what about having that show the active and pending tests as well? 

This is a little bit beyond my paygrade at the moment

> I think having some toggles on the main page (e.g. for LTC, 'my tests', passed etc) could be at the same time visually simpler, and more powerful..

This is way beyond my paygrade

server/fishtest/templates/elo_results.mak
```diff
@@ -73,7 +81,7 @@
         <div style="margin-left:90px;">
     % endif
 % endif
-<pre style="${get_run_style(run)}" class="rounded elo-results">
+<pre ${results_pre_attrs(run)}>
```
ppigazzini, line 84: see https://docs.makotemplates.org/en/latest/filtering.html#expression-filtering
```diff
- <pre ${results_pre_attrs(run)}>
+ <pre ${results_pre_attrs(run)|n}>
```

- before:

image


- after:

image

ppigazzini (commented):
DEV updated applying by hand the two bug fixes.

image

peregrineshahin:
fading the background color of the pre's with opacity is not readable to me in dark mode and messes up the contrast of the colors, I suggest reverting to the border idea or using a faded shadow with box-shadow property..

@dav1312 I remember you showed us a good design for this functionality, can you please just provide us with the design aspect of your proposal, probably the CSS classes you had?

peregrineshahin:
Also probably finished tests doesn't need changing style wise, active tests should be the target for such functionality imo

dav1312:
> @dav1312 I remember you showed us a good design for this functionality, can you please just provide us with the design aspect of your proposal, probably the CSS classes you had?

I added a class to each row (tr) and then used
```css
.class {
    opacity: .3;
}
```

peregrineshahin:
I definitely agree that applying additional style stuff over the table row as a whole is the way to go. Thanks

vdbergh:
What about simply sorting the table so that LTC tests are in front? Or provide a clickable header?

ppigazzini:
Added on the fly a sortable header on DEV (to be improved of course)

dav1312:
> Added on the fly a sortable header on DEV (to be improved of course)

Would need to be hidden for small screens

server/fishtest/templates/elo_results.mak
dubslow: fascinating

dubslow:
It is definitely *not* intended that finished tests are affected. If ppigazzini added the missing colon but then the finished results are still affected, that's a bug somewhere.

dubslow:
> I added a class to each row (tr) that wasn't an LTC test and then used

was that the entire design, no extra color or shapes or anything, just fading the non-LTCs a bit?

ppigazzini:
DEV updated with the new commit, this is the Green Test page (dark theme) PROD vs DEV

image

dubslow:
Yea I guess dark theme has no idea how to work with `.results-pre`, which at the moment fixing that is above my paygrade, but I'll try...

dubslow:
Holy shit whitespace is significant in css???? This patch fixes at least some of the night theme colors:

```diff --git a/server/fishtest/templates/elo_results.mak b/server/fishtest/templates/elo_results.mak
index 8d7d137..b0fa2d7 100644
--- a/server/fishtest/templates/elo_results.mak
+++ b/server/fishtest/templates/elo_results.mak
@@ -7,7 +7,7 @@
     ret = ''
     style = run['results_info'].get('style', '')
     if style:
-      ret = f'style="background-color: {style};"'
+      ret = f'style="background-color:{style};"'
 
     classes = 'rounded elo-results results-pre'
     if is_active_sprt_ltc(run):
```

dav1312:
> Holy shit whitespace is significant in css????

Hmmm no, it shouldn't matter

dubslow:
Well it worked in my browser dev tools, so I've pushed those fixes. The LTC highlight in dark theme is a fairly ugly color, but changing said color is easy enough.

ppigazzini:
DEV updated, dark theme seems fixed.

dubslow:
It does seem so. Still, I'm open to other designs.

>> I added a class to each row (tr) that wasn't an LTC test and then used

> was that the entire design, no extra color or shapes or anything, just fading the non-LTCs a bit?

@dav1312 for instance this? ^ I can try reimplementing this or something

ppigazzini:
> Holy shit whitespace is significant in css???? This patch fixes at least some of the night theme colors:
> 
> ```diff
> index 8d7d137..b0fa2d7 100644
> --- a/server/fishtest/templates/elo_results.mak
> +++ b/server/fishtest/templates/elo_results.mak
> @@ -7,7 +7,7 @@
>      ret = ''
>      style = run['results_info'].get('style', '')
>      if style:
> -      ret = f'style="background-color: {style};"'
> +      ret = f'style="background-color:{style};"'
>  
>      classes = 'rounded elo-results results-pre'
>      if is_active_sprt_ltc(run):
> ```

I confirm that this little change worked on DEV (tested before the last commit)

ppigazzini:
But that little change is wrong is applied after the last commit.

image

dubslow:
Yes, the commit fixed it the other way, by adding space to the night theme rules rather than removing space from the template. Don't apply the one liner atop the proper commit.

dubslow:
I've gotten little help on how to improve the background color, so I've pushed a version back to the green border.

However the key point of this change is to highlight that this PR refactors enough code that the style is now properly in the CSS, not hacked together in the Python/Mako. Note that this last commit only changes two lines, one for the light theme and one the dark theme, in the CSS, nothing in the code.

For this reason alone, the PR should be merged, even if the LTC highlighting is entirely removed. Style is no business of the backend, and this refactoring will make it much easier to apply any styling that might be desired in the future.

ppigazzini:
DEV updated, IMO this version looks good

image

ppigazzini (approved):
@dubslow LGTM, please squash the commits and write the final commit message :)

ppigazzini:

dubslow:
Thanks, I was a wee bit afk lol. My only followup thought is, should the `sprt @` bit also be included in the highlighting?

PR #1560, opened 2023-02-04
Final itp adjustments

dubslow:
1) remove itp from run table

2) add worker count to run table like so: `cores: 150 (2)` (open to feedback)

3) adjust the itp LLR bonus to a tighter window, with asymmetric bonus (reward good LLR more than punish bad) (see commit message)


ppigazzini:
DEV update LGTM

image

ppigazzini (approved):

ppigazzini:

PR #1561, opened 2023-02-11
Fix misalignment on mobile.

peregrineshahin:
before:
before
after:
after


ppigazzini (approved):
DEV updated, LGTM on mobile/desktop.

ppigazzini:

PR #1562, opened 2023-02-11
Add new line to big screens and temporarily change colors ltc highlig…

peregrineshahin:
…ht colors to suggested ones.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  border: 2px solid rgba(100, 200, 50, 0.4);
+  outline: 2px solid rgba(0, 100, 0, 0.8);
 }
 
 .time-odds {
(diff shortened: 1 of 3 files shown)


ppigazzini:
DEV updated.

peregrineshahin:
this will do the job, for the time being, will need feedback from users on colors and any suggested design enhancements

ppigazzini:
DEV updated

peregrineshahin:
the intended behavior is confirmed.

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @peregrineshahin for the fast fix :)

dubslow:
I don't fully follow the div change, but sure.

I do however think the original color was better. It's supposed to be a subtle eye attractor, not a big massive "LOOK AT ME" sort of thing which is what it feels like now. It should rather be the ~same shade as the link color (as it was before).

peregrineshahin:
The <div> change as opposed to a <br> fixes the responsiveness naturally because divs are displayed as a block adding a break line but in mobile the break line is not needed and handled otherwise via CSS and not HTML. 

Regarding the colors I agree that the color of the outline is too strong on the night-mode layout and is not in contrast with the other green which links have. But it was added temporarily based on 2 users feedback.
As of all, I'm in favor of the idea of making ltc easier to spot but I'm not sure about the design details before and after this commit.

PR #1564, opened 2023-02-13
Use get_tc_ratio more, add cache to estimate_game_duration

dubslow:
The former saves a bit of LOC, the latter should improve server performance.

Note that the tc_too_short comparison is now `<` rather than `<=`

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -3,6 +3,7 @@ import re
 import smtplib
 from datetime import datetime, timedelta
 from email.mime.text import MIMEText
+from functools import cache
 
 import fishtest.stats.stat_util
 import numpy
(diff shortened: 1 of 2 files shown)


server/fishtest/util.py
```diff
@@ -296,6 +297,7 @@ def format_results(run_results, run):
     return result
 
 
+@lru_cache # A single hash lookup is much cheaper than re-parsing a string every time
```
ppigazzini, line 300: We have only few TCs, so we can use safely the slightly faster `lru_cache(maxsize=None)` or the alias `cache` introduced with pyhton 3.9
https://docs.python.org/3/library/functools.html#functools.cache

ppigazzini:
DEV updated

peregrineshahin:
My suggestion is to do the opposite; Remove the outline around the `sprt @ 60+06s`
and keep the outline around the LLR box with: 
`outline: 2px solid #685210 !important; for dark mode`
and
`outline: 2px solid #d1b25d;` for white mode.

I believe this is the best way we can take advantage of the border/outline idea without applying a complex design idea on table rows as a whole.

ppigazzini:
The PR highlight works for time odds tests as well. A rare type of test, though.

image

ppigazzini (approved):

ppigazzini:

PR #1565, opened 2023-02-13
Some possibilities for LTC highlighting

dubslow:
Each commit is to be considered separately, independently:

1) Adjust color to match link color (still too dark/bright/attention-grabbing for my preference), reduce width for high-contrast color

2) Include `sprt @` in highlight box

3) don't highlight LLR box

@peregrineshahin thoughts?

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  outline: 2px solid rgba(0, 100, 0, 0.8);
+  outline: 1px solid var(--bs-link-color);
 }
 
 .time-odds {
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  outline: 1px solid var(--bs-link-color);
+  color: var(--bs-link-color);
 }
 
 .time-odds {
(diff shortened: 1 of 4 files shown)


ppigazzini:
DEV updated

peregrineshahin:
My suggestion is to do the opposite; Remove the outline around the sprt @ 60+06s
and keep the outline around the LLR box with:
outline: 2px solid #685210 !important; for dark mode
and
outline: 2px solid #d1b25d; for white mode.

I believe this is the best way we can take advantage of the border/outline idea without applying a complex design idea on table rows as a whole.

ppigazzini:
The PR highlight works for time odds tests as well. A rare type of test, though.

image

dav1312:
RFC
image

peregrineshahin:
Probably works also.

dubslow:
Pushed a version akin to dav's image, albeit still highlighting the `@`, altho that can of course be easily changed either way.

ppigazzini:
DEV updated, LGTM

image

dubslow:
My view in lightmode has much more contrast than your image: 
image

ppigazzini:
In effect lower contrast in Chrome and higher in Edge. I use only the dark theme, by the way.

dubslow:
My higher lighttheme contrast is in Chromium

ppigazzini (approved):

ppigazzini:

PR #1566, opened 2023-02-14
Add tooltip with itp and cores/itp to homepage

dubslow:
This allows more advanced users to keep a closer eye on the server without confusing new viewers


server/fishtest/rundb.py
```diff
@@ -522,18 +521,19 @@ def aggregate_unfinished_runs(self, username=None):
             if cores > 0:
                 eta = remaining_hours(run) / cores
                 pending_hours += eta
+                itp += run["args"].get("itp", 0)
```
ppigazzini, line 524: Change with zero practical effects but IMO it's more consistent if we use always the same default in case of missing key.

```python
itp += run["args"].get("itp", 100)
```

ppigazzini:
DEV updated, LGTM, just nit picking.

server/fishtest/rundb.py
dubslow: Ah, I think my first version had this line elsewhere, but behind the `cores > 0` guard it does make sense to have a positive default.

ppigazzini:
I fixed the conflicts, fixed 2 more errors in the original PR, rebased on master and ran on DEV.
There are not tooltips.

PR #1567, opened 2023-02-15
Discount tunes in time remaining

dubslow:
and misc cleanups in the function

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -344,16 +344,17 @@ def is_active_sprt_ltc(run):
 
 def remaining_hours(run):
     r = run["results"]
+    N = r["wins"] + r["losses"] + r["draws"]
+    remaining_games = max(0, run["args"]["num_games"] - N)
     if "sprt" in run["args"]:
         # current average number of games. Regularly update / have server guess?
         expected_games = 58000
         # checking randomly, half the expected games needs still to be done
         remaining_games = expected_games / 2
-    else:
-        expected_games = run["args"]["num_games"]
-        remaining_games = max(0, expected_games - r["wins"] - r["losses"] - r["draws"])
-    game_secs = estimate_game_duration(run["args"]["tc"])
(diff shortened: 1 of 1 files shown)


server/fishtest/util.py
```diff
@@ -343,17 +343,18 @@ def is_active_sprt_ltc(run):
 
 
 def remaining_hours(run):
+    if "spsa" in run["args"]: # Tunes are frequently stopped short, and with "autoselect" may have too many games anyways
+        return 0
     r = run["results"]
+    N = sum(r["pentanomial"]) if "pentanomial" in r else r["wins"] + r["losses"] + r["draws"]
```
ppigazzini, line 349: Check if this is the intent:
 - a run with "pentanomial" field has "wins", "losses", "draws" fields as well
 - `sum(r["pentanomial"])` is half the number of games played

server/fishtest/util.py
```diff
@@ -343,17 +343,18 @@ def is_active_sprt_ltc(run):
 
 
 def remaining_hours(run):
+    if "spsa" in run["args"]: # Tunes are frequently stopped short, and with "autoselect" may have too many games anyways
```
ppigazzini, line 346: What about adding a time scale factor for SPSA?

MinetaS:
To be honest I don't agree with this one.
It's named "remaining hours" not "expected remaining hours" ... for fixed games tests, this metric should be consistent.

server/fishtest/util.py
dubslow: ah, i was borrowing from Mineta's code lol, and did not stop to think about it

dubslow:
> It's named "remaining hours" not "expected remaining hours" ... for fixed games tests, this metric should be consistent.

Well fixed games tests have wildly differing behavior depending on if it's a tune or not. Nontunes finish their full count of games with high probability, whereas tunes fall well short of their full count with high probability.

However, as ppigazzini suggests, perhaps discounting tunes is better than ignoring them altogether.

ppigazzini:
Closed after merging #1545

PR #1570, opened 2023-02-16
Add itp tooltip to each run's core count

dubslow:
Allows site experts to better keep the finger on the pulse without overwhelming newbs

### DIFF
--- a/server/fishtest/templates/run_table.mak
+++ b/server/fishtest/templates/run_table.mak
@@ -119,7 +119,7 @@
                     ${run['args']['tc']} th ${str(run['args'].get('threads',1))}
                 </span>
                 % if not run['finished']:
-                    <div>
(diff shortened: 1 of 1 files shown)


server/fishtest/templates/run_table.mak
```diff
@@ -119,7 +119,7 @@
                     ${run['args']['tc']} th ${str(run['args'].get('threads',1))}
                 </span>
                 % if not run['finished']:
-                    <div>
+                    <div title="${f"itp: {run["itp"]:d}, c/i: {run["cores"]/run["itp"]:.2f}"}">
```
ppigazzini, line 122: wrong syntax.
```diff
- <div title="${f"itp: {run["itp"]:d}, c/i: {run["cores"]/run["itp"]:.2f}"}">
+ <div title="${f'itp: {run["itp"]:.0f}, c/i: {run["cores"]/run["itp"]:.2f}'}">
```
However the `run` dictionary has not the `itp` key.

PR #1571, opened 2023-02-16
Reject time losses mid-run, take 2

dubslow:
Considerably more sophisticated than my prior attempt some months ago. There's a lot going on here.

The first 6 commits are very nearly non functional, being solely cleanups and refactorings which were necessary for me to even hope to be able to understand what I was modifying (and even then I'm only like 80% sure). Even if the functionality of this PR is rejected, the first 6 commits of cleanups and refactoring should be considered/merged independently. (The third commit is slightly functional, but only the slightest bit, I believe only cosmetically.)

The last two commits are the meat of the proposed new functionality. Instead of waiting until the run is "over" to autopurge `crash_or_time` tasks (if autopurge is even set, which it usually is not), we purge them immediately, mid-run, if the worker has a recent history of spewing garbage.

My usual caveat applies: this is completely untested, and even more than that, I'm not even entirely sure that the last two commits are correct, in the sense of doing what I intended. In any case, I believe it should be easier to understand each commit individually than the grand sum diff. All kinds of review/feedback is requested and appreciated.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -78,8 +78,8 @@ def strip_run(run):
     run = copy.deepcopy(run)
     if "tasks" in run:
         run["tasks"] = []
-    if "bad_tasks" in run:
-        run["bad_tasks"] = []
+    if "purged_tasks" in run:
+        run["purged_tasks"] = []
     if "spsa" in run["args"] and "param_history" in run["args"]["spsa"]:
         run["args"]["spsa"]["param_history"] = []
     run["_id"] = str(run["_id"])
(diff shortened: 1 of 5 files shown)


dubslow:
I've stored the original, more detailed commit history here: https://github.com/glinscott/fishtest/compare/master...dubslow:fishtest:rejectTimeLosses2Orig

and instead have forcepushed a simplified history more suitable for merging.

dubslow:
I think it's probably a benefit to have "this run was purged of n tasks" as an event log, in parallel with the time or crashes log event, but I'll put that idea off until this one is complete in some form.

dubslow:
@vdbergh btw I would very much appreciate your review/feedback of this patch, since among other reasons you're the most familiar with the code.

PR #1573, opened 2023-02-22
LTC highlight back to outline (vote on discord!)

dubslow:
This is obviously somewhat contentious, being a subjective design and style choice.

My original intent was to keep it subtle and out of the way, so that those looking for it would find it, but that those *not* looking for it wouldn't be distracted by it. I felt some earlier iterations of this highlighting violated the second principle, hence the current form of subtle text coloring, but this current iteration was generally disliked on Discord.

Therefore, here is a small commit that switches back to using an outline, but on less elements than before, and with a thinner and overall less distracting form, in my opinion. Nevertheless, my opinion doesn't count more than anyone else's, so lets get more opinions.

Below in this message are sample images from my browser of the current and proposed version. Please vote for your preference. Thumbs up for the proposed outline version, thumbs down to keep the current text coloring. If prefer for right now to focus on this firm binary; if you prefer some third alternative, please still vote on the current two choices, but feel free to leave a message for future discussion/changes.

I assume it's better to vote on Discord than on Github, but whatever works. Here's the Discord voting link: https://discord.com/channels/435943710472011776/900771437177094156/1077938298900713502

Now the pictures:

Old/current style, using textcolor, if you prefer this, vote thumbs down:

image
 image

Proposed style, using 1px outline, if you prefer this, vote thumbs up:

image 
image

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -212,7 +212,7 @@ legend {
 }
 
 .ltc-highlight {
-  color: var(--bs-link-color);
+  outline: 1px solid var(--bs-link-color);
 }
 
 .time-odds {
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -297,7 +297,7 @@ pre[style*="background-color: #66CCFF"] {
 }
 
 .ltc-highlight {
-  color: var(--bs-link-color) !important;
+  outline: 1px solid var(--bs-link-color) !important;
 }
 
 /* Darken pink border */


vdbergh:
Personally I think that simply sorting the list with the LTC tests in front would clearer/easier.

dubslow:
I think that's the worst of all worlds, sorting by LLR is absolutely the best sort scheme. But alternate stylings, such as bolding text, are perfectly fine suggestions.

vdbergh:
> I think that's the worst of all worlds, sorting by LLR is absolutely the best sort scheme. But alternate stylings, such as bolding text, are perfectly fine suggestions.

Not sure why LLR would be objectively the best sorting scheme, given that it varies wildly during the life time of a test and no information can be deduced from its value. I prefer to sort the tests by importance. 

I actually I find all these decorations rather ugly in a non sorted list (and in a sorted list they are superfluous). The reason that I introduced a border for time odds test was to make sure that they would not accidentally lead to a PR. It was not to draw attention to them. With this philosophy other non standard tests should get a border too (e.g. different opening book, non standard TC etc...).

ppigazzini:
Closing because the negative votes won.

PR #1574, opened 2023-02-24
Update sprt calculator

dav1312:


ppigazzini:

PR #1575, opened 2023-03-14
A new option to override default platform version to remove sensitive information displayed to the public in the "System" column of the FishTest website workers list

maximmasiutin:
There are custom Linux builds that may contain sensitive information that is inappropriate to be sent to the server to be further displayed in the "System" column of the Fishtest website. An example is "Linux 5.15.0-53-contoso". The platform version here is "5.15.0-53-contoso". We have to patch the Fishtest code to remove this sensitive info, e.g.
```
sed -i 's/uname\[2\]/uname2/g' worker.py
sed -i '/    uname = platform.uname()/a\
    uname2 = uname[2].replace("-contoso", "")' worker.py
```
As a result of such patching, the version of Fishtest is displayed in the Worker column as one being modified, i.e. with an asterisk: "197*".

A better alternative would have been an option to override the platform version (one returned in uname[2] of platform.uname(), such as "5.15.0-53-contoso") to have the Python source code of the worker unmodified.

I propose the --platform_release command-line option (-R) and the same config file value, so that the users will be able to pass this information via command line, e.g. 

```
python3 worker.py --platform_release $(uname -r | sed s/-contoso//)
```


noobpwnftw:
Or an option to not sending this information at all since they are cosmetics anyways.

maximmasiutin:
> Or an option to not sending this information at all since they are cosmetics anyways.
@noobpwnftw 

Since the platform version in my example is is "5.15.0-53-contoso", the full string here is "Linux 5.15.0-53-contoso". I think that some information is still needed, e.g. "Linux 5.15" or "Darwin 21.6.0" or "Windows 10", but not the full Linux platform version with suffixes like "5.15.0-53-contoso".

PR #1576, opened 2023-03-14
itp: remove malus for bad LLR

dubslow:
I've concluded that the negative malus merely inflates the active test count rather than meaningfully spreading the love.

The positive bonus suffices to speed up passers, while clearing out the losers keeps the total test count lower, making the queue cleaner and easier to read.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -610,24 +610,20 @@ class RunDb:
             # => LTC base tp = 4 => log(4)/log(6) ~ 0.774
             itp *= tc_ratio**0.774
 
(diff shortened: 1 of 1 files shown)


dubslow:
In wake of current problems, I'd like to point out that this small tweak should have the effect of *slightly* lowering backend CPU load. By  increasing the average itp/run, we also effectively decrease the total number of active runs at any given time (for a fixed submission rate and nps supply), which should be a boon to both worker assignment and homepage generation. Not a large one, but perhaps a couple percent.

ppigazzini (approved):
LGTM on DEV

ppigazzini:
ps: only a heartless maintainer could reject a PR that `meaningfully spreads the love`

PR #1580, opened 2023-03-16
Proper engine paths for cutechess-cli 1.3.0 (When cutechess-cli 1.3.0 was already installed before running Fishtest, the engine binaries were searched in the same directory as the initial installation of cutechess-cli)

maximmasiutin:
When `cutechess-cli` 1.3.0 (the latest version at the moment) was already installed before running Fishtest, the engine binaries were searched in the same directory as the initial installation of `cutechess-cli`.

For example, when `cutechess-cli` was used on Ubuntu Linux for **ARM64**, and `cutechess-cli` was already installed before installing Fishtest, the engine binaries were searched in the same directory as the initial installation of `cutechess-cli` (such as `/usr/local/bin/`) rather than in the test directory `~/fishtest/worker/testing/`). 

Without this patch, we could not run Fishtest on Ubuntu Linux for **ARM64** where Fishtest could not download the proper binary for `cutechess-cli`, and relied on the `cutechess-cli` installed before running Fishtest.

It does not matter whether you run `cutechess-cli` 1.3.0 (latest version) as a local copy or via a symlink. This later version does not accept the engine binaries without `./`. Because earlier versions of `cutechess-cli` can accept engine paths with and without `./` but later versions require `./`, I suggest to modify FishTest to call it with `./`.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -2,14 +2,13 @@ import base64
 import hashlib
 from pathlib import Path
 
+from fishtest import helpers
 from fishtest.rundb import RunDb
 from pyramid.authentication import AuthTktAuthenticationPolicy
 from pyramid.authorization import ACLAuthorizationPolicy
 from pyramid.config import Configurator
 from pyramid.events import BeforeRender, NewRequest
 from pyramid.session import SignedCookieSessionFactory
-
-from fishtest import helpers
 
 
 def main(global_config, **settings):
--- a/worker/games.py
+++ b/worker/games.py
@@ -1392,7 +1392,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
                 "-engine",
                 "name=New-" + run["args"]["resolved_new"][:10],
                 "tc={}".format(scaled_new_tc),
-                "cmd={}".format(new_engine_name),
+                "cmd=./{}".format(new_engine_name),
                 "dir=.",
             ]
             + new_options
@@ -1401,7 +1401,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
                 "-engine",
                 "name=Base-" + run["args"]["resolved_base"][:10],
                 "tc={}".format(scaled_tc),
-                "cmd={}".format(base_engine_name),
+                "cmd=./{}".format(base_engine_name),
                 "dir=.",
             ]
             + base_options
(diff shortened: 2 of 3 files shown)


vdbergh:
I am curious. Does it work correctly if you copy the cutechess binary to the testing dir,  instead of making a symlink?

vdbergh:
I would be mildly in favour of your other PRs which detect the ARM architecture correctly and download the correct cutechess binary.

maximmasiutin:
@vdbergh - yes, if you copy the file instead of creating a symlink, than adding "./" is not needed and this pull request is also not needed. However, I thought that creating a symlink is a proper way of doing things rather than copying files. The other pull request that detects the ARM architecture correctly and downloads the correct cutechess binary also solves the issue and avoids the need to add "./" proposed by this pull request. However, that ARM cutechess binary is not yet available for download from glinscott github, we have to put it there first, still, you may use the version that I compiled or compile yourself.

maximmasiutin:
@vdbergh, can you anyway run the workflows (approve workflow runs) regardless of whether this pull request will be accepted or rejected?

MinetaS:
cutechess 1.3.0 also requires explicit relative path to current directory. This looks good to me.

maximmasiutin:
@vdbergh - as @MinetaS correctly pointed out, it does not matter whether you run `cutechess-cli` 1.3.0 (latest version) as a local copy or via a symlink. This later version does not accept the engine binaries without `./`. Because earlier versions of `cutechess-cli` can accept engine paths with and without `./` but later versions require `./`, I suggest to modify FishTest to specify paths to engine binaries with `./`, otherwise it will not work with later versions of `cutechess-cli`. 

The version `cutechess-cli` binary which Fishtest is downloading is 1.2.0. However, this version is not avaiable for certain platforms such as aarm64. Therefore, if you build latest `cutechess-cli` manually for one of such platforms, you will most likely be building the latest version (1.3.0) rather than an older one (1.2.0). Therefore, the pull request to "detect the ARM architecture correctly and download the correct cutechess binary" as you suggested may not help, because, in this case, we will need to find a cutechess binary 1.2.0 for ARM.

ppigazzini (approved):
LGTM after testing on DEV with linux and windows workers.

ppigazzini:
Triggered the workers update, thank you @maximmasiutin :)

A worker PRs require to raise the worker version in `server/fishtest/api.py` and `worker/worker.py`. The latter change requires a new `sri.txt`, so I merge the worker PRs from the cli:
```bash
git switch master
git fetch -p maximmasiutin
git fetch -p upstream
git remote add maximmasiutin ssh://github.com/maximmasiutin/fishtest.git
git switch -c cutechess-path maximmasiutin/cutechess-path
nvim server/fishtest/api.py
nvim worker/worker.py
cd worker/
python3 worker.py a a --only_config --no_validation
git commit -a -m "Raise worker version to 198 (also server side)" -m "closes https://github.com/glinscott/fishtest/pull/1580"
git diff upstream/master
git rebase -i upstream/master
# squashed some commits
git diff upstream/master
git push upstream HEAD:master
git switch master
git branch -D cutechess-path
git fetch -p upstream
git fetch -p origin
git push origin master
```

ppigazzini:
https://github.com/glinscott/fishtest/commit/af23f51e28456db2d3bd678a0dd8a38f0f20da74

PR #1581, opened 2023-03-17
Added support for other Linux architectures for cutechess-cli download. It didn't support other than x86_64 which was by default

maximmasiutin:
Fishtest didn't support other Linux machine (CPU) types to download proper binary of `cutechess-cli`. It used x86_64 by default, , omitting it, to form the URL to get the file from githubusercontent.com. Therefore, for Linux x86_64 64-bit it was looking for a file name `cutechess-cli-linux-64bit.zip`, but on other CPUs like ARM64 it was still downloading the binary for x86_64 which could not be executed and gave error due to instructions incompatibility, preventing Fishtest worker from running.

This pull request adds the machine name (Python's `platform.machine()`) to the file name if it is other than x86_64. Therefore, if the machine is x86_64, the machine name is not added to the file name, to maintain compatibility with earlier naming convention. Otherwise, the machine name is added before the architecture, so the full name is like cutechess-cli-linux-aarch64-64bit.zip 

P.S. Please make the file `cutechess-cli-linux-aarch64-64bit.zip` available to download from githubusercontent.com along with `cutechess-cli-linux-64bit.zip`. You can compile this aarch64 (ARM64) version yourself or use the archive that I prepared - get it from https://www.masiutin.net/cutechess-cli-linux-aarch64-64bit.zip -- I tested it on Ampere Altra.


vondele:
This PR makes sense, however the link to the cutechess binary doesn't seem to work any longer. Do you still have a copy @maximmasiutin ?

vondele:
Actually, I see we have this repo with binaries as well https://github.com/ppigazzini/static-cutechess-cli/releases

maximmasiutin:
> This PR makes sense, however the link to the cutechess binary doesn't seem to work any longer. Do you still have a copy @maximmasiutin ?

Sorry, don't have that any longer. I compiled the current version at that time from Github. Now we could do the same, but I see you found a repository with binaries already, so we should probably take from there.

vondele:
since we moved to fastchess, this has effectively been fixed

PR #1586, opened 2023-04-08
Save CPU on database backup

maximmasiutin:
Dump first all files to one tar and then use multithreaded gzip (pigz) with fastest compression to save CPU as much as possible, then move the file to aws instead of copying it.

The "--gzip" parameter of mongodump uses gzip compression level 6 by default. In this patch we call mongodump without "--gzip" and then use `pigz` (multithreaded gzip) with compression level 1 (`--fast`) to save CPU time.

Please make sure that `pigz` is installed. On Ubunty, run 'apt install pigz`.

### DIFF
--- a/server/utils/backup.sh
+++ b/server/utils/backup.sh
@@ -4,18 +4,18 @@
 # List available backups with:
 # ${VENV}/bin/aws s3 ls s3://fishtest/backup/archive/
 # Download a backup with:
-# ${VENV}/bin/aws s3 cp s3://fishtest/backup/archive/<YYYYMMDD>/dump.tar dump.tar
+# ${VENV}/bin/aws s3 cp s3://fishtest/backup/archive/<YYYYMMDD>/dump.tar.gz dump.tar.gz
 
 # Load the variables with the AWS keys, cron uses a limited environment
 . ${HOME}/.profile
 
 cd ${HOME}/backup
 for db_name in "fishtest_new" "admin" "fishtest" "fishtest_testing"; do
-  mongodump --db=${db_name} --numParallelCollections=1 --excludeCollection="pgns" --gzip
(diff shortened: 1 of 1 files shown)


ppigazzini:
The tiny OpenVZ VPS has few limits:
 - limited HD space
 - no swap
 - monthly bandwidth 
 - the IPS bans for over usage of the CPU (it happened few times in the past)

By the way the IPS plans are now all KVM based
https://iniz.com/


We scheduled to run the backup script with `cpulimit` to spread the CPU load wrt the time and we stripped off by the script any multithread computation (eg `pigz`) or concurrent operation.

```
# Backup mongodb database every 6 hours and upload to s3
3 */6 * * * /usr/bin/cpulimit -l 50 -f -m -- sh ${UPATH}/backup.sh
```

The script keeps the `dump.tar` on the VPS as first level of backup.

A couple of years ago the process with a direct `mongodump` compression was faster than using a second application (we should test the last `mongodump` version, of course).

https://github.com/glinscott/fishtest/commit/369a93cc1cdbfbf77cb50b45d97f8c26c7a747e3
https://github.com/glinscott/fishtest/commit/15460fd701af5e0497a954a0384ba7b27ad8f7ae

maximmasiutin:
Please add "--cpu" option for `cpulimit` and/or `taskset` to limit to just one CPU. In such case, pigz will just work with one thread as a regular gzip, but anyway the option for a fast compression (gzip level 1) will help reduce CPU load instead of the default 6.

maximmasiutin:
You can add nice -n 18 to set low priority to backup, so when CPU usage of other processes will be high, the backup will take even lower CPU resources; change the command to the following:
```
3 */6 * * * /usr/bin/cpulimit -l 50 -f -m -- sh nice -n 19 ${UPATH}/backup.sh
```

(19 is lowest priority)

PR #1587, opened 2023-04-09
Serve gzip compressed data, when already available, without decompression

maximmasiutin:
Serve compressed data as is without decompression. Even if client does not support gzip compression, nginx can decompress it, saving valuable traffic and CPU cycles on the server application.

Some clients do not accept responses with the gzip encoding method. To serve such clients, add `gunzip on;` in nginx, see https://docs.nginx.com/nginx/admin-guide/web-server/compression/#enabling-decompression

1) add `application/x-chess-pgn` to `gzip_types`
2) add `gzip on;` and `gunzip on;`

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -258,7 +258,7 @@ class RunDb:
         pgn = self.pgndb.find_one({"run_id": pgn_id})
         if pgn:
             response_body = pgn["pgn_zip"]
-            accept_encoding = request.headers.get("Accept-Encoding", "")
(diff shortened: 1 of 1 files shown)


maximmasiutin:
@noobpwnftw - can you share your opinion on that?

noobpwnftw:
Probably not very useful.

ppigazzini:
Implemented in https://github.com/official-stockfish/fishtest/pull/1823

Thank you for the suggestion @maximmasiutin :)

PR #1588, opened 2023-04-09
Allow caching responses of homepage and adjust cache_time.

peregrineshahin:
This reverts to allows caching of the home page response and adjusts cache_time for 4s, I think the benefit of that became more beneficial than solving this niche issue https://github.com/glinscott/fishtest/issues/1405 .


noobpwnftw:
this cache is purely client side, reverting has nothing beneficial.

maximmasiutin:
If there is nginx `proxy_cache` / `proxy_cache_path` configured, than nginx analyzes these headers too.

PR #1590, opened 2023-04-10
Idle

maximmasiutin:


maximmasiutin:
When you create a new server and go to the list of pending users, there is an "internal server error" because the index is missing in the dictionary on the line `del idle[u["username"]]` so we have to check whether the value exist.

ppigazzini:
Please rebase to master (I wrote some commands that you could already know)
```bash
git fetch upstream master
git switch idle
git rebase -i master
# this open an editor, you can keep "pick" for the first commit and replace "pick" with "squash" for the other commits to squash them, keep only logically different commits, save and exit
# this open an editor, edit the final commit message, save and exit
```

maximmasiutin:
I did these commands but it became 8 commits ahead instead of 6. I broken my "master" branch and maximum that I can do is for it to have 2 commits ahead.

maximmasiutin:
I failed to get the master to be the same as upstream unless it will be 2 commits ahead :-(

dubslow:
```
git checkout master
git reset --hard 38a858fd5d86c63f110f3eeee96078416aaf8cda
git push -f maximmasiutin # or whatever your name for your github repo is
```

maximmasiutin:
Made normal commit, thanks to @ppigazzini and @dubslow -- https://github.com/glinscott/fishtest/pull/1592

dubslow:
ideally the fix should be in this PR, not in a new one, no reason to spam the repo

PR #1594, opened 2023-04-13
Increase itp if game count > 200k.

xoto10:
High game counts add load to the server because of the number of loops through run["tasks"]. Increase itp in longer running sprts to finish them off quicker.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -617,6 +617,15 @@ class RunDb:
         a = 3 # max bonus 1.67x
         itp *= (llr + a) / a
 
+        # Gentle bonus for high game count (up to double the average)
+        if "sprt" in run["args"]:
+            r = run["results"]
+            n = r["wins"] + r["losses"] + r["draws"]
+            x = 200_000
+            if n > x:
+                bonus = min(n / x, 2)
+                itp *= bonus
+
         # Extra bonus for most promising LTCs at strong-gainer bounds
         # if (
         #    tc_ratio >= 3.0


dubslow:
Per Discord comments, at the present time I see no reason to believe that high run count, high task count or high game count have any particular relation with server instability.

That said, I actually did consider something of this nature months ago, when I first overhauled this section -- my overhaul was targeted at reducing feedback loop latency, and a change like this servers precisely that goal. I'm okay with it such a change in principle, with a caveat and a recommended tweak:

1) no expectations that this will improve server instability

2) "gentle" to me meant that the itp smoothly changes, with no sharp changes with small variations in the context. instead of flat cutoffs, where a run suddenly gets doubled from one instant to the next, it should be a smooth upgrade in itp... say perhaps:

```
# Gentle bonus for long SPRTs (loosely double the average, at present)
x = 200_000
if n > x:
    bonus = min(n / x, 2)
    itp *= bonus
```

Keep in mind that this combines with the LLR bonus, so long passers would get up to like 3.3x bonus. But I think this is reasonable.

xoto10:
I think it's a nudge towards improved server stability, perhaps only a very minor one. Still, it's easy to do, and I think it's good to get long-running jobs out of the way a bit quicker.
I don't mind what calc we use, so I can change it to your code, or you can create a PR, I don't mind, I just created the PR to get the ball rolling. I just pinched the gentle word from the previous code lines (I interpreted it more as "small").

dubslow:
> Still, it's easy to do, and I think it's good to get long-running jobs out of the way a bit quicker.

Agreed, even if for different reasons lol. Submitters or server, win-win.

> I don't mind what calc we use, so I can change it to your code, or you can create a PR, I don't mind, I just created the PR to get the ball rolling.

Well you've taken the initiative, and so I think you deserve it to be your commit. Feel free to pinch my code, or do something a bit different, my big recommendation is simply that itp is continuous on its inputs, discontinuities are what I want to avoid.

> (I interpreted it more as "small")

I admit I wasn't exactly clear with my comment lol. I suppose I meant it as both "gentle on the server and workers" (no discontinuities in itp) and "small".

xoto10:
I've changed it to your code (more or less). I agree discontinuities are bad, I was just too lazy to decide on a formula at the time.

xoto10:
This is untested code, I don't have cutechess on my chromebook, never mind fishtest, and my python is extremely rusty so there can be obvious mistakes.

dubslow:
LGTM!

nitpicks are that I would say the comment can also say `gentle`, and perhaps the 200k part should say `# roughly double the average`, and also I'm not sure about including crashes and timelosses in `n`, but these are minor indeed, it can be merged in any of these fashions.

maximmasiutin:
Why do we need a hardcoded 200000 value rather than a relative value? Do you think that calculating relative value would be a performance killer?

xoto10:
As a manual test, we could change the throughput on [this test](https://tests.stockfishchess.org/tests/view/6432798631feee5c6d306b07) (currently at 307k games) from 50 to 60 and see how the number of cores allocated changes ...

maximmasiutin:
Can't we calculate the median value and then give priority to the games that are above median?

xoto10:
My aim here is to reduce the amount of work on the server with simple changes. I guess we could calculate more info in e.g. the `self.get_unfinished_runs_id()` loop, store it somewhere and use it in `calc_itp()`, but I don't see any need. I like to "keep it simple".
(I do have some other thoughts on `calc_itp()`, but this seemed an easy place to start.)

maximmasiutin:
> .relative to what

relative to the number of games remaining in other tests. We may calculate mean value of the games remaining in other tests of the same priority and use that value instead of hardcoded 2000000.

xoto10:
I think the way itp works is that it is all relative anyway, the itp values are just arbitrary values that get scaled against each other. So perhaps we could simply scale all sprt tests with n? Something like:
```
            n = r["wins"] + r["losses"] + r["draws"] + r["crashes"] + r["time_losses"]
            k = min((n + 100_000) / 200_000, 2)
            itp *= k
```

\* I often find going for a walk helps coding. No access to keyboard => new ideas I want to check! :D

xoto10:
Another thought. I was using the number of games as a proxy for the size of the `tasks` list, but why not simply use ... the size of the tasks list! E.g.
```
            n = len(run["tasks"])
            k = min((n + 500) / 1000, 2)
            itp *= k
```
This addresses the issue more directly, allowing for the fact that some tasks have more games than others, depending on the power of the machine they were run on.

dubslow:
eh "games" a lot more predictable for human observers. as for the bonus, it shouldn't be less than 1, and indeed i like having it 1 beyond the average test length as well. making it kick in before 2*avg will be a hard sell for me.

xoto10:
I figure fishtest programmers should understand the task list, indeed fishtest should understand the list of tasks at the bottom of a test page, and `n = len(run["tasks"])` is a bit more accurate than counting games (and less code). But I doubt the difference is material, so not a huge deal.

I like the scaling of all tests because I believe the size of the task lists are important for performance, which you're not convinced by, so np if we differ on that.

In practical terms, I think it's the 200k/300k + tests that are important, so if we can get fishtest to nudge them on a bit with this code, and suggest to users that they encourage them to finish, rather than reducing the throughput / priority, then I think we're good.

xoto10:
Could an approver update [this test](https://tests.stockfishchess.org/tests/view/6432798631feee5c6d306b07) to throughput 60 please?

Then we can see how many extra cores it gets (currently 32). I suspect changing it to 100 will give it loads of extra cores, so 60 might be better to start with. This would move it on a bit, it would be nice to get this test finished.
I've asked sg (via github) to finish his paused tunes, but I don't know if he's seen my msg.

ppigazzini:
Done

xoto10:
Yeah, seems that way. I had the idea that the itp calc tended to overshoot when changing long-running jobs due to the long time at the lower throughput but I seem to be flat wrong on that. And the point about bigger workers and small cpu count on stc is good.

I'd like to change it to 100 now, but if you guys think we should ask bigpenor to do it, I'm ok with that. Fishtest would alter core count as conditions change anyway, and we're increasing not decreasing so I'm not sure there's any need to ask.

ppigazzini:
Throughput set to 100

dubslow:
> Yeah, seems that way. I had the idea that the itp calc tended to overshoot when changing long-running jobs due to the long time at the lower throughput but I seem to be flat wrong on that. And the point about bigger workers and small cpu count on stc is good.

I was not kidding when I said I did some monitoring a couple months ago as part of my little rewrite of this function:

the main overhaul https://github.com/glinscott/fishtest/pull/1557

then while we had itp monitoring directly on the rendered web pages, i also made this adjustment:
https://github.com/glinscott/fishtest/commit/c9020a7fdb26f28f7fc6306e646ac3368fc7f7c0
and thereafter recorded a bit of data plus more less formal monitoring to verify: https://github.com/glinscott/fishtest/pull/1558 (see last comments)

im honestly not quite sure what you mean with your overshoot hypothesis. there definitely was and is some overshooting at STCs, for the reasons laid in 1558 relating to worker size, but that reduced it noticeably, any further would have other adverse effects. each worker gets assigned new runs with almost no context about the history of the run or the worker, other than not two runs in a row. it's just LLR, base TP, and large-vs-small workers. 

i did notice last week or so that viz's omega-scaling test, at 150+1.5 x8 threads, had some issues with reducing its core count when itp was suddenly lowered, but that was explained solely by the fact that each individual task took for goddamned ever at that time control lol. a day later, after all those tasks finished, it was back to typical cores/itp. nothing to do with the assignment sorting itself.

xoto10:
> ... your overshoot hypothesis ...

A memory from years ago, I need to read the code more ...

xoto10:
> PR applied to PROD by hand

Ah, I was wondering why that test had so many cores!

I will have a look at the core allocation, all comments welcome ...

xoto10:
Obv there will be different opinions on what core count is desirable.

From my pov it looks good so far, clearly more cores on the few 300k+ tests, small (or invisible) change to ~200k tests.

ppigazzini:
Enabling all the paused SPSAs has raised **smoothly** the CPU load and MongoDB bytes out, as expected.
We should look for something than can trigger a sharp x2 jump or x0.5 dip of CPU load and MongoDB bytes out.

xoto10:
People are creating more tests (now 70 running+paused from a low of ~59 earlier), my belief is this will cause more cpu load as they accumulate more games.

My counter says there are currently 6.6m games in the unfinished tests, I've seen that vary from 7.8m down to 6.2m during the last 24 hours.
~5000 cores are presumably adding ~10k stc games per minute, but less than that since some tests are ltc, and obv tests finish occasionally, removing their count from the total games. Maybe increase will be ~300k / hour? (wild guess)
If that total gets back up to 7.5m+, then I think you will see the cpu load getting high. so maybe in 3 hours time or so?

xoto10:
For I/O, I can create a tuning job with several params and run it at 5+0.05 if you like?

xoto10:
I should say that I don't mind if we scrap this if maintainers are not keen on implementing it.

For me it's nice to have a subtle nudge in the code towards finishing the big tests a bit quicker, but I don't think that will make a huge difference to server performance. User behaviour will be the key to that - now we know to finish larger tests quicker, we'll do that, instead of pausing them or reducing throughput. That's where the main gain will be imo. (If we can keep the number of started but unfinished tests a bit lower as well, that would be nice, too.)

server/fishtest/rundb.py
```diff
@@ -617,6 +617,15 @@ def calc_itp(self, run):
         a = 3 # max bonus 1.67x
         itp *= (llr + a) / a
 
+        # Gentle bonus for high game count (up to double the average)
+        if "sprt" in run["args"]:
+            r = run["results"]
+            n = r["wins"] + r["losses"] + r["draws"] + r["crashes"] + r["time_losses"]
```
vondele, line 623: I think this should be just wins + losses + draws (crashes and time losses are part of those).

vondele:
apart from the comment I made, I think that's OK, as it will reduce the number of running tests with a large number of tasks.

server/fishtest/rundb.py
xoto10: ah yes. dub had that change and I didn't work out why so only got half-way there.
Changed, squashed and force pushed

ppigazzini (approved):
Worked fine for weeks on PROD

ppigazzini:

PR #1595, opened 2023-04-13
Don't acquire rundb.run_cache_lock when generating webpages

dubslow:
This lock is acquired every time we retrieve an individual run from the DB, so it's kind of an important lock.
    
On current master, while rendering the homepage, we might acquire and release this core lock dozens of times, once for each run,
and while acquired we prevent other threads from getting runs from the DB.
    
But we only do this while retrieving the run results, but we don't even need the lock to retrieve the results! It's just that
`rundb.get_results(run)` is used by both internal DB-updating code and by the webpage rendering code, and the former may need
to save its work with the lock, while rendering webpages is very much *not* an operation that writes to the DB -- no need to lock.
    
So instead we stick to "read-only mode", so to speak, which already exists, when using `rundb.get_results(run)` to render webpages.
And for the homepage in particular, we save dozens of acquire-release cycles, which could create myriad conflicts with other threads
... but no longer.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1142,14 +1142,14 @@ def get_page_title(run):
 @view_config(route_name="tests_live_elo", renderer="tests_live_elo.mak")
 def tests_live_elo(request):
     run = request.rundb.get_run(request.matchdict["id"])
-    request.rundb.get_results(run)
+    request.rundb.get_results(run, save_run=False)
     return {"run": run, "page_title": get_page_title(run)}
 
 
 @view_config(route_name="tests_stats", renderer="tests_stats.mak")
 def tests_stats(request):
     run = request.rundb.get_run(request.matchdict["id"])
-    request.rundb.get_results(run)
+    request.rundb.get_results(run, save_run=False)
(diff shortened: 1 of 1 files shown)


dubslow:
Untested.

This patch based on vondele's profile data, which suggested that rendering the home page was by far the most time consuming thing the server did. Digging through the code, it wasn't so hard to trace the call chain

`homepage --> get_paginated_finished_runs --> rundb.get_results(run) --> rundb.buffer(run) --> with self.run_cache_lock:`

...such that rendering the home page acquires and releases that lock once for each test on that page, and this is the lock used in `rundb.get_run()`, which perhaps not coincidentally also had some very large latency spikes sometimes. 

Hopefully this PR greatly reduces the burden of rendering the home page.

vdbergh:
I wonder if what you claim is correct. get_run first checks if the results are stale. If they are not, no lock is acquired. I think the case that the results are stale is rather rare.

vdbergh:
I would imagine staleness only applies to runs that are in the cache. Since rendering the main page does not use the cache, this could be a non-issue.

PR #1599, opened 2023-04-13
Avoid long task lists

vondele:
they make some operations in fishtest much slower, and the db bigger. Additionally larger batches means less idle time at the end of the batch for workers, which is good.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -654,7 +654,7 @@ class RunDb:
         assert concurrency >= 1
         # as we have more tasks done (>250), make them longer to avoid
         # having many tasks in long running tests
-        scale_duration = 1 + (len(run["tasks"]) // 250)
+        scale_duration = 1 + min(4, len(run["tasks"]) // 250)
         games = self.task_duration * scale_duration / game_time * concurrency
         if "sprt" in run["args"]:
             batch_size = 2 * run["args"]["sprt"].get("batch_size", 1)
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -93,7 +93,7 @@ class RunDb:
             print(f"Unable to obtain the port number. Error: {self.port}", flush=True)
         self.task_runs = []
 
-        self.task_duration = 900  # 15 minutes
+        self.task_duration = 1800  # 30 minutes
         self.ltc_lower_bound = 40  # Beware: this is used as a filter in an index!
         self.pt_info = {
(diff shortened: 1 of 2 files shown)


dubslow:
Another thing I'd like to see is to ensure that the batch size is a multiple of the worker concurrency.

For instance on the SMP VLTC tests, where I have a concurrency of 3, assigning a batch size of 7 or 8 is intensely silly and wasteful, not to mention the usual concerns about half-idle batches affecting frequency/hyperthreads and thus altering the effective TC.

vondele:
doesn't matter if the batch size is sufficiently long. The variations in game lengths are much larger.

ppigazzini:
Applied by hand on PROD.

dubslow:
On the other hand, as I stated, there are some cases where the batch size is not sufficiently long.

dubslow:
This PR does have an effect on the itp balancing, significantly increasing the latency of response to changing itp, but that should be a small effect outweighed by the benefit to the server. But we should keep an eye on it.

vondele:
> Applied by hand on PROD.

so, this works, larger tasks are being generated. Right now, as the tests have partially already >1000 tasks, the tasks generated are exceptionally large. For new tests, that will essentially never have that many tasks, the behavior will be OK.

xoto10:
How about the 800,000 limit, it was preventing more cores on some jobs with this set to 1000. Is there anything we can do about that?

vondele:
I think it might need a bit of a refinement. The quadratic scaling with task number gets too much, tasks get very large. So, I'll damp that down a bit. It will allow more workers to join a task. The 800k, maybe we can indeed increase it a bit. Let me have a look.

ppigazzini:
> I think it might need a bit of a refinement. The quadratic scaling with task number gets too much, tasks get very large. So, I'll damp that down a bit. It will allow more workers to join a task. The 800k, maybe we can indeed increase it a bit. Let me have a look.

Let me know if there is a change to be reloaded on PROD (or this was only a squash)

vondele:
This is somewhat different code. It doesn't need testing on PROD, but it is different from what we had on PROD before.

vondele:
so, based now on the experience gathered, I limit the growth of the batch size with task_id to a maximum of 5. This will help when many workers land to keep tasks available. 
Complements https://github.com/glinscott/fishtest/pull/1627

ppigazzini (approved):

ppigazzini:

PR #1600, opened 2023-04-13
A hand-written deep copy avoiding copying some very large lists

vondele:
The api call api_active_runs turns out to be pretty expensive (4s if the load is small). This PR writes the strip_run in a way that avoids the deep copy of the large task lists directy. Would need a measurement on the dev server to see if it improves performance. Basically a before / after of `time wget https://dfts-0.pigazzini.it/api/active_runs`

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -73,17 +73,30 @@ def validate_request(request):
 
 
 def strip_run(run):
-    run = copy.deepcopy(run)
-    if "tasks" in run:
-        run["tasks"] = []
-    if "bad_tasks" in run:
-        run["bad_tasks"] = []
-    if "spsa" in run["args"] and "param_history" in run["args"]["spsa"]:
-        run["args"]["spsa"]["param_history"] = []
-    run["_id"] = str(run["_id"])
-    run["start_time"] = str(run["start_time"])
-    run["last_updated"] = str(run["last_updated"])
-    return run
+    # a deep copy, avoiding copies of a few large lists.
+    stripped = {}
+    for k1, v1 in run.items():
+        if k1 in ["tasks", "bad_tasks"]:
+            stripped[k1] = []
+        elif k1 == "args":
+            stripped[k1] = {}
+            for k2, v2 in v1.items():
+                if k2 == "spsa":
+                    stripped[k1][k2] = {
+                        k3: [] if k3 == "param_history" else copy.deepcopy(v3)
+                        for k3, v3 in v2.items()
+                    }
+                else:
+                    stripped[k1][k2] = copy.deepcopy(v2)
+        else:
+            stripped[k1] = copy.deepcopy(v1)
+
+    # and some string conversions
+    stripped["_id"] = str(run["_id"])
+    stripped["start_time"] = str(run["start_time"])
+    stripped["last_updated"] = str(run["last_updated"])
+
+    return stripped
 
 
 @exception_view_config(HTTPBadRequest)


ppigazzini:
I'm loading on DEV the latest DB backup to have some dozens of active runs, no worker joined to have always the same active runs during the measurements.
I decorated `strip_run` with `profileit` as well.

ppigazzini (approved):

ppigazzini:
PROD updated, nice speedup, thank you @vondele :)

vdbergh:
It seems `/api/active_runs` became much faster! That's good since I plan to use it occasionally (every few hours) in the notification system as another line of defense against stale notifications (it is still possible to create them if you know what you are doing).

vondele:
it still is one of our most expensive api's so if you use it make sure it is indeed used with some guard in place. Contrary to the home page we don't have a cache for it. Current run profile:
```
# logging from [25/Apr/2023:03:22:36 +0000] to [25/Apr/2023:05:01:07 +0000] ########
# duration (seconds)            : 5911
# calls in total                : 50000
# calls per second              : 8
# calls not reaching the backend: 105
# calls handled by the backend  : 49895
                         /tests      337    257.823      0.765      0.020      2.004
               /api/update_task    30559    147.239      0.005      0.000      0.569
                       /actions       52    104.906      2.017      0.030      8.678
               /api/active_runs      104     87.565      0.842      0.341      2.348
                      /api/beat    11425     43.907      0.004      0.000      0.551
                /tests/finished       95     41.387      0.436      0.121      2.438
                   /api/actions     3041     37.180      0.012      0.002      1.252
                   /tests/user/       43     34.209      0.796      0.444      1.158
                   /tests/view/      239     20.734      0.087      0.003      3.286
              /api/request_task      924     19.665      0.021      0.006      0.477
           /api/request_version     1775     14.160      0.008      0.000      0.635
                /api/upload_pgn      438     13.421      0.031      0.003      0.345
                     /tests/run       30      7.456      0.249      0.001      2.153
                  /api/get_elo/      288      6.153      0.021      0.008      0.325
               /tests/live_elo/      112      2.148      0.019      0.003      0.402
              /api/request_spsa      208      0.872      0.004      0.000      0.062
                           /nns       28      0.704      0.025      0.007      0.177
                        /signup        8      0.546      0.068      0.001      0.373
               /api/failed_task       57      0.371      0.007      0.001      0.027
                         /users        3      0.361      0.120      0.087      0.167
                         /login       29      0.322      0.011      0.002      0.035
                      /api/pgn/        1      0.265      0.265      0.265      0.265
                       /api/nn/       51      0.211      0.004      0.003      0.014
                    /tests/stop        1      0.170      0.170      0.170      0.170
                  /tests/stats/        1      0.149      0.149      0.149      0.149
                 /tests/approve       12      0.109      0.009      0.006      0.012
                 /users/monthly        2      0.033      0.017      0.013      0.020
                       /pending        1      0.004      0.004      0.004      0.004

```

vdbergh:
>                       /actions       52    104.906      2.017      0.030      8.678

This line is a bit surprising. Consulting the event log should not take 2 seconds on average.

vondele:
well, these are measured timings, at nginx level, so probably quite accurate. Now, we have already accumulated 500k events if I see this correctly from the number of pages. I guess if there is any query that can bypass the index, it could be slow?

vdbergh:
> well, these are measured timings, at nginx level, so probably quite accurate. Now, we have already accumulated 500k events if I see this correctly from the number of pages. I guess if there is any query that can bypass the index, it could be slow?

I think it is not so easy to make a slow query since the search stops after 25 records. Pagination can be a bit slow (e.g. https://tests.stockfishchess.org/actions?page=22551 ) but I do not thing that people are doing that regularly.

vondele:
so we have the log of slow mongo operations (/var/log/mongodb/mongod.log)
```json
{"t":{"$date":"2023-04-25T07:19:10.339+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn8","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"aggregate":"actions","pipeline":[{"$match":{"action":{"$nin":["system_event","update_stats","dead_task"]}}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{},"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { action: 1, _id: -1 }","keysExamined":563763,"docsExamined":0,"cursorExhausted":true,"numYields":563,"nreturned":1,"queryHash":"60F6A5AC","planCacheKey":"55633155","queryFramework":"classic","reslen":133,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":565}},"Global":{"acquireCount":{"r":565}},"Mutex":{"acquireCount":{"r":2}}},"storage":{},"remote":"127.0.0.1:35618","protocol":"op_msg","durationMillis":783}}
{"t":{"$date":"2023-04-25T07:19:10.716+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn18","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"aggregate":"actions","pipeline":[{"$match":{"action":{"$nin":["system_event","update_stats","dead_task"]}}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{},"lsid":{"id":{"$uuid":"0adbcff4-ea3c-42ec-9006-f74607242562"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { action: 1, _id: -1 }","keysExamined":563763,"docsExamined":0,"cursorExhausted":true,"numYields":563,"nreturned":1,"queryHash":"60F6A5AC","planCacheKey":"55633155","queryFramework":"classic","reslen":133,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":565}},"Global":{"acquireCount":{"r":565}},"Mutex":{"acquireCount":{"r":2}}},"storage":{},"remote":"127.0.0.1:6662","protocol":"op_msg","durationMillis":778}}
{"t":{"$date":"2023-04-25T07:19:15.073+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn8","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"find":"actions","filter":{"action":{"$nin":["system_event","update_stats","dead_task"]}},"sort":{"_id":-1},"skip":563750,"limit":25,"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { _id: 1 }","keysExamined":947595,"docsExamined":947595,"fromMultiPlanner":true,"cursorExhausted":true,"numYields":1231,"nreturned":10,"queryHash":"1D081B0E","planCacheKey":"11DC8B40","queryFramework":"classic","reslen":1931,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":1232}},"Global":{"acquireCount":{"r":1232}},"Mutex":{"acquireCount":{"r":1}}},"storage":{},"remote":"127.0.0.1:35618","protocol":"op_msg","durationMillis":4700}}
{"t":{"$date":"2023-04-25T07:19:22.115+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn8","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"aggregate":"actions","pipeline":[{"$match":{"action":{"$nin":["system_event","update_stats","dead_task"]}}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{},"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { action: 1, _id: -1 }","keysExamined":563763,"docsExamined":0,"cursorExhausted":true,"numYields":563,"nreturned":1,"queryHash":"60F6A5AC","planCacheKey":"55633155","queryFramework":"classic","reslen":133,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":565}},"Global":{"acquireCount":{"r":565}},"Mutex":{"acquireCount":{"r":2}}},"storage":{},"remote":"127.0.0.1:35618","protocol":"op_msg","durationMillis":835}}
{"t":{"$date":"2023-04-25T07:19:40.692+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn8","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"aggregate":"actions","pipeline":[{"$match":{"action":{"$nin":["system_event","update_stats","dead_task"]}}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{},"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { action: 1, _id: -1 }","keysExamined":563763,"docsExamined":0,"cursorExhausted":true,"numYields":563,"nreturned":1,"queryHash":"60F6A5AC","planCacheKey":"55633155","queryFramework":"classic","reslen":133,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":565}},"Global":{"acquireCount":{"r":565}},"Mutex":{"acquireCount":{"r":2}}},"storage":{},"remote":"127.0.0.1:35618","protocol":"op_msg","durationMillis":753}}
```

vondele:
something likes querying the dead_tasks.

vondele:
no, that's right now in a quiet phase.

vondele:
In some of these events there seems to be a mongo hint / error message:
```json
{"t":{"$date":"2023-04-25T07:18:42.590+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn8","msg":"Slow query","attr":{"type":"command","ns":"fishtest_new.actions","command":{"find":"actions","filter":{"action":{"$nin":["system_event","update_stats","dead_task"]}},"sort":{"_id":-1},"skip":563750,"limit":25,"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"},"planSummary":"IXSCAN { _id: 1 }","keysExamined":947594,"docsExamined":947594,"fromMultiPlanner":true,"replanned":true,"replanReason":"cached plan was less efficient than expected: expected trial execution to take 711 works but it took at least 7110 works","cursorExhausted":true,"numYields":1239,"nreturned":10,"queryHash":"1D081B0E","planCacheKey":"11DC8B40","queryFramework":"classic","reslen":1931,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":1240}},"Global":{"acquireCount":{"r":1240}},"Mutex":{"acquireCount":{"r":1}}},"storage":{"data":{"bytesRead":1563166,"timeReadingMicros":34606}},"remote":"127.0.0.1:35618","protocol":"op_msg","durationMillis":4956}}
```

`"docsExamined":947594,"fromMultiPlanner":true,"replanned":true,"replanReason":"cached plan was less efficient than expected: expected trial execution to take 711 works but it took at least 7110 works","cursorExhausted"`

vdbergh:
Actually it is _not_ about dead tasks.  This is the query (see the `$nin`):
```json
{"find":"actions","filter":{"action":{"$nin":["system_event","update_stats","dead_task"]}},"sort":{"_id":-1},"skip":563750,"limit":25,"lsid":{"id":{"$uuid":"b42ea8ce-2a0d-4822-b35b-d05a745ea4e3"}},"$db":"fishtest_new"}
```
The `"skip":56375` is interesting. This is for pagination, which is indeed slow.

Other queries seems to contain a `$sum` aggregation which is necessary to count the total number of events (this in turn is necessary for pagination to work correctly).

vondele:
I looked for the error message, this seems to be explained here https://dba.stackexchange.com/questions/226405/why-mongo-is-choosing-the-wrong-index-execution-plan

vdbergh:
> so doing some counting on the log the one message appearing 50% of the time is `actions","command":{"aggregate":"actions","pipeline":[{"$match":{"action":{"$nin":["system_event","update_stats","dead_task"]}}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{}`

Yes this is counting the total number of documents satisfying the query (see the `$group` and `$sum`). This is necessary for pagination to work correctly.

I am thinking that the index that is designed for this is wrong (so mongo does not use it).
```python
 db["actions"].create_index([("action", ASCENDING), ("_id", DESCENDING)])
```
I think it should be
```
python
 db["actions"].create_index([("_id", DESCENDING), ("action", ASCENDING)])
```
Then mongodb could just scan the index and filter according to "action" (without having to retrieve the documents).

vdbergh:
So the counting takes 800ms. I think we can speed this up a lot by counting only up to a certain _id and then add a constant.

vondele:
might be, the slowest queries are the one with the error message however `"durationMillis":4956`

vdbergh:
> might be, the slowest queries are the one with the error message however `"durationMillis":4956`

Yes that's the one requesting a page at the end of the collection. I think this can be sped up by altering the compound index, as I explained above.

dav1312:
> we can speed this up a lot by counting only up to a certain _id and then add a constant.

Aren't you assuming that the database will have that _id and X amounts of entries? What if I want to run fishtest locally for development?

noobpwnftw:
So, won't it be easier to just write those numbers down somewhere and update the counters with each add/delete/edit?

vdbergh:
> So, won't it be easier to just write those numbers down somewhere and update the counters with each add/delete/edit?

Yes an option could be to do it in actiondb at every insert (the event log is immutable so there is no delete/edit).

However I would prefer not to pollute actiondb  with special case code to help with certain queries.

noobpwnftw:
While you'd prefer a special case index to do just that?

vdbergh:
> While you'd prefer a special case index to do just that?

Where did I say that? This has nothing to do with indices. 

What I am saying is that to know the total event count it is sufficient to count the events that happened after server startup. We could keep counters as you suggest, but this is unnecessary since just querying the db will be very fast.

noobpwnftw:
What is unnecessary is having to create indexes for pagination. In rundb there are indexes for green, yellow, tc, some combination of them but not all, sooner or later those indexes will be larger than the data itself and they are mostly there to make pagination work or if someone looks at page 10000, but are they really worth the effort than doing COLLSCAN anyways?

vdbergh:
> What is unnecessary is having to create indexes for pagination. In rundb there are indexes for green, yellow, tc, some combination of them but not all, sooner or later those indexes will be larger than the data itself and they are mostly there to make pagination work or if someone looks at page 10000, but are they really worth the effort than doing COLLSCAN anyways?

Now you are talking about the pagination itself and not the document count. So another issue. 

There is already an index that is supposed to help with pagination but it has the wrong definition so mongodb correctly ignores it (see the above discussion). So I was proposing to fix the index - not to create a new one.

Personally I think that correct pagination, also for high pages numbers, gives a professional feel to fishtest.

noobpwnftw:
Unfortunately professionals don't double cache DB content in RAM, so maybe it's just for the feel.

vdbergh:
> As for counting at startup and stay in memory, I completely ignored it due to bad practice, it limits active server instances that can write actions to one, or they won't be counting events written by others. 

It does no such things....  Of course determining the event count at startup needs to be done with a lock but that's it. After that there is no impact whatsoever.

noobpwnftw:
So what about having multiple server instances for API and webpage views?

vdbergh:
> So what about having multiple server instances for API and webpage views?

No problem. Each individual instance will have its own count (which do not have to be all the same).

noobpwnftw:
Then the total count seen on the webpage may not match the actual count of records in the database, how does this work with pagination? Some pages in the end disappearing?

vdbergh:
> Then the total count seen on the webpage may not match the actual count of records in the database, how does this work with pagination? Some pages in the end disappearing?

I really don't understand what you arguing about. 
```
total_count=count_strictly_after_X+count_before_X
```
At startup we fix X and determine count_before_X (under lock to account for multithreading). 
Afterwards we query the actiondb for count_strictly_after_X.

noobpwnftw:
I really don't understand how doing so improves anything, so for a document count without queries, there are already indexes for that, for a document count with query, without adding more indexes but having a `count_strictly_after_X` does what exactly? Do you not include those in the results?

vdbergh:
> I really don't understand how doing so improves anything, so for a document count without queries, there are already indexes for that, for a document count with query, without adding more indexes but having a `count_strictly_after_X` does what exactly? Do you not include those in the results?

Currently a document count without queries (except for the standard exclusions: dead tasks and system events) seems to take 800ms (it has to scan the whole index of > 500000 entries). This can be reduced to 0ms. Vondele's profile https://github.com/glinscott/fishtest/pull/1600#issuecomment-1521155630 shows that this makes a difference.

noobpwnftw:
Look, so that's with queries in the DB perspective, it'll have to go through every document to see whether they match, now, I agree that doing it once will reduce the scope(and you can even do incremental updates) for your next queries, but it will not fix the fact that they are slow to lookup due to lacking index on the actual queries. Excluding half-decent solutions like DB or runtime counters, can we actually move those exclusions to a separate database?

vdbergh:
> Look, so that's with queries in the DB perspective, it'll have to go through every document to see whether they match, now, I agree that doing it once will reduce the scope(and you can even do incremental updates) for your next queries, but it will not fix the fact that they are slow to lookup due to lacking index on the actual queries. Excluding half-decent solutions like DB or runtime counters, can we actually move those exclusions to a separate database?

Look,  Vondele has pinpointed two bottlenecks by logging the db  and I have proposed correct solutions for both.  If other problems appear they can be dealt with when they appear. I don't think there is a point continuing this discussion.

PR #1603, opened 2023-04-14
Increase cache time for user creditials

vondele:
increases the time to 2min, which should halve the time spent in authetication. Authentication happes quite often, i.e. each worker connects every minute, and is one of the more expensive calls for the critical pserve.

The PR also clears the cache on writes to the database, so that e.g. a password change takes effect immediately.

It appears possible to just use the cache independent of the age of the entry, as long as the cache is cleared on every db write. This has not yet been implemented, but seems straightforward.


server/fishtest/userdb.py
```diff
@@ -92,6 +93,7 @@ def create_user(self, username, password, email):
                 }
             )
             self.last_pending_time = 0
+            clear_cache()
```
xoto10, line 96: is this needed?

vondele: the function writes to the db, so it makes it safe. If we assume that create user does always create a new user, that new user can't be in the cache yet.

vondele: since it actually checks, it is not needed.

xoto10: I was thinking that the new user will be added to the cache if/when it is searched for, no need to clear all the other users that are already in the cache

vondele: right, just wanted to avoid the case where create_user would create an existing user with a new password, but that can't happen as the check is explicit.

ppigazzini:
Added by hand on PROD.

The cache invalidations seem to be safe.

```bash
git grep -n save_user
server/fishtest/userdb.py:100:    def save_user(self, user):
server/fishtest/views.py:505:        request.userdb.save_user(user_data)
server/fishtest/views.py:733:            request.userdb.save_user(u)
server/tests/test_api.py:103:        self.rundb.userdb.save_user(user)
server/tests/test_api.py:483:        self.rundb.userdb.save_user(user)
server/tests/test_users.py:73:        self.rundb.userdb.save_user(user)
server/utils/delta_update_users.py:213:            rundb.userdb.save_user(u)
```

`delta_update_users.py` calls `save_user()` only in this dead code (that could be dropped), used to clean the admin group (in the past all the users had this group, appended wrongly many times).
https://github.com/glinscott/fishtest/blob/90029373f6a20b1b6106afbd32acc11c2e3b3318/server/utils/delta_update_users.py#L207-L213

`views.py` when creating a user or resheduling a test (update of the git repo)

https://github.com/glinscott/fishtest/blob/0df452710c472cc23e8510c07d05788e17318d2f/server/fishtest/views.py#L505
https://github.com/glinscott/fishtest/blob/0df452710c472cc23e8510c07d05788e17318d2f/server/fishtest/views.py#L731-L733

vondele:
the only thing that is not 100% safe is that the DB is accessible outside of the Class, so if somebody adds code that directly access the DB, the cache might be wrong. The CI is one place doing that.

ppigazzini (approved):

ppigazzini:
> It appears possible to just use the cache independent of the age of the entry, as long as the cache is cleared on every db write. This has not yet been implemented, but seems straightforward.

All these cache are a show stopper for the load balancing. Without a major refactoring of the code (eg memcached), we can use another `pserve` instance only adding another upstream in nginx config and distributing the costly api on two different upstream.

xoto10:
wherever possible we shouldn't be hand-coding caching, the db should be capable of serving data fast enough in most cases.

ppigazzini (approved):
Every pserve has his copy of the users cache, so we could have a synchronization problem.
Not an issue introduced by this PR.

PR #1604, opened 2023-04-14
Build main page in background.

vdbergh:
The build happens every 20 seconds.

I am not sure it will be faster since I have to deepcopy the cached data since it is being modified by the client.


dubslow:
I believe it should not be necessary for `tests`  to not need to copy `last_tests` when returning (same as master), altho it is admittedly not a big deal since there's only a dozen-or-so keys in this dict.

(Setting `machines_shown` is done on every fresh request anyways, so that doesn't affect the fact that `last_tests` need not be copied.)

vdbergh:
I have converted this to draft since there is a serious bug.

vdbergh:
Closing this since it should be done better/differently.

PR #1606, opened 2023-04-15
Background homepage build.

vdbergh:
The build happens every 20s.


vdbergh:
I have now used a much better approach.

peregrineshahin:
> The build happens every 20s.
I think that previous profiling shows that Homepage building took 6s at most and now it seems to take much less, could you please enlighten me on why increasing the time would be an improvement?

vondele:
I have a question about the `clear_cache()` ... the function is still there, and called from cached_flash(). This means that approving a test doesn't necessarily trigger a homepage rebuild (so the test still is pending). 

This is actually one of the things that doesn't seem to be working since the move of the homepage to the new pserve, i.e. when approving a test the view of the main page doesn't show that change.

vdbergh:
I leave the choice of an appropriate refresh period up for discussion. The point of this PR is to show that it possible to build the main page in the background.

One thing that should be taken into account is that the build now happens after _every_ refresh period and no longer on demand.

vdbergh:
> I have a question about the `clear_cache()` ... the function is still there, and called from cached_flash(). This means that approving a test doesn't necessarily trigger a homepage rebuild (so the test still is pending).

Hmm. I am not sure how this cached_flash does its thing but I agree that when there is a popup the main page should not come from the cache.

> 
> This is actually one of the things that doesn't seem to be working since the move of the homepage to the new pserve, i.e. when approving a test the view of the main page doesn't show that change.

This should be a missing buffer(run,True).

vdbergh:
> > This is actually one of the things that doesn't seem to be working since the move of the homepage to the new pserve, i.e. when approving a test the view of the main page doesn't show that change.
> 
> This should be a missing buffer(run,True).

Actually it is not clear to me why this would happen now and not before.

vdbergh:
Unfortunately I do not see a simple way to solve this. Note that it is not an issue of this PR but a general issue. But of course the longer the cache time, the worse it becomes. Some approaches I see:

-  the primary instance to writes some indicator in the db that the homepage needs to be refreshed, which is then checked by the secondary instance (I don't like this).

- There is a clear_cache api served by the secondary instance, which is then invoked by the primary instance (this I like more).

Other suggestions?

vondele:
I was thinking about the db solution, but I like the refresh api more.

vdbergh:
I am thinking that while the api solution seems to be more elegant, the error handling becomes a spaghetti (the api will want to show its own flash messages which then invokes clear_cache again etc...).

So the db solution seems more realistic.

vondele:
well, we don't need to connect it to the flash message. 
The api could be 'ensure_homepage_update' and a call to this could block until the homepage is updated. 
We could call that at any point we make a critical update.

vdbergh:
But what if the ensure_homepage_update api call fails (not unrealistic because of firewalls)? I think it would be a bad idea to make it fail silently.  The appropriate way to show the error is a flash message (we have flash messages for all kinds of errors). Of course we can give an optional argument to cached_flag to avoid recursive behavior. Maybe not so difficult after all.

vondele:
I would let it fail silently, or to the fishtest log, it is not something the user can act on.

vdbergh:
There are other errors the user cannot act on (e.g. with nn_upload). But the user can act still act on it by posting on Discord or making an issue about it.

I general I believe that things should never fail silently.

vondele:
actually, concerning the primary and secondary instance... if the main page would use two instances to serve the main page (as  @ppigazzini had while testing), the idea would still not necessarily work, i.e. you could signal one but not the other instance.

vondele:
right now, having two instances is probably not very important..

vondele:
actually a db based implementation would work, also with 2 instances, but not the api one.

vdbergh:
> actually a db based implementation would work, also with 2 instances, but not the api one.

Well the db based solution also needs some thought with more than one secondary instance. I would make the primary instance write a timestamp to the db and the secondary instances should check that this timestamp has changed since the last refresh.

vdbergh:
I will try to hack something tonight or tomorrow (in a separate PR).

vondele:
concerning the 20s wait time. What about measuring the time needed to build the home page, and sleep for e.g. 2x that time. In this way, the load due to building on that process is about 30%. If we have a fast build, we will have up-to-date data displayed.

ppigazzini:
With all the cache distributed in the code, it's not very safe to set a load balancer.
It's more simple to set another upstream/pserve to server the home page and the history of tests:

```nginx
upstream backend_tests {
    server 127.0.0.1:6543;
}

upstream backend_hist {
    server 127.0.0.1:6544;
}

upstream backend_all {
    server 127.0.0.1:6545;
}
... 

       location = /tests {
            proxy_pass http://backend_hist;
        }
```

vondele:
I think the main advantage of this scheme is that there will never be a long wait for the viewer, it will always serve a page immediately if one lands on the main page.

vdbergh:
As @vondele points out, the idea is to improve the user experience. It seems this cannot be achieved by simply devoting more instances to homepage building since the many required db accesses make homepage building intrinsically slow.

In the latest push I incorporated Vondele's nice idea for auto load balancing the homepage build (see https://github.com/glinscott/fishtest/pull/1606#issuecomment-1509853135 ) but not yet #1607 since this is a fix for an independent issue which also affects master.

ppigazzini:
I simply wanted to inform you not to start with the DB synchronization work because we have some caches scattered in the code. To start a load balancer we should get rid of all those caches, the simpler solution to use another `pserve` is to use nginx.

vdbergh:
@ppigazzini But even when using the nginx cache the user would sometimes have to wait for a homepage rebuild. The aim of this PR is to avoid this entirely. I thought it should nice to test how users would react to this experience, but I do not feel strongly about it.

The synchronization is already done in #1607 . It is necessary also for master. Of course if the plan is to eliminate caching altogether inside Fishtest then it is not necessary (but Fishtest should be able to invalidate the nginx cache).

ppigazzini:
nginx cache?

I partitioned the URIs on 3 `pserve`.
6543 - URIs that need the cached runs
6544 - API (and catch all)
6545 - home page and the history of the tests

I have nothing against this PR, I was sharing some thoughts on the shortcomings of using a load balancer.

```c
14617 mongodb   20     3475052 700120  59316 S 215.3 13.4   1:20.47  `- /usr/bin/mongod --config /etc/mongod.conf
14665 fishtest  20     1374812 331460  33784 S   4.3  6.3   1:46.36  `- /home/fishtest/fishtest/server/env/bin/python3 /home/fishtest/fishtest/server/env/bin/pserve production.ini http_port=6543
14666 fishtest  20     1216532 354328  32700 S  57.8  6.8   1:15.34  `- /home/fishtest/fishtest/server/env/bin/python3 /home/fishtest/fishtest/server/env/bin/pserve production.ini http_port=6544
14667 fishtest  20     1631648 788328  32776 S  30.2 15.0   2:49.70  `- /home/fishtest/fishtest/server/env/bin/python3 /home/fishtest/fishtest/server/env/bin/pserve production.ini http_port=6545
```

I'm not against this PR, I was sharing some thoughts on the shortcomings of using a load balancer.

The `pserve` serving the home page needs a lot of RAM.
With a load balancer we have: 2 `pserve` using a lot of RAM and few unsynchronized caches (a problem when clearing the users cache), see at examples:
https://github.com/glinscott/fishtest/blob/90029373f6a20b1b6106afbd32acc11c2e3b3318/server/fishtest/userdb.py#L17-L19
https://github.com/glinscott/fishtest/blob/90029373f6a20b1b6106afbd32acc11c2e3b3318/server/fishtest/userdb.py#L51-L54

The flag_cache is the only one coded to be used with a load balancer.
https://github.com/glinscott/fishtest/blob/90029373f6a20b1b6106afbd32acc11c2e3b3318/server/fishtest/api.py#L274-L284

xoto10:
> So the db solution seems more realistic.

Sure, the whole point of a db is it's a central data repository for others to look at.

xoto10:
I don't know if this could be useful at some point, but one approach to the db access that python makes relatively easy is to have one process doing (most of the) db access and use Queue to send requests and receive data to/from that process. This makes caching easy because only this one process is doing (most of) the db access and any intensive cpu calcs are being done in other processes helping to spread the cpu load.
Just a thought.

vdbergh:
> > So the db solution seems more realistic.
> 
> Sure, the whole point of a db is it's a central data repository for others to look at.

Well a db is for data. It should not be for control flow.

xoto10:
Perhaps so.

I'm a database programmer, so I tend to look at things from a data pov.
"Time of last update to data underlying cache" sounds like a piece of data to me. Easy for multiple processes to overwrite and multiple processes to read, with no worry about contention.

vdbergh:
The homepage takes only 2s to build now. So closing this.

PR #1607, opened 2023-04-15
Ensure that the homepage is properly refreshed when served from another instance

vdbergh:
We add a db collection "comms" which currently contains one record {'type': 'ensure_refresh', 'timestamp': `<timestamp>`}. The function clear_hash() updates the timestamp.

The function check_db_refresh() may be used to verify if the timestamp of the last refresh is different from the timestamp in the comms collection.

If it is then a refresh should be performed.

Fixes https://github.com/glinscott/fishtest/pull/1606#issuecomment-1509706955

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -87,6 +87,7 @@ class RunDb:
         self.nndb = self.db["nns"]
         self.runs = self.db["runs"]
         self.deltas = self.db["deltas"]
+        self.comms = self.db["comms"]
         self.port = get_port()
         if self.port < 0:
             print(f"Unable to obtain the port number. Error: {self.port}", flush=True)
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -28,18 +28,41 @@ from requests.exceptions import ConnectionError, HTTPError
 HTTP_TIMEOUT = 15.0
 
 
-def clear_cache():
+def clear_cache(request):
     global last_time, last_tests
     building.acquire()
     last_time = 0
     last_tests = None
+    request.rundb.comms.replace_one(
+        {"type": "ensure_refresh"},
+        {"type": "ensure_refresh", "timestamp": datetime.datetime.utcnow()},
+        upsert=True,
+    )
     building.release()
 
 
 def cached_flash(request, requestString):
-    clear_cache()
+    clear_cache(request)
     request.session.flash(requestString)
     return
+
+
+last_refresh = None
+
+
+def check_db_refresh(request):
+    global last_refresh
+    refresh_record = request.rundb.comms.find_one({"type": "ensure_refresh"})
+    if refresh_record is not None:
+        new_refresh = refresh_record["timestamp"]
+        if new_refresh != last_refresh:
+            last_refresh = new_refresh
+            ret = True
+        else:
+            ret = False
+    else:
+        ret = False
+    return ret
 
 
 def pagination(page_idx, num, page_size):
@@ -1470,6 +1493,10 @@ def tests(request):
         return get_paginated_finished_runs(request)
 
     global last_tests, last_time
+    if check_db_refresh(request):
+        # A clear_cache happened in another instance
+        last_time = 0
+        last_tests = None
     if time.time() - last_time > cache_time:
         acquired = building.acquire(last_tests is None)
         if not acquired:


vdbergh:
I made this draft since there is a logical error.

vdbergh:
Undrafted.

vdbergh:
Hmm it seems master has a race condition.
https://github.com/glinscott/fishtest/blob/0df452710c472cc23e8510c07d05788e17318d2f/server/fishtest/views.py#L1448-L1453

Before returning, another thread can have set `last_tests` to `None`. The solution is to make a copy of `last_tests` while the lock is still being held.

In general I think the whole caching code is a bit convoluted.

vdbergh:
Although the issue https://github.com/glinscott/fishtest/pull/1606#issuecomment-1509706955 is real, people do not seem to be bothered by it. So closing.

vondele:
actually, I think this is pretty important. It is quite confusing that after approving a test it still shows as being not approved.

vdbergh:
Ok I'll reopen it then.

ppigazzini:
Rebased on master to trigger the CI

ppigazzini:
In the last couple of months, we have switched from a load balancing configuration, with several backend instances serving the same URLs, to a strictly reverse proxy server configuration, with every backend instance serving his own subset of URLs.
The home page is served only by one backend instance.
```nginx
map $uri $backends {
    /upload_net/                             backend_8000;
    /tests                                   backend_6544;
    ~^/api/(actions|active_runs|calc_elo)    backend_6545;
    ~^/api/(download_pgn|download_pgn_100)/  backend_6545;
    ~^/tests/(finished|user/)                backend_6545;
    ~^/(users|actions)/                      backend_6545;
    ~^/tasks/                                backend_6545;
    ~^/tests/view/                           backend_6543;
    ~^/(api|tests)/                          backend_6543;
    default                                  backend_6544;
}
```
I'm closing the PR, please reopen if I missed something.

vdbergh:
Are you sure? I think this PR was supposed to alert the main page to changes of the DB (e.g. approval) which happened in another instance.

ppigazzini:
Thanks for the explanation, I will test the behavior on DEV without and with PR.

vondele:
yes, approval of pending tests is the prime use case.
Right nowthe approved test is still shown as pending, which can be confusing.

ppigazzini:
Let me know if the process is correct to replicate the issue on DEV:
- opened browser_0 on home page
- opened browser_1 to create/approve the runs
- created 10 runs on browser_1
- for every run:
  - approve the run on broswer_1
  - refresh immediately the home page in browser_0 and check the list of pending runs

I always viewed the up to date list of pending runs, either with or without the PR, now I'm reloading a clean DB backup on DEV to redo the tests from scratch.

However, on DEV I viewed the problem only with several instances serving the home page, perhaps the load of PROD is biasing the behavior.

vondele:
hmm, for whatever reason right now prod works correctly, but that's recent behavior.

ppigazzini:
I'm very used to create and approve runs on DEV, I never viewed the issue with only one instance serving the home page.
I always viewed the issue with more than one instance serving the home page, though.
Now DEV is running on master.
Feel free to suggest tests to trigger the issue.

vdbergh:
To be honest, I do not really like this PR, so if it is solved otherwise then I am glad it is not necessary.

ppigazzini:
@vdbergh yesterday we have moved the fishtest repository under the Stockfish organization, GitHub is redirecting the `glinscott/fishtest` to the `official-stockfish/fishtest`, but I have updated my remotes/scripts.

vondele:
Would be nice to understand why it works now, it definitely was the case on prod until recently. It was the api to approve the calls that runs on a different instance than the one that serves the homepage that was the issue IIRC.

ppigazzini:
I create/approved 30 runs on DEV (master) in browser_1, the approved run always disappeared from the pending list either in the browser_1 when auto returning at the home page, and in the browser_0 (kept on the home page) after a fast refresh.

vondele:
similar experience with 3 tests on prod. This is new though.

vondele:
Ok, captured an event that shows this is not yet fixed on prod:

image

Shows, the 'run approved' message, while at the same time showing the 'Pending approval' of the same test.

No other events of relevance (I think) in the same time window:

image

ppigazzini:
Little issue, this PR uses the function `datetime.datetime.utcnow()` that is deprecated in python 3.12, see
https://docs.python.org/3.12/library/datetime.html?highlight=utcnow#datetime.datetime.now
> Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling `datetime.now(timezone.utc)`."

With the switch to the new python we will have a flood of warnings in the server logs and some warnings in the worker as well.
Yesterday I started replacing the deprecated function, but this triggered errors when comparing offset-naive and offset-aware datetimes. I have two alternatives:
- upgrade the datetime when embedded in historical data
```python  
   diff = datetime.now(timezone.utc) - date.replace(tzinfo=timezone.utc)
```
- update with a script all the datetimes in the DB (fishtest down for maintenance)
https://github.com/official-stockfish/fishtest/commit/704842caa022e2d25bccaf6a84b7322fc227b0e3

PR #1608, opened 2023-04-17
Increase batch size

vondele:
api_update_task and api_request_spsa are among the most called api functions, they limit the scalability of the framework.

With the large number of workers, we have frequent updates also with these new defaults. LTC SPRT tests update at the same rate, STC SPRT tests slower.

SPSA tests load the framework quickly, increase the batch size. This also make sense with highly concurrent workers that otherwise update very frequently.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1286,6 +1286,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
     factor = 1328000 / base_nps
 
     # Adjust CPU scaling.
+    scaled_tc_ltc, tc_limit_ltc = adjust_tc("60+0.6", factor)
     scaled_tc, tc_limit = adjust_tc(run["args"]["tc"], factor)
     scaled_new_tc = scaled_tc
     if "new_tc" in run["args"]:
@@ -1312,7 +1313,8 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
         tc_limit *= 2
 
     while games_remaining > 0:
-        batch_size = games_concurrency * 4  # update frequency
+        # update frequency: every 4 games at LTC, or a similar time interval at other TCs
+        batch_size = games_concurrency * 4 * max(1, round(tc_limit_ltc / tc_limit))
 
         if spsa_tuning:
             games_to_play = min(batch_size, games_remaining)
--- a/worker/sri.txt
+++ b/worker/sri.txt
@@ -1,1 +1,1 @@
-{"__version": 199, "updater.py": "PHFUVXcoxBFiW2hTqFN5q5WdAw2pK8uzFC7hyMUC3cLY5bGZPhL8TBGThtqDmcXd", "worker.py": "uFRfeWMhIxsSBppqLLWpNKfYd6VAq8MGRZUjtu+XKWE1RhM4cACqLhXA9jP+92Z1", "games.py": "wzzUVbeYyxLGRQTxx0RyCh67lHVadBe6HIQi84GM2W27cF3+aJIXLZkZmysfxTv5"}
+{"__version": 199, "updater.py": "PHFUVXcoxBFiW2hTqFN5q5WdAw2pK8uzFC7hyMUC3cLY5bGZPhL8TBGThtqDmcXd", "worker.py": "uFRfeWMhIxsSBppqLLWpNKfYd6VAq8MGRZUjtu+XKWE1RhM4cACqLhXA9jP+92Z1", "games.py": "0kqs7uzEyvg9AlU/55YvF7SPzohdPsIlkNf3zFWsRWxK5gJDdEZIBg0BZXRAIhsp"}
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -781,7 +781,8 @@ def validate_form(request):
     # Integer parameters
 
     if stop_rule == "sprt":
-        sprt_batch_size_games = 8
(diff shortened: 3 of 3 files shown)


worker/games.py
```diff
@@ -1312,7 +1313,8 @@ def make_player(arg):
         tc_limit *= 2
 
     while games_remaining > 0:
-        batch_size = games_concurrency * 4  # update frequency
+        # update frequency: every 4 games at LTC, or a similar time interval at other TCs
+        batch_size = games_concurrency * 4 * max(1, round(tc_limit_ltc / tc_limit))
```
ppigazzini, line 1317: the factor works only for TC < LTC, if it's intended better to improve the comment
```python
# update frequency: every 4 games at LTC, or a similar time interval at lower TCs
```

worker/games.py
```diff
@@ -1286,6 +1286,7 @@ def parse_options(s):
     factor = 1328000 / base_nps
 
     # Adjust CPU scaling.
+    scaled_tc_ltc, tc_limit_ltc = adjust_tc("60+0.6", factor)
```
ppigazzini, line 1289: we need only `tc_limit_ltc`:
```python
_, tc_limit_ltc = adjust_tc("60+0.6", factor)
```

worker/sri.txt
```diff
@@ -1 +1 @@
-{"__version": 200, "updater.py": "PHFUVXcoxBFiW2hTqFN5q5WdAw2pK8uzFC7hyMUC3cLY5bGZPhL8TBGThtqDmcXd", "worker.py": "bYP5sBennlrsURZiN0CwotQLYapRMDL4CIG/2sfyzQL/1fuzl1ZsQDdawOM/b92j", "games.py": "wzzUVbeYyxLGRQTxx0RyCh67lHVadBe6HIQi84GM2W27cF3+aJIXLZkZmysfxTv5"}
+{"__version": 200, "updater.py": "PHFUVXcoxBFiW2hTqFN5q5WdAw2pK8uzFC7hyMUC3cLY5bGZPhL8TBGThtqDmcXd", "worker.py": "bYP5sBennlrsURZiN0CwotQLYapRMDL4CIG/2sfyzQL/1fuzl1ZsQDdawOM/b92j", "games.py": "wzzUVbeYyxLGRQTxx0RyCh67lHVadBe6HIQi84GM2W27cF3+aJIXLZkZmysfxTv5"}
```
ppigazzini, line 1: I fixed the conflict from GitHub GUI to be able to use the PR in the test workers.

ppigazzini (approved):
Tests on DEV LGTM.

To be merged by hand because the required version raise -> new sri.txt
I can do myself the requested changes after your approval.

ppigazzini (approved):

ppigazzini:

xoto10:
How often are updates in clock time with this change? About every 10 mins or so??
(Just wondering, not critical. Thought I'd seen a comment somewhere giving a number and looked here)

vondele:
still rather quickly. If I understand the code correctly, a worker running on a SPRT test will update a task once it has computed 32 games. So at STC, 8 cores worker, maybe 30s per game, every 2 minutes (previously every 30s). If we have the fleet with 20k workers, and 20 tests, a test is updated 10x per second.

xoto10:
Ah, I did 4 * 2.5 mins (for ltc) but didn't allow for multiple cores per worker.

So, you did updates per test, but with 20 tests that would be a total of ~200 updates a second?
And this would scale approx linearly with the number of workers.

vondele:
yes, the rate of update_task scales linearly with the number of workers.

xoto10:
Just realised you used 20k workers?! So that would be 160k cores?
Ambitious! :-)

vondele:
we had this on the framework over the weekend :-)

xoto10:
Ok. so 8 cores per worker for the fleet.

I remember seeing 60k cores, maybe a mention of towards 100k, didn't know we wanted to allow more than that.

For updates, I've realised 16mins is ~1000s, so 160k cores is roughly 160 updates / second. This is getting fairly quick.

Then again, if we're caching the updates it's lower than that, and back down to numbers that should be easy. 20 tests, each cached for 60s is only one update every 3s, or 1/s for 60 tests. Am I right in thinking we cache the updates to tasks?

vondele:
there are caches, I haven't fully figured out what the actual update frequency to the db is.

xoto10:
I've now seen 
https://github.com/glinscott/fishtest/blob/d015b3b5a602712646d97a1860ff5d6befbe2c26/server/fishtest/rundb.py#L335

which suggests a write after 1 second, so 20 tests means 20 writes / second, etc. Seems like a reasonable setting.

Suggests fishtest may run smoother with fewer active tests.

So we get: 
- fewer total tasks across all unfinished tests => less cpu use
- fewer active tests => less I/O from db updates

vondele:
well, actually that's potentially an issue. 

If for whatever reason this flush starts taking 1s or more (i.e. various other load), we essentially deadlock? There are a few places we call this function. At the very least, this should be called after the work is finished (i.e. a 1s pause is made in flushing).

noobpwnftw:
Timer is created only once, and they are restarted after finished doing their work. No problem.

vondele:
OK, I was confused by this one https://github.com/glinscott/fishtest/blob/d015b3b5a602712646d97a1860ff5d6befbe2c26/server/fishtest/rundb.py#L341 
which is started before the work, but at most once indeed.

PR #1609, opened 2023-04-18
Throttle pgn purging

vondele:


ppigazzini:
To be honest I will prefer to use cpulimit in cron. Easier to fine tuning the CPU usage, easier to use the script at full speed when requested without hacking the script code.
I started a pgn purge right now, with `nice` 10 and `cpulimit` 25, the script is using 15.25% CPU with PROD unloaded.
```c
 7104 fishtest  30  10   82516    932    844 S   0.3  0.0   0:00.20          `- /usr/bin/cpulimit -l 25 -f -m -- /home/fishtest/fishtest/server/env/bin/python3 fishtest/server/utils/purge_pgn.py
 7105 fishtest  30  10  894640 238880  32652 S  23.3  4.6   0:09.34              `- /home/fishtest/fishtest/server/env/bin/python3 fishtest/server/utils/purge_pgn.py
```

vondele:
The reason to limit form within the script is that even at low load of the script, it can load mongodb to its highest peak. That's a different process that isn't affected by cpulimit or nice.

ppigazzini:
`cpulimit` does somewhat the same thing, with a easier way to setup the exact CPU usage.
mongodb is not a known bottleneck.

image

vondele:
OK, I'm fine with setting a proper cpulimit so we reduce the spike on mongo.

ppigazzini:
`cpulimit` set to 10% to test the result this night.

vondele:
that's good. I was looking at the bytes out during that test, which is about 80x the usual volume.
Anyway, right now framework running smoothly with 30k cores

ppigazzini:
30k? Paper fleet joined I suppose...

PR #1610, opened 2023-04-18
Introduce a scheduled start

vondele:
in case a fleet of workers lands at simultaneously, the server can be overloaded.

To have a more gradual start, schedule worker start times,
ensuring at least 200ms spacing between worker starts.

200ms is a rough estimate, might need refinement after tests

The scheduling is per user name (and not per IP or globally),
to ensure different use cases (fleet behind single IP, or with different IPs) work,
and to ensure that malicious or single workers can not block other user's workers
to join the framework.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -104,6 +104,7 @@ def main(global_config, **settings):
     config.add_route("api_failed_task", "/api/failed_task")
     config.add_route("api_stop_run", "/api/stop_run")
     config.add_route("api_request_version", "/api/request_version")
+    config.add_route("api_request_start", "/api/request_start")
     config.add_route("api_beat", "/api/beat")
     config.add_route("api_request_spsa", "/api/request_spsa")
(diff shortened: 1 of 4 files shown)


vdbergh:
The idea looks good but I wonder if an ill behaving worker which calls `api_request_start` in a loop would not block the server (last_scheduled_start would keep increasing forever).

vdbergh:
I what I am saying is an issue then it could perhaps be solved by having `last_scheduled_start` depend on the user.

vondele:
I think that's a valid concern.

vondele:
I've made the schedule one per user. BTW, had some exchange with ppigazzini to understand if this should instead be implemented via some rules in nginx. So that's still an open question. 

I have also a little bit extended the commit message on this one to make more  what the considerations of the PR are.

noobpwnftw:
Shaping of bursts is already implemented with nginx, will look at fine tuning the numbers.

vondele:
great, will close.

PR #1614, opened 2023-04-18
Fix mongodb reconnect causing CPU hog

noobpwnftw:


noobpwnftw:
Note, problem observed is that somehow after a mongodb restart, python process start to have CPU hog without doing any work. It does still respond to requests, but seems to just running off stale cached data.

This fix is to make failed operations handled on spot, and have iterators carried over in such events.

What I don't understand is that if we don't handle this:
https://pymongo.readthedocs.io/en/stable/api/pymongo/errors.html#pymongo.errors.AutoReconnect
Then any failed attempt will throw exceptions anyways, until Pyramid handles it as generic error? It may or may not cause the process to end(which then will be restarted by systemd), in case that it didn't, what could be hogging the CPU?

vdbergh:
Are you sure mongod would die when running on localhost? And if it does, wouldn’t that leave some trace in the logs (when dying or when restarting)?

vdbergh:
Interesting. So maybe @noobpwnftw is onto something!

vondele:
just for reference, the monitor shows (top memory, bottom CPU):
image
image

vdbergh:
I wonder if the log and the monitoring use the same timezone.

vondele:
I think so, you can even see the service going down (1min resolution):

image

vdbergh:
But then the CPU hogging occurred before the db went down. So the db going down could not have been the cause.

vondele:
different kinds of hogging going on. At least 3 phases:
1) fleet joins, but server handling things fine 4:10 pm
2) server goes overloaded                               4:45pm
3) some hogging with normal fishtest load.   7:22pm (visible).

ppigazzini:
`mongod` RAM usage (30%) never grows over the cache size set in the config.
With a normal sized worker pool and 2 `pserve` the total RAM usage is at 60%
With 3 `pserve` and  load balancing (2 `pserve` making the same work) the total RAM usage is by higher (75%), so in that case I lower the mongod cache size.
With 3 `pserve` serving different APIs the total RAM usage is still at 60% (56.4% right now).

With the previous 96k cores attempt we had a 2 `pserve` configuration, from the top copied in the discord channel the total RAM usage was 40%. The working set seems smaller than the mongo cache.

noobpwnftw:
Is there a better way to wrap the cursors? This code is ugly.

noobpwnftw:
Not this problem, it was probably caused by running out of fd.

PR #1615, opened 2023-04-19
Fix flags

noobpwnftw:


noobpwnftw:
I'm not so confident that the original code ran with multiple threads won't corrupt the dict element.

noobpwnftw:
Close & reopen due to CI network error.

ppigazzini:
```python
from concurrent.futures import ThreadPoolExecutor
import time

flag_cache = {}
clean_flag_data = {"cc": None, "dt": None}
flag_cache["127.0.0.0"] = clean_flag_data

# add a range of values to the dictionary
def add_items(i):
    ip = "127.0.0.1"
    clean_flag_data = {"cc": None, "dt": None}
    try:
#        if i%2 == 0:
#            time.sleep(1)
        print(i) if i%1000 == 0 else None
        flag_cache[ip] = clean_flag_data
        flag_cache[ip]["cc"] = "aa"
        flag_cache[ip]["dt"] = i
    except Exception as e:
        print("Exception: ", e)

n = 1000000
with ThreadPoolExecutor(max_workers=n) as executor:
    executor.map(add_items, range(n))

# print the dictionary
print(flag_cache)
```

noobpwnftw:
Yeah, I tested, old code does race but will not cause corruption but new code is nicer.

ppigazzini:
I didn't read yet the PR code.
Simply I wanted to answer your question. At the time I tested locally the code with a threaded setting to verify the python documentation claim about the thread safety of the built-in. The master code comment declares the race condition, but in that specific case, the threads are writing the same information.

noobpwnftw:
Even better: https://github.com/glinscott/fishtest/pull/1618

PR #1616, opened 2023-04-19
Unfollowed test before trying to delete or stop.

peregrineshahin:
needs testing on DEV.

### DIFF
--- a/server/fishtest/templates/run_table.mak
+++ b/server/fishtest/templates/run_table.mak
@@ -66,7 +66,12 @@
                         <i class="fas fa-trash-alt"></i>
                       </button>
                       <div class="dropdown-menu" role="menu">
-                        <form action="/tests/delete" method="POST" style="display: inline;">
+                        <form 
+                          action="/tests/delete"
+                          method="POST"
+                          style="display: inline;"
+                          onsubmit="unfollow_run('${run['_id']}'); return true;"
+                        >
                           <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}">
                           <input type="hidden" name="run-id" value="${run['_id']}">
                           <button type="submit" class="btn btn-danger btn-mini">Confirm</button>
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -158,30 +158,32 @@ async function main_follow_loop() {
     }
     // I won the race, other tabs should skip their fetch
     save_timestamp(current_time);
-    let json;
+    let json = [];
     let notifications = get_notifications();
     try {
-      json = await fetch_post("/api/actions", {
-        action: "finished_run",
-        run_id: { $in: notifications.content },
-      });
+      if (notifications.content?.length) {
+        json = await fetch_post("/api/actions", {
+          action: "finished_run",
+          run_id: { $in: notifications.content },
+        });
+      }
     } catch (e) {
       console.log(e);
       continue;
     }
     notifications = get_notifications();
     let work = [];
-    json.forEach((entry) => {
+    for (const entry of json) {
       let run_id = entry["run_id"];
       if (notifications.contains(run_id)) {
         work.push(entry);
         notifications.remove(run_id);
       }
-    });
+    }
     save_notifications(notifications); // make sure other tabs see up to date data
     // Instrumentation
     console.log("active notifications: ", JSON.stringify(notifications));
-    work.forEach(async (entry) => {
+    for (const entry of work) {
       run_id = entry["run_id"];
       disable_notification(run_id);
       set_notification_status(run_id);
@@ -197,7 +199,7 @@ async function main_follow_loop() {
         console.log(e);
         notify_elo(null, entry);
       }
-    });
+    }
   }
 }
 
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -149,7 +149,11 @@
     <div class="row g-2 mb-2">
       % if not run['finished']:
         <div class="col-12 col-sm">
-          <form action="/tests/stop" method="POST">
+          <form
+            action="/tests/stop"
+            method="POST"
+            onsubmit="unfollow_run('${run['_id']}'); return true;"
+          >
             <input type="hidden" name="run-id" value="${run['_id']}">
             <button type="submit" class="btn btn-danger w-100">
               Stop


vondele:
Will that stop the following for everybody watching that test, or only for the person that presses stop/delete.

peregrineshahin:
only for the person that presses stop/delete.
The 'followed' tests are stored in localStorage of the browser as of the current implementation. meaning you can't force the requestor/client who uses the API to not request the following.

I will try to dig deep into how we can improve the notification system as a whole but this line, in particular, is only touching what runs "should the client" ask the server if finished or not to notify about or not at first.

P.s.: We cannot prevent such API's requests to be requested as of the current implementation of the server (i.e one can still ask about what he wants).

vdbergh:
I have thought about this issue. The correct fix for this is to do it in notifications.js by not only looking for finished_run events but also for stop_run and delete_run events. 

However then there would still be an issue with this approach. If the browser unsubscribes from tests that have been stopped then it cannot resubscribe when they are unstopped (because the actions query would still see the old stop run event and unsubscribe again).

So to really fix this (and to also handle purging correctly which can even unfinish a test) the run_id's in the notifications data structure maintained by the browser should be a given an insertion timestamp so that events which happened before the insertion timestamp are ignored.

Since the current code working sufficiently well I propose to leave it as it is, until the fix as in the previous paragraph is implemented.

vondele:
The problem of the current code is that it could be at the source of some DoS on the server. The api call can be pretty expensive (>1s depending on the payload) and with many people following tests, there seems to be an issue.

peregrineshahin:
> The correct fix for this is to do it in notifications.js by not only looking for finished_run events but also for stop_run and delete_run events.

We are not targeting the same issue here and in general, looking for `finished_run` is the only thing we want to ask about but I believe in the current form there is no guard against abuse because the whole notification system is client-sided

On another note What this PR does is to correctly fix what should be asked about, i.e in the current implementation the notification is stopped being asked if finished or not if it already notified to the client, Well there is another two cases that was a bit overlooked, which is when the user stops their tests and when the user deletes them.

vdbergh:
The actions call is very lightweight. It uses and index to fulfill the query and only returns a few records. I would expect it to take only a couple of ms (of course if the db is doing something else it might seem to take longer, but that's not the fault of the query). 

Also note that the query only happens every 20s and moreover this period is independent of the number of tabs open (the tabs share the information).  

I would expect get_elo to be much heavier.

vondele:
see https://discord.com/channels/435943710472011776/900771437177094156/1097117723659599892 for timings of all our apis a while ago.

vdbergh:
Hmm this is does not agree with my timings (which was only a few ms IIRC). I do not believe such a lightweight query can take 1.7s on average (if that were the case then people would not be using mongod). I will check again tonight.

vdbergh:
That the view "actions" would take 1.7s on average to render seems more believable (although this is still a lot).

vdbergh:
This is interesting data since we can infer from it how many user were connected.  A browses call api-actions around 180 times in one hour and there were 15297 calls. So about 80 users connected.

vondele:
I have updated timing data for the clean run today with the fleet on. Just copying here the numbers shared on discord:
```
# logging from [19/Apr/2023:13:00:00 +0000] to [19/Apr/2023:13:59:59 +0000] ########
                      /api/beat   180369   2411.841      0.013      0.000      1.799
               /api/update_task   113830   1709.715      0.015      0.000      1.748
                         /tests      315    650.689      2.066      0.120      9.526
                   /api/actions    11174    184.817      0.017      0.001      5.269
               /api/active_runs       60    153.639      2.561      1.825      4.582
           /api/request_version     2719    149.030      0.055      0.000      1.709
              /api/request_task     2380    148.228      0.062      0.011      6.231
                   /tests/user/       48     83.463      1.739      0.080      4.189
                /api/upload_pgn     2258     81.587      0.036      0.002      2.110
                       /actions       55     50.627      0.920      0.074      1.544
                  /api/get_elo/      420     45.695      0.109      0.017      1.217
                /tests/finished       68     28.863      0.424      0.151      1.622
                   /tests/view/       71     26.069      0.367      0.006      8.662
              /api/request_spsa      802     10.346      0.013      0.000      0.897
                     /tests/run        7      7.709      1.101      0.001      2.105
                       /api/nn/       55      1.334      0.024      0.003      0.483
               /tests/live_elo/       77      1.315      0.017      0.002      0.055
                  /tests/stats/       23      1.248      0.054      0.004      0.671
               /api/failed_task       29      1.244      0.043      0.004      0.824
                         /users        4      0.946      0.236      0.070      0.541
                 /users/monthly        6      0.582      0.097      0.014      0.330
                           /nns        4      0.261      0.065      0.013      0.216
                        /signup        1      0.134      0.134      0.134      0.134
                 /tests/approve        3      0.099      0.033      0.015      0.067
                  /tests/modify        1      0.086      0.086      0.086      0.086
                         /login        1      0.015      0.015      0.015      0.015

```

vdbergh:
We can see that the server is multi threaded!  The sum of the top two totals is already > 3600...

peregrineshahin:
TODO: after this is tested on DEV, put a line in PROD that removes all old localStorage entries from notifications for a while until most users open the site, so we can fix this retroactively.

vdbergh:
Do you have reference for your claim that an async function cannot be used in for each?

vdbergh:
The foreach async is a pretty bad bug! Thanks for noticing it. Perhaps this can explain some mysterious errors with duplicate notifications which I could not fix with locks.

vondele:
BTW, timing measurements during a CPU spike yesterday looked like the below, and are almost certainly triggered by the notification code:
```
# logging from [19/Apr/2023:20:26:00 +0000] to [19/Apr/2023:20:32:59 +0000] ########
                   /api/actions     1497   3971.695      2.653      0.001      9.465
                /tests/finished      569   2332.340      4.099      0.123     10.228
                       /actions      402   1623.356      4.038      0.818      5.401
                         /users      373   1461.685      3.919      0.050      9.435
                 /users/monthly      421   1368.207      3.250      0.006      9.431
                         /tests      767    289.778      0.378      0.016      6.791
                       /pending      355     75.501      0.213      0.001      1.617
               /api/active_runs        7     52.400      7.486      2.183     13.037
               /api/update_task     2340     51.006      0.022      0.001      1.554
                  /api/get_elo/      116     31.273      0.270      0.020      1.373
                      /api/beat     1823     31.045      0.017      0.000      1.565
                   /tests/user/        2     11.882      5.941      1.958      9.924
                   /tests/view/        3      9.868      3.289      0.462      8.712
                /api/upload_pgn       27      1.314      0.049      0.004      0.366
              /api/request_task       29      0.606      0.021      0.010      0.097

```

vdbergh:
> BTW, timing measurements during a CPU spike yesterday looked like the below, and are almost certainly triggered by the notification code:

Why are you saying that the CPU spike is "almost certainly caused by the notification code"? I am still not convinced.

The number of calls seems to be for about 70 people connected, which is similar to what we had before.

It is true that in this case the calls take a long time to complete but I think they are just waiting for the db (thus using no CPU time). The api call itself is very light weight. 

The notification code has been running for several months and the CPU spikes are a relatively recent phenomenon...

vondele:
some more analysis on the fishtest server (see discord, including the 10s of calls from the same IP at the same time in 20s intervals, which are the reason for some of the other PRs that you see). Right, they might be waiting on the db or other threads holding locks, but I guess that's the point?

vdbergh:
Isn't it true that the frequent spikes are a recent phenomenon? If the notification code would be that troublesome it should have been noticed much earlier...

> some more analysis on the fishtest server (see discord, including the 10s of calls from the same IP at the same time in 20s intervals, which are the reason for some of the other PRs that you see). Right, they might be waiting on the db or other threads holding locks, but I guess that's the point?

There is indeed an issue with the synchronization code between tabs which occasionally can cause multiple threads to make an api call at the same time (especially at browser restart). I do not think the impact of this is particularly large but it should be fixed by noob's PR.

vondele:
we have had phases before of large load, but things have probably been masked by other inefficiencies, and the fact that we had no setup/tools to analyze. Without fleet load, the server just handles this, but if at the same time the fleet arrives, the server collapses.

vondele:
I think it is a bit of a generic issue with providing api calls that are potentially very demanding (db searches etc). It is easy to load inadvertently the server.

vdbergh:
> I think it is a bit of a generic issue with providing api calls that are potentially very demanding (db searches etc). It is easy to load inadvertently the server.

I agree but /api/actions is not demanding... Of course if the db is otherwise occupied (with truly demanding queries) then the action query may have to wait for the db to clear. But then it is not consuming any CPU time so it is not loading the system.

vondele:
so, do we know if and how we query `/tests/finished` ?

vdbergh:
I wonder if those api measurements are not fundamentally flawed. Once one query starts hogging the db all the other apis will be delayed (and seemingly take a long time) but they will not necessarily consume any CPU time.

I think one should identify the queries that most stress the db, but I do not know how to do this. I guess all queries these days have the right indexes, so ideally one needs to have a log with the db queries and the amount of data returned. I don't if mongodb can do this.

vondele:
yes, one needs to be careful with the interpretation. However, we also have some more cProfile date on discord. Here a profile from building the homepage:
```
         1    0.016    0.016    7.485    7.485 fishtest/views.py:1379(homepage_results)
     18/10    0.000    0.000    6.937    0.694 pymongo/_csot.py:98(csot_wrapper)
         9    0.000    0.000    6.935    0.771 pymongo/mongo_client.py:1418(_retryable_read)
       280    0.001    0.000    6.861    0.025 pymongo/cursor.py:1244(next)
        12    0.000    0.000    6.861    0.572 pymongo/cursor.py:1128(_refresh)
         8    0.000    0.000    6.860    0.857 pymongo/cursor.py:1037(__send_message)
         8    0.000    0.000    6.859    0.857 pymongo/mongo_client.py:1302(_run_operation)
         8    0.000    0.000    6.854    0.857 pymongo/mongo_client.py:1324(_cmd)
         8    0.001    0.000    6.854    0.857 pymongo/server.py:76(run_operation)
         1    0.071    0.071    6.772    6.772 fishtest/rundb.py:485(aggregate_unfinished_runs)
        10    0.000    0.000    6.466    0.647 pymongo/message.py:1391(unpack_response)
        10    0.000    0.000    6.466    0.647 bson/__init__.py:1184(_decode_all_selective)
        10    0.019    0.002    6.466    0.647 bson/__init__.py:1093(decode_all)
         8    0.000    0.000    6.466    0.808 pymongo/cursor.py:1117(_unpack_response)
        10    6.445    0.644    6.447    0.645 {built-in method bson._cbson._decode_all}
         2    0.041    0.020    3.117    1.559 fishtest/rundb.py:240(get_machines)
```
https://discord.com/channels/435943710472011776/900771437177094156/1098544090804650005 
aggregate_unfinished_runs really seems costly, and most of this goes to unpack_response.

vdbergh:
It seems mongod can log execution time and "documents examined/returned" https://www.mongodb.com/docs/manual/reference/database-profiler/ . "Documents returned" is a bit misleading for Fishtest since the run documents are humongous because of the task lists (I think this is a design flaw, unfortunately encouraged by mongodb, but impossible to fix now).

ppigazzini:
> I think this is a design flaw, unfortunately encouraged by mongodb, but impossible to fix now

https://www.mongodb.com/developer/products/mongodb/mongodb-schema-design-best-practices/

server/fishtest/templates/run_table.mak
```diff
@@ -66,7 +66,12 @@
                         <i class="fas fa-trash-alt"></i>
                       </button>
                       <div class="dropdown-menu" role="menu">
-                        <form action="/tests/delete" method="POST" style="display: inline;">
+                        <form 
+                          action="/tests/delete"
+                          method="POST"
+                          style="display: inline;"
+                          onsubmit="unfollow_run('${run[\'_id\']}'); return true;"
```
ppigazzini, line 73: The quote escaping doesn't work.

peregrineshahin: > The quote escaping doesn't work
I would appreciate it if hinted I need the escaping at the first place?

vdbergh:
> > I think this is a design flaw, unfortunately encouraged by mongodb, but impossible to fix now
> 
> https://www.mongodb.com/developer/products/mongodb/mongodb-schema-design-best-practices/

Yes that link confirms what I wrote. With mongodb you are suppose to throw overboard all established principles of database design, simply because doing queries over multiple tables is cumbersome (and probably inefficient as well).

This is of course a recipe for creating unscalable designs.

vondele:
If I recall correctly, at at given point we we ran into the issue that the task list exceeded 16MB. It is much smaller now, but that makes retrieving a single run pretty expensive. If there are 80 active runs, getting just that list is roughly 1GB of data transfered.

vdbergh:
> If I recall correctly, at at given point we we ran into the issue that the task list exceeded 16MB. It is much smaller now, but that makes retrieving a single run pretty expensive. If there are 80 active runs, getting just that list is roughly 1GB of data transfered.

Yes in a normal design the tasks would be a in a separate (flat) table so retrieving runs would be very fast (no caching necessary).

ppigazzini:
> If I recall correctly, at at given point we we ran into the issue that the task list exceeded 16MB. It is much smaller now, but that makes retrieving a single run pretty expensive. If there are 80 active runs, getting just that list is roughly 1GB of data transfered.

We surely ran in the 16MB limit few times with the SPSA param_history used to draw the graph. The last time we unstuck fishtest with a light mongodb surgery (dropping the inflated SPSA run and reloading a smaller sized version from a db backup).
This commit set a size bound for the param_history. 

https://github.com/glinscott/fishtest/commit/33d35ba3fa9e61981e132f1bd55a8a35fc154b72

ppigazzini:
@peregrineshahin DEV updated, this time no errors in the logs, please check if the code is working as intended.

ppigazzini (approved):
Tested working for both "Stop" and "Delete" actions.

image

ppigazzini:

vondele:
let me add one note: tests could be stopped by any approver, while they could be followed by anybody. So somehow this is just a local solution, right?

peregrineshahin:
> let me add one note: tests could be stopped by any approver, while they could be followed by anybody. So somehow this is just a local solution, right?

Yes it is a local solution, because the notification system is a local system by now... Unless the notification system is connected to backend and the user has actual profile and his notification data is saved in the database we can't do anything about that.

vdbergh:
Yes we can, and I explained to you how. https://github.com/glinscott/fishtest/pull/1616#issuecomment-1514237494 . It is on my TODO list.

PR #1617, opened 2023-04-19
Send heartbeat at a somewhat slower rate

vondele:
send every 2min instead of every 1min. Scavenge after 6 instead of 3min.

Based on the profile that shows this is the top api call: https://github.com/glinscott/fishtest/pull/1616#issuecomment-1515224083 when the fleet is on.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -438,11 +438,11 @@ class RunDb:
             self.start_timer()
 
     def scavenge(self, run):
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):

ppigazzini:
Merged with https://github.com/glinscott/fishtest/commit/77f7d76ae2f1ca41f5dff664bb17b431b32d3217
Workers updated.
Thank you @vondele :)

PR #1618, opened 2023-04-19
Use country code produced from nginx side

noobpwnftw:


noobpwnftw:
Fishtest server has nginx configs which will send country code of every request in HTTP header `X-Country-Code` field.

server/fishtest/api.py
```diff
@@ -247,62 +245,13 @@ def stats(self):
...
-            ip_cc = flag_cache[ip]["cc"]
-            ip_dt = flag_cache[ip]["dt"]
-        except KeyError:
-            pass
-        else:
-            if ip_cc is None and (now_dt - ip_dt).total_seconds() > 3600 * 4:
-                flag_cache[ip] = clean_flag_data
-                flag_cache[ip]["cc"] = get_flag_cc(self, ip)
-        return flag_cache[ip].get("cc")
+    def get_country_code(self):
```
ppigazzini, line 248:
```python
    def get_country_code(self):
        country_code = self.request.headers.get("X-Country-Code")
        return None if country_code == "ZZ" else country_code
```

server/fishtest/api.py
```diff
@@ -220,9 +218,9 @@ def pgn(self):
     def worker_info(self):
         worker_info = self.request_body["worker_info"]
         worker_info["remote_addr"] = self.request.remote_addr
-        flag = self.get_flag()
-        if flag:
-            worker_info["country_code"] = flag
+        country_code = self.get_country_code()
```
ppigazzini, line 221: LGTM on DEV https://dfts-0.pigazzini.it/tests
I need to review/test the indexes script change before merging.

BTW this is the pythonic way (I can refactor it in the weekend)
```python
worker_info["country_code"] = (country_code := self.get_country_code()) if country_code else "?"
```

ppigazzini (commented):
Nice!

server/fishtest/api.py
noobpwnftw: I made it more pythoner.

server/fishtest/api.py
```diff
@@ -247,62 +241,9 @@ def stats(self):
...
-        except KeyError:
-            pass
-        else:
-            if ip_cc is None and (now_dt - ip_dt).total_seconds() > 3600 * 4:
-                flag_cache[ip] = clean_flag_data
-                flag_cache[ip]["cc"] = get_flag_cc(self, ip)
-        return flag_cache[ip].get("cc")
+    def get_country_code(self):
+        country_code = self.request.headers.get("X-Country-Code")
+        return "?" if not country_code or country_code == "ZZ" else country_code
```
ppigazzini, line 246: This is working code, so just nitpicking here (aka trolling):
If `get("X-Country-Code")` can be Falsy (`None`, ""):
```python
return "?" if country_code in (None, "", "ZZ") else country_code
```
otherwise:
```python
return "?" if country_code in (None, "ZZ") else country_code
```

ppigazzini:
DEV updated, LGTM, very nice clean up :)

server/fishtest/api.py
noobpwnftw: HTTP headers cannot and will not have an empty value, therefore I'll adopt the second line.

ppigazzini:
After last commit:
```python
$ git grep -n country_code
server/fishtest/api.py:221:        worker_info["country_code"] = self.get_country_code()
server/fishtest/api.py:244:    def get_country_code(self):
server/fishtest/api.py:245:        country_code = self.request.headers.get("X-Country-Code")
server/fishtest/api.py:246:        return "?" if country_code in (None, "ZZ") else country_code
server/fishtest/templates/machines_table.mak:28:            % if 'country_code' in machine:
server/fishtest/templates/machines_table.mak:29:                <div class="flag flag-${machine['country_code'].lower()}"
server/utils/upgrade.py:81:    "country_code": "?",
```

So we use that field only to show the flag in the workers table:
https://github.com/glinscott/fishtest/blob/0df452710c472cc23e8510c07d05788e17318d2f/server/fishtest/templates/machines_table.mak#L28-L31

with:
https://github.com/glinscott/fishtest/blob/master/server/fishtest/static/css/flags.css

It's seems that we don't need to deal with "ZZ",
I will check why we wrote "?" (I suspect before having the flag) and if the code can be further simplified, eg:

```diff
diff --git a/server/fishtest/api.py b/server/fishtest/api.py
index 4a821d0..2c7ca39 100644
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -242,8 +242,7 @@ class ApiView(object):
         return self.request_body.get("spsa", {})

     def get_country_code(self):
-        country_code = self.request.headers.get("X-Country-Code")
-        return "?" if country_code in (None, "ZZ") else country_code
+        return self.request.headers.get("X-Country-Code")

     @view_config(route_name="api_active_runs")
     def active_runs(self):
diff --git a/server/fishtest/templates/machines_table.mak b/server/fishtest/templates/machines_table.mak
index deeb2be..b345efe 100644
--- a/server/fishtest/templates/machines_table.mak
+++ b/server/fishtest/templates/machines_table.mak
@@ -25,7 +25,7 @@
         <tr>
           <td>${machine['username']}</td>
           <td>
-            % if 'country_code' in machine:
+            % if 'country_code' in machine and machine['country_code'] is not None:
                 <div class="flag flag-${machine['country_code'].lower()}"
                      style="display: inline-block"></div>
             % endif
```

ppigazzini:
PROD restarted with this PR as well.

ppigazzini:
The "?" has not a real meaning, https://github.com/glinscott/fishtest/commit/25ffbcd93331e6707b7e776a2ac679bf30488b79
> For consistency always set the country_code field in
worker_info, even if it is unknown.

noobpwnftw:
Without it, layout is weird, I prefer keeping the `?`.

ppigazzini (approved):
Works fine on PROD

PR #1619, opened 2023-04-19
Update notifications.js

noobpwnftw:


vdbergh:
I don’t understand this. The time stamp is supposed to record when the latest fetch happened. But now you updating it always?

vdbergh:
I think this change is ok. With multiple tabs open this change makes a difference at start up and when the api call fails (that should be rare).

noobpwnftw:
Sleep is 20s+~0.5s, if tabs were opened about the same time and fetch takes more than 1.5s, next loop every tab fetches(> 19s).

vdbergh:
@noobpwnftw Vondele’s profiling show that even under heavy load the api call only takes a few milliseconds. I agree with this PR but I think the impact will be low.

noobpwnftw:
Under heavy load things get slower and the old behavior would amplify it. Seen many individuals having 10+ tabs on creating spikes.

noobpwnftw:
Add initial spread, should average into the same delays.
Sleeps are now moved to one common place.

vdbergh:
> Under heavy load things get slower and the old behavior would amplify it. Seen many individuals having 10+ tabs on creating spikes.

Who are these “many individuals”? You cannot get an average of a few a few ms (under heavy load) when loading takes 1.5s all the time…

Anyway as I said I agree with this change.

Personally I would also keep the extra delay when the api call fails (e.g. when the server is down) but that’s a matter of taste.

noobpwnftw:
Problem:
Measurements were made from a pure server perspective, it did not account for the time a client and the server took during communication of the request and response, network quality, distance, etc.

Timestamps are already written before calling API, so in any case next loop will check if it is either 19 seconds later or it'll sleep.

vdbergh:
> Problem: Measurements were made from a pure server perspective, it did not account for the time a client and the server took during communication of the request and response, network quality, distance, etc.
> 

I guess this I agree with.

vdbergh:
Anyway the code looks cleaner now!

ppigazzini:
PROD update with this PR to test.

ppigazzini (approved):
Worked fine on PROD.

PR #1625, opened 2023-04-22
Remove stopped/deleted test by approver from notifications

peregrineshahin:
This leaves us with if the approver continued the test we lost the notification for the user-end which I don't actually imagine a good solution for with the current implementation

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -163,7 +163,7 @@ async function main_follow_loop() {
     try {
       if (notifications.content?.length) {
         json = await fetch_post("/api/actions", {
-          action: "finished_run",
+          action: { $in: ["finished_run", "stop_run", "delete_run"] },
           run_id: { $in: notifications.content },
         });
       }
@@ -176,7 +176,8 @@ async function main_follow_loop() {
     for (const entry of json) {
       let run_id = entry["run_id"];
       if (notifications.contains(run_id)) {
-        work.push(entry);
+        if (entry["action"] === "finished_run")
+          work.push(entry);
         notifications.remove(run_id);
       }
     }


vdbergh:
Sigh: I explained to you that this is not correct for stopped runs ( https://github.com/glinscott/fishtest/pull/1616#issuecomment-1514237494 ) because they can be unstopped and then the end cannot resubscribe (because the browser will unsubscribe again seeing the old stop_run event).

peregrineshahin:
I don't understand the solution proposed by this
```
So to really fix this (and to also handle purging correctly which can even unfinish a test) the run_id's in the notifications data structure maintained by the browser should be given an insertion timestamp so that events which happened before the insertion timestamp are ignored.
```

vdbergh:
> I don't understand the solution proposed by this
> 
> ```
> So to really fix this (and to also handle purging correctly which can even unfinish a test) the run_id's in the notifications data structure maintained by the browser should be given an insertion timestamp so that events which happened before the insertion timestamp are ignored.
> ```

Assume a run is stopped. So all the browsers unsubscribe.  Some time later the run is unstopped and a user resubscribes. His browser now sees the old stop event again and currently will unsubscribe again.

Hence the browser somehow needs to know that the old stop event is not relevant. The way to do this is to insert in the LRU data structure a pair `(run_id, insertion_timestamp)` (instead of currently just `run_id`). So if a browser sees an event with a timestamp older than the `insertion_timestamp` it knows that it should ignore it.

I hope I have explained it better now. If it is still unclear please ask again (or suggest another solution that deals correctly with stopped runs).

peregrineshahin:
Assume a run is stopped. So all the browsers unsubscribe. Some time later the run is unstopped and a user resubscribes.

That's already covered in this case, if a user subscribes then this will notify him also.
The problem isn't when a user resubscribed to an old stopped test
the problem is that the user already subscribed and they are obliged to do it again to get notified.

vdbergh:
Ah your aim is to send notifications to the user if a run is stopped or deleted.  But if you do not unsubscribe then you will see the same stop/delete event over and over again. So that should not work.

peregrineshahin:
Oh so this is searching through events and not the run itself if it is stopped of finished or deleted
Now I see how thanks.
So a resubscribtion will not work damn..

Okay in that case yes I believe there should be another primary key for the entries of course.. Something like timestamp probably works.

I wasn't actually thinking about sending the user if the test is deleted or stopped but this idea also seems the best because we need more than this datastructre, implementation to let him subbed after unstopped the test without him resubing

Closing this... Feel free to implement your fix.

vdbergh:
> Closing this... Feel free to implement your fix.

Yes it queries the event log. That's why the api is super light weight (the records in the event log are small). If it would query the rundb instead it would be terribly heavy.

PR #1626, opened 2023-04-23
Unsubscribe from stopped and deleted tests.

vdbergh:
Since the format of the LRU data structure has changed, it is very important to reload all tabs, since otherwise the results will be unpredictable.

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -8,6 +8,9 @@ function LRU(capacity, content) {
   if (!(content instanceof Array)) {
     throw "content is not an Array";
   }
+  if (content.length != 0 && !(content[0] instanceof Array)) {
+    throw "content entry has the wrong type";
+  }
   return {
     capacity: capacity,
     content: content,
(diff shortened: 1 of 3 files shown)


ppigazzini:
DEV updated, please start testing yourself the PR, I'm in the middle of another code review right now.

vdbergh:
> DEV updated, please start testing yourself the PR, I'm in the middle of another code review right now.

Sure!

vdbergh:
I did some testing and things seem to be working satisfactorily.

ppigazzini (approved):

ppigazzini:
If I'm not wrong, you are not on Discord: this afternoon fishtest dealt with 100k cores, but we should have hit some bugs in the code.
image

server/fishtest/static/js/notifications.js
```diff
@@ -175,7 +201,9 @@ async function main_follow_loop() {
     let work = [];
     for (const entry of json) {
       let run_id = entry["run_id"];
-      if (notifications.contains(run_id)) {
+      const ts = notifications.timestamp(run_id);
+      // ignore events that happened more than 1 minute before subscribing
```
peregrineshahin, line 205: Here instead of ignoring returned data in the loop why not do this in the query itself
```
        json = await fetch_post("/api/actions", {
          action: { $in: ["finished_run", "stop_run", "delete_run"] },
          run_id: { $in: notifications.toArray() },
          time: { $gt: ts / 1000 - 60}
        });
      }
 ```

peregrineshahin: My general take is that there are lots of manipulation/operations to data that we can use Mongo for instead of querying stuff and manipulating returned data with Python/JS code, this applies to most places around the code.

vdbergh: > My general take is that there are lots of manipulation/operations to data that we can use Mongo for instead of querying stuff and manipulating returned data with Python/JS code, this applies to most places around the code.

I agree with this particular change but in general if you make more complicated queries you are shifting work from the client to the server, which is not always a good idea (there are many clients but only one server).

peregrineshahin: Yeah I'm more referring to Python in the server here probably does not hold, but complex queries given a fixed number of queries are better IMO than looping through a general query, the thing stays that we might sometimes need general queries if this data is shared with many functions else I don't expect MongoDB at worst case is any slower looping instead of python code.

PR #1627, opened 2023-04-23
Increase the default maximum number of games

vondele:
- tests take longer, we hit 800k from time to time
- in the case the fleet lands, the number of tasks created per run is limited by 800k games. This limits the number of cores that can participate to a single run.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -795,7 +795,7 @@ def validate_form(request):
             batch_size=sprt_batch_size_games // 2,
         )  # game pairs
         # Limit on number of games played.
-        data["num_games"] = 800000
+        data["num_games"] = 1600000
     elif stop_rule == "spsa":
         data["num_games"] = int(request.POST["num-games"])
         if data["num_games"] <= 0:


dubslow:
Eh, I'm not a big fan. 1) is a good, because it encourages us to double check the sanity of test when it runs longer than it ought to, statistically speaking. If anything, I would be in favor of reducing the default maximum to 600k or 500k. Re-upping that default when the rare test hits it requires approximately negligible time, and is worth the extra sanity check imo.

Sacrificing 1) as a hack/workaround for 2) is a bad idea imo. I suggest that 2) rather reflects that the fishtest code is broken in other ways, and can therefore be fixed in other ways without impacting 1).

maximmasiutin:
Small batches are bad, but batches of five or ten hours is not good also. The batches of 1514 games for 32 cores aren't good for tc=160+1.6 nodestime=600. Maybe we can introduce a column "batch duration" or "batch started" (near the "Last updated" column) so we could make sure that batches of five or ten hours aren't very good also.

maximmasiutin:
The disadvantage of tasks of that duration is that when I change priority of a test, this change only applies as soon as the batch finishes. Therefore, the priority is almost static.

maximmasiutin:
Or when I don't change priority but create a new task with hire priority, that test does not release workers (for many hours) to the more urgent task.

ppigazzini:
@maximmasiutin if your instance is for development/test, on DEV I set the `task_duration` to 3 minutes merging a branch with a script (view the server script under Test a PR/branch section).
https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup#test-a-prbranch
https://github.com/glinscott/fishtest/compare/master...ppigazzini:fishtest:task_duration_180

peregrineshahin:
This PR has nothing to do with Batch size it's the maintainer of SF deciding if the test should go longer or not.
Daily reminder some linrock net passing after extending it to more than 1M games, IMO no need for delay merging this.

The arguments that this is a hack if it really was are no longer valid.

PR #1628, opened 2023-04-23
Fix race

noobpwnftw:


noobpwnftw:
https://github.com/glinscott/fishtest/blob/5e25a0cc9a7e22db980056b35eca4427b274d65d/server/fishtest/rundb.py#L880
https://github.com/glinscott/fishtest/blob/5e25a0cc9a7e22db980056b35eca4427b274d65d/server/fishtest/rundb.py#L432

vdbergh:
Locking all task lists seems rather expensive. We could just take the length of the task list and then do a regular for loop. 

BTWI suspect that is what Python does internally anyway, so it could be that the original code was already safe.

noobpwnftw:
OTOH, this only fixes the obvious bug, there are more places where the same pattern can potentially have data race, second degree elements returned by `get_run` reference should not be accessed without a lock, do you think it'd be a good idea to add a lock per run in run_cache?

vdbergh:
> OTOH, this only fixes the obvious bug, there are more places where the same pattern can potentially have data race, second degree elements returned by `get_run` reference should not be accessed without a lock, do you think it'd be a good idea to add a lock per run in run_cache?

I think doing
```python
tasks=run["tasks"]
l=len(run["tasks"])
for i in range(0,l):
  task=tasks[i]
```
is not relying on any python features except the global interpreter lock.

But yes if we want to use a lock in scavenge then it would probably be best to have it per run.

dubslow:
I really don't see the point of that. That doesn't change the fact that `task[i]` still changes while the loop is running. In other words, it's effectively no different from master.

noobpwnftw:
GIL is a Python "feature", and it lets people get away with bad coding styles related to threads.
`get_run` is one of them. It is equivalent to returning a wild pointer in most languages when threaded.

server/fishtest/rundb.py
```diff
@@ -427,10 +427,9 @@ def scavenge(self, run):
         # print("scavenge ", run["_id"])
         dead_task = False
         old = datetime.utcnow() - timedelta(minutes=6)
-        task_id = -1
         run_id = str(run["_id"])
-        for task in run["tasks"]:
-            task_id += 1
+        for task_id in list(run["tasks"]):
```
ppigazzini, line 431: https://github.com/glinscott/fishtest/blob/16d52ff19438f7af513ddf17df633dc9ebad314f/server/fishtest/rundb.py#L184-L194

`run["tasks"]` is already a list. In the proposed code `task_id` is one of the tasks (a dict with all the info about the task), so the code should break.

vdbergh: This is not what I was proposing....

ppigazzini:
BTW the calls chain looks like: `buffer() -> start_timer() -> flush_buffers() -> scavenge()`

buffer grabs the `run_cache_lock`

https://github.com/glinscott/fishtest/blob/d015b3b5a602712646d97a1860ff5d6befbe2c26/server/fishtest/rundb.py#L338-L362

noobpwnftw:
Does it grab `task_lock` where append took place?

server/fishtest/rundb.py
```diff
@@ -427,10 +427,9 @@ def scavenge(self, run):
         # print("scavenge ", run["_id"])
         dead_task = False
         old = datetime.utcnow() - timedelta(minutes=6)
-        task_id = -1
         run_id = str(run["_id"])
-        for task in run["tasks"]:
-            task_id += 1
+        for task_id in range(0, len(run["tasks"])):
```
ppigazzini, line 431:
```python
    old = datetime.utcnow() - timedelta(minutes=6)
    for task_id, task in enumerate(run["tasks"]):
        if task["active"] and task["last_updated"] < old:
            task["active"] = False
            dead_task = True
```

noobpwnftw: Is `enumerate` thread safe when other threads add elements to the enumerated without locking?

ppigazzini: The run is locked, I'm expecting that nobody can add a task in that run.

ppigazzini:
> Does it grab `task_lock` where append took place?

I'm expecting that a thread that wants to add info in a field of a run is waiting to grab the `run_cache_lock`

noobpwnftw:
Well, is it?

ppigazzini:
> Well, is it?

We need to review the code to find an access to the run without the run lock.

noobpwnftw:
You know, like I said above, there are a good amount of such races scattering around.
I may not have caught all of them, but this is wrong and I'm going to close this.

> We need to review the code to find an access to the run cache without the lock.

Like, you did not read https://github.com/glinscott/fishtest/pull/1628#issuecomment-1519193027?

server/fishtest/rundb.py
noobpwnftw: The run is locked how?

noobpwnftw:
Looking at the diff now, do you realize why I whined?

ppigazzini:
`enumerate` is a built-in, so it should be thread safe. With enumerate you are sure to iterate along the appended elements, not with your code.
- `enumerate()`:
```python
from concurrent.futures import ThreadPoolExecutor

a = []
def func(id):
   # print("thread =", id)
    global a
    c = 0
    for i, v in enumerate(a):
        c = max(c, i + 1)
    a.append(c)

if __name__ == "__main__":
    mw = 1000
    with ThreadPoolExecutor(max_workers=mw) as executor:
        executor.map(func, range(mw))

    print(len(a))
    print(len(set(a)))
    print(a)
```

- `range(len())`
```python
from concurrent.futures import ThreadPoolExecutor

a = []
def func(id):
   # print("thread =", id)
    global a
    c = 0
    for i in range(len(a)):
        c = max(c, i + 1)
    a.append(c)

if __name__ == "__main__":
    mw = 1000
    with ThreadPoolExecutor(max_workers=mw) as executor:
        executor.map(func, range(mw))

    print(len(a))
    print(len(set(a)))
    print(a)
```

noobpwnftw:
Sure, how about this?
```
from concurrent.futures import ThreadPoolExecutor

a = []
def func(id):
   # print("thread =", id)
    global a
    c = 0
    for i in a:
        c = max(c, i + 1)
    a.append(c)

if __name__ == "__main__":
    mw = 1000
    with ThreadPoolExecutor(max_workers=mw) as executor:
        executor.map(func, range(mw))

    print(len(a))
    print(len(set(a)))
    # print(a)
    ```

**Comment by ppigazzini** [comment]
It works, iterators are thread safe, and list (all the built-ins) is thread safe in python. Data written in the built-ins are not thread safe, but we are looking for crashes IIRC, not data corruption.

ps: you "vps sux", me "server code sux"

**Comment by vdbergh** [comment]
> enumerate is a built-in, so it should be thread safe. With enumerate you are sure to iterate along the appended elements, not with your code (sometime yes, sometime no).

First of all it does not matter if we do not interate over the appended item. It will be iterated over next time.

Secondly `enumerate(list)` is a view of `list`. So I am pretty sure that it is illegal to change `list` while enumerating over it although it probably works if you append things.

**Comment by ppigazzini** [comment]
I wrote these scripts to test the different syntax, tbh I was expecting some unexpected results.

**Comment by ppigazzini** [comment]
> Secondly `enumerate(list)` is a view of `list`. So I am pretty sure that it is illegal to change `list` while enumerating over it although it probably works if you append things.

`enumerate()` returns an enumerate object, nearly an iterator, it's not a view as at example `dict.keys()`.

```python
a = list(range(10))
b = enumerate(a)

for i, v in b:
    print(i, v)

# exausted
for i, v in b:
    print(i, v)

d = {str(i): i for i in range(10)}

v = d.keys()

for k in v:
    print(k, v[k])

for k in v:
    print(k, v[k])
```

vdbergh:
> enumerate is an iterator, obviously is has a lazy evaluation.
> 
> enumerate is not a view, as you wrongly stated.

Well this is just a question of terminology. If you iterate over enumerate(list) then it accesses the underlying list. That's why I am thinking it is illegal to modify the underlying list while enumerating over it.

vdbergh:
```python
l=[1,2,3]

for k,v in enumerate(l):
    print(k,v)
    if k==0:
        del l[0]
```
```
0 1
1 3
```
So it is possible to confuse python.

ppigazzini:
A view is a memory proxy, `dict.keys() `is iterable (not an iterator) and has several methods of `set` type.

```python
d = {"a": 1}

# this works because enumerate is an iterator
e = enumerate(a)
next(e)

# this not works because keys is not an iterator
v = d.keys()
next(v)
```


```python
d1 = dict(a=1, b=2, c=3, d=4)
d2 = dict(b=20, d=40, e=50)
d1.keys() & d2.keys()
```

I wrote those scripts expecting to found problems, but appending a new element to a list while another thread is iterating seems to be safe.

vdbergh:
What I am is saying that using enumerate does not provide additional security over directly iterating over the list. It is equivalent.

ppigazzini:
I never, ever wrote that enumerate added safety. I simply wrote in pythonic way the same thing already in master.

vdbergh:
Well it is written everywhere that you should not modify a list while iterating over it. This is not considered pythonic.

ppigazzini:
To be clear, noob thinks that "vps sux", I that "server code sux"

It seems that the iteration on the appended list doesn't crash (I thought that it should crash)

ps: `enumerate` is not a view :)

vdbergh:
> 
> ps: `enumerate` is not a view :)

Well in a database sense it is a view... You are just adding an extra computed field.

ppigazzini:
Ok, in that sense I agree, but in python a `view` is a very different thing, it's an object that has ~ 70% of the methods of a `frozenset`

noobpwnftw:
So although wrong, this for iteration was not the cause of a deadloop occurred during `flush_buffers`. What can be observed is log ends at `dead task`, `run_cache_lock` taken, CPU at 100% for more than 30 minutes with close to none API requests externally.

vdbergh:
Perhaps there were just a lot of dead tasks (a task is declared death if it has send no heartbeats for 3 minutes)? I haven't looked at the code so I don't know if this makes sense.

vondele:
there were definitely 1000s of dead_tasks.

ppigazzini:
It seems that the VPS has slowly recovered the normal CPU load while finishing the runs polluted with the 1000s of dead_tasks.

image

noobpwnftw:
Logs regarding dead tasks stopped before the server still doing deadloops for another 30 minutes, while restarting the server recovers it on spot. Workers were long gone.

vondele:
but could it be we have these tasks marked as dead or dirty and we do something like cleaning it up, flushing to DB once for each dead task?

noobpwnftw:
I still suspect some race instead of just too many work, as they don't appear the same. Mongo was also not busy at all around the time.

vondele:
I'm not going to exclude it, the code looks quite 'tricky' there. What I think could speak against the race is that we're seemingly not triggering it during days/weeks of stable operation under moderate load. If we could deadloop with small probability, I would expect it to happen.

vondele:
BTW, I looked into some tools, and it seems one could get the thread sanitizer to work under python, but I haven't tried it https://github.com/raminfp/sanitizers_python_native_module

ppigazzini:
> BTW, I looked into some tools, and it seems one could get the thread sanitizer to work under python, but I haven't tried it https://github.com/raminfp/sanitizers_python_native_module

From the description they should be sanitizers to debug a python module written in C/C++

PR #1633, opened 2023-04-26
Adjust batch size for SPRT tests.

vondele:
an earlier commit made the batch size for SPRT test 32. This is a good number of single-threaded STC tests, but not for multi-threaded LTC tests. Tests running 60+0.6@8t take too long on an 8core worker to update results. The following patch adjust the number of games, keeping it the same for single threaded STC, but scaling down with TC and threading.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -779,10 +779,27 @@ def validate_form(request):
     data["base_net"] = get_net(data["resolved_base"], data["tests_repo"])
 
     # Integer parameters
+    data["threads"] = int(request.POST["threads"])
+    data["priority"] = int(request.POST["priority"])
+    data["throughput"] = int(request.POST["throughput"])
+
+    if data["threads"] <= 0:
+        raise Exception("Threads must be >= 1")
 
     if stop_rule == "sprt":
         # Too small a number results in many API calls, especially with highly concurrent workers.
-        sprt_batch_size_games = 32
+        # This expression results in 32 games per batch for single threaded STC games.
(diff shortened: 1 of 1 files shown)


xoto10:
Looks reasonable to me.

Should the worker setting for sprt: https://github.com/glinscott/fishtest/blob/0d4736423cbea49fb8ed1c4698cc68746b3943c0/worker/games.py#L1328 use games_concurrency like the line for numgames and spsa does: ?https://github.com/glinscott/fishtest/blob/0d4736423cbea49fb8ed1c4698cc68746b3943c0/worker/games.py#L1318

vondele:
no I think the point is the SPRT tests need to use a fixed batch size, not depending on the concurrency.

xoto10:
Looking at a test, it seems the batch size is already being adjusted for worker thread count somewhere:

https://tests.stockfishchess.org/tests/view/6449328f5d29c10f4f219a0c
image

vondele:
that's a different batch size, i.e. the task size. The batch size this PR adjusts is the one that influence the update_task content. The task size can dependent on the concurrency.

xoto10:
How about `2 * max(5, ...` giving a min of 10 seconds between updates?

Not sure we need 56-core workers to update every 2 seconds? (and higher in future?)

vondele:
I think a 56 cores worker will update about every 15+ seconds with this patch, 32 games each taking roughly 30s divided by 56 cores.

ppigazzini:
Tests on DEV (started this afternoon) look good to me.

ppigazzini (approved):

ppigazzini:

PR #1637, opened 2023-04-27
Fix illegible batch size calc

dubslow:
master is so hard to read that it was faster to fix it than to verify the correctness of the fix

no functional change

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -19,6 +19,7 @@ from fishtest.util import (
     format_results,
     get_chi2,
     get_hash,
+    get_tc_ratio,
     password_strength,
     update_residuals,
 )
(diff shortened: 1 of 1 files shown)


vondele:
convention is that you format your patches with black...

dubslow:
if the autoformatter insists on such nonsense, then better to use intermediate variables to clear it up for humans as well as silly autoformatters.

but that's a moot debate anyways, since the utility function for exactly this purpose already existed, this stupid PR only uses it where it already should have been used

ppigazzini (approved):
TBB (to be blackened)

ppigazzini:

I confess that in the past I was a bit lazy to perform the autoformatting, thanks to your periodic gentle reminders the code is always in a great shape now :)

PR #1644, opened 2023-04-28
Retry failed net downloads

vondele:
on fleet arrival cdn might have troubles to serve the net. This issue only affects some downloads, and usually corrects itself relatively quickly. Retry up to 5 times before throwing the exception.


ppigazzini (approved):
LGTM on DEV and workers.

ppigazzini:
Workers update triggered, thank you @vondele :)

PR #1645, opened 2023-04-29
Fix race

noobpwnftw:
https://github.com/mongodb/mongo-python-driver/blob/bc9029a22879c772e6049b962435fd563636ea92/bson/_cbsonmodule.c#L1465-L1490

I don't think these are GIL bound. So, make a deepcopy?

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -348,7 +348,7 @@ class RunDb:
                     "run": run,
                 }
                 with self.run_cache_write_lock:
-                    self.runs.replace_one({"_id": ObjectId(r_id)}, run)
+                    self.runs.replace_one({"_id": ObjectId(r_id)}, copy.deepcopy(run))
             else:
                 if r_id in self.run_cache:
                     ftime = self.run_cache[r_id]["ftime"]
@@ -396,7 +396,7 @@ class RunDb:
                         if self.scavenge(self.run_cache[r_id]["run"]):
                             with self.run_cache_write_lock:
                                 self.runs.replace_one(
-                                    {"_id": ObjectId(r_id)}, self.run_cache[r_id]["run"]
+                                    {"_id": ObjectId(r_id)}, copy.deepcopy(self.run_cache[r_id]["run"])
                                 )
                     if self.run_cache[r_id]["rtime"] < now - 300:
                         del self.run_cache[r_id]
@@ -412,7 +412,7 @@ class RunDb:
                 # print("SYNC")
                 with self.run_cache_write_lock:
                     self.runs.replace_one(
-                        {"_id": ObjectId(oldest)}, self.run_cache[oldest]["run"]
+                        {"_id": ObjectId(oldest)}, copy.deepcopy(self.run_cache[oldest]["run"])
                     )
         except:
             print("Flush exception", flush=True)


vdbergh:
C-extensions are still under the  GIL unless they give it up explicitly. I don’t know if pymongo does that (not an expert on C-extensions).

noobpwnftw:
GIL is only taken when bytecode is running, and this is not bytecode. Modules are expected to lock it for themselves, and I don't see it anywhere. So, do you think we should whine at them with an issue for not doing the Python object serialization with native code inside GIL so that all assumptions regarding bytecode holds?

vdbergh:
Releasing the GIL is opt in I think

https://docs.python.org/3/c-api/init.html#thread-state-and-the-global-interpreter-lock

One should do
```c++
Py_BEGIN_ALLOW_THREADS
... Do some blocking I/O operation ...
Py_END_ALLOW_THREADS
```

noobpwnftw:
Or Python does it for you https://github.com/python/cpython/blob/738c226786997262b76557d2dadd2beb89ea3fd1/Objects/stringlib/join.h#L120.

vdbergh:
The code you link to is manipulating python objects. It would be a major bug to not do this under the GIL. I believe the pymongo developers are not stupid. So if you think they made a mistake you should produce some evidence for that.

noobpwnftw:
So you obviously haven't read it. Their serialization function sure is called from within GIL, done by Python, to run a bytecode that does the method call, yet it also has plenty of call-outs back to Python methods, like `__get_attr__` as shown, this opens a wide range of possibilities to how serializations can get done and involving things that are not written in their C module, some of them may have implementations that will allow preempting, again, as shown. Now, this is probably fine during execution of orthodox Python bytecodes but probably not in the middle of such a complex serialization.

As for the stupidity, I just couldn't help myself with the desire see how they will react after we whine that their function got stuck due to what we are calling it to write is some volatile data that may see edits during serialization.

xoto10:
If we're editing some data and trying to write it at the same time that seems like our problem?

noobpwnftw:
Additional read:
https://stackoverflow.com/a/42023362

noobpwnftw:
> If we're editing some data and trying to write it at the same time that seems like our problem?

This I don't understand, it's Python, what can go wrong? If in Python C extensions, calling any Python code will never yield the GIL, then this would not have happened. If in PyMongo, they lock things for themselves, or make their C code to be preempt safe, then this would not have happened. Why are we fixing our code for this?

vdbergh:
I think I see your point. The problem is that pymongo is serializing a run object which is potentially being modified at the same time by Python code (the critical part is the task list).

Pymongo itself uses the GIL so it does not expect the objects it is dealing to be changed by Python code (which also uses the GIL).

However potentially the GIL can be released somewhere in the external code that it called by Pymongo, which creates a window during which the run object can indeed be changed by Python code while it is being serialized.

It is worth checking if your deepcopy fixes this potential issue.

noobpwnftw:
I don't know Python enough to tell whether this deepcopy is deep enough, although there should be nobody else editing this object though.

ppigazzini:
PROD updated by hand with the PR, time to summon the p-fleet for a test?

vdbergh:
But the race is fundamentally in the python code, not in pymongo...

noobpwnftw:
We of course can argue that if no C extension is used(they do have a pure python serializer), then this will not cause problems. Their C extension is not preempt safe compared to Python bytecode.

vdbergh:
> We of course can argue that if no C extension is used(they do have a pure python serializer), then this will not cause problems. Their C extension is not preempt safe compared to Python bytecode.

I guess if the c-code hangs then that would be a serious bug. I have not studied the c-code enough to see if that is possible. 

But the result of saving an object that is being modified at the same time is fundamentally undefined. One cannot blame Pymongo for that. I don't think that Pymongo ever claims that database operations are atomic.

noobpwnftw:
So, what is causing load peaks right after 1 hour after workers joined? I don't seem to find relevant code that works at this interval.
Other than that, this fix seems successful if given 100k cores.

noobpwnftw:
After checking logs, it looks like the peaks are workers normally finishing their batch(scaled up) at about the same time, calling flushes more often. And about deepcopy, I haven't read the full C extension code to see why it could get stuck, although it doesn't appear to get stuck now.

maximmasiutin:
Maybe the workers are finishing their jumbo batches. What a pity that an almost entire hour(s) progress of 100000 cores was lost.

noobpwnftw:
I believe this was the cause of previous non-recovering hangs, and now with the fix not reproduced anymore.

Regardless, I'll say again that such bugs are very hard to find due to the fact that our schematics relied on too many GIL assumptions, I think someday if we do the locking ourselves like how people do it in every other language, it will make life easier for everyone.

vdbergh:
Hmm why is this closed?

noobpwnftw:
Because I think people have found a better way: just restart the server whenever shit happens.

PR #1648, opened 2023-04-30
Avoid Useless calculations for /tests_user pages

peregrineshahin:
Since Only 'runs' are taken from
```
       request.rundb.aggregate_unfinished_runs(
            username=username,
        )[0]
```
with index 0, it makes no sense to calculate workers related stuff for this page since they are not shown in the first place.


ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1650, opened 2023-04-30
Regularly purge stale notifications.

vdbergh:
Stale notifications are notifications for finished runs. They are harmless but they needlessly load the server (forever with the current implementation).

Stale notifications can for example be created by subscribing to a finished run in a browser tab that was not refreshed.

Purging will happen every 5.5 hours on average and uses the active_runs api to detect finished but still subscribed runs.

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -169,9 +169,38 @@ function save_timestamp(ts) {
   localStorage.setItem(fishtest_timestamp_key, JSON.stringify(ts));
 }
 
+async function purge_notifications() {
+  let json = {};
+  try {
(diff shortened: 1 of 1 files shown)


vdbergh:
I tested this as follows.

Set the interval to 20s instead of 5.5 hours.
Open Firefox and Chrome and create a new test in one of the browsers. 
Refresh the browsers.
Delete the test in Chrome.
Subscribe in Firefox.
Watch the console to see the purging message appear.

vdbergh:
Drafted. Just saw a bug.

vdbergh:
Undrafted.

ppigazzini (approved):
LGTM on my tests.

ppigazzini:

PR #1653, opened 2023-05-01
Use timer instead of counter for purging stale notifications.

vdbergh:
Using a counter does not really work since it is reset with every browser refresh.


vdbergh:
It was tested in the same way as the previous PR.

vdbergh:
You can see the effect of the PR on dev. There were two stale notifications (they became stale because of the db reload).

Screenshot from 2023-05-01 22-03-31

They have been purged successfully.

ppigazzini (approved):

ppigazzini:
PROD updated, thnak you @vdbergh :)

PR #1654, opened 2023-05-01
Speedup in tests_view for long list of tasks

ppigazzini:
The code was very CPU-intensive for runs with long task lists:
- checking that a dictionary is contained in a list of dictionaries is expensive
- the code performed O(n^2) checks with n the length of the list

Improvements:
- check very rare events first (bad task)
- check if the task is contained in the first list by comparing idx and
  the length of the first list
- some code refactoring

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -395,12 +395,9 @@
     <tbody>
       % for idx, task in enumerate(run['tasks'] + run.get('bad_tasks', [])):
           <%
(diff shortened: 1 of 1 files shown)


vdbergh:
Hmm. So it was my code causing the hang :(

vdbergh:
About your first item. I always thought that `a in lst` does the comparison of `a` with the list elements using `is`. Not so apparently. It uses `==`.

I was thinking how to do the comparison with `is`. Maybe
```python
any(x is a for x in lst)
```

PR #1655, opened 2023-05-01
A try to render tasks table in client side.

peregrineshahin:
A starter try to move the load of this rendering from the server to the client. If successful, this can be improved with on demand rendering or lazy loading.

Note this needs testing and debugging in dev enviroment as I haven't compiled the code.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -8,6 +8,196 @@
     spsa_data = json.dumps(run["args"]["spsa"])
 %>
 
+<script>
+
+  function worker_name(worker_info) {
+    // A user friendly name for the worker.
+    const username = worker_info["username"];
+    const cores = String(worker_info["concurrency"]);
+    const uuid = worker_info["unique_key"];
+    const modified = worker_info["modified"] || false;
+    let name = `${username}-${cores}cores`;
+    
+    if (uuid.length !== 0) {
+      const uuid_split = uuid.split("-");
+      if (uuid_split.length >= 1) {
+        name += "-" + uuid_split[0];
+      }
+      if (uuid_split.length >= 2) {
+        name += "-" + uuid_split[1];
+      }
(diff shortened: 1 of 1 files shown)


ppigazzini:
Mixing Mako and Javascript templating seem to require this syntax.
https://stackoverflow.com/questions/57360613/javascript-template-literals-in-a-mako-template

```html
<script>
    //This is just an example
    let entry = {title: "${some_py_value['some_key']}"}
    //Mako templating
    <%text>
        $('body').append(`$ {entry.title}`);
        //JS templating
    </%text>
</script>
```

I would like to have found this reference some time ago, I remember to have tried some random syntaxes to have this code working...

https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/server/fishtest/templates/tests_live_elo.mak#L11

ppigazzini:
Every task contains information that should be not sent to the users (e.g. IP, unique_key).
I am afraid that the work required to clean up the data (looping the tasks list) defeats the purpose to unload the server.

ppigazzini:
I read more references about how to pass a python object to javascript in mako template.
For a complex python object, the safest way is to serialize the object in json and skip all mako filtering.

https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/server/fishtest/templates/tests_view.mak#L3-L9

https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/server/fishtest/templates/tests_view.mak#L35-L37

These equivalent shortcuts work right now, but they could fail for some type of data in the object (eg if it contains a boolean for the `True/False` vs `true/false` difference)

```js
<script>
    const spsa_data = ${run["args"]["spsa"] | str, n}
<script>

<script>
    const spsa_data = ${str(run["args"]["spsa"]) | n}
<script>
```

PR #1657, opened 2023-05-02
Alleviate race in stale notifications purging.

vdbergh:
Do not purge notifications with recent activity. They should be dealt with in the normal way.


ppigazzini:
DEV updated, but I'm commuting right now. I will be at the keyboard in 45 minutes.

vdbergh:
It was tested in the usual way with an instrumented version….

ppigazzini (approved):

ppigazzini:

PR #1658, opened 2023-05-03
Go through task lists in reverse order to find active runs.

vdbergh:
Active tasks are near the end of the task list so this should give a small speed up.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -470,7 +470,7 @@ class RunDb:
         runs = {"pending": [], "active": []}
         for run in unfinished_runs:
             state = (
-                "active" if any(task["active"] for task in run["tasks"]) else "pending"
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:
I timed this improvement yesterday. The relative speedup of the single instruction  is very high with 25 atcive tests (2-3 magnitude order), unfortunately the db read takes much time (0.1-0.6 seconds). These are the cProfile of the `tests` and `tests_view` ruotes, took some hours ago.
[cprofile_tests.zip](https://github.com/glinscott/fishtest/files/11379873/cprofile_tests.zip)
[cprofile_tests_view.zip](https://github.com/glinscott/fishtest/files/11379883/cprofile_tests_view.zip)

ppigazzini:

PR #1659, opened 2023-05-03
Config option to set worker cap

### LINKED ISSUE #1635: Bug in batch size calculation introduced recently: when there are many workers but few tasks, the task is not properly split between the workers which results in idle workers
There was a bug introduced in /worker/games.py in one of latest commits which results in a problem that a task is not properly split between the workers which results in idle workers when there are many workers but few tasks. The tasks are split to the batches which becomes too large, therefore they are taking very long time to proceed when there are plenty idle workers available. Please consider reverting the code or have this option configured on the server and transmitted to workers.

[        batch_size = games_concurrency * 4 * max(1, round(tc_limit_ltc / tc_limit))](https://github.com/glinscott/fishtest/blob/0d4736423cbea49fb8ed1c4698cc68746b3943c0/worker/games.py#L1318)

```
     # Adjust CPU scaling.
-    _, tc_limit_ltc = adjust_tc("60+0.6", factor)
     scaled_tc, tc_limit = adjust_tc(run["args"]["tc"], factor)
     scaled_new_tc = scaled_tc
     if "new_tc" in run["args"]:
@@ -1313,9 +1312,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file):
         tc_limit *= 2

     while games_remaining > 0:
-        # Update frequency for NumGames/SPSA test:
-        # every 4 games at LTC, or a similar time interval at shorter TCs
-        batch_size = games_concurrency * 4 * max(1, round(tc_limit_ltc / tc_limit))
+        batch_size = games_concurrency * 4  # update frequency
```

P.S. Too large batch sizes issue was also noticed by @Technologov 

The bug is reproducable with SPRT tests, but you can also probably reproduce this bug with an SPSA test:, create an SPSA task with `nodestime=600` and `160+1.6` as described in fishtest FAQ at https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#spsa-tests; have num_games 3000000 and have 1000 machines of workers. It will allocate less than 600 machines, the other machines will be idle unless there are other tasks. With just that single task remaining machines will be idle. However, with the old version of this code when there were no `max(1, round(tc_limit_ltc / tc_limit))` all machines were allocated.

maximmasiutin:
Fixes #1635

Added config option to set number of games cap of a single task for a worker.

There is a `fishtest.ini` config file at `~/`.
```
[worker]
GamesCap = 100
```
This file and this `GamesCap` parameter is not mandatory. By default (if not GamesCap parameter is specified in the config), it is set to 9000 games per task. I.e. the tasks larger than 9000 games will be limited to 9000 games. However, if you set `GamesCap` for example to 100, then the largest batch a worker will receive is 100.

This option is configured on the server, not on the worker.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -11,9 +11,11 @@ import time
 import zlib
 from datetime import datetime, timedelta
 
-import fishtest.stats.stat_util
 from bson.binary import Binary
 from bson.objectid import ObjectId
+from pymongo import DESCENDING, MongoClient
+
+import fishtest.stats.stat_util
 from fishtest.actiondb import ActionDb
 from fishtest.stats.stat_util import SPRT_elo
 from fishtest.userdb import UserDb
@@ -32,13 +34,27 @@ from fishtest.util import (
     worker_name,
 )
 from fishtest.views import del_tasks
-from pymongo import DESCENDING, MongoClient
 
 DEBUG = False
 
 boot_time = datetime.utcnow()
 
 last_rundb = None
+
+server_config = None
+
+
+def get_server_config():
+    global server_config
+    if server_config is None:
+        server_config = configparser.ConfigParser()
+        server_config["worker"] = {
+            "GamesCap": "9000",
+        }
+        config_path = os.path.expanduser("~/fishtest.ini")
+        if os.path.exists(config_path):
+            server_config.read(config_path)
+    return server_config
 
 
 def get_port():
@@ -470,7 +486,9 @@ class RunDb:
         runs = {"pending": [], "active": []}
         for run in unfinished_runs:
             state = (
-                "active" if any(task["active"] for task in reversed(run["tasks"])) else "pending"
+                "active"
+                if any(task["active"] for task in reversed(run["tasks"]))
+                else "pending"
             )
             if state == "pending":
                 run["workers"] = run["cores"] = 0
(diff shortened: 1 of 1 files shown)


Sopel97:
what's the motivation for this feature

maximmasiutin:
1) Whenever you wish to change that on tests.stockfishchess.org, you would not need to modify the source code;
2) Different instances of fishtest would not need the code to be modified; what works well on tests.stockfishchess.org might not work well on other instances, as explained in https://github.com/glinscott/fishtest/issues/1635

maximmasiutin:
You can just set the following in config.ini
```
[worker]
GamesCap = 3200000
```
And you will get the current behavior as on tests.stockfishchess.org, but it will be configurable. I already submitted a few pull requests to get rid of hardcoded values such as captcha key or tests.stockfishchess.org URL which are inapplicable for the other instance and are not compatible with https://github.com/glinscott/fishtest/wiki/Fishtest-server-setup

maximmasiutin:
> but what we have right now works well on fishtest,

at <https://tests.stockfishchess.org/tests/view/642df05c77ff3301150e87f2> there are workers with 56 cores that got 29616 games each. With TC=150+1.5 it will probably takes 7 days to complete 29616 games. So a computer with 56 cores will be dedicated exclusively to one test for the whole week. For example, if you change priority or throughput, the changes will take place only after 7 days.

Also, why should a worker be so greedy to get 29616 games for the whole week in advance? Cannot it get smaller batches, say, every hour?

Sopel97:
This looks to me like an issue with how batch size is calculated with respect to the available memory, not with the absolute limit on the batch size.

Sopel97:
https://github.com/glinscott/fishtest/blob/68faa547232ad9a99f4a8d5acf13f216a744f0c3/server/fishtest/rundb.py#L668

This should not use the number of tasks, because as we can see for the test you linked empty tasks count the same as tasks with thousands of games. The logic is just flawed.

vdbergh:
The reason for this scaling is that mongodb documents have a size upperbound, which was hit in the past. So one wants to limit the total number of tasks, not the total number of games.

PR #1660, opened 2023-05-03
Introduce a signal handler for printing thread stack traces

vondele:
as a debugging aid, introduce a signal handler for printing stack traces

`kill -USR1 <pid>`

will result in a stack dump

needs testing on dev. The signal handler must be installed from the main python thread.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -1,5 +1,9 @@
 import base64
 import hashlib
+import signal
+import sys
+import threading
+import traceback
 from pathlib import Path
 
 from fishtest.rundb import RunDb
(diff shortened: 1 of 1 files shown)


server/fishtest/__init__.py
```diff
@@ -8,12 +8,23 @@
 from pyramid.config import Configurator
 from pyramid.events import BeforeRender, NewRequest
 from pyramid.session import SignedCookieSessionFactory
-
 from fishtest import helpers
+import traceback, signal, sys
+
+
+def thread_stack_dump(sig, frame):
+    for th in threading.enumerate():
```
ppigazzini, line 16: Reading from mobile, if I'm not wrong we should import `threading`

vondele:
updated to current master. Add a flush for better understandable output, and catch exceptions that might happen if threads do no longer exist. 

Good for merge IMO.

ppigazzini (approved):
Works fine on PROD

ppigazzini:
PROD updated, thank you @vondele

PR #1663, opened 2023-05-03
Inject tasks.

peregrineshahin:
Try implement alternative solution (vs https://github.com/glinscott/fishtest/pull/1655) for task list rendering via injecting html.

drafted at the moment until I can build this

### DIFF
--- a/server/fishtest/templates/tasks.mak
+++ b/server/fishtest/templates/tasks.mak
@@ -1,0 +1,82 @@
+<%!
+  from fishtest.util import worker_name
+%>
+
+% for idx, task in enumerate(run['tasks'] + run.get('bad_tasks', [])):
+          <%
+            if 'bad' in task and idx < len(run['tasks']):
+              continue
+            task_id = task.get('task_id', idx)
+            stats = task.get('stats', {})
+            if 'stats' in task:
+              total = stats['wins'] + stats['losses'] + stats['draws']
+            else:
+              continue
+
+            if task_id == show_task:
+              active_style = 'highlight'
+            elif task['active']:
+              active_style = 'info'
+            else:
+              active_style = ''
+          %>
+          <tr class="${active_style}" id=task${task_id}>
+            <td>
+              <a href=${f"/api/pgn/{run['_id']}-{task_id:d}.pgn"}>${task_id}</a>
+            </td>
+            % if 'bad' in task:
+                <td style="text-decoration:line-through; background-color:#ffebeb">
+            % else:
+                <td>
+            % endif
+            % if approver and task['worker_info']['username'] != "Unknown_worker":
+                <a href="/user/${task['worker_info']['username']}">
+                  ${worker_name(task['worker_info'])}
+                </a>
+            % elif 'worker_info' in task:
+                ${worker_name(task["worker_info"])}
+            % else:
+                -
+            % endif
+            </td>
+            <td>
+            <%
+               gcc_version = ".".join([str(m) for m in task['worker_info']['gcc_version']])
+               compiler = task['worker_info'].get('compiler', 'g++')
+               python_version = ".".join([str(m) for m in task['worker_info']['python_version']])
+               version = task['worker_info']['version']
+               ARCH = task['worker_info']['ARCH']
+            %>
+               os: ${task['worker_info']['uname']};
+               ram: ${task['worker_info']['max_memory']}MiB;
+               compiler: ${compiler} ${gcc_version};
+               python: ${python_version};
+               worker: ${version};
+               arch: ${ARCH}
+            </td>
+            <td>${str(task.get('last_updated', '-')).split('.')[0]}</td>
+            <td>${f"{total:03d} / {task['num_games']:03d}"}</td>
+            % if 'pentanomial' not in run['results']:
+                <td>${stats.get('wins', '-')}</td>
+                <td>${stats.get('losses', '-')}</td>
+                <td>${stats.get('draws', '-')}</td>
+            % else:
+                <%
+                  p=stats.get('pentanomial',5*[0])
+                %>
+                <td>[${p[0]},&nbsp;${p[1]},&nbsp;${p[2]},&nbsp;${p[3]},&nbsp;${p[4]}]</td>
+            % endif
+            <td>${stats.get('crashes', '-')}</td>
+            <td>${stats.get('time_losses', '-')}</td>
+
+            % if 'spsa' not in run['args']:
+                % if 'residual' in task and task['residual']!=float("inf"):
+                    <td style="background-color:${task['residual_color']}">
+                      ${f"{task['residual']:.3f}"}
+                    </td>
+                % else:
+                    <td>-</td>
+                % endif
+            % endif
+          </tr>
+% endfor
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -1,8 +1,6 @@
 <%inherit file="base.mak"/>
 
 <%
-  from fishtest.util import worker_name
-  
   if 'spsa' in run['args']:
     import json
     spsa_data = json.dumps(run["args"]["spsa"])
@@ -346,22 +344,11 @@
   <pre id="diff-contents"><code class="diff"></code></pre>
 </div>
 
-<script>
-    function toggle_tasks() {
-      const button = document.getElementById("tasks-button");
-      const div = document.getElementById("tasks");
-      const active = button.textContent.trim() === 'Hide';
-      button.textContent = active ? 'Show' : 'Hide';
-      document.cookie =
-        'tasks_state' + '=' + button.textContent.trim() + ";max-age=${60 * 60 * 24 * 365 * 10};SameSite=Lax;";
-    }
-</script>
-
 <h4>
     <a
       id="tasks-button" class="btn btn-sm btn-light border" 
       data-bs-toggle="collapse" href="#tasks" role="button" aria-expanded="false"
-      aria-controls="tasks" onclick="toggle_tasks()"
+      aria-controls="tasks"
     >
       ${'Hide' if tasks_shown else 'Show'}
     </a>
@@ -392,80 +379,7 @@
         % endif
       </tr>
     </thead>
-    <tbody>
-      % for idx, task in enumerate(run['tasks'] + run.get('bad_tasks', [])):
-          <%
-            if 'bad' in task and idx < len(run['tasks']):
-              continue
-            task_id = task.get('task_id', idx)
-            stats = task.get('stats', {})
-            if 'stats' in task:
-              total = stats['wins'] + stats['losses'] + stats['draws']
-            else:
-              continue
-
-            if task_id == show_task:
-              active_style = 'highlight'
-            elif task['active']:
-              active_style = 'info'
-            else:
-              active_style = ''
-          %>
-          <tr class="${active_style}" id=task${task_id}>
-            <td><a href=${f"/api/pgn/{run['_id']}-{task_id:d}.pgn"}>${task_id}</a></td>
-            % if 'bad' in task:
-                <td style="text-decoration:line-through; background-color:#ffebeb">
-            % else:
-                <td>
-            % endif
-            % if approver and task['worker_info']['username'] != "Unknown_worker":
-                <a href="/user/${task['worker_info']['username']}">${worker_name(task['worker_info'])}</a>
-            % elif 'worker_info' in task:
-                ${worker_name(task["worker_info"])}
-            % else:
-                -
-            % endif
-            </td>
-            <td>
-            <%
-               gcc_version = ".".join([str(m) for m in task['worker_info']['gcc_version']])
-               compiler = task['worker_info'].get('compiler', 'g++')
-               python_version = ".".join([str(m) for m in task['worker_info']['python_version']])
-               version = task['worker_info']['version']
-               ARCH = task['worker_info']['ARCH']
-            %>
-               os: ${task['worker_info']['uname']};
-               ram: ${task['worker_info']['max_memory']}MiB;
-               compiler: ${compiler} ${gcc_version};
-               python: ${python_version};
-               worker: ${version};
-               arch: ${ARCH}
-            </td>
-            <td>${str(task.get('last_updated', '-')).split('.')[0]}</td>
-            <td>${f"{total:03d} / {task['num_games']:03d}"}</td>
-            % if 'pentanomial' not in run['results']:
-                <td>${stats.get('wins', '-')}</td>
-                <td>${stats.get('losses', '-')}</td>
-                <td>${stats.get('draws', '-')}</td>
-            % else:
-                <%
-                  p=stats.get('pentanomial',5*[0])
-                %>
-                <td>[${p[0]},&nbsp;${p[1]},&nbsp;${p[2]},&nbsp;${p[3]},&nbsp;${p[4]}]</td>
-            % endif
-            <td>${stats.get('crashes', '-')}</td>
-            <td>${stats.get('time_losses', '-')}</td>
-
-            % if 'spsa' not in run['args']:
-                % if 'residual' in task and task['residual']!=float("inf"):
-                    <td style="background-color:${task['residual_color']}">${f"{task['residual']:.3f}"}</td>
-                % else:
-                    <td>-</td>
-                % endif
-            % endif
-          </tr>
-      % endfor
-    </tbody>
+    <tbody id="tasks-body"></tbody>
   </table>
 </div>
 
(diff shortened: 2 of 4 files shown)


vdbergh:
https://github.com/glinscott/fishtest/blob/dcb092d2208784843e5a2778f1c1ccc629afe51b/server/fishtest/rundb.py#L324-L330

There is an issue with this code I think. As the above code snipped shows, the projected run will be saved in the cache and then it will even find its way to the db.

I think that the first thing to do, before making any other architectural changes, is to make `tests/view` query the db directly (i.e. eliminate `get_run()`). In that way `tests/view` can be served by a secondary instance so that it will not interfere with the primary function of fishtest.

peregrineshahin:
> https://github.com/glinscott/fishtest/blob/dcb092d2208784843e5a2778f1c1ccc629afe51b/server/fishtest/rundb.py#L324-L330
> 
> There is an issue with this code I think. As the above code snipped shows, the projected run will be saved in the cache and then it will even find its way to the db.
> 
> I think that the first thing to do, before making any other architectural changes, is to make `tests/view` query the db directly (i.e. eliminate `get_run()`). In that way `tests/view` can be served by a secondary instance so that it will not interfere with the primary function of fishtest.

Hmm In that since I assume my mistake is excluding tasks and bad_tasks from the projection, so I should keep them.
Now If get_results depend on replacing db results I don't see how one can query the db with no stale results, we are now depending on the user to help us get the results by them requesting the render and then saving to db.

If this process isn't done in the background we will for sure get stale results.

vdbergh:
There is no such thing as stale results. Only delayed results. The worker accumulates at least 15 min of work before informing the server. What does it matter if the user sees the results with a few extra seconds of delay?

vondele:
> There is no such thing as stale results. Only delayed results. The worker accumulates at least 15 min of work before informing the server. What does it matter if the user sees the results with a few extra seconds of delay?

I'm not sure where the 15min estimate comes from. update_task is called quite frequently (I estimate workers call this every 2min on average).

ppigazzini:
> > There is no such thing as stale results. Only delayed results. The worker accumulates at least 15 min of work before informing the server. What does it matter if the user sees the results with a few extra seconds of delay?
> 
> I'm not sure where the 15min estimate comes from. update_task is called quite frequently (I estimate workers call this every 2min on average).

On DEV (with master code) a SPRT STC test gets updates every 6 minutes (every 32 games with a batch of 160 games)

vdbergh:
> > There is no such thing as stale results. Only delayed results. The worker accumulates at least 15 min of work before informing the server. What does it matter if the user sees the results with a few extra seconds of delay?
> 
> I'm not sure where the 15min estimate comes from. update_task is called quite frequently (I estimate workers call this every 2min on average).

Sorry not 15 minutes but say 2 minutes.

vdbergh:
> > There is no such thing as stale results. Only delayed results. The worker accumulates at least 15 min of work before informing the server. What does it matter if the user sees the results with a few extra seconds of delay?
> 
> What matters is people will see updated results in tests/ and test/user but stale results in tests/view I cannot handle the load of whines then for the delays in discord lol.

tests and tests/view would  see the same results.

server/fishtest/templates/tasks.mak
```diff
@@ -0,0 +1,77 @@
+<%
```
ppigazzini, line 1: The import doesn't depend on the context. better to use the module level block instead than the python block, see:
https://docs.makotemplates.org/en/latest/syntax.html#module-level-blocks
https://docs.makotemplates.org/en/latest/syntax.html#python-blocks

```html
<%!
  from fishtest.util import worker_name
%>
```

server/fishtest/templates/tests_view.mak
```diff
@@ -347,21 +345,58 @@
...
+        await renderTasks();
+    }
+
+    async function renderTasks() {
+      await DOM_loaded();
+      const tasksBody = document.getElementById("tasks-body");
+      if (tasksBody.innerHTML.length > 0)
+         return;
+
+      fetch('/tasks/${str(run['_id'])}?show_task=${show_task}')
```
ppigazzini, line 364: this is mako or js templating?

ppigazzini:
DEV updated.

server/fishtest/templates/tests_view.mak
peregrineshahin: Mako

server/fishtest/templates/tests_view.mak
```diff
@@ -347,21 +345,58 @@
...
 
 <script>
-    function toggle_tasks() {
+    async function handleRenderOnLoad(){
+      await DOM_loaded();
+      const tasksButton = document.getElementById("tasks-button");
+      tasksButton?.addEventListener("click", async () => {
+        await toggle_tasks();
+      })
+      if (${tasks_shown})
```
ppigazzini, line 354: python `False` vs js `false`
perhaps this works:
```js
 if (${str(tasks_shown).lower()})
```

ppigazzini:
```
May 04 12:56:58 dfts-0 pserve[29250]:   File "/home/usr00/fishtest/server/fishtest/templates/tasks.mak", line 55, in render_body
May 04 12:56:58 dfts-0 pserve[29250]:     % if 'pentanomial' not in run['results']:
May 04 12:56:58 dfts-0 pserve[29250]: KeyError: 'results'

```

ppigazzini:
@peregrineshahin I wonder if we are exposing on the new route the tasks without before dropping the reserved info (eg IP, unique_key)

peregrineshahin:
I don't understand your question.

ppigazzini:
DEV updated, it's starting to work
https://dfts-0.pigazzini.it/tasks/644bfb62f609e0c9f1579ec0

ppigazzini:
DEV updated, it seems to work :)

https://dfts-0.pigazzini.it/tests/view/644bfb62f609e0c9f1579ec0

ppigazzini:
DEV updated, next step lazy loading the table content while scrolling :P

ppigazzini:
If the new route `tasks` depends on `get_run()` we must move it on 6543 in nginx config.

ppigazzini:
DEV updated.

vondele:
> If the new route `tasks` depends on `get_run()` we must move it on 6543 in nginx config.

we should really try to separate the worker related apis from the 'human' apis, so that even if an expensive api is triggered by a human the workers are not really impacted. Of course, things can only be moved as they get ready.

ppigazzini:
DEV updated, same binomial distribution unfortunately.

ppigazzini:
DEV updated, LGTM

vdbergh:
> > If the new route `tasks` depends on `get_run()` we must move it on 6543 in nginx config.
> 
> we should really try to separate the worker related apis from the 'human' apis, so that even if an expensive api is triggered by a human the workers are not really impacted. Of course, things can only be moved as they get ready.

Yes we should make `tests/view` `get_run` free so that it can be run in a separate instance (like `tests` and I think `tests/users`). Then the primary function if fishtest would never be impacted.

ppigazzini:
DEV updated, still LGTM

vondele:
> Yes we should make `tests/view` `get_run` free (a trivial thing to do) so that it can be run in a secondary instance (like `tests` and I think `tests/users`). Then the primary function of fishtest would never be impacted.

It would be nice if we could somehow introduce an assert for this, i.e. ensure in `get_run` that it is only called via APIs that explicitly declare they need 'live' data.

vdbergh:
It still seems to be the case that the projected run is saved in the cache. I am I misreading the code?

peregrineshahin:
> It still seems to be the case that the projected run is saved in the cache. Am I misreading the code?

No I don't think wo the projected run for the tasks is not saved in the cache because there no get_result to call around this run for the new tasks route.

peregrineshahin:
For the test_view route everything stays the same as before

peregrineshahin:
Oh oh you are correct, so I should add if statement

vdbergh:
What about just getting the data directly from the db?

peregrineshahin:
> What about just getting the data directly from the db?

Yes that's cleaner

peregrineshahin:
But what if the results are stale/delayed?
I think not, I just read without writing to cache anything for now

vdbergh:
> But what if the results are stale/delayed? I think not, I just read without writing to cache anything for now

Stale/delayed compared to what?

peregrineshahin:
> > But what if the results are stale/delayed? I think not, I just read without writing to cache anything for now
> 
> Stale/delayed compared to what?

compared the cached values.

vondele:
I think getting tasks directly from the db to display on the web pages is OK. That's what will allow us to decouple this from the 'main pserve'.

vdbergh:
Who uses the cached values besides tests/view now?

vdbergh:
Did anyone ever complain about the main page being stale/delayed?

peregrineshahin:
I pushed to only read from db, please review if this is the way to do it.

ppigazzini:
DEV updated.

vondele:
seems to work well for me, with very similar performance to the cache based version for 'the bad case' https://dfts-0.pigazzini.it/tests/view/642df05c77ff3301150e87f2

vdbergh:
I wonder if there could be some kind of wait cursor (or another visual clue) while the tasks are loading? If you click the show/hide button nothing seems to be happening for several seconds https://dfts-0.pigazzini.it/tests/view/642df05c77ff3301150e87f2 which feels a bit strange compared to the earlier behaviour.

ppigazzini:
> I wonder if there could be some kind of wait cursor (or another visual clue) while the tasks are loading? If you click the show/hide button nothing seems to be happening for several seconds https://dfts-0.pigazzini.it/tests/view/642df05c77ff3301150e87f2 which feels a bit strange compared to the earlier behaviour.

For the SPSA graph we have a preloader

https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/server/fishtest/static/js/spsa.js#L198-L204

https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/server/fishtest/static/css/application.css#L470-L474

peregrineshahin:
If we were waiting for the fetch (i.e the response from the server) it would be okay to add a loading gif or something along these lines, but here we are waiting for the browser process to add/show the very long task list to/from the DOM and adding the giff would even overload it even more, so in general, I don't like loading cursors because they mean accepting defeat, I would like to introduce a lazy loading render instead in the near future.

vdbergh:
But the current version on DEV is unusable on an iPad. The button does not react, or is maybe in the wrong state … I don’t know. It never seems to get to a state where the button reacts instantly as before.

vdbergh:
I checked on my laptop and there the button reacts a bit quicker (so it is usable there). But iPad seems  a disaster.

vondele:
The link above links to a test with 10k+ tasks, the server needs at least 2s to send the task list, which happens after pressing the button. After that the browser still needs some time to display the list...

vdbergh:
Ok it works on iPad too but it is extremely slow. This is what happens

- Push button "Show".
- Nothing happens for a while ... (the button does not change into `Hide`).
- Finally after about 5 the button changes and the tasks are there.
- Push button "Hide"
- Sometimes the iPad reacts instantly but sometimes not.
- The button changes in the "Hide" and the tasks are hidden.

If you repeat this, the rendering always takes a long time (once the tasks are loaded you would at least expect rendering to be instant).

I was using this test https://dfts-0.pigazzini.it/tests/view/644bfb62f609e0c9f1579ec0 .

vdbergh:
Actually this test  https://tests.stockfishchess.org/tests/view/644bfb62f609e0c9f1579ec0 does not do well on iPad on PROD either (buttons react very slowly). It is funny that I never noticed this before.

I guess iPad is just very slow in rendering long tables...

ppigazzini:
DEV updated.

peregrineshahin:
> DEV updated.

I think this should be it for this PR, I see behaviors as intended.

vdbergh:
It is funny that people talk a lot about mobile but never about iPads. But everyone I know has an iPad. Am I in a different universe?

peregrineshahin:
> It is funny that people talk a lot about mobile but never about iPads. But everyone I know has an iPad. Am I in a different universe?

I'm afraid you convinced them to buy it :D.
But yeah iPad I think is no different than mobile both have a little slowdown rendering long lists than a PC of course.

Though perhaps PROD will be faster than DEV, as I said I will introduce lazy loading eventually.

vdbergh:
> > It is funny that people talk a lot about mobile but never about iPads. But everyone I know has an iPad. Am I in a different universe?
> 
> I'm afraid you convinced them to buy it :D. 

No. Actually on principle I don't buy Apple products  (so no Iphone, Mac, AppleWatch etc...) because I don't like their walled garden philosophy.

But for the note taking apps on iPad there appears to be no viable substitute on other platforms. 


But yeah iPad I think is no different than mobile both have a little slowdown rendering long lists than a PC of course (Assuming both are the same age :P).
> 
> Though perhaps PROD will be faster than DEV, as I said I will introduce lazy loading eventually.

peregrineshahin:
I need to format some js inside the mako also.

peregrineshahin:
Done.

ppigazzini:
PR applied by hand to PROD to monitor the CPU usage with the direct DB read

vdbergh:
On PROD it seems very fast, even on an iPad.

vdbergh:
Maybe it is sensitive to network speed? I think I have a fast network at home. But probably at work it is even faster.

vdbergh:
No it is not on my end.

 https://tests.stockfishchess.org/tests/view/644bfb62f609e0c9f1579ec0 

is slow on DEV on iPad.

is fast on PROD on iPad.

Does PROD have a faster network connection perhaps?

peregrineshahin:
Sorry not db but maybe the mako in anyways I will further improve upon this with lazy fetching

vondele:
The task API is below 2s typically
```
                         /tests       66     36.088      0.547      0.037      1.367
                   /tests/view/      189     32.385      0.171      0.008      0.829
               /api/update_task     3778     22.850      0.006      0.001      0.457
                      /api/beat     4527     17.792      0.004      0.000      0.399
                        /tasks/       18     15.159      0.842      0.004      2.104
                   /api/actions      472      6.873      0.015      0.002      0.212
               /api/active_runs       22      6.402      0.291      0.114      0.876
                /api/upload_pgn      188      4.828      0.026      0.003      0.444
                /tests/finished       14      3.850      0.275      0.158      0.373
           /api/request_version      356      3.485      0.010      0.000      0.412
              /api/request_task      238      2.683      0.011      0.003      0.227
                   /tests/user/        4      1.565      0.391      0.353      0.463
```

vdbergh:
> It's insanely slow here though https://tests.stockfishchess.org/tests/view/644bfb62f609e0c9f1579ec0
> 
> Db takes eternity to send the response......

This takes 5s for me on an iPad. But it has an insane amount of tasks.

vondele:
```
$ time wget https://tests.stockfishchess.org/tasks/644bfb62f609e0c9f1579ec0
--2023-05-05 16:58:04--  https://tests.stockfishchess.org/tasks/644bfb62f609e0c9f1579ec0
Resolving tests.stockfishchess.org (tests.stockfishchess.org)... 185.83.217.134
Connecting to tests.stockfishchess.org (tests.stockfishchess.org)|185.83.217.134|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 6024025 (5.7M) [text/html]
Saving to: ‘644bfb62f609e0c9f1579ec0’

644bfb62f609e0c9f1579ec0                         100%[========================================================================================================>]   5.74M  --.-KB/s    in 0.1s    

2023-05-05 16:58:07 (49.7 MB/s) - ‘644bfb62f609e0c9f1579ec0’ saved [6024025/6024025]


real	0m2.225s
user	0m0.013s
sys	0m0.017s
```
It is a 5.7MB document, so yeah, could take some time to display.

vdbergh:
It’s strange that querying a single run takes about the same amount of time as querying all the unfinished runs.

ppigazzini:
> No it is not on my end.
> 
> https://tests.stockfishchess.org/tests/view/644bfb62f609e0c9f1579ec0
> 
> is slow on DEV on iPad.
> 
> is fast on PROD on iPad.
> 
> Does PROD have a faster network connection perhaps?

The IPS of DEV is mostly down today, a DDOS attack I suppose.

ppigazzini:
PROD updated.

ppigazzini:
PROD updated.

vdbergh:
The PR seems to be working well. I see no difference with the prior behavior.

vondele:
yes, for the record, /tests/view/ must be on the primary (worker) pserve, while the tasks/ are now served from the auxiliary pserve. Now, loading the page puts 'minimal load' on the worker pserve, while the expensive tasks part can run without causing issues for the workers..

ppigazzini:
@peregrineshahin PROD updated. Let me know when this PR can be merged, I think that last commits were only about code formatting.

peregrineshahin:
I think it can be merged as is now.

ppigazzini (approved):

ppigazzini:
Thank you @peregrineshahin :)

PR #1664, opened 2023-05-04
Extra defense against race after wake up from sleep.

vdbergh:
I noticed that after sleep all tabs wake up which occasionally may lead to multiple fetches and possibly duplicate notifications, if many tabs are open.
    
We use a standard trick to defend against this.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -146,6 +146,8 @@ function supportsNotifications() {
 }
 
 function notify(title, body, link, fallback) {
+  // Instrumentation
+  console.log("Notification: title=" + title + " body=" + body);
   if (supportsNotifications() && Notification.permission === "granted") {
     const notification = new Notification(title, {
       body: body,
@@ -157,8 +159,6 @@ function notify(title, body, link, fallback) {
     };
   } else if (fallback) {
     fallback(title, body, link);
-  } else {
-    console.log("Notification: title=" + title + " body=" + body);
   }
 }
 
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -129,7 +129,6 @@ function notify_elo(entry_start, entry_finished) {
     const message = `<a class="notification-message" href=${link}>${escapeHtml(
       title
     )} ${escapeHtml(body)}</a>`;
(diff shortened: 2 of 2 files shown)


vdbergh:
I managed to trigger this code by opening and closing my laptop every ~ 30s with many tabs open.  It is triggered rarely but this is to be expected.

vdbergh:
I added a bit more console instrumentation:

1.  A message when a wake up from sleep is detected (this normally happens in all tabs).
2.  A message when the main code of this PR is triggered for a tab (this will happen rarely).
3.  A message when the tab wants to display a notification (if everything goes well, this should only happen in a single tab).

ppigazzini (approved):
Tested saving/restoring a checkpoint of a VM.

ppigazzini:

vdbergh:
Some things I am seeing on Chrome on Linux.

1. Tabs that are not active are severely throttled. A 20ms second (asynchronous) sleep can be much longer.
2. The milliseconds part of the time (obtained by Date.now()) is severely quantized.

1. Does not seem to be a problem but I will have to do something about 2. since I was using the milliseconds part as a source of randomness.

vdbergh:
https://developer.chrome.com/blog/timer-throttling-in-chrome-88/

vdbergh:
That page has some interesting alternatives for the current polling model. E.g. "server events" which I had never heard off.

ppigazzini:
All browsers seems to have a sleeping mode for inactive tabs:
https://www.ghacks.net/2021/01/26/which-browser-has-the-best-tab-snoozing-sleeping-freezing-implementation/

vdbergh:
Well if they are really sleeping then it is as if they aren't there. The effect of severe throttling is a bit more difficult to understand.

Server events seem tempting but it looks as if one needs a connection per tab which does not look so good.

vdbergh:
> Well if they are really sleeping then it is as if they aren't there. The effect of severe throttling is a bit more difficult to understand.
> 
> Server events seem tempting but it looks as if one needs a connection per tab which does not look so good.

Actually there is only one tab that needs the connection. The server can refuse the connection from other tabs on the same browser. 

Still one would need one connection per browser. I don't know if that is acceptable.

PR #1665, opened 2023-05-04
Use unittest as test runner for github CI

ppigazzini:
Setuptools is deprecated as test runner, see

https://setuptools.pypa.io/en/latest/deprecated/commands.html#test-build-package-and-run-a-unittest-suite

Retire all unnecessary files for github CI


ppigazzini:
@vdbergh @vondele this version of the PR is github CI oriented, so it retires all the unnecessary files. For the worker tests it uses the local packages. However, it's easy to refactor the PR to keep the helper scripts and/or create a worker/tests folder. Let me know your point of view.

vondele:
Also I don't run the local tests, and rely on github CI instead. Not sure this is the optimal workflow, but it is quite convenient.

ppigazzini:
The server packages installation has started randomly failing in the CI, in this PR I use the standard way to setup a project `python -m pip install .`, that I use every time to setup DEV/PROD (in editable mode to change the files and restart the service `python -m pip install -e .`)
The deprecated `setup.py test` under the hood already uses `unittest`, I simply removed the unnecessary code.
IMO `tox` or `nox` are over killer for our test framework.

vdbergh:
I don’t think I ever used the scripts. So I guess the change will not affect me.

Issue #1666, opened 2023-05-05
Delete binaries cache on worker startup


vondele:
As people have now the option to change the compiler used by the worker, it is possible that SF binaries generated by different compilers are kept in the testing_dir. This could result an incorrect test with binaries generated by different compilers. The easiest solution is to delete the cache of binaries at worker startup.

I'll have a look at that solution.

vdbergh:
A binary generated by a different compiler is still a valid binary IMHO. If the binary is broken it will be detected further down the line.

Also: the server remembers which binaries the worker has (because of the rate limit). If the worker systematically deletes binaries at start up, this feature will not work anymore.

vondele:
The binary is not broken, but might have a different speed. E.g. master could be a gcc binary that is 10% slower than a clang binary. Any test would pass on this worker. This would be just a one-time clean at worker startup.

vdbergh:
Ok. Actually my other concern isn't valid because if the worker is restarted its UUID changes. The prefix stays the same but this is not used as the key.

Doesn't the SF binary know the compiler it was compiled with? Or could the compiler command line be incorporated in the hash? That would allow for a less incisive change.

vondele:
some information about the compiler is in the binary, but we might not have the details. It seems less error prone to just reset the cache. Having to compile a binary is a relatively small overhead compared to how infrequent it is to restart the worker.

PR #1667, opened 2023-05-05
Remove cached binaries on worker startup

vondele:
as the user could change the compiler at worker startup time (e.g. clang vs gcc), the binary cache might contain binaries compiled with a different compiler than the ones generated by the worker. As different compilers lead to binaries of different speed, test results could be biased, unless we assure that the same compiler is used for both base and new test.

This patch just clears the binary cache once at worker startup, leaving the caching mechanism unchanged otherwise.

Possibly the binary cache should also be flushed on worker update (e.g. the architecture selection logic might change). This is not implemented.

fixes https://github.com/glinscott/fishtest/issues/1666

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 210
+WORKER_VERSION = 211
 
 
 def validate_request(request):
--- a/worker/games.py
+++ b/worker/games.py
@@ -1094,7 +1094,7 @@ def launch_cutechess(
     return task_alive
 
 
-def run_games(worker_info, password, remote, run, task_id, pgn_file):
+def run_games(worker_info, password, remote, run, task_id, pgn_file, clear_binaries):
     # This is the main cutechess-cli driver.
     # It is ok, and even expected, for this function to
     # raise exceptions, implicitly or explicitly, if a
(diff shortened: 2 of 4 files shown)


vondele:
If was looking at flushing the cache also after a worker update, but was somewhat confused by https://github.com/glinscott/fishtest/blob/52a42301c2214b5c8004362fcef95ec46f9596c0/worker/worker.py#L1282-L1283

will this always print the failure message? I might not know how exactly the worker update goes.

vdbergh:
No because a successful update does not return (as written in the comment…).

vondele:
OK, so for the purpose of this patch, a worker update can be treated as a fresh start of the worker.

vdbergh:
What if the user updates his system compiler while the worker is running? Not an unlikely scenario with auto updates…

vondele:
not impossible, but usually these are rare and patch level updates, i.e. wrong code bug fixes that have typically no real performance impact. I'm mostly worried about the clang / gcc switch that the user can access.

We could look at the compiler version, but that wouldn't capture the assembler or the linker, or libc, or lpthreads, or anything else that might have an impact on the generated binary.

ppigazzini (approved):
LG with DEV and windows/linux workers

ppigazzini:
> Possibly the binary cache should also be flushed on worker update (e.g. the architecture selection logic might change). This is not implemented.

The worker update moves the `testing` folder to start from scratch with engines, books, nets and `cutechess-cli` binary. It even deletes the old engine binaries and nets to free up the HD space.

https://github.com/glinscott/fishtest/blob/d00af797fb4cc1b813a81af21b8d119acf799b47/worker/updater.py#L65-L85

ppigazzini:
I triggered the workers update, I will keep on eye on noob's workers, last time they got stuck during the update.
Thank you @vondele :)

Issue #1668, opened 2023-05-06
Update_task..


vondele:
Note sure if this is an issue or expected behavior.

Once a run finishes, workers that call update_task leave a message in /var/syslog that they are not active:
```
May  6 06:10:54 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/213 is not active
May  6 06:11:04 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/216 is not active
May  6 06:11:24 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/211 is not active
May  6 06:11:37 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/212 is not active
May  6 06:15:03 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/214 is not active
May  6 06:15:23 tests pserve[21941]: Update_task: task 645536ffa3695638d17a6003/217 is not active
```

I believe this is just expected behavior, and probably should not be logged. The comment in the code above suggests this might be something wrong:
```python
        # First some sanity checks on the update
        # If something is wrong we return early.

        # task["active"]=True means that a worker should be working on this task.
        # Tasks are created as "active" and become "not active" when they
        # are finished, or when the worker goes offline.
```

vdbergh:
Yes it is expected behavior. When an SPRT run finishes all tasks are set to non-active but there are usually still workers processing tasks. When they try to send their updates to the server, the server will reject them. There seems to be no need to log this.

We could try to upgrade the beat api to make the workers stop faster but since the beat api is now called only every two minutes, it is probably not worth it.

vdbergh:
Of course when the fleet visits and all runs get massive numbers of workers assigned to them, it would be good if we could make these workers stop in a way that does not require them to bombard the server with useless updates.

vondele:
If think the number of updates is fine. Probably just not logging this if the run is finished would be enough. Let me see if I can get a PR in.

PR #1669, opened 2023-05-06
Avoid logging a frequent, expected event

vondele:
once a run is finished we expect workers to update_task, where they will be signalled that they can quit working on this task.

fixes https://github.com/glinscott/fishtest/issues/1668

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -470,7 +470,9 @@ class RunDb:
         runs = {"pending": [], "active": []}
         for run in unfinished_runs:
             state = (
-                "active" if any(task["active"] for task in reversed(run["tasks"])) else "pending"
+                "active"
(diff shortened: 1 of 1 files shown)


ppigazzini:
Tested on DEV, applied to PROD with `x/update_fishtest.sh`

vondele:
Thanks... looking at this, should I update to also get rid of `Request_task: refresh queue` ? 

On the other hand, that could serve as 'a heartbeat' of the server.

ppigazzini:
It's up to you, I search the `pserve` logs only for exceptions.
However, the VPS rotates the logs, so there is not a space issue.

vondele:
I guess that term is nice to keep as a heartbeat. So the PR is good as is.

ppigazzini:
Merged, thank you @vondele :)

PR #1670, opened 2023-05-06
First shot at incremental results updates.

vdbergh:
For each task_update this PR computes an incremental update and compares the result to get_results.

If there is a difference it throws an assertion error.


vdbergh:
I closed this PR since apparently @xoto10 is already working on this.

PR #1671, opened 2023-05-06
Update results stats incrementally

xoto10:
Looping through all the tasks in get_results() gets very expensive when called from update_task() with a large fleet of workers on fishtest.
This patch removes the get_results() call from update_task() and only uses the info in the updated task to increment the run results.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1091,7 +1091,20 @@ class RunDb:
             task["active"] = False
             return {"task_alive": False, "error": error}
 
-        # The update seems fine. Update run["tasks"][task_id] (=task).
+        # The update seems fine.
+
+        # Increment run results before overwriting task["stats"]
+
+        for key, value in stats.items():
+            if key == "pentanomial":
+                run["results"][key] = [
+                    x + y - z
+                    for x, y, z in zip(run["results"][key], value, task["stats"].get(key, [0] * 5))
+                ]
+            else:
+                run["results"][key] += value - task["stats"].get(key, 0)
+
+        # Update run["tasks"][task_id] (=task).
 
         task["stats"] = stats
         task["last_updated"] = update_time
@@ -1115,15 +1128,9 @@ class RunDb:
                 run["cores"] -= task["worker_info"]["concurrency"]
                 assert run["cores"] >= 0
 
-        run["results_stale"] = True  # force recalculation of results
-        updated_results = self.get_results(
-            run, False
-        )  # computed from run["tasks"] which
-        # has just been updated. Sets run["results_stale"]=False.
-
         if "sprt" in run["args"]:
             sprt = run["args"]["sprt"]
-            fishtest.stats.stat_util.update_SPRT(updated_results, sprt)
+            fishtest.stats.stat_util.update_SPRT(run["results"], sprt)
             if sprt["state"] != "":
                 task_finished = True
                 task["active"] = False
@@ -1139,7 +1146,7 @@ class RunDb:
         # Check if the run is finished.
 
         run_finished = False
-        if count_games(updated_results) >= run["args"]["num_games"]:
+        if count_games(run["results"]) >= run["args"]["num_games"]:
             run_finished = True
         elif "sprt" in run["args"] and sprt["state"] != "":
             run_finished = True
@@ -1147,7 +1154,8 @@ class RunDb:
         # Return.
 
         if run_finished:
-            self.buffer(run, True)
+            self.check_results(run, run_id, task_id)
+
             self.stop_run(run_id)
             # stop run may not actually stop a run because of autopurging!
             if run["finished"]:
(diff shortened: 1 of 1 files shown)


vdbergh:
I think it would be a good idea to check that the result of the incremental updates is the same as the result of summing the tasks (when the run finishes).

server/fishtest/rundb.py
```diff
@@ -1163,6 +1175,46 @@ def count_games(d):
...
+            or len(new["pentanomial"]) < 5
+        ):
+            info = "Check_results: task {}/{} pentanomial length results mismatch: {}/{}".format(
+                run_id,
+                task_id,
+                len(old.get("pentanomial", [])),
+                len(new.get("pentanomial", [])),
+            )
+            print(info, flush=True)
+        else:
```
ppigazzini, line 1210: This should be equivalent.
```python
        else:
            for i, (old_value, new_value) in enumerate(
                zip(old["pentanomial"], new["pentanomial"])
            ):
                if old_value != new_value:
                    info = "Check_results: task {}/{} pentanomial value {} results mismatch: {}/{}".format(
                        run_id, task_id, i, old_value, new_value
                    )
                    print(info, flush=True)
```

server/fishtest/rundb.py
```diff
@@ -1091,7 +1093,22 @@ def count_games(d):
             task["active"] = False
             return {"task_alive": False, "error": error}
 
-        # The update seems fine. Update run["tasks"][task_id] (=task).
+        # The update seems fine.
+
+        # Increment run results before overwriting task["stats"]
```
ppigazzini, line 1098: This should be equivalent (btw iterate with the built-in dict iterator `for s in stats:` it's faster than iterate on the view of the keys `for s in stats.key():`).
```python
for key, value in stats.items():
    if key == "pentanomial":
        run["results"][key] = [
            x + y - z
            for x, y, z in zip(run["results"][key], value, task["stats"].get(key, [0] * 5))
        ]
    else:
        run["results"][key] += value - task["stats"].get(key, 0)
```

ppigazzini (commented):
IIRC this PR is a workout with python :)

xoto10:
Yes, I was aware there are several different ways to code these bits, and probably my code was not the most pythonic. I will look at your suggestions and probably come back with some questions ...

server/fishtest/rundb.py
xoto10: Yes, I like this.

I should have remembered that `.items()` was an option, and it works nicely here, e.g. just makes that last line a bit simpler.

I was looking at zip() more when I coded `check_results()`. When I was coding this pentanomial branch, it seemed that zip would be more complicated so I just did my loop thing.

server/fishtest/rundb.py
xoto10: Yes, nice again.

I misread it when I looked earlier and thought I might disagree, but ofc you are right.

Thanks for your help! :-)

ppigazzini:
DEV has been running with this PR since last night, no errors in the logs.

vondele:
I had a further look in the code, I believe we can almost get rid of the `run["results_stale"]`  flag. I think the only place where there is some non-trivial code related to it would be `purge_run`, which can indeed recompute the results based on the task list without harm, but this wouldn't need this flag.

vdbergh:
> I had a further look in the code, I believe we can almost get rid of the `run["results_stale"]` flag. I think the only place where there is some non-trivial code related to it would be `purge_run`, which can indeed recompute the results based on the task list without harm, but this wouldn't need this flag. This change should not necessarily be part of this PR, just mentioning it.

Yes it would be much better if get_results were a simple function to recompute the results from scratch. When that is required.

vondele:
well, get_results() could be a simple getter function, with a new recompute_results() being explicit in what it does (and essentially equivalent to the current get_results() with the stale check removed).

peregrineshahin:
> well, get_results() could be a simple getter function, with a new recompute_results() being explicit in what it does (and essentially equivalent to the current get_results() with the stale check removed).
> 
> EDIT: I agree with not updating the run object in place... this would facilitate using it on the non-primary pserves.

BTW we got rid of using get_results() for finished runs also in this PR
https://github.com/glinscott/fishtest/pull/1675/files#diff-b39bc06fda86fb55166cd3acd5f75bc21428b8d1bf95d6e4b268efdef2300c11L1369

vdbergh:
@vondele Yes I think it looks correct  (I think there is only one place where the results are updated and it is done under lock). And since the result of the incremental updates is verified when the run is finished, a bug will be quickly discovered.

vondele:
So, no errors so far on PROD. Timings already look a little better:
image
maybe 20% improvement now. Right now, there are not really tests with long task lists, and since this is now constant cost with task list size, we know that this gain will be easily 10x under fleet load.

peregrineshahin:
Did anybody test purging runs with this patch?
beware purge_run uses get_results to update the new results after the run is finished

peregrineshahin:
Oh I see get_results function wasn't remove/changed, but would be nice if all cases tested.

vondele:
I did try that on PROD, this looks fine. Meanwhile the log is still clean on PROD.

There is a pre-existing issue (i.e. both master and patch), restarting very old runs that don't have pentanomial results stored. I.e. update_task tries to update those.

vdbergh:
> I did try that on PROD, this looks fine. Meanwhile the log is still clean on PROD.
> 
> There is a pre-existing issue (i.e. both master and patch), restarting very old runs that don't have pentanomial results stored. I.e. update_task tries to update those. This might not be worth the effort to fix it.

There should be a time limit on restarting runs (like there is for purging),

vondele:
I wasn't aware of the purge time limit, but yes, I think that makes sense also for 'modify'.

peregrineshahin:
@xoto10 is the function that checks results only for debugging purposes?
If so would you mind removing it?

vondele:
It might not be bad to have it as a sanity check for a while. Since it is only called when the run stops, the overhead of it is very small.

server/fishtest/rundb.py
```diff
@@ -1163,6 +1171,48 @@ def count_games(d):
 
         return ret
 
+    def check_results(self, run, run_id, task_id):
+        old = run["results"]
```
peregrineshahin, line 1175: It seems that 'old' and 'new' refer to the same dict "results", if we are comparing the thing with itself it surely will bring up clean logs :)

vondele: so we better repeat the testing with a copy.deepcopy() somewhere.

vdbergh: get_results makes a new copy of run["results"].

peregrineshahin: No it doesn't 

        run["results_stale"] = False
        run["results"] = results
        if save_run:
            self.buffer(run, True)

        return results
        
        you are returning a new copy but you assigned results already here to the same old reference

vdbergh:
```
run["results"] = results
```
The right hand side is a freshly built object.

peregrineshahin: which we reflect on the old variable old = run["results"]
So you are comparing new vs new

vdbergh:
```python
old = run["results"]    # old contains reference to current results object
run["results"] = results # RHS has been built from scratch
new = run["results"]  # new contains reference to new results object
```

vdbergh: Run this test program
```python
run={}
run["results"]={'a':1}
old=run["results"]
run["results"]={'a':2}
new=run["results"]
print(old,new)
```
The result is
```python
{'a': 1} {'a': 2}
```

peregrineshahin: Okay now I understand as long as we assign the run["results"] to be a different object no reflection, but if we modify it they there will be reflection.
I'm new to python so bare with me a bit :P

vdbergh: So this has nothing to do with Python. It is general computer language semantics.

peregrineshahin: Well, yes but new is a reserved keyword ;P

peregrineshahin: I seem to delete the first comment instead of marking this resolved (which it seems I can't).
Well, all in all if this works let's move forward with it..

ppigazzini:

PR #1673, opened 2023-05-07
Fix api_active_runs

### LINKED ISSUE #1672: api_active_runs does extra unnecessary stuff.
Now that get_unfinished_runs always queries stuff from the db, it makes no sense to query all fields from db and then strip the data to only wanted fields, rather the logic should be to no stripping runs and query only the stuff we want to show from the first place.

peregrineshahin:
Fixes #1672

I would like further to remove the addition below as the existence of such a key with empty data is very improbable that is used by people who use this API, please let me know what you think.
 
```
            run["tasks"] = []
            run["bad_tasks"] = []
            if "args" in run and "spsa" in run["args"]:
               if "param_history" not in run["args"]["spsa"]:
                 run["args"]["spsa"]["param_history"] = []
```

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -260,9 +260,16 @@ class ApiView(object):
 
     @view_config(route_name="api_active_runs")
     def active_runs(self):
+        runs = self.request.rundb.runs.find(
+             {"finished": False},
+             {"tasks": 0, "bad_tasks": 0, "args.spsa.param_history": 0})
         active = {}
-        for run in self.request.rundb.get_unfinished_runs():
-            active[str(run["_id"])] = strip_run(run)
+        for run in runs:
+            # some string conversions
+            run["_id"] = str(run["_id"])
+            run["start_time"] = str(run["start_time"])
+            run["last_updated"] = str(run["last_updated"])
+            active[str(run["_id"])] = run
         return active
 
     @view_config(route_name="api_actions")


vdbergh:
```
sort=[("last_updated", DESCENDING)])
```
The corresponding index is no longer sorted by "last_updated". So this seems to just create  unnecessary work for Mongodb as the users of this api typically will want to sort the list themselves.

peregrineshahin:
> ```
> sort=[("last_updated", DESCENDING)])
> ```
> 
> The corresponding index is no longer sorted by "last_updated". So this seems to just create unnecessary work for Mongodb as the users of this api typically will want to sort the list themselves.

Makes sense, removed.

vondele:
some of the conditionals in the python code could be to deal with fields that don't exist in the database. Will Mongo be fine if "args.spsa.param_history" doesn't exist?

peregrineshahin:
> some of the conditionals in the python code could be to deal with fields that don't exist in the database. Will Mongo be fine if "args.spsa.param_history" doesn't exist?

I'm only 96% percent sure that it is the case.
chatGPT says:
```
MongoDB will simply ignore any fields that are specified in the projection but do not exist in the documents being queried. In this case, if args.spsa or args.spsa.param_history does not exist in the documents, MongoDB will simply skip them in the projection and return the remaining fields.
```

server/fishtest/api.py
```diff
@@ -260,9 +260,16 @@ def get_country_code(self):
 
     @view_config(route_name="api_active_runs")
     def active_runs(self):
+        runs = self.request.rundb.runs.find(
+             {"finished": False},
+             {"tasks": 0, "bad_tasks": 0, "args.spsa.param_history": 0})
```
ppigazzini, line 265: IMO better to return or the expected empty container or None:

```python
# empty container
{"tasks": [], "bad_tasks": [], "args.spsa.param_history": []})
# None
{"tasks": None, "bad_tasks": None, "args.spsa.param_history": None})
```

ppigazzini:
I read in MongoDB documentation that that is the projection to skip fields and not the returned values, code comment deleted.

peregrineshahin:
Yes, my idea of not returning those is that the API does not query or return these fields as opposed to the API queries' unfinished runs but gives a wrong piece of information that the run has no tasks while this is not true.

ppigazzini:
I wonder if, after testing this PR, this can be done in a more general way in `apy.py` module or `rundb.py` module.

peregrineshahin:
Thus generally REST APIs are to get the desired/correct data as opposed to knowing the structure of the database behind the scene.

peregrineshahin:
> Yes, my idea of not returning those is that the API does not query or return these fields as opposed to the API queries' unfinished runs but gives a wrong piece of information that the run has no tasks while this is not true.

Sure.

ppigazzini:
DEV updated.

vdbergh:
Roughly twice as fast as before here.
```sh
time wget https://dfts-0.pigazzini.it/api/active_runs

real	0m0,416s
user	0m0,009s
sys	0m0,019s
```

ppigazzini:
3x id benched on DEV averages: PR: 0m0.065s vs  master: 0.0m0.200s

vdbergh:
I saw you editing out the extra zero in 0.0200...

ppigazzini:
> I saw you editing out the extra zero in 0.0200...

False! This was the extra zero edited.
> vs master: 0.0m0.200s

peregrineshahin:
> I wonder if, after testing this PR, this can be done in a more general way in `apy.py` module or, better, in `rundb.py` module.

Hmm after some thought I think the current state is the cleanest way to write it. Others, feel free to override me.

ppigazzini (approved):

ppigazzini:

vondele:
I'm not really observing any difference on prod. Did the server get restarted after this update?

ppigazzini:
Yes, server restarted 52 minutes ago (UTC time).

```
May 07 17:32:29 tests.stockfishchess.org pserve[18767]: flush
May 07 17:32:29 tests.stockfishchess.org systemd[1]: Stopping Fishtest Server port 6543...
May 07 17:32:30 tests.stockfishchess.org pserve[18767]: .......................done
May 07 17:32:30 tests.stockfishchess.org systemd[1]: Stopped Fishtest Server port 6543.
May 07 17:33:04 tests.stockfishchess.org systemd[1]: Started Fishtest Server port 6543.
May 07 17:33:05 tests.stockfishchess.org pserve[21673]: Starting server in PID 21673.
May 07 17:33:06 tests.stockfishchess.org pserve[21673]: /api/update_task: request is not json encoded
May 07 17:33:06 tests.stockfishchess.org pserve[21673]: Failed_task: failure for: https://tests.stockfishchess.org/tests/view/64578924eb75932ccfec061d, task_id: 37, worker: ChessDBCN-16cores-57fd25b6-cc16, reason: 'Too many failed update
May 07 17:33:08 tests.stockfishchess.org pserve[21673]: Request_task: no useful cached runs left
```

vondele:
looks like the PR is not merged?

ppigazzini:
My fault, I spproved but not merged the PR...

vondele:
nice 
image
look for the last few accesses in the graph

ppigazzini:
Who is using this API?

vdbergh:
I am. To purge stale notifications.

vondele:
some external scripts are pulling this (e.g. to keep track of the LLR of running tests) . (And the last final points were some wgets from myself).

vondele:
some more before and after:

```
# logging from [07/May/2023:16:34:55 +0000] to [07/May/2023:17:01:46 +0000] ########
               /api/active_runs       45     15.976      0.355      0.132      0.927
# logging from [07/May/2023:18:30:26 +0000] to [07/May/2023:18:50:17 +0000] ########
               /api/active_runs       42      2.133      0.051      0.024      0.099
```

BTW, if we could do something similar for finished_runs this would be great. I think the main requirement would be to ensure that get_results() would never need to go through the task list for runs coming from the db (or in general in fact).

PR #1674, opened 2023-05-07
Add a bit of randomness to current_time.

vdbergh:
We deal with some issues observed on Chrome/Linux.

- Tabs that are not active are severely throttled. A 20 second (asynchronous) sleep can last much longer. Thus the console becomes flooded with "Wake up from sleep" messages. This should be fixed in this PR.

- The milliseconds part of the time (obtained by `Date.now()`) is quantized. So it cannot be used as a simple source of randomness (see the code how it is used).

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -241,16 +241,20 @@ async function main_follow_loop() {
   await DOM_loaded();
   await async_sleep(5000 + 10000 * Math.random());
   while (true) {
-    const current_time = Date.now();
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

ppigazzini:

The new wiki server setup script writes the updated nginx configuration. On PROD nginx is doing the redirection of the FQDN to the home page, I'm waiting to merge this branch that drop that redirection from the python server.
https://github.com/glinscott/fishtest/compare/master...ppigazzini:fishtest:retire_home_route

PR #1676, opened 2023-05-09
More robust wake up from sleep.

vdbergh:
I occasionally still observe duplicate notifications on ipad after wake up. I think the reason is that one tab may sleep before a fetch and another thread may sleep during the "async_sleep(1500)". This presents a race.

We catch this by checking the actual time slept. If it is > 90s then we cancel the fetch.

One would like to reduce the 90s but this would be defeated by the fact that timers on Chrome fire only once per minute in inactive tabs.

There is also some nice refactoring. The function async_sleep can now be used as
```
let t = await async_sleep(delay_ms);
```
where t is the amount of time (in ms) that is actually slept.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -228,7 +228,11 @@ function DOM_loaded() {
 
 function async_sleep(ms) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


vdbergh:
Thx!

PR #1677, opened 2023-05-09
Use fetch_text() from application.js instead of raw fetch.

vdbergh:
This shortens the code a bit and fetch_text() has moreover a bit more error checking built in.


ppigazzini (approved):

ppigazzini:

PR #1681, opened 2023-05-10
Disable modifying old runs

vondele:
as for purging, limit modifying to runs not older than 30 days. As suggested https://github.com/glinscott/fishtest/pull/1671#issuecomment-1542600011


ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1686, opened 2023-05-13
Remove most console.log messages.

vdbergh:
Also: fix possible race. Currently if the browser sleeps after fetching and before purging (unlikely, but...) then some notifications may be lost.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -1,3 +1,14 @@
+// A console log with time stamp and optional stack trace
+function console_log(message, trace) {
+  const d = new Date().toISOString();
+  const message_ = `${d}: ${message}`;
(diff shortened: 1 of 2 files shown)


ppigazzini:
DEV updated.

ppigazzini:
PROD updated by hand.

vdbergh:
Thanks! Is it possible to keep it on DEV for a bit? I created a stale notification and I am waiting for it to be purged!

ppigazzini:
Yes, I'm playing on my PC with the new pyramid configuration.

vdbergh:
Works!

vdbergh:
Latest push was just a small tweak.

ppigazzini (approved):

ppigazzini:

PR #1690, opened 2023-05-16
Make follow/unfollow atomic on platforms supporting web locks.

vdbergh:


vdbergh:
Web locks are promises. They must be since javascript cannot block. So you must `await` for the lock. However this means that if you use web locks to protect a low level shared data structure, then virtually every function in your script must become async. Nothing wrong with that but it feels a bit strange.

ppigazzini:
DEV updated.

vdbergh:
To my surprise it does not work. Following/unfollowing works correctly but no notifications are generated. I am drafting this.

vdbergh:
I tried a quick fix (I seem to have misunderstood how web locks return values).

vdbergh:
Working!

vdbergh:
Web locks allow for an easy, fast and robust leader election. It seems tempting to go from the current distributed model (all tabs are equivalent) to a leader model where only the leader manages the shared state, communicates with the server and creates notifications. The other tabs communicate with the leader via a form of asynchronous message passing.

ppigazzini:

PR #1694, opened 2023-05-19
Simplify html escaping in mako

ppigazzini:
- the string is already html escaped in "view.py", simplify the escaping in mako template
- python3 string supports unicode, drop unnecessary try except


ppigazzini:
Tested working creating some tests with:
 - unicode characters in branch name, commit message and test info field
 - html tags in commit message and test info field

image

Issue #1693, opened 2023-05-19
game counts on users page not correct?


vondele:
I wanted to update the CPU contributors list based https://tests.stockfishchess.org/users (by the same script I used previously) and it seems for some users (e.g. technologov) the numbers of games played decreased. In principle the number of games played by a user should be exact?

```diff
diff --git a/Top CPU Contributors.txt b/Top CPU Contributors.txt
index 30c963d7c..b13f79dc3 100644
--- a/Top CPU Contributors.txt  
+++ b/Top CPU Contributors.txt  
@@ -1,161 +1,166 @@
-Contributors to Fishtest with >10,000 CPU hours, as of 2022-11-19.
+Contributors to Fishtest with >10,000 CPU hours, as of 2023-05-19.
 Thank you!
 
 Username                                CPU Hours     Games played
 ------------------------------------------------------------------
-noobpwnftw                               36475307       2748033975
-technologov                              14570711        760073590
+noobpwnftw                               37190604       2821544067
+technologov                              13071604        697078252
+linrock                                   3973304        269633305
 mlang                                     3026000        200065824
-dew                                       1689222        100034318
-grandphish2                               1442171         86798057
-okrout                                    1439985        133471766
-pemo                                      1405374         44189811
-linrock                                   1299003         28382783
-TueRens                                   1163420         71159522
-JojoM                                      897158         55177114
+dew                                       1689162        100033738
+okrout                                    1515907        142725352
+grandphish2                               1459364         91364265
+pemo                                      1458782         46775413
+TueRens                                   1171537         69505744
+JojoM                                      931638         60288112
 tvijlbrief                                 796125         51897690
 mibere                                     703840         46867607
-gvreuls                                    635982         40652394
-oz                                         590763         41201352
-sebastronomy                               581517         23307132
-cw                                         517915         34865769
-fastgm                                     504266         30264740
```

peregrineshahin:
Yes somene introduced a bug and they don't want to look what caused this for at least 3 months now.....

peregrineshahin:
Current solution is to just rebuild it again some "just resta(r)t" thing ;)

ppigazzini:
> Yes somene introduced a bug and they don't want to look what caused this for at least 3 months now.....

Yeah, at example someone changed few days ago the return values of the `aggregate_unfinished_runs()` breaking `delta_update_users.py` again...

```diff
- return (runs, machines, pending_hours, cores, nps)
[...]

PR #1695, opened 2023-05-19
Fixes the part that got altered in delta users.

peregrineshahin:
Fixes: https://github.com/glinscott/fishtest/issues/1693

(needs confirming if this is the only part that creates the issue).

### DIFF
--- a/server/utils/delta_update_users.py
+++ b/server/utils/delta_update_users.py
@@ -189,13 +189,7 @@ def update_users():
         rundb.deltas.delete_many({})
         rundb.deltas.insert_many([{k: v} for k, v in new_deltas.items()])
 
-    (
-        __runs,
-        machines,
-        __pending_hours,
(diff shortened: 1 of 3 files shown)


server/utils/delta_update_users.py
```diff
@@ -189,13 +189,7 @@ def update_users():
...
         rundb.deltas.insert_many([{k: v} for k, v in new_deltas.items()])
 
-    (
-        __runs,
-        machines,
-        __pending_hours,
-        __cores,
-        __nps,
-    ) = rundb.aggregate_unfinished_runs()
+    machines = rundb.get_machines()
```
ppigazzini, line 192: It's used two times in the script:
```diff
- machines = rundb.get_machines()
+ machines = list(rundb.get_machines())
```

ppigazzini:

PR #1697, opened 2023-05-20
Wrap saving/loading objects from localstorage.

vdbergh:
This makes the code in notifications.js a bit more readable.

Functions:

load_object(key)
save_object(key, value)

The functions JSON.stringify() and JSON.parse() are applied under the hood.

When saving to localstorare, keys are give a prefix __fishtest__. This makes upgrading easier since we can just change the prefix to something else, e.g. __fishtest_v2__.

After this PR all tabs should be reloaded.


ppigazzini:
DEV updated.

peregrineshahin:
Hi vdberg, 
probably unrelated to this PR but related to notification.js.. Could we benifit from the idea of nginx cache we applied on machines route and use Get instead of the Posts requests here?

If it works. Then certainly we have a new speedup.

vdbergh:
@peregrineshahin 

/api/active_runs could indeed be a GET request. I will change it. but it is only called every 4.5 hours so will not make a lot of difference.

/api/actions is harder because it just forwards mongodb queries. The queries depends on the runs that the user (=browser) is following.  I suspect that this is different for every user so there is little to be gained from caching with the current setup.

vdbergh:
Ok /api/active_runs is now a GET request.

vondele:
> /api/actions is harder because it just forwards mongodb queries. The queries depend on the runs that the user (=browser) is following. I suspect that this is different for every user so there is little to be gained from caching with the current setup.

looks like the number of requests is quite a bit bigger than the number of runs, so I suspect that some people follow quite a bit. E.g. following LTC might be quite common. If there is no need for a post (i.e. modify server state), I think it would be better to use a get.

ppigazzini:
DEV updated.

vdbergh:
The query is not per run but for set of runs. Every user is following a set of run_ids and this set is part of the mongodb query. I think most users will have a different set so caching will not help with the current setup.

vdbergh:
This is how the api is called (it is used at other places to with different queries, but this is the main engine).
```js
 if (notifications.count()) {
        json = await fetch_post("/api/actions", {
          action: { $in: ["finished_run", "stop_run", "delete_run"] },
          run_id: { $in: notifications.toArray() },
        });
      }
```

vondele:
ok, I see, yes, that's likely to be unique per user.

vdbergh:
I guess one could write a wrapper which % encodes the JSON and passes it as a GET parameter but it will look quite ugly.

vondele:
I guess if one would query per run_id one could probably get a cachable response. It might not be worth the effort.

vdbergh:
To make it cacheable one should probably periodically bulk download a recent part of the actiondb (which can then be cached) instead of doing queries tailored per user. But it requires some serious rethinking.

vdbergh:
Actually I am reading here that POST requests that do not change server state _are_ cacheable. 
https://snippets.bentasker.co.uk/page-2003170906-Caching-POST-requests-in-NGinx-NGinx.html

One reason for preferring POST over GET is that the former is more convenient if the parameters are not easily converted to strings.

vondele:
yes, it appears possible, but it is not the default. I think also for good reason (i.e. post requests are supposed to change the server state). Nevertheless, even if cacheable, only identical calls would be cached.

IMO, I think one could asks the actions db to return e.g. all recently (say last 25 or 100, or last 2h) finished (stopped/deleted/etc) runs and client side use that data.

vdbergh:
> yes, it appears possible, but it is not the default. I think also for good reason (i.e. post requests are supposed to change the server state). 

Well there are other good reasons for using POST requests. They make it more convenient to send data to the server (the data is sent as body rather than being shoehorned into URL format with all the resulting quoting issues). One should also use them if some arguments are secret (this does not apply here).
> 
> IMO, I think one could asks the actions db to return e.g. all recently (say last 25 or 100, or last 2h) finished (stopped/deleted/etc) runs and client side use that data.
 
Yes something like that. For mobile users one should be careful not to download too much data on a regular basis. This is something to keep in mind. With the current setup `/api/actions` returns `[]` most of the time.

One also needs to take into account that the user may go off line for a longer time (e.g. sleeping). So at start up the user would still have to make a non-cached call to synchronize. `/api/actions` uses few resources as it is (and is served by a secondary instance) so I am rather hesitant to uglify the code for little gain.

vondele:
yes, I agree we can leave as is... a lot of time already went in these notifications. However, I guess keeping in mind what we've learned designing this feature for the next one is probably good.

vdbergh:
I drafted this since I discovered a bug (unrelated to this PR) with stopped/deleted runs. The user is given as the user who did the stopping/deleting rather than the owner of the run (usually these are the same but not always). I will fix this tonight.

vdbergh:
I think I fixed it.

vdbergh:
@ppigazzini  Can you maybe put the new version on DEV? Then I can have a look to see if it is fixed.

ppigazzini:
> @ppigazzini Can you maybe put the new version on DEV? Then I can have a look to see if it is fixed.

Done.

ppigazzini:

PR #1698, opened 2023-05-21
Fix wrong colors after purging.

peregrineshahin:
See
https://tests.stockfishchess.org/tests/view/644d1ddb68e01d8194cddf35?show_task=90

https://tests.stockfishchess.org/api/get_run/644d1ddb68e01d8194cddf35

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1410,10 +1410,9 @@ class RunDb:
                 run["is_yellow"] = False
             else:
                 # Copied code. Must be refactored.
-                if run["results_info"]["style"] == "#44EB44":
-                    run["is_green"] = True
-                elif run["results_info"]["style"] == "yellow":
-                    run["is_yellow"] = True
+                style = run["results_info"]["style"]
+                run["is_green"] = style == "#44EB44"
+                run["is_yellow"] = style == "yellow"
             self.buffer(run, True)
 
         return message


vdbergh:
Seems a legit bugfix.

But probably, as the comment says, we should refactor determining the state of a run - this is needed in several places. Obtaining it from the style of the Elo widget is very fragile (the style sometimes changes).  

Not per se this PR of course.

vdbergh:
There are several ` tasks = copy.copy(run["tasks"])` which appear to be not necessary in `purge_run()`. 

I think this is my fault. In the past this code would delete bad tasks from the task list (and hence one had to iterate over a copy).

But deleting bad tasks is bad because then the task_id's change and hence the task links in the event log become incorrect.

So I changed it, but it seems I forgot to delete some copy statements. Needs to be double checked of course.

server/fishtest/rundb.py
```diff
@@ -1412,8 +1412,13 @@ def purge_run(self, run, p=0.001, res=7.0, iters=1):
                 # Copied code. Must be refactored.
                 if run["results_info"]["style"] == "#44EB44":
                     run["is_green"] = True
+                    run["is_yellow"] = False
```
ppigazzini, line 1415: The comment called for a refactoring :)

```diff
-                if run["results_info"]["style"] == "#44EB44":
-                    run["is_green"] = True
-                    run["is_yellow"] = False
-                elif run["results_info"]["style"] == "yellow":
-                    run["is_green"] = False
-                    run["is_yellow"] = True
-                else:
-                    run["is_green"] = False
-                    run["is_yellow"] = False

+                style = run["results_info"]["style"]
+                run["is_green"] = style == "#44EB44"
+                run["is_yellow"] = style == "yellow"
```

ppigazzini:
pythonic gone wrong.
```python
 u=(t:=time.time(),b:=copy.deepcopy(a),print(time.time()-t),print(b == a),print(b is a))
```

peregrineshahin:
I mean the hexa-colors are saved into DB for runs ("very unwise practice"), and we are doomed in refactoring this to actual text ... it's a pity also to have a PROD downtime while converting the database just for test colors.....

vdbergh:
> I mean the hexa-colors are saved into DB for runs ("very unwise practice"), and we are doomed in refactoring this to actual text ..

Well we can ignore the style that's on the db and we can compute directly the is_green/is_yellow status (I don't like these flags BTW). That's what I meant by refactoring.

peregrineshahin:
> > I mean the hexa-colors are saved into DB for runs ("very unwise practice"), and we are doomed in refactoring this to actual text ..
> 
> Well we can ignore the style that's on the db and we can compute directly the is_green/is_yellow status (I don't like these flags BTW). That's what I meant by refactoring.

That probably doable for runs, but not sure about residual colors having no flags.
Recomputing might be pain.

vdbergh:
> > > I mean the hexa-colors are saved into DB for runs ("very unwise practice"), and we are doomed in refactoring this to actual text ..
> > 
> > 
> > Well we can ignore the style that's on the db and we can compute directly the is_green/is_yellow status (I don't like these flags BTW). That's what I meant by refactoring.
> 
> That probably doable for runs, but not sure about residual colors having no flags. Recomputing might be pain.

Well I can look into it (it should not be difficult).

vdbergh:
Of course meanwhile the current PR can be committed since it fixes a genuine bug...

ppigazzini (approved):

ppigazzini:

vdbergh:
Actually `is_yellow` and `is_green` really do refer to the styling of the results widget. So it is ok to get it from there. The bad thing is, as @peregrineshahin  pointed out above, is that this styling has hard coded hex colors. 

What are the drawbacks of replacing the hex colors by the strings `"green","yellow","red","blue","none"` and then to use classes (for example) to do the styling?

It means (until db conversion) that we cannot use `run["results_info"]` for old runs when displaying the results widget, but we have to call `format_results()` instead. But this is not a big deal.

PR #1699, opened 2023-05-22
Replace +- by ± in Elo output

UniQP:
Also use consistent spacing around ±.

When loading finished runs, we now always re-format the results because the stored results might still contain +-.

### DIFF
--- a/server/fishtest/templates/elo_results.mak
+++ b/server/fishtest/templates/elo_results.mak
@@ -64,7 +64,7 @@
   % endfor
   % if elo_ptnml_run:
     <br>
-    ${f"nElo: {nelo5:.2f} +-{nelo5_delta:.1f} (95%) PairsRatio: {results5_pairs_ratio:.2f}"}
+    ${f"nElo: {nelo5:.2f} ± {nelo5_delta:.1f} (95%) PairsRatio: {results5_pairs_ratio:.2f}"}
   % endif
 </%def>
 
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -266,7 +266,7 @@ def format_results(run_results, run):
             elo, elo95, los = fishtest.stats.stat_util.get_elo([WLD[1], WLD[2], WLD[0]])
 
         # Display the results
-        eloInfo = "Elo: {:.2f} +-{:.1f} (95%)".format(elo, elo95)
+        eloInfo = "Elo: {:.2f} ± {:.1f} (95%)".format(elo, elo95)
         losInfo = "LOS: {:.1%}".format(los)
 
         result["info"].append(eloInfo + " " + losInfo)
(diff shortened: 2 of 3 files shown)


UniQP:
Current Elo output on master:
image

Elo output with this change (manually edited HTML for demonstration purpose):
image

peregrineshahin:
Cool one!

UniQP:
Thanks for testing. I adapted the commit to use the unicode characters directly.

> Here it's working only for the last line, to be investigated, https://dfts-0.pigazzini.it/tests/finished?ltc_only=1
> 
> image

Sounds like the master version of `util.py` was used in this case.

ppigazzini:
I restarted the 3 instances of pserve but still the same result. I will check the code this night when at the keyboard.

vondele:
will we still be able to copy and paste this in the git commit comments, and in all tools that read and parse those? Quite a few things might still rely on ascii.

ppigazzini:
I don't know about parsing scripts, but the regression wiki page already use the unicode character:

image

and markdown github works fine:
```
Elo: -0.37 ± 1.2 (95%) LOS: 27.9%
Total: 60000 W: 16101 L: 16165 D: 27734
Ptnml(0-2): 24, 5900, 18212, 5844, 20
nElo: -0.83 ± 2.8 (95%) PairsRatio: 0.99
```

ppigazzini:
git and github support unicode, I wrote a branch name and a git message using only unicode character to test a PR. see the page on DEV https://dfts-0.pigazzini.it/tests/view/6468d0d8cb62647546b4f732 while sill live and here the screenshot https://github.com/glinscott/fishtest/pull/1694#issuecomment-1555920590

UniQP:
Yeah, the commit of this PR uses the `±` character as well.

@vondele Do you know any scripts that parse the output?

vondele:
I was thinking about for example https://abrok.eu/stockfish/ (the red added Elo is after parsing I assume):
image
but that parsed line is not affected.

ppigazzini:
???

image

image

vondele:
yes, the Elo calculation for SPRT runs on abrok is not using the SPRT-adjusted formula, and is thus somewhat inaccurate.

UniQP:
> Here it's working only for the last line, to be investigated, https://dfts-0.pigazzini.it/tests/finished?ltc_only=1
> 
> image

I took a closer look. The Elo is part of the result info string that is stored in the database. Thus, it currently shows "+-" for older runs. Would it be OK to re-format the result info every time? If so, we could just remove the condition here: https://github.com/glinscott/fishtest/blob/88ece9fc73e6adc8a0ffc5af1db2c09cbc2830d6/server/fishtest/views.py#L1374-L1375

For the test view, we already re-format the result info (but we have only one result, not 25): https://github.com/glinscott/fishtest/blob/88ece9fc73e6adc8a0ffc5af1db2c09cbc2830d6/server/fishtest/views.py#L1204

ppigazzini (approved):
LGTM on DEV with earlier and latest tests.

ppigazzini:
PROD updated, updated AUTHORS as well, thank you @UniQP :)

PR #1702, opened 2023-05-23
Rewrite "delta_update_users.py"

ppigazzini:
The previous version of the script had two major issues:
1. it loaded only one run ID from the deltas collection without any sorting
   criterion. Since there is no guarantee that the returned natural order
   is the insertion order, this could result in a very inaccurate contribution
   calculation in the first update scan after a full scan, when the deltas
   collection contains all the finished run IDs.
2. the run, with the ID loaded from the deltas collection, could switch
   to active status (e.g., after a purge).
   This resulted in failing to stop the processing and double counting
   of already processed finished runs from the last 30 days.

Changes:
- load and save the IDs of all the processed finished runs in
  deltas collection, python checks if a key is in a dict in O(1)
- in update scan mode, for the total contribution:
  - skip processing of any finished run with ID in deltas
  - stop processing after hitting 10 finished runs with ID in deltas
- lazy load the data from collections to lower the RAM footprint
- simplify the algorithms
- add comments

Note that the update scan is not perfect:
- it does not account for the additional contribution of runs that have
  switched from finished to unfinished and back to finished in the meantime.
- it underestimates the top month contribution compared to the full scan
  due to the exit condition.

### DIFF
--- a/server/utils/delta_update_users.py
+++ b/server/utils/delta_update_users.py
@@ -7,35 +7,77 @@ from fishtest.rundb import RunDb
 from fishtest.util import delta_date, diff_date, estimate_game_duration
 from pymongo import DESCENDING
 
-new_deltas = {}
-skip = False
 
+def initialize_info(rundb, clear_stats):
+    info_total = {}
+    info_top_month = {}
+    diff_ini = diff_date(datetime.min).total_seconds()
+    last_ini = delta_date(diff_date(datetime.min))
 
-def process_run(run, info, deltas=None):
-    global skip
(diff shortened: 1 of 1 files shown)


ppigazzini:
Tested for few days:
- running the scheduled delta scans on PROD
- loading the DB on DEV after a S3 backup
- running a delta scan on DEV to compute the just finished runs (open the contributors and top month pages)
- running a full scan on DEV to compute all the runs from zero (open the contributors and top month pages)
- check that the contributors and top month pages have the same numbers after the delta scan and the full scan

PR #1703, opened 2023-05-23
Fix alignment of flags

peregrineshahin:
current flags have messed up alignments due to the align center when injected machines.


ppigazzini (approved):
LGTM

PR #1704, opened 2023-05-25
Fix workers count for no machines

ppigazzini:


peregrineshahin:
Ah I see, my bad I checked the if of the parent tr instead of the td
That's because the html is wrong, it's non standard to add a td without a tr wrapping it

Try keep the js code as is but wrap the td with a tr that has the id

PR #1707, opened 2023-05-26
Remove some extra work in sync_request_task

peregrineshahin:


vdbergh:
The new code seems not equivalent to the old code since we will exit the loop early if a run is found.

peregrineshahin:
> The new code seems not equivalent to the old code since we will exit the loop early if a run is found.

Oops, that's a miss! That means 2 loops should still be there. (probably incremental update would be an overkill)
I will keep the connections limit checking inside the loop though, there is no need to continue looping if we actually reach the limit.

peregrineshahin:
and should be better to ask the db for the machines limit of the user only once instead of with every iteration.

vondele:
interesting, so that's again quadratic load on the server for in fleet size  (i.e. number of requests per second to the request_task api is linear in fleet size, and the work it does each call is linear (iterate through all tasks)).

peregrineshahin:
> interesting, so that's again quadratic load on the server for in fleet size (i.e. number of requests per second to the request_task api is linear in fleet size, and the work it does each call is linear (iterate through all tasks)).

Yes, it looks like a very unfortunate scenario in a fleet visit to loop over all tasks in all unfinished runs just to check if the machine reached an IP address limit... and then loop over all the tasks in unfinished runs again to see if this run is suitable to add work.. any thoughts?

vdbergh:
Of course `request_task` occurs relatively rarely compared to `update_task`. So one has to see if it is really a serious issue.

If it is, one can try to collect the data incrementally. But one has to careful with the proper locking since the data is affected both by `request_task` and `update_task`.

@ppigazzini reported some time ago that the number of cores assigned to a run can become negative (triggering an assertion error). In theory that should not happen.

vondele:
request_task is not so rare, about 10x less than update_task, but maybe not critical. If we have 10x more tasks than we have currently, it could become dominant:
```
               /api/update_task     3954     10.557      0.003      0.000      0.087
                      /api/beat     3309      7.602      0.002      0.000      0.032
                  /api/get_elo/      251      5.115      0.020      0.009      0.078
                /tests/finished       20      4.443      0.222      0.075      0.344
              /api/request_task      428      3.149      0.007      0.002      0.033
                /api/upload_pgn      134      2.855      0.021      0.003      0.147
                   /api/actions      557      2.180      0.004      0.002      0.020
           /api/request_version     1101      2.001      0.002      0.000      0.021
```

vdbergh:
Probably it would be best that all incremental data is only updated in `update_task`. This probably requires a zero update (no games yet) to initialize some things but this would be useful anyway, like for reporting the nps of the worker (currently one has to wait for the first update before this can be displayed in the machines table).

ppigazzini (approved):
LGTM

ppigazzini:

PR #1708, opened 2023-05-27
No height overflow if we have info box

peregrineshahin:
This makes the right panel accept as many features one can add there; by making it scrollable instead of the height overflowing leaving empty space when extra widgets are added.

See:
Capture
vs after scroll:
Capture2

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -4,6 +4,11 @@
   import json
 %>
 
+<%
+  args_row_count = 0
+  args_row_height = "33px"
+%>
+
 <%namespace name="base" file="base.mak"/>
 
 % if show_task >= 0:
@@ -72,6 +77,7 @@
           <thead></thead>
           <tbody>
             % for arg in run_args:
+              <% args_row_count += 1 %>
               % if len(arg[2]) == 0:
                 <tr>
                   <td>${arg[0]}</td>
@@ -133,11 +139,13 @@
                 </tr>
               % endif
             % endfor
+            <% args_row_count += 1 %>
             <tr>
               <td>events</td>
               <td><a href="/actions?run_id=${str(run['_id'])}">/actions?run_id=${run['_id']}</a></td>
             </tr>
             % if 'spsa' not in run['args']:
+              <% args_row_count += 1 %>
               <tr>
                 <td>raw statistics</td>
                 <td><a href="/tests/stats/${str(run['_id'])}">/tests/stats/${run['_id']}</a></td>
@@ -148,7 +156,10 @@
       </div>
     </div>
 
-    <div class="col-12 col-lg-3">
+    <div 
+      class="col-12 col-lg-3"
+      style="height: calc(${args_row_count + 2} * ${args_row_height}); overflow: auto;"
+    >
       <h4>Actions</h4>
       <div class="row g-2 mb-2">
         % if not run['finished']:


server/fishtest/templates/tests_view.mak
```diff
@@ -4,6 +4,11 @@
   import json
 %>
 
+<%
```
ppigazzini, line 7: Independent from template arguments, move in this tag.
```
<%!
```

vondele:
don't we have now two parallel scroll bars? That's usually a bit painful, IMO. 

BTW, personally, I really would like to have the diff next to the 'approve' button, i.e. above the details. It's good to have the diff visible without scrolling, I think.

peregrineshahin:
> don't we have now two parallel scroll bars? That's usually a bit painful, IMO.
> 
> BTW, personally, I really would like to have the diff next to the 'approve' button, i.e. above the details. It's good to have the diff visible without scrolling, I think.

as the diffs are fetched from external server rendering on initial load would make the page bumpy if we move it above, I have another idea since we have 3 different places we go to the github diff url I think it is best if we make the `diff` for each test in the home page when clicked instead of going to github it can scroll for you to the diff box when clicked? what do you think?

server/fishtest/templates/tests_view.mak
```diff
@@ -72,6 +77,7 @@
           <thead></thead>
           <tbody>
             % for arg in run_args:
+              % args_row_count += 1
```
ppigazzini, line 80: replace everywhere:
```diff
- % args_row_count += 1
+ <% args_row_count += 1 %>
```

ppigazzini: DEV updated with the fixes.

ppigazzini:
DEV updated

server/fishtest/templates/tests_view.mak
ppigazzini: Ooops, it seems that I was wrong, please revert this change, it seems that in that way the names become const. 

```yaml
May 27 15:45:29 dfts-0 pserve[27882]:   File "/home/usr00/fishtest/server/fishtest/templates/tests_view.mak", line 80, in render_body
May 27 15:45:29 dfts-0 pserve[27882]:     <% args_row_count += 1 %>
May 27 15:45:29 dfts-0 pserve[27882]: UnboundLocalError: cannot access local variable 'args_row_count' where it is not associated with a value
```

server/fishtest/templates/tests_view.mak
```diff
@@ -2,6 +2,8 @@
 
 <%!
   import json
+  args_row_count = 0
+  args_row_height = "33px"
```
ppigazzini, line 6: Ooops, it seems that I was wrong, please revert this change, it seems that in that way the names become const. 

```yaml
May 27 15:45:29 dfts-0 pserve[27882]:   File "/home/usr00/fishtest/server/fishtest/templates/tests_view.mak", line 80, in render_body
May 27 15:45:29 dfts-0 pserve[27882]:     <% args_row_count += 1 %>
May 27 15:45:29 dfts-0 pserve[27882]: UnboundLocalError: cannot access local variable 'args_row_count' where it is not associated with a value
```

peregrineshahin:
> Ooops, it seems that I was wrong, please revert this change, it seems that this way the names become const.

Fixed.

peregrineshahin:
Superseded by #1709 .

PR #1709, opened 2023-05-27
Move the diff above details.

peregrineshahin:


ppigazzini:
@vondele DEV updated

ppigazzini:
DEV updated

ppigazzini:
DEV updated

vondele:
I like it.

ppigazzini:
DEV updated.

peregrineshahin:
right, now I think we should make hiding it with cookies possible (like other places) for users who keep refreshing to see LLR.

vondele:
It already has some logic to hide it when the diff is large, so it would be kind of a 3 state thing?

peregrineshahin:
> 

No, I will remove the large diff hide by default.

ppigazzini:
> No, I will remove the large diff hide by default.

IMO the auto hide for large diff is a good solution.

vondele:
Auto hide for large diffs is a good thing, I agree.

peregrineshahin:
> > No, I will remove the large diff hide by default.
> 
> IMO the auto hide for large diff is a good solution.

it's a good solution but you can see in current DEV if the large diff is hidden we cannot predict the first text inside the button resulting in kind of flash if you refresh, try https://dfts-0.pigazzini.it/tests/view/6471a2ade549d9cf2fb213cd

ppigazzini:
@peregrineshahin I view the change from " comments" to "0 comments".

peregrineshahin:
> I view

Yeah just pushed that before you comment in 0s

so what do you think about being able to predict the first text?

ppigazzini:
I think that we should collect the comments by the approvers using the diff feature.

peregrineshahin:
We can of course transition the text change (make a delay) so that it has less flashing.

peregrineshahin:
Okay the way people hack this in general is by having an icon instead of text, imagine an eye icon when show that gets crossed out when it's hidden.
I will claim this as LGTM maybe we can change the texts for Hide/Show to the whole website if some icon looks good.

peregrineshahin:
Drafted I have an idea.

ppigazzini:
A ghostly effect on auto hide :)

peregrineshahin:
> A ghostly effect on auto hide :)

Yeah, it's just we can't predict the future due to the time nature of flowing linearly. I will revert the ghost until we invent a time-machine or we stop relying on the client to fetch stuff for us.

ppigazzini:
Could auto-show behave better than auto-hide?

peregrineshahin:
> Could auto-show behave better than auto-hide?

Great idea, let's see.

ppigazzini:
LGTM :)

ppigazzini:
DEV updated.

peregrineshahin:
Ready for shipping.

ppigazzini:
DEV updated.

ppigazzini:
> Ready for shipping.

I must wait to have back some GitHub API rate to made the last tests, I'm getting 503 right now.

ppigazzini (approved):

ppigazzini:

PR #1711, opened 2023-05-28
Remove bad alignments in the the users page.

peregrineshahin:


ppigazzini:
IMO, the firts version of the PR introduced 4 alignment issues.

image

peregrineshahin:
Yes they fixed in the last versions. I was experimenting and pushing.

peregrineshahin:
What do you think about making cards/tiles for the above table info? just like home page?

vdbergh:
> What do you think about making cards/tiles for the above table info? just like home page?

Sounds good.

ppigazzini:
I'm loading the last DB backup to try to have "minutes ago" as well.

peregrineshahin:
> > What do you think about making cards/tiles for the above table info? just like home page?
> 
> Sounds good.

I will look into that in a different PR, now closing as per the feedback of two admins.

PR #1712, opened 2023-05-28
Allow users to purge there own runs.

peregrineshahin:


peregrineshahin:
Consistent with the user being able to modify their own runs.

ppigazzini:
image

server/fishtest/views.py
```diff
@@ -1091,12 +1091,10 @@ def tests_approve(request):
...
 @view_config(route_name="tests_purge", require_csrf=True, request_method="POST")
 def tests_purge(request):
-    if not request.has_permission("approve_run"):
-        request.session.flash("Please login as approver")
-        return HTTPFound(location=request.route_url("login"))
-    username = request.authenticated_userid
-
     run = request.rundb.get_run(request.POST["run-id"])
+    if not request.has_permission("approve_run") and not is_same_user(request, run):
+        request.session.flash("Only approvers or the same user can purge the run.")
```
ppigazzini, line 1096:
```diff
-        request.session.flash("Only approvers or the same user can purge the run.")
+        request.session.flash("Only approvers or the submitting user can purge the run.")
```

ppigazzini: Lint failure :)

ppigazzini (approved):

PR #1713, opened 2023-05-28
Add cards/tiles to the users/monthly pages.

peregrineshahin:
Consistent with Home page cards UX.

### DIFF
--- a/server/fishtest/templates/users.mak
+++ b/server/fishtest/templates/users.mak
@@ -15,48 +15,69 @@
   % endif
 </h2>
 
-<div class="row" style="padding: 1em 0">
-  <div class="col-sm">
-    <div class="row">
-      <div class="col text-end"><b>Testers</b></div>
-      <div class="col text-start">
-        ${sum(u['str_last_updated'] != 'Never' for u in users)}
-      </div>
-    </div>
-    <div class="row">
-      <div class="col text-end"><b>Developers</b></div>
-      <div class="col text-start">
-        ${sum(u['tests'] > 0 for u in users)}
+<div class="row g-3 mb-3">
+  <div class="col-6 col-sm">
+    <div class="card card-lg-sm text-center">
+      <div class="card-header text-nowrap" title="Testers">Testers</div>
+      <div class="card-body">
+        <h4 class="card-title mb-0 monospace">
+          ${sum(u['str_last_updated'] != 'Never' for u in users)}
+        </h4>
       </div>
     </div>
   </div>
 
-  <div class="col-sm">
-    <div class="row">
-      <div class="col text-end"><b>Active testers</b></div>
-      <div class="col text-start">
-        ${sum(u['games_per_hour'] > 0 for u in users)}
-      </div>
-    </div>
-    <div class="row">
-      <div class="col text-end"><b>Tests submitted</b></div>
-      <div class="col text-start">
-        ${sum(u['tests'] for u in users)}
+  <div class="col-6 col-sm">
+    <div class="card card-lg-sm text-center">
+      <div class="card-header text-nowrap" title="Developers">Developers</div>
+      <div class="card-body">
+        <h4 class="card-title mb-0 monospace">
+          ${sum(u['tests'] > 0 for u in users)}
+        </h4>
       </div>
     </div>
   </div>
 
-  <div class="col-sm">
-    <div class="row">
-      <div class="col text-end"><b>Games played</b></div>
-      <div class="col text-start">
-        ${sum(u['games'] for u in users)}
+  <div class="col-6 col-sm">
+    <div class="card card-lg-sm text-center">
+      <div class="card-header text-nowrap" title="Active testers">Active testers</div>
+      <div class="card-body">
+        <h4 class="card-title mb-0 monospace">
+          ${sum(u['games_per_hour'] > 0 for u in users)}
+        </h4>
       </div>
     </div>
-    <div class="row">
-      <div class="col text-end"><b>CPU time</b></div>
-      <div class="col text-start">
-        ${f"{sum(u['cpu_hours'] for u in users)/(24*365):.2f} years"}
+  </div>
+  
+  <div class="col-6 col-sm">
+    <div class="card card-lg-sm text-center">
+      <div class="card-header text-nowrap" title="CPU years">CPU years</div>
+      <div class="card-body">
+        <h4 class="card-title mb-0 monospace">
+          ${f"{sum(u['cpu_hours'] for u in users)/(24*365):.2f}"}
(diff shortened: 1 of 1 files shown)


peregrineshahin:
Follow up of <https://github.com/glinscott/fishtest/pull/1711#issuecomment-1565979074>

ppigazzini:
DEV updated, I would like @dav1312 review :)

vdbergh:
All cards are 1 line except this one:

Screenshot from 2023-05-29 11-28-42

I also think that perhaps a 2x3 table will look better.

peregrineshahin:
smaller font will do it

ppigazzini:
What about moving "years" in the title?
"CPU time (years)"

peregrineshahin:
> What about moving "years" in the title? "CPU time (years)"

Brilliant.

ppigazzini:
Brilliant if the number is always computed in years, though.

peregrineshahin:
> Brilliant if the number is always computed in years, though.

it is always in years
years is already hard coded

peregrineshahin:
> I also think that perhaps a 2x3 table will look better.

would have a bit of an empty space or bloated boxs I tried them locally, doesn't look good

ppigazzini:
> > I also think that perhaps a 2x3 table will look better.
> 
> would have a bit of an empty space or bloated boxs I tried them locally, doesn't look good

On mobile/portrait the distribution of the cards is in 3 rows 2 columns, this could be another issue for a 2x3 table.

ppigazzini:
DEV updated, it looks good on my mobile/portrait.

server/fishtest/templates/users.mak
```diff
@@ -15,48 +15,66 @@
   % endif
 </h2>
 
-<div class="row" style="padding: 1em 0">
-  <div class="col-sm">
-    <div class="row">
-      <div class="col text-end"><b>Testers</b></div>
-      <div class="col text-start">
-        ${sum(u['str_last_updated'] != 'Never' for u in users)}
+<div class="row row g-3 mb-3">
```
dav1312, line 18: Duplicated row class?

peregrineshahin: that's unintended, fixed.

server/fishtest/templates/users.mak
```diff
@@ -15,48 +15,66 @@
...
-    <div class="row">
-      <div class="col text-end"><b>CPU time</b></div>
-      <div class="col text-start">
-        ${f"{sum(u['cpu_hours'] for u in users)/(24*365):.2f} years"}
+  </div>
+  <div class="col-6 col-sm">
+    <div class="card card-lg-sm text-center">
+      <div class="card-header text-nowrap" title="CPU time">CPU time (years)</div>
+      <div class="card-body">
+        <h4 class="card-title mb-0 monospace" style="font-size: 20px;">
```
dav1312, line 75: With the "years" removed u can also try removing the hardcoded font-size

ppigazzini:
DEV updated.

server/fishtest/templates/users.mak
peregrineshahin: well, that's intended but as you wish.

ppigazzini:
DEV update after https://github.com/glinscott/fishtest/commit/3b0689090ba7643dde7743b5eb7f1152d5a30752

vdbergh:
Perhaps "CPU years" (instead of "CPU time (years)")?

ppigazzini:
I propose to sort the tiles like the table headers, so
Testers"; "Developers"; "Active testers"; "CPU time (years)"; "Games played", "Tests submitted";

ppigazzini:
DEV updated after https://github.com/glinscott/fishtest/pull/1713/commits/47318f4311aea73c12affecfd303e885b0d40805

peregrineshahin:
> DEV updated after [47318f4](https://github.com/glinscott/fishtest/commit/47318f4311aea73c12affecfd303e885b0d40805)

Looks good to me.

peregrineshahin:
On a different note I'm looking at the tables and some users have last updated Never which we use the negation of that to determine the number of testers. could you please explain the logic?

peregrineshahin:
How sg last active: never?
Capture

peregrineshahin:
Okay now I understand, it refers to the last_active for a worker not user.
But still linrock is not considered tester bc he does both?

ppigazzini:
It's the latest `task["time_updated"]` for all runs, so developers that never contributed CPU have "Never".

https://github.com/glinscott/fishtest/blob/d5f26eb5257b8e5d2720515b1028d7c7a8b8e26b/server/utils/delta_update_users.py#L227-L236

https://github.com/glinscott/fishtest/blob/d5f26eb5257b8e5d2720515b1028d7c7a8b8e26b/server/utils/delta_update_users.py#L146-L157

ppigazzini:
> Okay now I understand, it refers to the last_active for a worker not user. But still linrock is not considered tester bc he does both?

linrock should be counted as "Tester" and "Developer"

https://github.com/glinscott/fishtest/blob/d5f26eb5257b8e5d2720515b1028d7c7a8b8e26b/server/fishtest/templates/users.mak#L21-L23

https://github.com/glinscott/fishtest/blob/d5f26eb5257b8e5d2720515b1028d7c7a8b8e26b/server/fishtest/templates/users.mak#L27-L30

peregrineshahin:
> linrock should be counted in both "Tester" and "Developer" classes.
Linrock does not have `Never` so how is he counted in testers?

peregrineshahin:
What I mean is some people run their workers from their user account, and if they do that they are not counted as testers

peregrineshahin:
Okay, my bad it's the opposite a tester is someone that has last_updated?
What does that mean?

A tester is a person who contributes CPU time?
Why a tester is not the opposite?

peregrineshahin:
it makes no sense. Developers are the testers. not the cpu contributers

ppigazzini:
A developer writes a patch on his pc, opens a PR on GithHub SF, submits a run on fishtest testing framework, fishtest distributes the tasks of the run to the workers, the worker starts cutechess to play a match and get a result, the worker send the match results to fishtest, fishtest update the SPRT/NumGames result and check the exit conditions (tests).
So the only tester is fishtest.

peregrineshahin:
Okay between you and me: Changing Testers to CPU contributors?

peregrineshahin:
Okay never mind, now everybody knows what testers refer to in fishtest thanks to the debate :P
feel free to merge as is.

ppigazzini (approved):

ppigazzini:
PROD updated, thank @peregrineshahin :)

PR #1718, opened 2023-06-02
Only strip runs if they came from cache.

peregrineshahin:


peregrineshahin:
Since get_run api is used by humans (not workers) externally I make it consistent like active_runs to only get runs from DB
This will allow moving from worker/server pserve.

peregrineshahin:
That leaves get_elo to be the only place where stripping run is applied
I like to keep it gives consistency with tests_view results, same LLR and stuff with no delay.

peregrineshahin:
Okay this is so wrong for get_elo
get_run is used elsewhere I will close this for now to get a better idea how to minimize it

PR #1721, opened 2023-06-05
Reduce concurrency if needed

vondele:
In case the worker has little memory (noob's workers), it might be impossible to run any tasks at the full concurrency specified by the user.

In that case, reduce concurrency to be able to run at least STC tests.


ppigazzini (approved):
It looks good with linux/msys2 workers with normal and SMP tests.

ppigazzini:

PR #1729, opened 2023-06-12
Consistent bench extraction with CI

peregrineshahin:


ppigazzini:
I'm failing to view an improvement, please explain.

peregrineshahin:
> I'm failing to view an improvement, please explain.

commit message of a recent stockfish commit picked up bench for base signatures to be 16 in PROD,  because the commit message had the text `bench 16 1 13` for demonstration purposes.

https://github.com/official-stockfish/Stockfish/commit/38e61663d836e062af0bc002814ad5149c4b7729

peregrineshahin:
Could you please elaborate on your suggestion?
As far as I understand, the change is range from 5 digits to 9 digits and only picks the last match?

ppigazzini:
Changes were:
- find the last bench signature in the text
- signature range 6-10 digits
- first digit not zero

However according to the official instructions for the commit message the bench signature must be on the last empty line.

https://github.com/glinscott/fishtest/wiki/Creating-my-first-test#im-ready-for-my-pull-request
- What will be the next signature of Stockfish if my patch is accepted?
  - Both the commit message and the pull request comment on GitHub must mention if the patch is a 'No functional change' or changes the search. The last line of the commit message should be either 'No functional change' or 'Bench: XXXXXXX' where 'Nodes searched : XXXXXXX' can be found in the output of a ./stockfish bench invocation.

This code should work as intended by the official instructions (running on DEV right now):

```diff
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index 67b966f..58c5df6 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -530,8 +530,7 @@ def users_monthly(request):


 def get_master_info():
-    message_search = re.compile(r"^.*$", re.MULTILINE)
-    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
+    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9][0-9]{5,9})(?!\d)")
     for idx, commit in enumerate(
         requests.get(
             "https://api.github.com/repos/official-stockfish/Stockfish/commits"
@@ -539,16 +538,17 @@ def get_master_info():
     ):
         if "commit" not in commit:
             return None
-        bench = bench_search.search(commit["commit"]["message"])
+        message_idx = commit["commit"]["message"].strip().split("\n")
+        bench = bench_search.search(message_idx[-1])
         if idx == 0:
-            message = message_search.search(commit["commit"]["message"])
+            message = "\n".join(line.strip() for line in message_idx)
             date = datetime.datetime.strptime(
                 commit["commit"]["committer"]["date"], "%Y-%m-%dT%H:%M:%SZ"
             )
         if bench:
             return {
                 "bench": bench.group(2),
-                "message": message.group(),
+                "message": message,
                 "date": date.strftime("%b %d"),
             }
     return None
```

server/fishtest/views.py
```diff
@@ -530,25 +530,25 @@ def users_monthly(request):
...
         ).json()
     ):
         if "commit" not in commit:
             return None
-        bench = bench_search.search(commit["commit"]["message"])
+        message_idx = commit["commit"]["message"].strip().split("\n")
+        bench = bench_search.search(message_idx[-1])
         if idx == 0:
-            message = message_search.search(commit["commit"]["message"])
+            message = "\n".join(line.strip() for line in message_idx)
```
ppigazzini, line 544: My fault, the regex of the original code writes only the first line of the message (but it misses to strip blank spaces)

```diff
- message = "\n".join(line.strip() for line in message_idx)
+ message = message_idx[0].strip()
```

server/fishtest/views.py
```diff
@@ -530,25 +530,25 @@ def users_monthly(request):
...
+    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
     for idx, commit in enumerate(
         requests.get(
             "https://api.github.com/repos/official-stockfish/Stockfish/commits"
         ).json()
     ):
         if "commit" not in commit:
             return None
-        bench = bench_search.search(commit["commit"]["message"])
+        message_idx = commit["commit"]["message"].strip().split("\n")
```
ppigazzini, line 541: here the strip can be dropped:
```diff
- message_idx = commit["commit"]["message"].strip().split("\n")
+ message_idx = commit["commit"]["message"].split("\n")
```

ppigazzini: `message_lines` seems to be a better name wrt `message_idx`

server/fishtest/views.py
```diff
@@ -530,25 +530,25 @@ def users_monthly(request):
 
 
 def get_master_info():
-    message_search = re.compile(r"^.*$", re.MULTILINE)
-    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
+    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
     for idx, commit in enumerate(
         requests.get(
```
ppigazzini, line 535: We are doing over and over the same request to GitHub, please move it outside the enumerate loop.

```python
commits = requests.get(........).json()

for idx, commit in enumerate(commits):
```

ppigazzini:
This should be the final version of the function:
```python
def get_master_info():
    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
    commits = requests.get(
        "https://api.github.com/repos/official-stockfish/Stockfish/commits"
    ).json()
    for idx, commit in enumerate(commits):
        if "commit" not in commit:
            return None
        message_lines = commit["commit"]["message"].strip().split("\n")
        bench = bench_search.search(message_lines[-1].strip())
        if idx == 0:
            message = message_lines[0].strip()
            date_str = commit["commit"]["committer"]["date"]
            date = datetime.datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%SZ")
        if bench:
            return {
                "bench": bench.group(2),
                "message": message,
                "date": date.strftime("%b %d"),
            }
    return None
```

PR #1732, opened 2023-06-14
Fix validating Master branch in the form.

peregrineshahin:
needs testing on dev.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -529,14 +529,15 @@ def users_monthly(request):
     return {"users": users_list}
 
 
-def get_master_info():
-    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
-    commits = requests.get(
-        "https://api.github.com/repos/official-stockfish/Stockfish/commits"
-    ).json()
-    for idx, commit in enumerate(commits):
-        if "commit" not in commit:
-            return None
+def get_master_info(url):
+    try:
+        commits = requests.get(url)
+        commits.raise_for_status()
+    except Exception as e:
+        print(f"Exception getting commits:\n{e}")
+        return None
+    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)")
+    for idx, commit in enumerate(commits.json()):
         message_lines = commit["commit"]["message"].strip().split("\n")
         bench = bench_search.search(message_lines[-1].strip())
         if idx == 0:
@@ -664,7 +665,11 @@ def validate_form(request):
         data["rescheduled_from"] = request.POST["rescheduled_from"]
 
     def strip_message(m):
-        s = re.sub(r"[Bb]ench[ :]+[0-9]+\s*", "", m)
+        lines = m.strip().split("\n")
+        last_line = lines[-1].strip()
+        if re.match(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)", last_line):
+            lines[-1] = ""
+        s = "\n".join(lines)
         s = re.sub(r"[ \t]+", " ", s)
         s = re.sub(r"\n+", r"\n", s)
         return s.rstrip()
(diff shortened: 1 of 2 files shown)


server/fishtest/views.py
```diff
@@ -529,11 +529,9 @@ def users_monthly(request):
     return {"users": users_list}
 
 
-def get_master_info():
+def get_master_info(url):
```
ppigazzini, line 532:
```python
def get_master_info(url):
    try:
        commits = requests.get(url)
        commits.raise_for_status()
    except Exception as e:
        print(f"Exception getting commits:\n{e}")
        return None
    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
    for idx, commit in enumerate(commits.json()):
        message_lines = commit["commit"]["message"].strip().split("\n")
        bench = bench_search.search(message_lines[-1].strip())
        if idx == 0:
```

peregrineshahin:
added an exception as requested, unfortunately, I'm not able to run black formatter.

server/fishtest/views.py
ppigazzini:
```diff
    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,9})(?!\d)")
    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)")
```

peregrineshahin: We made Stockfish CI to match to at max 10 digits I think it's a bit late for returning to 8.

ppigazzini: SF CI code is still broken, try this
"Fix bench

With old code produced bench + 1, at example for this commit we had
bench 123456

close 00000
bench 123457
"

peregrineshahin: > 

Yeah log could be splitted by each commit and each commit stripped to from ending white spaces and only the last line checjed for bench afterwards.

That's the fishtest way of doing it, I just thought it wasn't worth the complexity added for this case.

peregrineshahin: in any ways, I will change it back to original Joost's suggestion and now yours (to a max of 7.)
Feel free to overwrite me on that CI patch in Stockfish repo.
Hoping we haven't bothered Sf devs that much.

peregrineshahin: max of 8

ppigazzini:
```yaml
- name: Extract the bench number from the commit history
  run: |
    N=0
    while true; do
        commit_message=$(git log HEAD -n 1 --skip $N --pretty=format:%B)
        if [[ -z "$commit_message" ]]; then
            break
        fi
        bench_value=$(echo "$commit_message" | tail -n 1 | grep -oP "\b[Bb]ench[ :]\+\K[1-9][0-9]{5,7}\b")
        if [[ -n "$bench_value" ]]; then
            echo "benchref=$bench_value" >> $GITHUB_ENV
            echo "Reference bench: $bench_value"
            break
        fi
        N=$((N+1))
    done
    if [[ -z "$bench_value" ]]; then
        echo "No bench found"
    fi
```

ppigazzini:
Formatted, DEV updated.

peregrineshahin:
it seems to work taking aside the current master net not yet in the backup of dev.

peregrineshahin:
Ive noticed there will be still an issue with current implementation if we test master vs master

I try to keep changes as minimal as possible but neccessity exists.. I will make all regrex the same tomorrow(it's around 1am now). Sorry for the inconvience.

dav1312:
Nitpicking but if the same regex will be used multiple times (3) can't we store it somewhere once?

server/fishtest/views.py
```diff
@@ -664,7 +665,11 @@ def validate_form(request):
         data["rescheduled_from"] = request.POST["rescheduled_from"]
 
     def strip_message(m):
-        s = re.sub(r"[Bb]ench[ :]+[0-9]+\s*", "", m)
+        lines = m.split("\n")
```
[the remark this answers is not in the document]
peregrineshahin: Pushed.

peregrineshahin:
> Nitpicking but if the same regex will be used multiple times (3) can't we store it somewhere once?

You could argue the links should not be redeclared each time
This should be refactored along other places in the code via a config file.

peregrineshahin:
> > Nitpicking but if the same regex will be used multiple times (3) can't we store it somewhere once?
> 
> You could argue the links should not be redeclared each time This should be refactored along other places in the code via a config file.

A config file refactor would be a cool PR for all the code not only here I think we're a bit in a hurry including the fix.

server/fishtest/views.py
```diff
@@ -682,8 +687,10 @@ def strip_message(m):
         if "commit" not in c:
             raise Exception("Cannot find branch in developer repository")
         if len(data["new_signature"]) == 0:
-            bs = re.compile(r"(^|\s)[Bb]ench[ :]+([0-9]+)", re.MULTILINE)
-            m = bs.search(c["commit"]["message"])
+            bs = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)")
+            lines = c.split("\n")
```
ppigazzini, line 691:
```diff
-            lines = c.split("\n")
+            lines = c["commit"]["message"].split("\n")
```

ppigazzini (approved):
Tested fine on DEV

ppigazzini:

PR #1735, opened 2023-06-24
Set the SF16 reference Nps using SF11 as base

ppigazzini:
Fishtest with Stockfish 11 had 1.6MNps as reference Nps and 0.7MNps as threshold for the slow worker.
Set the new reference Nps according to the 26% slowdown of the SF 16 wrt SF 11 on the ARCH=x86-64-bmi2.
Set the new threshold for slow worker according to the 34% slowdown of the SF 16 wrt SF 11 on the ARCH=x86-64-modern (x86-64-sse41-popcnt).

compiler clang++ 16.0.6

- arch=bmi2 (Intel Xeon CPU E5-2680 v3)
```
sf_base =  1580934 +/-  12410 (95%)
sf_test =  1163656 +/-   7679 (95%)
diff    =  -417278 +/-  11547 (95%)
speedup = -26.394% +/- 0.730% (95%
```

- arch=modern (Intel core i7 3770k)
```
sf_base =  1865978 +/-  14061 (95%)
sf_test =  1228350 +/-  11701 (95%)
diff    =  -637627 +/-   5230 (95%)
speedup = -34.171% +/- 0.280% (95%)
```

Raise worker version to 206, also server side.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 205
+WORKER_VERSION = 206
 
 
 def validate_request(request):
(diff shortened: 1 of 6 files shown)


ppigazzini:
By the way benches done using the SF 16 pre-release
https://github.com/official-stockfish/Stockfish/releases/tag/stockfish-dev-20230622-a49b3ba7

ppigazzini:
Triggered the workers update.

PR #1739, opened 2023-07-05
Adjust settings for the SMP PT

vondele:
perform SMP PT at 60+0.6T. The reason is that we're doing quite a few tests at those settings today, so it might be nice to capture that.


ppigazzini (approved):
LG on DEV

ppigazzini:

PR #1740, opened 2023-07-06
Clarify message

vondele:


vondele:
not sure what the lint CI test wants to get green.

ppigazzini:
> not sure what the lint CI test wants to get green.

The latest version of prettier must have introduced some changes.

```
Run prettier --check "server/fishtest/static/{css/*.css,html/*.html,js/*.js}"
Checking formatting...
[warn] server/fishtest/static/html/maintenance.html
[warn] server/fishtest/static/js/application.js
[warn] server/fishtest/static/js/calc.js
[warn] server/fishtest/static/js/live_elo.js
[warn] server/fishtest/static/js/notifications.js
[warn] server/fishtest/static/js/spsa.js
[warn] server/fishtest/static/js/toggle_password.js
[warn] Code style issues found in 7 files. Run Prettier to fix.
Error: Process completed with exit code 1.
```

ppigazzini:
"Moving to the next task." is wrong
 
```
Score of New-e87e103ca9 vs Base-e87e103ca9: 46 - 49 - 79  [0.491] 174
Post request https://dfts-0.pigazzini.it:443/api/update_task handled in 176.40ms (server: 0.23ms)
The server told us that no more games are needed for this task. Moving to the next task.

Uploading compressed PGN of 204640 bytes
Post request https://dfts-0.pigazzini.it:443/api/upload_pgn handled in 255.93ms (server: 19.85ms)
Task exited
Stopped by 'fish.exit' file
Heartbeat stopped
```

vondele:
In that case, I don't really understand what is happening. Why would it stop by 'fish.exit' ?

ppigazzini:
In my opinion, the log of a step should have the information about the context of that step, and not about the context of the next step:
 - we can have several next steps depending on certain conditions (if the current step is not the final one)
 - the next step could change in the time due to changes of the workflow logic
 
 `fish.exit` is used to signal to the worker to exit after completing the current task.

vondele:
Updated based on the feedback.

ppigazzini (approved):

ppigazzini:
PROD updated to trigger the worker update, thnak you @vondele :)

PR #1743, opened 2023-07-10
Fix design overflow issue.

peregrineshahin:
image
https://tests.stockfishchess.org/tests/view/64abb2afe0c0e4e9fd848b02


ppigazzini:
image

ppigazzini:

PR #1744, opened 2023-07-11
Remove Use NNUE which is no longer an option in master SF

vondele:


ppigazzini:
DEV updated.

vondele:
hmm, looks wrong. Hash disappears.

ppigazzini:
DEV updated

vondele:
OK, that appears correct

ppigazzini (approved):
LG on DEV

ppigazzini:

PR #1745, opened 2023-07-11
Fix overflow on both mobile and desktop

peregrineshahin:
mobile version looks unreadable now, do an auto fix.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -451,14 +451,6 @@ legend {
   display: none;
 }
 
-.spsa-table {
-  table-layout: fixed;
-}
-
-.spsa-table td {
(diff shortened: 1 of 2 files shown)


ppigazzini:
LG on DEV

image

dav1312:
Note that `.table-responsive` already adds `overflow-x: auto;`
image

ppigazzini:
LG on DEV the desktop version too.

image

peregrineshahin:
> Note that `.table-responsive` already adds `overflow-x: auto;` image

Yeah makes sense. pushed

ppigazzini:
DEV updated, LGTM

ppigazzini (approved):

ppigazzini:

PR #1747, opened 2023-07-12
Reduce ITP with number of active tests per user

vondele:
given some excessive number of active tests by some users, this patch reduces itp based on the number of active tests.

The formula used is `itp *= 36.0 / (36.0 + count * count)`

This reduces slowly the itp for a number of tests smaller than 6, and rapidly afterwards. For 3 active tests, the itp is 80%, for 6 active tests, the itp is 50%, for 15 active tests it is down to 13%.

For a maximum share of fishtest... have 6 tests in queue.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -612,7 +612,7 @@ class RunDb:
 
         return results
 
-    def calc_itp(self, run):
+    def calc_itp(self, run, count):
         itp = run["args"]["throughput"]
         itp = max(min(itp, 500), 1)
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
user01 has 9 active tests with IT=59.02
user02 has 1 active test with ITP=100.00 

image

vondele:
The precise numbers might need some time to establish, i.e. once the itp has been recomputed. With 9 tests, it should be around 30%

vondele:
which it is now:
image
user 2 is close to the 98%
image

cj5716:
I have a question, can we regard tests already set to 50% TP as something like 0.5 tests
So that we can transfer throughput to 1 test over another that we have

vondele:
no, not easily. To transfer throughput, one should be able to pause the others (i.e. use prio -1)

ppigazzini:

ps: nitpicking, 36.0 was mandatory in python2

PR #1749, opened 2023-07-16
Do not use deprecated x86-64-modern

vondele:
explicitly use x86-64-sse41-popcnt


ppigazzini:
Please raise the worker version and create the sri.txt
https://github.com/glinscott/fishtest/blob/803f2c52cca9cce149c3118aea3a2d2e0ddf7b5c/server/fishtest/api.py#L28
https://github.com/glinscott/fishtest/blob/803f2c52cca9cce149c3118aea3a2d2e0ddf7b5c/worker/worker.py#L57

ppigazzini:

PR #1753, opened 2023-08-15
Fix string representation of signals for Python 3.11.

vdbergh:
See 

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1692105137.772177&run_id=

vs

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1692102104.224252&run_id=


ppigazzini (approved):
Tested working on DEV with workers with python 3.6 and python 3.11 (linux and windows)

ppigazzini:

PR #1759, opened 2023-08-24
Try to avoid workers with the same name.

### LINKED ISSUE #1360: Workers with issues.

vdbergh:
During request_task, we check if there are active tasks for workers with the same name, which have recently been updated (i.e. they are not dead). If so then we return an error.

Should fix https://github.com/official-stockfish/fishtest/issues/1360#issuecomment-1689436054 .

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -769,27 +769,44 @@ class RunDb:
 
         self.task_runs.sort(key=priority)
 
-        # We go through the list of unfinished runs to see if the worker
+        # get the list of active tasks
+
+        active_tasks = [
+            task for run in self.task_runs for task in run["tasks"] if task["active"]
+        ]
+
+        # We go through the list of active tasks to see if a worker with the same
+        # name is already connected.
+
+        my_name = "-".join(worker_name(worker_info).split("-")[0:3])
+        now = datetime.utcnow()
+        for task in active_tasks:
+            task_name = "-".join(worker_name(task["worker_info"]).split("-")[0:3])
+            if my_name == task_name:
+                last_update = (now - task["last_updated"]).seconds
+                # 120 = period of heartbeat in worker.
+                if last_update <= 120:
+                    error = 'Request_task: There is already a worker running with name "{}" which sent an update {} seconds ago'.format(
+                        my_name,
+                        last_update,
+                    )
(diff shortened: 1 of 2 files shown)


vdbergh:
Drafted.

vdbergh:
Hmm CI fails with
```
ERROR: test_auto_purge_runs (test_api.TestRunFinished.test_auto_purge_runs)
  File "/home/runner/work/fishtest/fishtest/server/tests/test_api.py", line 527, in test_auto_purge_runs
    self.assertEqual(response["run"]["_id"], str(run["_id"]))
                     ~~~~~~~~^^^^^^^
KeyError: 'run'
```
I did not touch autopurge...

ppigazzini:
I restarted a couple of times the CI, same error, the raw log doesn't show further information.

ppigazzini:
The PR changes the code of `def sync_request_task(self, worker_info):` and that function is called over and over in test_api.py


https://github.com/official-stockfish/fishtest/blob/7363126d7953e896058ee034a26f01d0d15f274f/server/tests/test_api.py#L508-L518

https://github.com/official-stockfish/fishtest/blob/940e460a3e91368b980926ef759cbd0f92db23b0/server/fishtest/api.py#L426-L434

https://github.com/official-stockfish/fishtest/blob/940e460a3e91368b980926ef759cbd0f92db23b0/server/fishtest/rundb.py#L689-L698

vdbergh:
I see. With this PR a worker can no longer have two active tasks at the same time (roughly). I guess this condition will not be satisfied during some tests. I will look at it tonight.

vdbergh:
Fixed!

ppigazzini:
> I will look at it tonight.

I wonder what is your timezone right now..

ppigazzini:
Right now DEV is running with the time aware datetime PR, I will test this PR this evening after loading a clean DB backup.

ppigazzini (approved):
Looks good on DEV, tested with a couple of workers started with `python3 worker.py --uuid_prefix 123456`

ppigazzini:

PR #1761, opened 2023-08-25
Replace deprecated `datetime.datetime.utcnow()`

ppigazzini:
The function is deprecated in python 3.12, see
https://docs.python.org/3.12/library/datetime.html?highlight=utcnow#datetime.datetime.now

"Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling `datetime.now(timezone.utc)`."


ppigazzini:
PROD updated, triggered the workers update.

vdbergh:
Thank you @ppigazzini :)

PR #1762, opened 2023-08-27
Fix a small bug. Add a comment.

vdbergh:


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1768, opened 2023-08-28
Fix event listeners are not removing.

peregrineshahin:
to reproduce bug:
step 1 - Click SPSA
step 2 - Click LTC
step 3 - Change branch name

base_branch changes when you change test_branch to be same (bug)

`base-signature` had an unintended bad value assigned to it in case of SPSA.

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -830,19 +830,22 @@
     })
   );
 
+  function testBranchHandler() {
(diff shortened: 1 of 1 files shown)


peregrineshahin:
Yes dev different than Prod, It is solved no changing of base_branch based on new_branch if we hit SPSA then choose anything else.

ppigazzini:
Ok, replicated.

ppigazzini (approved):
LGTM on DEV

PR #1769, opened 2023-08-28
Refine some tables.

peregrineshahin:


server/fishtest/templates/nns.mak
```diff
@@ -96,6 +96,11 @@
           </tr>
         % endif
       % endfor
+      % if nns is None or len(nns) == 0:
```
ppigazzini, line 99: nns is a generator, you could pass the num_nns in the view:

https://github.com/official-stockfish/fishtest/blob/fae6bcc7499fb04d11602e6f0ab975984db6ef03/server/fishtest/rundb.py#L285-L291

https://github.com/official-stockfish/fishtest/blob/fae6bcc7499fb04d11602e6f0ab975984db6ef03/server/fishtest/views.py#L305-L328

ppigazzini (approved):
LGTM on DEV

ppigazzini:
PROD updated with all the PRs (but the last one, still open waiting my test), thank you @peregrineshahin :)

PR #1770, opened 2023-08-28
Fix response is cached if the user uses same branch for a new commit.

peregrineshahin:


ppigazzini:
I hit the GitHub API limit, I will test after the reset.

PR #1771, opened 2023-08-28
Replace deprecated `distutils` package

ppigazzini:
The package `distutils` is deprecated since python 3.10 and it is scheduled to be removed with python 3.12, see: https://docs.python.org/3.12/whatsnew/3.12.html
https://github.com/python/cpython/issues/92584

`shutil.copytree()` is able to copy the source directory tree to an existing destination directory only from python 3.8 (parameter `dirs_exist_ok=True`), for python < 3.8 still use `distutils.dir_util.copy_tree()`

While here:
- create in a safe way the temporary directory used for the update
- deal with possible download problems

Raise worker version to 214 (also server side)

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 213
+WORKER_VERSION = 214
 
 
 def validate_request(request):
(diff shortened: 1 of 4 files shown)


ppigazzini:
PR tested working with linux worker (python: 3.6.8, 3.8.0, 3.11.5, 3.12.0rc1) and windows worker (python: 3.11.5).

ppigazzini:
Triggered the workers update.

PR #1772, opened 2023-08-28
Prefer the stateless approach over the stateful one for submitting te…

peregrineshahin:
…sts.

After experimenting with the approach proposed, it appears that the new approach break many workflows that submitter adapted for, which cannot be trivially fixed due to the stateful nature of this approach.

an example of that is the test run page requires another user interaction to reflect a new change on the branch after fetching first time occurred.


ppigazzini:
DEV updated

ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1773, opened 2023-08-29
Add log message to #1759.

vdbergh:


ppigazzini:

PR #1775, opened 2023-08-30
Refactor #1759 slightly and add a unittest.

vdbergh:


ppigazzini (approved):
Looks good:
- running 2 workers with `python3 worker.py --uuid_prefix 123456`:
```
Fetching task...
Error from remote: Request_task: There is already a worker running with name "user00-2cores-123456" which sent an update 46 seconds ago
```
- github actions:
```
Request_task: refresh queue
Request_task: There is already a worker running with name "JoeUserWorker-7cores-unique key" which sent an update 0 seconds ago
test_duplicate_workers (test_api.TestRunFinished.test_duplicate_workers) ... ok
```

ppigazzini:

PR #1776, opened 2023-08-30
send full commit SHAs to cutechess-cli

robertnurnberg:
This PR uses the full commit sha of `Base` for the engine name sent to cutechess-cli.

This would make it easier for scripts to identify the SF version that played games on fishtest.

Possible applications would be the tracking of the normalizing constant in SF's WDL_model, in the spirit of @vondele 's matetrack.

I do not know if the longer names could break things for other third party tools that make use of fishtest pgn's, so feedback would be appreciated.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-# List of authors for fishtest, as of April 12, 2023
+# List of authors for fishtest, as of August 31, 2023
 
 Gary Linscott (glinscott)
 
@@ -37,6 +37,7 @@ Noé Costa
 Onur Zungur (zungur)
 Pasquale Pigazzini (ppigazzini)
 Raimund Heid (RaimundHeid)
+Robert Nürnberg (robertnurnberg)
 Ronald de Man (syzygy1)
 Sebastian Buchwald (UniQP)
 sf-x
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 214
+WORKER_VERSION = 215
 
 
 def validate_request(request):
--- a/worker/games.py
+++ b/worker/games.py
@@ -780,7 +780,7 @@ def update_pentanomial(line, rounds):
     current = {}
 
     # Parse line like this:
-    # Finished game 4 (Base-5446e6f vs New-1a68b26): 1/2-1/2 {Draw by adjudication}
+    # Finished game 4 (Base-SHA vs New-SHA): 1/2-1/2 {Draw by adjudication}
     line = line.split()
     if line[0] == "Finished" and line[1] == "game" and len(line) >= 7:
         round_ = int(line[2])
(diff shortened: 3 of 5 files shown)


Disservin:
> I do not know if the longer names could break things for other third party tools that make use of fishtest pgn's, so feedback would be appreciated.

I dont think that's the case, at least I havent seen anyone parse the player names for stockfish.

ppigazzini:
DEV updated, this is one fo the runs approved with some PGNs ready to be downloaded.
https://dfts-0.pigazzini.it/tests/view/64ef820ce1778be0689d3f83

robertnurnberg:
Sample output
```
[White "New-4c4cb185aa"]
[Black "Base-4c4cb185aaaa0b3175ca35ab6473f17e9ec64055"]
```
Looks good to me, thanks.

ppigazzini:
The PR code is fine, I'm not directly involved in Stockfish development so I don't know if this PR can break some tools. I wait the green light from the other Stockfish maintainers before merging.
@robertnurnberg while waiting, please add your name in https://github.com/official-stockfish/fishtest/blob/master/AUTHORS :)

vondele:
I think this will be fine. Now, I'm not sure this asymmetric choice will be really a solution for what @robertnurnberg wants to achieve, in some cases (e.g. progression tests, the master code might be 'New' and 'Base' would be SF16). I think having it symmetric would be more natural.

robertnurnberg:
Agreed. When I push the change to make it symmetric, should I also update the code comments of the form
```python
    # Parse line like this:
    # Finished game 4 (Base-5446e6f vs New-1a68b26): 1/2-1/2 {Draw by adjudication}
```
to something like `Base-SHA New-SHA` ? I think they were left at 7 digits in https://github.com/official-stockfish/fishtest/commit/019519668c21e619101a391f317bb1e5a09f1913, so are not current anyway. But might be the time now to update them?

ppigazzini:
> `Base-SHA New-SHA` ? I think they were left at 7 digits in [0195196](https://github.com/official-stockfish/fishtest/commit/019519668c21e619101a391f317bb1e5a09f1913), so are not current anyway. But might be the time now to update them?

Ok

ppigazzini:
Unfortunately, the symmetric change disrupts the worker console output.
I propose this change to print the short hash in the console.
```diff
diff --git a/worker/games.py b/worker/games.py
index e851475..ea51db7 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -835,6 +835,11 @@ def validate_pentanomial(wld, rounds):
 def parse_cutechess_output(
     p, remote, result, spsa_tuning, games_to_play, batch_size, tc_limit
 ):
+    hash_pattern = re.compile(r"(Base|New)-[a-f0-9]+")
+    def shorten_hash(match):
+        word = match.group(0).split("-")
+        return "-".join([word[0], word[1][:10]])
+
     saved_stats = copy.deepcopy(result["stats"])
     rounds = {}

@@ -856,7 +861,7 @@ def parse_cutechess_output(
             time.sleep(1)
             continue

-        print(line, flush=True)
+        print(hash_pattern.sub(shorten_hash, line), flush=True)

         # Have we reached the end of the match? Then just exit.
         if "Finished match" in line:
 ```

robertnurnberg:
Thanks for spotting and providing a fix. I applied your fix verbatim, and then used black.

ppigazzini:
I force pushed to trigger the CI.

ppigazzini:
Triggered the workers update, thank you @robertnurnberg :)

PR #1778, opened 2023-09-01
some views fixes

peregrineshahin:
* Navigate to the previous page in case login is required.
* Don't show options that a user can't use.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -143,9 +143,7 @@ def workers(request):
     if not userid:
         request.session.flash("Please login")
         return HTTPFound(
-            location=request.route_url(
-                "login", _query={"next": f"/workers/{worker_name}"}
-            )
+            location=request.route_url("login", _query={"next": request.path})
         )
     username = worker_name.split("-")[0]
     if not request.has_permission("approve_run") and userid != username:
@@ -188,7 +186,10 @@ def upload(request):
     userid = request.authenticated_userid
     if not userid:
         request.session.flash("Please login")
-        return HTTPFound(location=request.route_url("login"))
+        return HTTPFound(
+            location=request.route_url("login", _query={"next": request.path})
+        )
+
     if request.method != "POST":
         return {}
     try:
@@ -504,7 +505,9 @@ def user(request):
     userid = request.authenticated_userid
     if not userid:
         request.session.flash("Please login")
-        return HTTPFound(location=request.route_url("login"))
+        return HTTPFound(
+            location=request.route_url("login", _query={"next": request.path})
+        )
     user_name = request.matchdict.get("username", userid)
     profile = user_name == userid
     if not profile and not request.has_permission("approve_run"):
@@ -1140,7 +1143,6 @@ def tests_stop(request):
 @view_config(route_name="tests_approve", require_csrf=True, request_method="POST")
 def tests_approve(request):
     if not request.authenticated_userid:
-        request.session.flash("Please login")
         return HTTPFound(location=request.route_url("login"))
     if not request.has_permission("approve_run"):
         request.session.flash("Please login as approver")
@@ -1308,6 +1310,7 @@ def tests_view(request):
         "auto_purge",
         "priority",
         "itp",
+        "throughput",
         "username",
         "tests_repo",
         "adjudication",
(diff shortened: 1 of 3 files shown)


ppigazzini:
DEV updated.

server/fishtest/views.py
```diff
@@ -1082,7 +1103,12 @@ def tests_stop(request):
 def tests_approve(request):
     if not request.authenticated_userid:
         request.session.flash("Please login")
-        return HTTPFound(location=request.route_url("login"))
+        next_page = request.path
```
ppigazzini, line 1106: I wonder if this code is reachable after exposing only the options to which the user is entitled.

peregrineshahin: Well, in case an approver logged out from one tab and clicked approve on the other tab I guess that's the case.

peregrineshahin:
I forgot about the delete icon on home page.

peregrineshahin:
> I forgot about the delete icon on home page.

Hmm we can live without hiding/disabling it since the page will look like 'broken teeth' if hide the delete button for some tests or even lower down the opacity.
Capture

server/fishtest/views.py
ppigazzini: Already tried but I got "check_csrf_token(): Invalid token"

peregrineshahin: Hmmm right, we require the token here which is removed if a user logged out I missed that.

peregrineshahin: > Already tried but I got "check_csrf_token(): Invalid token", same error using the back button. I'm not able to trigger that condition.

I'm keeping the return though for security reasons if the csrf was hacked/broken somehow.

ppigazzini:
- no user logged in
- click on "red bin" button to delete a pending test
- click on "confirm" button
- redirected to the login page https://dfts-0.pigazzini.it/login?next=/tests/delete
- login as admin
- redirected to https://dfts-0.pigazzini.it/tests/delete
- error message:
```
404 Not Found
The resource could not be found.

predicate mismatch for view tests_delete (request_method = POST)
```

peregrineshahin:
I think stuff should be ready now. two ideas was mixed together earlier.

ppigazzini:
For worker management links we use this code to redirect to the page after a required login:

https://github.com/official-stockfish/fishtest/blob/13297926d25a497d4ceefcc9c32bbdfa823e2971/server/fishtest/views.py#L132-L149

ppigazzini:
https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_url

this seems work on DEV:

```diff
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index e645899..cbf6768 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -143,9 +143,7 @@ def workers(request):
     if not userid:
         request.session.flash("Please login")
         return HTTPFound(
-            location=request.route_url(
-                "login", _query={"next": f"/workers/{worker_name}"}
-            )
+            location=request.route_url( "login", _query={"next": request.path})
         )
     username = worker_name.split("-")[0]
     if not request.has_permission("approve_run") and userid != username:
@@ -188,11 +186,8 @@ def upload(request):
     userid = request.authenticated_userid
     if not userid:
         request.session.flash("Please login")
-        next_page = request.path
-        if "id" in request.params:
-            next_page += "?id={}".format(request.params["id"])
         return HTTPFound(
-            location="{}?next={}".format(request.route_url("login"), next_page)
+            location=request.route_url( "login", _query={"next": request.path})
         )

     if request.method != "POST":
@@ -512,9 +507,9 @@ def user(request):
         request.session.flash("Please login")
         next_page = request.path
         if "id" in request.params:
-            next_page += "?id={}".format(request.params["id"])
+            next_page += "id={}".format(request.params["id"])
         return HTTPFound(
-            location="{}?next={}".format(request.route_url("login"), next_page)
+            location=request.route_url( "login", _query={"next": next_page})
         )
     user_name = request.matchdict.get("username", userid)
     profile = user_name == userid
```

server/fishtest/views.py
```diff
@@ -129,7 +129,13 @@ def upload(request):
     userid = request.authenticated_userid
     if not userid:
         request.session.flash("Please login")
-        return HTTPFound(location=request.route_url("login"))
+        next_page = request.path
+        if "id" in request.params:
```
ppigazzini, line 133: I don't understand what it is the "id" params, here and in the next function.

ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1779, opened 2023-09-01
Workers page and icon

dav1312:


ppigazzini:
DEV update, nice :)

vdbergh:
However it raises usual problem: does a checkbox refer to a state or to an action? For me it refers to a state (buttons are for actions). That's why I wrote "blocked" instead of "block".

dav1312:
I just made a mistake and added the name and check to the label instead of the checkbox, that's why it doesn't work when you try to change the state, give me a minute

ppigazzini:
DEV updated.

ppigazzini:
```diff
- <h2>Workers Management</h2>
+ <h2>Workers Administration</h2>
```
I used "Management" because ther was already "Worker admin"
We have "User Administration" that has the checkbox with "Blocked"
https://dfts-0.pigazzini.it/user/user01

image

dav1312:
Okaaay I'll change it... But I like "management" and "block" more and I would much rather change it for the user page

ppigazzini:
I'm fine with either Administration or Management, simply I would like to have consistency..

dav1312:
Okay changed back to Management, same for the user page

ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1782, opened 2023-09-02
Avoid saving the host_url in the db with every task.

vdbergh:


ppigazzini:
DEV updated.

vdbergh:
It seems to have been fixed.

ppigazzini:
Two different runs, without and with the PR
image

ppigazzini (approved):
LGTM on DEV

ppigazzini:

PR #1783, opened 2023-09-02
Simplify page redirection after required login

ppigazzini:


vdbergh:
Would it be a good idea to do this login redirection by default?

peregrineshahin:
> Would it be a good idea to do this login redirection by default?
not sure what that means?

redirecting to login even if the user is logged in and then redirect back to the page? no

ppigazzini:
This https://github.com/official-stockfish/fishtest/pull/1778 has hidden nearly all the buttons not usable if not logged in, the redirection works mostly when having several tabs open with the buttons shown, logout in one tab, press on a button (new run, net upload, user management, worker management etc.) in another tab.

peregrineshahin:
Why removed this btw?
this was intended for reschedule.......

vdbergh:
> > Would it be a good idea to do this login redirection by default?
> 
> not sure what that means?
> 
> redirecting to login even if the user is logged in and then redirect back to the page? no

No just using a wrapper around request.login that sends you automatically to request.path after login (this functionality is used in several places now).

peregrineshahin:
I think this simplification is incorrect because it was dealing with recheduling when you are not logged in.

peregrineshahin:
i.e. click reschedule when you are not logged in -> takes you to login -> login -> the test run is now empty as opposed to being filled and ready for reschedule

vdbergh:
Perhaps path_qs works? https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.path_qs

peregrineshahin:
> Perhaps path_qs works? https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.path_qs

That's very interesting one.

peregrineshahin:
> Perhaps path_qs works? https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.path_qs

perfect finding indeed! just what we wants.

PR #1784, opened 2023-09-02
Remember query strings

peregrineshahin:
latest simplifications omits the query string, add them natively with `path_qs`

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -87,6 +87,15 @@ def home(request):
     return HTTPFound(location=request.route_url("tests"))
 
 
+def ensure_logged_in(request):
+    userid = request.authenticated_userid
+    if not userid:
+        request.session.flash("Please login")
+        raise HTTPFound(
+            location=request.route_url("login", _query={"next": request.path_qs})
+        )
+
+
 @view_config(
     route_name="login",
     renderer="login.mak",
@@ -139,13 +148,10 @@ def workers(request):
             "show_admin": False,
             "blocked_workers": blocked_workers,
         }
-    userid = request.authenticated_userid
-    if not userid:
-        request.session.flash("Please login")
-        return HTTPFound(
-            location=request.route_url("login", _query={"next": request.path})
-        )
+    ensure_logged_in(request)
+
     username = worker_name.split("-")[0]
+    userid = request.authenticated_userid
     if not request.has_permission("approve_run") and userid != username:
         cached_flash(request, f"Only owners and approvers can block/unblock", "error")
         return {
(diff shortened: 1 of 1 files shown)


vdbergh:
So exactly the same code is used in 4 places… Time to refactor! ensure_logged_in() …

peregrineshahin:
> So exactly the same code is used in 4 places… Time to refactor! ensure_logged_in(request) …

done

vdbergh:
One can perhaps even do a bit more?
```python
    if not userid:
```
This could be refactored as well (that was the `ensure` part in my suggestion...).

vdbergh:
We could possibly return `user_id` here

https://github.com/official-stockfish/fishtest/blob/16ca7d20dfb58d08a76ca8c2529e0432a72f9922/server/fishtest/views.py#L90-L97

So 
```
user_id = ensure_logged_in(request)
```
in case we need the `user_id` later on.

peregrineshahin:
> 

wouldn't that continue the code in the other case where user_id is None, we want to break off the route funtion?

vdbergh:
I don't think so. If the user is logged in this will return `user_id`. If he is not then `ensure_logged_in()` does not return but the browser will fetch the same page again, now with the user logged in, so that this time `ensure_logged_in()` does return the `user_id`.

peregrineshahin:
do you mean that at any level of nesting functions `return HTTPFound()` will exit all root level function calls?
I'm not sure that how it works. that will work for the user sure, but it will generate server error since the excution will continue after returning from `ensure_logged_in()`?

vdbergh:
Ah yes I think you are right. I guess `return HTTPFound()` will only work if it returns from the view. This makes things more tricky. Sorry about that :frowning_face:

vdbergh:
I think an alternative to return may be to raise an exception. Not sure. Have to run.

vdbergh:
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html#using-a-view-callable-to-do-an-http-redirect

ppigazzini:
DEV updated, no good.
```yaml
Sep 03 11:15:31 dfts-0 pserve[10392]: 2023-09-03 11:15:31,295 ERROR [waitress][waitress-0] Exception while serving /tests/run
Sep 03 11:15:31 dfts-0 pserve[10392]: Traceback (most recent call last):
...
Sep 03 11:15:31 dfts-0 pserve[10392]:   File "/home/usr00/fishtest/server/fishtest/views.py", line 1005, in tests_run
Sep 03 11:15:31 dfts-0 pserve[10392]:     "tests_repo": u.get("tests_repo", ""),
Sep 03 11:15:31 dfts-0 pserve[10392]:                   ^^^^^
Sep 03 11:15:31 dfts-0 pserve[10392]: AttributeError: 'NoneType' object has no attribute 'get'
```

image

vdbergh:
Yes the return needs to be replaced by raise.

ppigazzini:
DEV updated, my first tests look good.

ppigazzini (approved):
Looks good on DEV in my tests

ppigazzini:

PR #1790, opened 2023-09-05
Fix wrong DOM selector

peregrineshahin:
production shows  `Workers - undefined machines` instead of an actual count.


ppigazzini (approved):
LGTM on DEV

ppigazzini:
PROD updated, thank you @peregrineshahin and @dav1312 :)

PR #1794, opened 2023-09-07
fix links colors on hover as well

peregrineshahin:
thanks david for the fix.


ppigazzini (approved):
Looks Dark Green on DEV

ppigazzini:
PROD updated, thank you @peregrineshahin @dav1312 :)

PR #1797, opened 2023-09-09
Add a schema to userdb.

vdbergh:
This serves as documentation but we also enforce the schema using asserts to avoid inadvertent database corruption.

Unfortunately this was not tested. I could not create a new user on my local instance (captcha failure).

### DIFF
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -3,7 +3,21 @@ import threading
 import time
 from datetime import datetime, timezone
 
+from fishtest.util import optional_key, validate
 from pymongo import ASCENDING
+
+schema = {
+    "username": str,
+    "password": str,
+    "registration_time": datetime,
+    "blocked": bool,
+    "email": str,
+    "groups": list,
+    "tests_repo": str,
+    optional_key("machine_limit"): int,
+}
+
+DEFAULT_MACHINE_LIMIT = 16
 
 
 class UserDb:
@@ -73,6 +87,7 @@ class UserDb:
     def add_user_group(self, username, group):
         user = self.find(username)
         user["groups"].append(group)
+        assert validate(schema, user, "user", strict=False) == ""
         self.users.replace_one({"_id": user["_id"]}, user)
         self.clear_cache()
 
@@ -81,17 +96,18 @@ class UserDb:
             if self.find(username):
                 return False
             # insert the new user in the db
-            self.users.insert_one(
-                {
-                    "username": username,
-                    "password": password,
-                    "registration_time": datetime.now(timezone.utc),
-                    "blocked": True,
-                    "email": email,
-                    "groups": [],
-                    "tests_repo": "",
-                }
-            )
+            user = {
+                "username": username,
+                "password": password,
+                "registration_time": datetime.now(timezone.utc),
+                "blocked": True,
+                "email": email,
+                "groups": [],
+                "tests_repo": "",
+                "machine_limit": DEFAULT_MACHINE_LIMIT,
+            }
+            assert validate(schema, user, "user", strict=False) == ""
+            self.users.insert_one(user)
             self.last_pending_time = 0
 
             return True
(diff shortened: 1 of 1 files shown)


peregrineshahin:
Topic related: we should not accept usernames with spaces/begin with numbers/have (-)/ nor weak passwords..

vdbergh:
Weak passwords are a semantic thing so rejecting them is a task of the front end IMHO (it is already being done to some extent).

The validation framework in `util.py` is extensible. So it is possible define a schema `username` if we agree what constitutes a valid username,

peregrineshahin:
> Weak passwords are a semantic thing so rejecting them is a task of the front end IMHO (it is already being done to some extent).
> 
> The validation framework in `util.py` is extensible. So it is possible define a schema `username` if we agree what constitutes a valid username,

I find this one good enough <https://stackoverflow.com/questions/12018245/regular-expression-to-validate-username#:~:text=%5E(%3F%3D.%7B8%2C20%7D%24)(%3F!%5B_.%5D)(%3F!.*%5B_.%5D%7B2%7D)%5Ba%2DzA%2DZ0%2D9._%5D%2B(%3F%3C!%5B_.%5D)%24>

Although tbh for the validation of both user name and password it should be done on both frontend and backend because front-end can be bypassed -- CSRF is a security measures for attack on users, it does not work when the user is the attacker. i.e. we should not allow a user to bypass our password rules if we agree on one, even if he finds it convenient to use `0000`. we cn have the backend validations if you wish.. and I can implement same regexes for front end then.

vdbergh:
Note that the username validation must fit the usernames in the the db.

peregrineshahin:
> Note that the username validation must fit the usernames in the the db.

I more inclined to fix this retroactively - i.e. if we really haven't had users who do that and regester with spaces, I wouldn't have noticed that we need to do this.

vdbergh:
Concerning weak passwords: if @ppigazzini wants to create an account on PROD with password 0000 then there is nothing we can do about it since he can put the record directly into the db. So validation code is useless in an adversarial setting, bypassing the front end. But is very useful in a non adversarial setting.

The only thing we have to make sure of is that `create_user()` is not accessible via an external api.

vdbergh:
> > Note that the username validation must fit the usernames in the the db.
> 
> I more inclined to fix this retroactively - i.e. if we really haven't had users who do that and regester with spaces, I wouldn't have noticed that we need to do this.

Yes of course the fit can be achieved in two ways. But it needs to be done.

peregrineshahin:
hmm after looking at the users table I think these cases are pretty rare maybe less than 10 username of all users.
I think it would be much easier to manually edit their usernames to fit the regex and then send them an email of their new username policy.

apart from that let's accept whatever password they have choosen at the moment since that would need a whole lot of work asking all users to put a new password again on their first login.

one think that could be editted about that regex I mentioned is maybe the length of the username could be limited to at least 2 characters not 8 unless we want to banter with `vdv` and `sg` :D

ppigazzini:
I wonder if in the future we could replace our hand made validation code with Pydantic:
https://docs.pydantic.dev/latest/

ppigazzini (approved):
Looks good on DEV

vdbergh:
> I wonder if in the future we could replace our hand-made validation code (that I barely know) with Pydantic: https://docs.pydantic.dev/latest/

I had never heard of Pydantic :frowning: 

It would definitely be an option but from looking at the documentation there seems to be a substantial learning curve.

ppigazzini:
I stumbled in Pydantic when writing the net server with FastAPI (skipping Pydantic).

ppigazzini (approved):
Looks good on DEV

ppigazzini:

vdbergh:
I would propose for a valid username any combination of letters/digits and "_". So both camel case and snake case would work. I would also propose to impose a minimum length. If there are a few important usernames not satisfying the length condition then they can be hard coded as legacy exceptions. 

Allowing other special characters than "_" is dangerous for parsing (in fact there is currently a bug in the parsing of worker names if the username contains "-").

vdbergh:
Actually for the schema an arbitrary length condition seems not useful (except `>0`). The aim of the schema is to prevent bugs in Fishtest, not to police the users.

PR #1798, opened 2023-09-09
Enable last page in pagination

peregrineshahin:
Access to last pages for most API's is no longer slow.


peregrineshahin:
There seem to be inconsistent behavior with opening up paginations
A speed up of 108x (17s to 157ms) is noticed on all last pages of all users after at least one last page of a user is opened, to be investigated later on.

PR #1799, opened 2023-09-10
Add mailto links to worker management.

vdbergh:
A draft message is provided (which includes the message in the worker management panel) which of course can be edited.

<hr>

Hi user00,

It seems your worker user00-7cores-1234567 has some issue(s):

Your worker is producing excessive timelosses! 

Can you take a look?

You may possibly find more information at
https://34.205.74.97/actions?text=%22user00-7cores-1234567%22.

Note: it may be that your worker is currently blocked. In that case
you can unblock it at https://34.205.74.97/workers/user00-7cores-1234567.

Thanks in advance!
The Fishtest team

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -3,6 +3,7 @@ import hashlib
 import html
 import os
 import re
+import textwrap
 import threading
 import time
 from datetime import datetime, timezone
@@ -139,22 +140,75 @@ def login(request):
     return {}
 
 
+# Note that the allowed length of mailto URLs on Chrome/Windows is severely
+# limited.
+
+
+def worker_email(worker_name, blocker_name, message, host_url, blocked):
+    owner_name = worker_name.split("-")[0]
+    wrapped_message = textwrap.fill(message, width=70)
+    body = f"""\
+Dear {owner_name},
+
+Thank you for contributing to the development of Stockfish. Unfortunately,
+it seems your Fishtest worker {worker_name} has some issue(s). More
+specifically the following has been reported:
+
+{wrapped_message}
+
+You may possibly find more information about this in our event log at 
+{host_url}/actions
+
+Feel free to reply to this email if you require any help, or else contact
+the #fishtest channel on the Stockfish Discord server:
+https://discord.com/invite/awnh2qZfTT
+
+Enjoy your day,
+
+{blocker_name} (Fishtest approver)
+
+"""
+
+    return body
+
+
 @view_config(route_name="workers", renderer="workers.mak", require_csrf=True)
 def workers(request):
+    is_approver = request.has_permission("approve_run")
+
     blocked_workers = request.rundb.workerdb.get_blocked_workers()
+    blocker_name = request.authenticated_userid
+
+    # If we are approver then we are logged in, so blocker_name is not None
+    if is_approver:
+        for w in blocked_workers:
+            owner_name = w["worker_name"].split("-")[0]
+            owner = request.userdb.find(owner_name)
+            w["owner_email"] = owner["email"] if owner is not None else ""
+            w["body"] = worker_email(
+                w["worker_name"],
+                blocker_name,
+                w["message"],
+                request.host_url,
+                w["blocked"],
+            )
+            w["subject"] = f"Issue(s) with worker {w['worker_name']}"
+
     worker_name = request.matchdict.get("worker_name")
     # TODO. Do more validation of worker names
     if len(worker_name.split("-")) != 3:
         return {
             "show_admin": False,
+            "show_email": is_approver,
             "blocked_workers": blocked_workers,
         }
-    user_id = ensure_logged_in(request)
-    username = worker_name.split("-")[0]
-    if not request.has_permission("approve_run") and user_id != username:
-        cached_flash(request, f"Only owners and approvers can block/unblock", "error")
+    ensure_logged_in(request)
+    owner_name = worker_name.split("-")[0]
+    if not is_approver and blocker_name != owner_name:
+        request.session.flash(f"Only owners and approvers can block/unblock", "error")
         return {
             "show_admin": False,
+            "show_email": is_approver,
             "blocked_workers": blocked_workers,
         }
 
@@ -162,26 +216,31 @@ def workers(request):
         button = request.POST.get("submit")
         if button == "Submit":
             blocked = request.POST.get("blocked") is not None
-            message = request.POST.get("message")[0:500]
+            message = request.POST.get("message")
+            if len(message) > 400:
+                request.session.flash(
+                    "Warning: your description of the issue has been truncated to 400 characters",
+                    "error",
+                )
+                message = message[:400]
+            was_blocked = request.workerdb.get_worker(worker_name)["blocked"]
             request.rundb.workerdb.update_worker(
                 worker_name, blocked=blocked, message=message
             )
-            cached_flash(
-                request,
-                f"Worker {worker_name} {'blocked' if blocked else 'unblocked'}!",
-            )
+            if blocked != was_blocked:
+                request.session.flash(
+                    f"Worker {worker_name} {'blocked' if blocked else 'unblocked'}!",
+                )
         return HTTPFound(location=request.route_url("workers", worker_name="show"))
 
     w = request.rundb.workerdb.get_worker(worker_name)
-    blocked = w["blocked"]
-    message = w["message"]
-    last_updated = w["last_updated"]
     return {
         "show_admin": True,
         "worker_name": worker_name,
-        "blocked": blocked,
-        "message": message,
-        "last_updated": last_updated,
+        "blocked": w["blocked"],
+        "message": w["message"],
+        "show_email": is_approver,
+        "last_updated": w["last_updated"],
         "blocked_workers": blocked_workers,
     }
 
--- a/server/fishtest/templates/workers.mak
+++ b/server/fishtest/templates/workers.mak
@@ -2,6 +2,7 @@
 
 <%!
   import datetime
+  from urllib.parse import quote
   from fishtest.util import delta_date, diff_date
 %>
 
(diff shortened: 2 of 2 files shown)


vdbergh:
It works well on my desktop, but on my cell phone the spaces in the draft are replaced by +'es.

ppigazzini:
On my desktop, the "+" replaces all the spaces in the draft of Thunderbird v 115.2.0

vdbergh:
Maybe solved now. + -> %20

dav1312:
Will this show the email of users in dev?

ppigazzini:
> Will this show the email of users in dev?

For the approver.

vondele:
let's please be careful about these things. Also, in case sending email is automatic, this needs to be approached very very carefully. If emails are sent by mistake or too frequently that's a real issue. Similarly, if this is a no-reply email, I think that's potentially an issue.

ppigazzini:
> let's please be careful about these things. Also, in case sending email is automatic, this needs to be approached very very carefully. If emails are sent by mistake or too frequently that's a real issue. Similarly, if this is a no-reply email, I think that's potentially an issue.

Try on DEV, there is a link that opens the email client with a draft message.

vondele:
OK, that's good.

vondele:
Should there be a checkbox or a date, that the approver could click if an email has been sent, to avoid repeated emails?

vdbergh:
With the new version the only thing that does not work for me is gmail on ipad as handler of mailto links (the usual mail app works fine). In gmail all spaces in the draft are suppressed. 

Ipad is always a bit strange so maybe no big deal....

vdbergh:
Hmm with the new version Thunderbird as a handler in firefox no longer shows the draft (blank body).

It would be nice to know how firefox actually calls thunderbird.

vdbergh:
It seems that thunderbird simply ignores the body argument of a mailto link. E.g.
```
/usr/lib/thunderbird/thunderbird mailto:user00@user00.user00?body=test
```
give a blank body. That's quite shocking.

vondele:
probably version dependent, since it works with ` Thunderbird 102.15.0`

vdbergh:
That's the version also have, but it seems it was badly packaged (Ubuntu 22.04). Using command line arguments to compose a message also does not work.

vondele:
Ok, on Ubuntu 20.04 here, so yeah could be, or could be some security concern?

Disservin:
I dont think so, it's rather common to prefill a mail with some html template, but js cant be added and html alone is not a security concern

vdbergh:
@vondele Should there be a checkbox or a date, that the approver could click if an email has been sent, to avoid repeated emails?

Perhaps but it requires some thought what is the correct UI. In principle a worker can have two issues with some time in between which would require sending emails with different messages.

vondele:
yes, a checkbox / date was my initial idea, but maybe it indeed needs some more thinking.

ppigazzini (approved):
Looks good on DEV

server/fishtest/views.py
```diff
@@ -139,14 +140,50 @@ def login(request):
     return {}
 
 
+def worker_email(worker_name, message, host_url, blocked):
+    owner = worker_name.split("-")[0]
+    wrapped_message = textwrap.fill(message, width=70)
+    body = f"""\
+Hi {owner},
+
+It seems your worker {worker_name} has some issue(s):
```
vondele, line 149: Maybe something like:

Thank you for contributing to the development of Stockfish. Unfortunately, it seems your fishtest worker {worker_name} has some issue(s):

ppigazzini: I write something like this:

Dear XXX,

  I'm YYY, one of the fishtest maintainers.

Thank you for contributing to the Stockfish/fishtest projects!

Unfortunately, your worker is having many time losses, see the Events log:

https://tests.stockfishchess.org/actions?action=crash_or_time&user=SimonDemel&text=

Feel free to contact me or by email or, better, on the Discord Stockfish server, #fishtest channel.

https://discord.com/invite/awnh2qZfTT

Enjoy your day,

vdbergh:
@ppigazzini  I like your proposal but it needs to integrate the message in the panel (there may be other things than time losses).

So perhaps
```
Dear <owner>,

I'm <approver>, one of the fishtest maintainers.

Thank you for contributing to the development of Stockfish. Unfortunately, it seems your fishtest worker <worker_name> 
has some issue(s):

<message>

See also https://tests.stockfishchess.org/actions?text=<worker_name>

Feel free to reply to this email for more information, or else contact the Discord Stockfish server, #fishtest channel.

https://discord.com/invite/awnh2qZfTT

It may be that your worker is currently blocked which means that it will not play any games. In that case you can unblock it at https://tests.stockfishchess.org/workers/<worker_name> after resolving the issue(s).

Enjoy your day,

<approver>
```
Of course it is only a template. One still has a chance to modify the text before actually sending the email.


There is one inconvenient issue that we do not know the real names of the users. We could add them to the db for the people that want to. It makes sense to make it a requirement for approvers.

ppigazzini:
Usually, I search for the real name on GitHub, in the email address, in the AUTHOR file etc.
When applicable, I report other similar cases, write some suggestions etc.
But these are steps to be made by the reporter case for case.

vdbergh:
I added a commit with an alternative template.

vdbergh:
> Usually, I search for the real name on GitHub, in the email address, in the AUTHOR file etc. When applicable, I report other similar cases, write some suggestions etc. But these are steps to be made by the reporter case for case.

Still it would make sense to add them to the db, at least for approvers. Their names are likely already in the commit log anway. So they are hardly a secret.

ppigazzini:
Some contributors don't want to have their real name in AUTHORS file, use a real email in commits etc.

vdbergh:
> Some contributors don't want to have their real name in AUTHORS file, or use a not incognito email in git commits etc.

Still many real names are public (e.g. from the author file). Why not add these to the db? Then we can use the real name if it is available and otherwise the username.

vdbergh:
In principle the db is more private than the AUTHORs file.

server/fishtest/views.py
```diff
@@ -139,22 +140,76 @@ def login(request):
     return {}
 
 
+def worker_email(worker_name, blocker_name, message, host_url, blocked):
+    owner = worker_name.split("-")[0]
+    wrapped_message = textwrap.fill(message, width=70)
+    body = f"""\
+Dear {owner},
+
+I'm {blocker_name}, one of the fishtest maintainers.
```
vondele, line 149: probably approvers instead of maintainers ?

ppigazzini:
DEV updated.
The emal link stop working with a blocking message with more than 259 characters, I wonder if there is a max length of the link.

vdbergh:
gmail does not seem to have this problem. Neither does the mail app on ipad. What mail application are you using?

ppigazzini:
I was using the combo Edge (chromium) + Thunderbid, with max supported length of the mailto link of 1810 characters.

This question hints to a Chromium limit:
https://stackoverflow.com/questions/4473959/effective-maximum-mailto-body-lengths

vdbergh:
I tried Chromium on Ubuntu 22.04 and it opens the mailto links without problem (so does Chrome). They are less than 2000 characters (I estimate) so there should be no problem.

I wonder if it is not Thunderbird that is the problem. Can you open the mailto link directly with Thunderbird?
```sh
 thunderbird <mailto link>
```
I can test it myself since thunderbird does not work at all for me.

ppigazzini:
I tried on a VM with the Windows Mail client.
The second link provided states that it's a long-time known problem with chromium on Windows.

vdbergh:
The mailto link is indeed 1810 characters.  Which is less than 2000 (which according to the 2nd link is when the problems start). But you say there is a problem anyway. Does it also occur with Chrome? Chromium is not really a consumer browser.

Perhaps we can make the text area smaller.

ppigazzini:
Chrome, Edge, Opera are chromium based and they don't work with a link bigger than 1810 characters.

vdbergh:
On Linux they do...

Is 1810 the hard limit on Windows? Then enforcing the message to be smaller will fix the problem.

ppigazzini:
I found the max working length of the mailto link adding and removing a character at time. However, the StackOverflow post reports a max length for the **encoded** link, and in the code the threshold is set to 1995, not 2000.

vdbergh:
By coincidence the technologov link is exactly 1810 characters (with encoding).

vdbergh:
I now cut the message off at 400 characters. Of course one needs to resubmit the form, otherwise the message will not change on the db.

We need to be smarter than just cutting the message off but perhaps not this PR.

ppigazzini:
> By coincidence the technologov link is exactly 1810 characters (with encoding).

Yes, by coincidence :D

DEV updated.

vdbergh:
Now a message is printed if the description is truncated.

ppigazzini:
DEV updated.

ppigazzini:
The Issue text message (with the technologov worker) must be max 256 chars to work on Windows with a chromium based browser. We should rework the emal content to save some of the other 1554 chars.

vdbergh:
Hmm. So the restriction is much more severe than 1810 characters. This is weird since on Linux there seems to be no restriction and it is the same code base.

We could include a link to the message instead of including the message directly in the email.

Alternatively we could try to make the email shorter.

vdbergh:
Ok I misunderstood. The length of 1810 of the technologov link was no accident but corresponded to a message of 256 characters (I thought the message was 500 characters).

That’s a serious problem. Perhaps dropping the last sentence (about unblocking yourself) might help.

vdbergh:
In the latest commit I made the email shorter.

ppigazzini:
DEV updated. With the new email template, an Issue message of 400 chars results in a mailto link of 1370 chars. Dropping the ***** helped a lot to save chars.

ppigazzini:

I raised the max_chars clamp to 500 in https://github.com/official-stockfish/fishtest/pull/1808, the technologov mailto link after that is 1503 chars, so ve have ~ 200 chars for a future revision of the mail text.

PR #1800, opened 2023-09-10
Update book to UHO_4060_v3

vondele:
this book adds back the move counters to UHO_4060_v2, restored from UHO_4060_v1, but is otherwise identical.


ppigazzini (approved):
Looks good on DEV runs

ppigazzini:

PR #1802, opened 2023-09-10
cleanup the cutechess setup code

Disservin:
refactor the cutechess setup code a bit to make it more readable

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 219
+WORKER_VERSION = 220
 
 
 def validate_request(request):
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):
Looks good on linux/windows workers, tested working:
- `cutechess-cli` clean download
- `cutechess-cli` download on `cutechess-cli` error
- `cutechess-cli` restore from backup

ppigazzini:
Triggered the workers update, thank you @Disservin :)

vdbergh:
The output of the worker now looks a bit strange
```
Post request https://dfts-0.pigazzini.it:443/api/request_version handled in 170.02ms (server: 0.09ms)
Obtaining version info for /home/vdbergh/fishtest/worker/testing/cutechess-cli ...
Found:  cutechess-cli 1.2.0
Obtaining version info for /home/vdbergh/fishtest/worker/testing/cutechess-cli ...
Found:  cutechess-cli 1.2.0
Downloading https://raw.githubusercontent.com/official-stockfish/fishtest/master/worker/sri.txt
```
I wonder if this is caused by this PR?

Disservin:
Ahh yes, current:

```python
    # Download cutechess-cli if missing or overwrite if there are issues.
    if not cutechess_path.exists() or not verify_required_cutechess(cutechess_path):
        download_cutechess(cutechess, testing_dir)

    ret = True

    if not verify_required_cutechess(cutechess_path):
```

this should fix it right?
```python
    # Download cutechess-cli if missing or overwrite if there are issues.
    if not cutechess_path.exists():
        download_cutechess(cutechess, testing_dir)
   
    if verify_required_cutechess(cutechess_path)
         os.chdir(curr_dir)
         return True
    else:
        download_cutechess(cutechess, testing_dir)

    ret = True

    if not verify_required_cutechess(cutechess_path):
```

vdbergh:
Sort of. But now it would do a pointless redownload if the downloaded cutechess-cli does not work. So the first else clause should go I think. 

Also the if clause after that becomes redundant.

vdbergh:
Hmm but if an existing cutechess-cli does not work then it has to do the download. So my comment above is not quite right.

vdbergh:
I think the code can also be simplified a bit by moving the check for the existence of `cutechess-cli` inside `verify_required_cutechess()`. Morally it belongs there I think.

Disservin:
Yeah that's a good idea ;) just keep the download in the setup part. I think then it's just?
```python
    if not verify_required_cutechess(cutechess_path):
        download_cutechess(cutechess, testing_dir)
    else:
         os.chdir(curr_dir)
         return True

    ret = True

    if not verify_required_cutechess(cutechess_path):
```

vdbergh:
Like that it seems ok to me....

vdbergh:
Please make a PR!

PR #1803, opened 2023-09-10
Use Git trees to look for books

dav1312:
This allows the use of folders in the books repo


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1804, opened 2023-09-10
Introduce a pgn bulk downloader

peregrineshahin:
Should be tested as I'm very new to this topic in python.

fixes: https://github.com/official-stockfish/fishtest/issues/1801

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -116,7 +116,7 @@ def main(global_config, **settings):
     config.add_route("api_get_task", "/api/get_task/{id}/{task_id}")
     config.add_route("api_upload_pgn", "/api/upload_pgn")
     config.add_route("api_download_pgn", "/api/pgn/{id}")
-    config.add_route("api_download_pgn_100", "/api/pgn_100/{skip}")
+    config.add_route("api_download_test_pgns", "/api/test_pgns/{id}")
     config.add_route("api_download_nn", "/api/nn/{id}")
     config.add_route("api_get_elo", "/api/get_elo/{id}")
     config.add_route("api_actions", "/api/actions")
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -491,13 +491,20 @@ class ApiView(object):
             self.request.response.content_type = "application/x-chess-pgn"
         return pgn
 
-    @view_config(route_name="api_download_pgn_100")
-    def download_pgn_100(self):
-        skip = int(self.request.matchdict["skip"])
-        urls = self.request.rundb.get_pgn_100(skip)
-        if urls is None:
-            raise exception_response(404)
-        return urls
+    @view_config(route_name="api_download_test_pgns")
+    def download_test_pgns(self):
+        test_id = self.request.matchdict["id"]
+        archive = self.request.rundb.get_test_pgns_as_archive(test_id)
+        if archive:
+            response = Response(content_type="application/zip")
+            response.app_iter = archive
+            response.content_length = archive.getbuffer().nbytes
+            response.headers[
+                "Content-Disposition"
+            ] = f'attachment; filename="{test_id}_pgns.zip"'
+            return response
+        else:
+            return Response("No data found", status=404)
 
     @view_config(route_name="api_download_nn")
     def download_nn(self):
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1,5 +1,6 @@
 import configparser
 import copy
+import io
 import math
 import os
 import random
@@ -9,6 +10,7 @@ import sys
 import textwrap
 import threading
 import time
+import zipfile
 import zlib
 from datetime import datetime, timedelta, timezone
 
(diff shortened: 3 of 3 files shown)


vondele:
works for me, one question, is there an upper limit that nginx or other settings enforce that might need fixing. These pgn zips will probably sometimes be really large.

peregrineshahin:
for some reason this downloads empty zip
https://dfts-0.pigazzini.it/tests/view/6511e363d17ea78d961ff26d
https://dfts-0.pigazzini.it/api/test_pgns/6511e363d17ea78d961ff26d

what am I missing?

peregrineshahin:
> works for me, one question, is there an upper limit that nginx or other settings enforce that might need fixing. These pgn zips will probably sometimes be really large.

@noobpwnftw

peregrineshahin:
Ah I'm assuming that whenever there is a pgn record in the database there must be a file, maybe that's okay?

> for some reason this downloads empty zip https://dfts-0.pigazzini.it/tests/view/6511e363d17ea78d961ff26d https://dfts-0.pigazzini.it/api/test_pgns/6511e363d17ea78d961ff26d
> 
> what am I missing?

vondele:
> for some reason this downloads empty zip https://dfts-0.pigazzini.it/tests/view/6511e363d17ea78d961ff26d https://dfts-0.pigazzini.it/api/test_pgns/6511e363d17ea78d961ff26d
> 
> what am I missing?

pgns are only available for specific tests started on the dev system, as pasquale mentioned higher up.

The largest zip I get from looking at a few tests (not exhaustive), is 492MB (1.6GB unzipped), for 61e48afe17407209066c58fc

peregrineshahin:
Yeah I don't know, it feels slow on DEV for 45MB, maybe we should update PROD to test before merging.

ppigazzini:
> Yeah I don't know, it feels slow on DEV for 45MB, maybe we should update PROD to test before merging.

Way faster simply storing the already zipped pgns.

```diff
diff --git a/server/fishtest/rundb.py b/server/fishtest/rundb.py
index 1a6baaa..a31d4f2 100644
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -251,10 +251,10 @@ class RunDb:
             return None

         archive = io.BytesIO()
-        with zipfile.ZipFile(archive, "w", zipfile.ZIP_DEFLATED) as zipf:
+        with zipfile.ZipFile(archive, "w", zipfile.ZIP_STORED) as zipf:
             for pgn in pgns:
-                pgn_data = zlib.decompress(pgn["pgn_zip"]).decode()
-                pgn_filename = f"{pgn['run_id']}.pgn"
+                pgn_data = pgn["pgn_zip"]
+                pgn_filename = f"{pgn['run_id']}.zip"
                 zipf.writestr(pgn_filename, pgn_data)

         archive.seek(0)
```

peregrineshahin:
Hmm, so this requires the user to unzip every pgn.zip again?
> Way faster simply storing the already zipped pgns.

ppigazzini:
Hmm, so that requires the server to unzip and zip every pgn again for all users?

> Hmm, so this requires the user to unzip every pgn.zip again?
> 
> > Way faster simply storing the already zipped pgns.

peregrineshahin:
I mean we still haven't tested how slow it is on PROD, we can nginx cache on the drive, How much drive we have on the server if we think about caching the reponse for some time?

> Hmm, so that requires the server to unzip and zip every pgn again for all users?
> 
> > Hmm, so this requires the user to unzip every pgn.zip again?
> > > Way faster simply storing the already zipped pgns.

peregrineshahin:
TBH, the PR is already an improvement over what we have already even if we zip the zipped pgns into big one, so feel free to edit it.

ppigazzini:
My proposed change doesn't work, the files within the zip are not really zip files, I think that they need a decoding.

ppigazzini:
> TBH, the PR is already an improvement over what we have already even if we zip the zipped pgns into big one, so feel free to edit it.

Right now, scraping the pgns the server must only unzip the files, with this PR the server must unzip and then zip the files doubling the load on the server.

peregrineshahin:
If the user wants to download all PGN's of the same test then they will scrape the available pgns, in that sitiuation which is the only real-world case why one would use this API, will recieving and serving 400 request per test be better than 1 request for the server load?

ppigazzini:
Way better only 1 request that does not require to unzip and zip 400 files.

peregrineshahin:
k.

ppigazzini:
@peregrineshahin if I'm reading correctly the code, the worker saves encoded text and then compresses it, the code decompresses the data and then decodes it to get back the text. We have not zipped text in the DB, the load on the server seems unavoidable without changing the worker/server code and converting the pgns collection.

https://github.com/official-stockfish/fishtest/blob/e53389f22878e0c7698ba58b8efd61d99820d999/worker/worker.py#L1454-L1466

https://github.com/official-stockfish/fishtest/blob/e53389f22878e0c7698ba58b8efd61d99820d999/server/fishtest/rundb.py#L244C1-L249C20

worker and server code should changed to this

```python
from pathlib import Path
import zipfile
import io

pgn_file = "file.pgn"
pgn_path = Path.cwd() / pgn_file
#data = bytes(pgn_path.read_text(), "utf-8").decode("utf-8", "ignore")
data = pgn_path.read_bytes().decode("utf-8", "ignore")

# create a ZipFile object in write mode and write the string to it
zip_buffer = io.BytesIO()
with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zipf:
    # Write the data to the zip file three times under different names
    for i in range(3):
        zipf.writestr(f'file{i}.pgn', data)

# get the zip data as a bytes object
zip_data = zip_buffer.getvalue()

# write the zip data to a file on the filesystem
with open('compressed_pgn.zip', 'wb') as f:
    f.write(zip_data)

# create another ZipFile object in read mode and read the string from it
zip_buffer = io.BytesIO(zip_data)
with zipfile.ZipFile(zip_buffer, 'r') as zipf:
    text = zipf.read("file0.pgn").decode()

# write the string to a new text file
with open('decompressed.pgn', 'w') as f:
    f.write(text)
```

ppigazzini:
> works for me, one question, is there an upper limit that nginx or other settings enforce that might need fixing. These pgn zips will probably sometimes be really large.

I artificially raised 15x the number of pgns on DEV, no problem downloading a zipped file of 450GB

peregrineshahin:
Superseded with https://github.com/official-stockfish/fishtest/pull/1823

PR #1809, opened 2023-09-21
Encode \n as crlf in email body.

vdbergh:
This is per the spec. It makes gmail on ipad work (before, all linefeeds were suppressed).

### DIFF
--- a/server/fishtest/templates/workers.mak
+++ b/server/fishtest/templates/workers.mak
@@ -54,7 +54,7 @@
           <td>${delta_date(diff_date(w["last_updated"])) if last_updated is not None else "Never"}</td>
           <td><a href="/actions?text=%22${w['worker_name']}%22">/actions?text="${w['worker_name']}"</a></td>
 % if show_email:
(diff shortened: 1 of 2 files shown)


vdbergh:
Most email clients are smart enough to recognize \n as line separator. So for those clients there is no difference. The only exception I know is gmail on iPad (gmail on android is fine).

Let me also try to fix the formatting.

vdbergh:
The formatting should be fixed now.

ppigazzini:
DEV updated, In thunderbid the formatting is mostly fixed, a bit strange the Issue text formatting, though.

image

vdbergh:
So it seems that on windows the string returned from a textarea already contains \r\n. And it does not end in a line feed.

vdbergh:
Hopefully fixed.

vdbergh:
Converted to draft since I think it is better to convert the linefeeds before the message is stored in the db.

vdbergh:
Undrafted. You have to resubmit the message (the one in the db still contains \r\n).

ppigazzini:
DEV updated, looks good! The mailto link has 1600 chars, with a Issue of 500 chars.

image

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1811, opened 2023-09-23
Various validation fixes.

vdbergh:
- Use f-strings in `validate()`.
- Fix bug which prevented strict validation with optional keys.
- Tighten up the user schema and use strict validation (strict validation means that one cannot introduce keys which are not in the schema).
- Add some unittests for the validation code. There is in particular a test that shows how to create a type `lower_case_string`.

### DIFF
--- a/server/fishtest/userdb.py
+++ b/server/fishtest/userdb.py
@@ -3,10 +3,12 @@ import threading
 import time
 from datetime import datetime, timezone
 
+from bson.objectid import ObjectId
 from fishtest.util import optional_key, validate
 from pymongo import ASCENDING
 
 schema = {
+    optional_key("_id"): ObjectId,
     "username": str,
     "password": str,
     "registration_time": datetime,
@@ -14,7 +16,7 @@ schema = {
     "email": str,
     "groups": list,
     "tests_repo": str,
-    optional_key("machine_limit"): int,
+    "machine_limit": int,
 }
 
 DEFAULT_MACHINE_LIMIT = 16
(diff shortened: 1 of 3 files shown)


vdbergh:
In recent year Python has developed a type annotation system which can be used by external tools such as mypy. However I do not know how to use this from within python.

For example in python you can use the type annotation `list[str]` for a list of strings. However I do not know a fast way to verify that an object is an instance of `list[str]`... E.g. the following does not work
```python
>>> T=list[str]
>>> a=['a']
>>> isinstance(a,T)
TypeError: isinstance() argument 2 cannot be a parameterized generic
>>> 
```

ppigazzini:
DEV updated.

ppigazzini:
The output of the server tests seems odd:
```
...done
ok
test_keys (test_validate.TestValidation.test_keys) ... ok
test_missing_keys (test_validate.TestValidation.test_missing_keys) ... ok
test_strict (test_validate.TestValidation.test_strict) ... ok
test_union (test_validate.TestValidation.test_union) ... ok
test_validate (test_validate.TestValidation.test_validate) ... ok

Ran 26 tests in 10.652s

OK
my_object['b'] is missing
my_object['c'] is not in the schema
my_object['c'] is not in the schema
my_object['b'] is not equal to 2 and my_object['b'] is not equal to 3
my_object is not a string
A, contained in the string my_object, is not a lower case letter
A, contained in the string my_object['a'], is not a lower case letter
```

vdbergh:
Afk. Wil have a look later.

vdbergh:
Actually this is as intended. What is printed are the error messages of the failed validations. I can delete them if you want.

vdbergh:
After thinking it over I decided to deleted the printouts.

ppigazzini (approved):
Looks good on DEV.
I created few users, ran a test with some workers, checked the logs.

ppigazzini:

PR #1815, opened 2023-09-25
Tighten up parsing of "Finished game" lines in cutechess-cli output.

vdbergh:
See https://tests.stockfishchess.org/actions?max_actions=3&text=%22update_pentanomial%22&before=1695591050.683729

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -25,7 +25,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 217
+WORKER_VERSION = 218
 
 
 def validate_request(request):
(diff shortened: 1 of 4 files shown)


vdbergh:
The third of these messages is a bit disturbing. 

- In UCI an engine cannot claim a result. So this suggest an internal bug in cutechess-cli.
- Apparently sometimes error messages and genuine output message are not interleaved cleanly (this is maybe not unexpected as they are written to different streams which are merged by Python).

vdbergh:
Actually the other cases recorded in the event log URL appear to be of similar nature. Unclean interleaving of error and non-error messages.

vdbergh:
Independently of this PR. I wonder if it would not be better to poll both p.stderr and p.stdout in the worker, instead of relying on Python to merge the streams for us.

ppigazzini (approved):

ppigazzini:

PR #1817, opened 2023-09-26
Add monospace fixed size font for numbers.

peregrineshahin:
closes: https://github.com/official-stockfish/fishtest/issues/1542


ppigazzini:
DEV updated.
On Windows I don't view any improvement in readability using the monospace, the combination normal font + monospace looks a bit odd as well.

peregrineshahin:
Yeah it seems useless for readability IMO.

> DEV updated. On Windows I don't view any improvement in readability using the monospace, the combination normal font + monospace looks a bit odd as well.

PR #1818, opened 2023-09-28
Implement efficient PGN upload and bulk download

ppigazzini:
The previous worker code posted the PGN to the server in an encoded and
 zlib-compressed format. This required the server to decompress and decode
 the data before serving the PGN as text, resulting in significant CPU
 and network overhead. Furthermore, a bulk download of all PGNs from a run
 necessitated additional CPU overhead for compression.

The updated worker code posts a zip file containing the PGN text
 to the server, leading to more efficient handling on the server side:
- for single PGN downloads, the server simply serves
 the zipped PGN retrieved from MongoDB
- for bulk PGN downloads, the server writes a zip file in store mode
 containing all the zipped PGNs retrieved from MongoDB

@peregrineshahin wrote the bulk download code in https://github.com/official-stockfish/fishtest/pull/1804

closes https://github.com/official-stockfish/fishtest/issues/1801

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -116,7 +116,7 @@ def main(global_config, **settings):
     config.add_route("api_get_task", "/api/get_task/{id}/{task_id}")
     config.add_route("api_upload_pgn", "/api/upload_pgn")
     config.add_route("api_download_pgn", "/api/pgn/{id}")
-    config.add_route("api_download_pgn_100", "/api/pgn_100/{skip}")
+    config.add_route("api_download_run_pgns", "/api/run_pgns/{id}")
     config.add_route("api_download_nn", "/api/nn/{id}")
     config.add_route("api_get_elo", "/api/get_elo/{id}")
     config.add_route("api_actions", "/api/actions")
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -482,14 +482,14 @@ class ApiView(object):
         )
         return self.add_time(result)
 
-    def download_archive(self, archive_name, archive):
-        if archive:
+    def download_archive(self, zip_name, zip_buffer):
+        if zip_buffer:
             response = Response(content_type="application/zip")
-            response.app_iter = archive
-            response.content_length = archive.getbuffer().nbytes
+            response.app_iter = zip_buffer
+            response.content_length = zip_buffer.getbuffer().nbytes
             response.headers[
                 "Content-Disposition"
-            ] = f'attachment; filename="{archive_name}"'
+            ] = f'attachment; filename="{zip_name}"'
             return response
         else:
             return Response("No data found", status=404)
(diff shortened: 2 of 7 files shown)


vdbergh:
```python
    if pgn_file[0] is not None:
        pgn_file = pgn_file[0]
```

Perhaps not relevant for this PR, but there is no real need for these gymnastics. The name of the pgn_file
can be computed before calling `run_games()` and passed as a parameter (in other words it does not have to be computed inside `run_games()` and passed back somehow).

ppigazzini:
> ```python
>     if pgn_file[0] is not None:
>         pgn_file = pgn_file[0]
> ```
> 
> Perhaps not relevant for this PR, but there is no real need for these gymnastics. The name of the pgn_file can be computed before calling `run_games()` and passed as a parameter (in other words it does not have to be computed inside `run_games()` and passed back somehow).

I agree as discussed yesterday on Discord;
> The weird code is `pgn_file[0]`. `game()` should simply return `pgn_file` (or `None`) and not change the content of a `pgn_file` list passed as parameter. That is C/C++ coding style.

vdbergh:
> I agree as discussed yesterday on Discord;
> 
> > The weird code is `pgn_file[0]`. `game()` should simply return `pgn_file` (or `None`) and not change the content of a `pgn_file` list passed as parameter. That is C/C++ coding style.

That is tricky since `run_games()` typically returns via an exception.

vdbergh:
But even for a function that does not use exceptions but has many early return statements, it may be tricky to return auxilliary data because you have to include that data with every return statement, which is ugly. In that case it may be easier to return the data via a parameter, which only has to be set once.

ppigazzini:
The PR is a draft because I must still add:
- a script to convert the format of the documents in the `pgns` collection
- a temporary function that check if the compressed data is not a zip and convert to zip, to be used in `upload_pgn()`. The function could be temporary used in `download_pgn()` and `download_run_pgns()` as well to be extremly safe.

ppigazzini:
With a fast test (not checked the code) the name seems to be (pseudo python) `pgn_file = testing_dir / f"result-{UUID}.pgn"`

ppigazzini:
By the way using a unique name like `pgn_file = testing_dir / f"{UUID}-{run_id}-{task_id}.pgn"` (the UUID seems not really necessary) should guarantee to upload the right file and use simply the check `if pgn_file.exists()`. We can pass the name to the nested calls or simply use the same naming convention in `worker.py` and `games.py`

vdbergh:
> By the way using a unique name like `pgn_file = testing_dir / f"{UUID}-{run_id}-{task_id}.pgn"` (the UUID seems not really necessary) should guarantee to upload the right file and use simply the check `if pgn_file.exists()`. We can pass the name to the nested calls or simply use the same naming convention in `worker.py` and `games.py`

I would pass the name as an argument to `run_games()` (`cutechess-cli` needs it). Choices should preferably be made in only one place.

vdbergh:
I agree that UUID does not seem necessary. The pgn is a property of the task which is characterized by `run_id/task_id`.

ppigazzini:
Superseded by #1823

PR #1819, opened 2023-09-29
fix double cutechess verification

Disservin:
as discussed in https://github.com/official-stockfish/fishtest/pull/1802, this should fix the double cutechess verification

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -27,7 +27,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 221
+WORKER_VERSION = 222
 
 
 def validate_request(request):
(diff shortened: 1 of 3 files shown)


vdbergh:
It is still not 100% robust. I replaced the `cutechess-cli` binary by `/bin/ps` and I got an uncaught exception
```
Obtaining version info for /home/vdbergh/fishtest/worker/testing/cutechess-cli ...
Found:  cutechess-cli from procps-ng 3.3.17
  File "/home/vdbergh/fishtest/worker/worker.py", line 1668, in <module>
    sys.exit(worker())
  File "/home/vdbergh/fishtest/worker/worker.py", line 1570, in worker
    if not setup_cutechess(worker_dir):
  File "/home/vdbergh/fishtest/worker/worker.py", line 471, in setup_cutechess
    if not verify_required_cutechess(cutechess_path):
  File "/home/vdbergh/fishtest/worker/worker.py", line 428, in verify_required_cutechess
    major = int(m.group(1))
ValueError: invalid literal for int() with base 10: ''
Deleting lock file /home/vdbergh/fishtest/worker/worker.lock
```

Disservin:
Well that's outside the scope of this pr, since that never worked, right?

vdbergh:
If we replace `cutechess-cli` by `/bin/ps` then `cutechess-cli --version` yields
```
cutechess-cli from procps-ng 3.3.17
```
So `/bin/ps` does some funky stuff with its zeroth argument.

vdbergh:
> Well that's outside the scope of this pr, since that never worked, right?

I guess so. But tightening the regular expression a little bit fixes it
```python
pattern = re.compile("cutechess-cli ([0-9]+)\.([0-9]+)\.([0-9]+)")
```

ppigazzini (approved):
Looks good on linux/msys2 worker.
Tightened the regex as well.

ppigazzini:
Triggered the workers update, thank you @Disservin :)

PR #1821, opened 2023-09-30
Bugfix in worker management.

vdbergh:
If the admin panel is shown and the worker in the admin panel has never been updated then all the last_updated fields in the list of blocked workers are shown as "Never".


ppigazzini:
DEV updated, let me know if the bug is fixed.

vdbergh:
It seems to be fixed.

ppigazzini:

PR #1822, opened 2023-10-01
Prevent same email registration.

peregrineshahin:


vdbergh:
```
    result = request.userdb.create_user(
        username=signup_username, password=signup_password, email=validated_email
    )
    if not result:
        request.session.flash("Error! Invalid username or password", "error")
```
This error message in `signup()` should be adapted.

peregrineshahin:
Done
> ```
>     result = request.userdb.create_user(
>         username=signup_username, password=signup_password, email=validated_email
>     )
>     if not result:
>         request.session.flash("Error! Invalid username or password", "error")
> ```
> 
> This error message in `signup()` should be adapted.

vdbergh:
If the Fishtest server could send email (i.e. if the DNS issues are sorted out) then it would be possible to send the user a confirmation link.

Then the life cycle of the user state would be
```
pending-->confirmed-->accepted-->blocked-->unblocked-->blocked-->...
```

peregrineshahin:
indeed, a confirmation email is neccessary IMO. We use the email for contacting concerning worker issues and we should know that we are not sending someone else.

server/fishtest/userdb.py
```diff
@@ -114,6 +120,7 @@ def create_user(self, username, password, email):
 
             return True
         except:
+            request.session.flash("Error! Invalid username or password", "error")
```
vdbergh, line 123: This will not work no? `create_user()` does not know about `request`. Besides `userdb` is for handling the db, not for user interaction.

peregrineshahin: > This will not work no? `create_user()` does not know about `request`. Besides `userdb` is for handling the db, not for user interaction.

Please check now

peregrineshahin:
Does that replace manual accepting users?

> If the Fishtest server could send email (i.e. if the DNS issues are sorted out) then it would be possible to send the user a confirmation link.
> 
> Then the life cycle of the user state would be
> 
> ```
> pending-->blocked-->unblocked-->blocked-->unblocked-->blocked-->...
> ```
> 
> EDIT: I shortened the life cycle even more. The transition pending-->blocked should happen after the confirmation link has been clicked.

server/fishtest/userdb.py
vdbergh: This seems ok!

vdbergh:
> Does that replace manual accepting users?
> 
That's a policy choice. One could decide that after receiving the confirmation the user is automatically unblocked. Or one could require manual intervention by an approver to unblock the user.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1823, opened 2023-10-01
Implement efficient PGN upload and bulk download - gzip and tar

ppigazzini:
The previous worker code posted the PGN to the server in a zlib-compressed format. This required the server to decompress the data before serving the PGN as text, resulting in significant CPU and network overhead. Furthermore, a bulk download of all PGNs from a run necessitated additional CPU overhead for compression.

The updated worker code posts a gzip file containing the PGN text to the server, leading to more efficient handling on the server side:
- for single PGN downloads, the server simply serves the gzipped PGN retrieved from MongoDB. All modern browsers decompress the gzip file client side, so the user will get the pgn as text file. nginx can be configured to decompress on the fly the gzip content for a client that is not declaring his compression capability.
- for bulk PGN downloads, the server writes a tar file containing all the gzipped PGNs retrieved from MongoDB ``` wget https://test.stockfishchess.org/api/run_pgns/<run_id>.pgns.tar ```
- gzip module uses zlib, no change for the compression ratio.

@peregrineshahin wrote the bulk download code
in https://github.com/official-stockfish/fishtest/pull/1804

Retire "pgn_100" api.

closes https://github.com/official-stockfish/fishtest/issues/1801

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -116,7 +116,7 @@ def main(global_config, **settings):
     config.add_route("api_get_task", "/api/get_task/{id}/{task_id}")
     config.add_route("api_upload_pgn", "/api/upload_pgn")
     config.add_route("api_download_pgn", "/api/pgn/{id}")
-    config.add_route("api_download_pgn_100", "/api/pgn_100/{skip}")
+    config.add_route("api_download_run_pgns", "/api/run_pgns/{id}")
     config.add_route("api_download_nn", "/api/nn/{id}")
     config.add_route("api_get_elo", "/api/get_elo/{id}")
     config.add_route("api_actions", "/api/actions")
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -1,5 +1,7 @@
 import base64
 import copy
+import io
+import re
 from datetime import datetime, timezone
 
 from fishtest.stats.stat_util import SPRT_elo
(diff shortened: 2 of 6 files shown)


ppigazzini:
This is a run on DEV with all gzipped pgns:
https://dfts-0.pigazzini.it/api/run_pgns/65182ad6bf7d207571a5883
```
curl https://dfts-0.pigazzini.it/api/run_pgns/65182ad6bf7d207571a5883.tar | tar -x && gunzip *.gz
```

peregrineshahin:
Looks good. what advantages do we have in using `tar` vs `zip` for bulk download, sorry I'm a Windows user; I had to ask..

vondele:
tar as a more native tool for linux users, no need to compress a second time?

ppigazzini:
> tar as a more native tool for linux users, no need to compress a second time?

No real difference between the two 2 PRs from a performance point of view, the container zip is in store mode (same as the tar).

I have a little bias for this tar gz PR (I do like the browser decompression for the single pgn download), in my opinion the choice must be done from the users of the bulk download process.

ppigazzini:
This is the companion script to convert the pgns collection from the old zlib format to the new gzip format.

```python
#!/usr/bin/env python3
import gzip
import io
import zlib

from bson.binary import Binary


def process_pgn(pgn, rundb):
    run_id = pgn["run_id"]
    pgn_zip = pgn["pgn_zip"]

    try:
        with gzip.open(io.BytesIO(pgn_zip), "rb") as test_f:
            test_f.read(1)
    except OSError:
        try:
            pgn_text = zlib.decompress(pgn_zip).decode()
            with io.BytesIO() as gz_buffer:
                with gzip.GzipFile(
                    filename=f"{run_id}.pgn.gz", mode="wb", fileobj=gz_buffer
                ) as gz:
                    gz.write(pgn_text.encode())
                pgn_zip = gz_buffer.getvalue()
                rundb.pgndb.update_one(
                    {"_id": pgn["_id"]}, {"$set": {"pgn_zip": Binary(pgn_zip)}}
                )
            print(f"converted: {run_id}")
        except Exception as e:
            print(f"exception converting: {run_id}, {e}")
    else:
        print(f"skipped: {run_id}")


def main():
    rundb = RunDb()

    for pgn in rundb.db.pgns.find():
        process_pgn(pgn, rundb)

    msg = rundb.db.command({"compact": "pgns"})
    print(msg)


if __name__ == "__main__":
    main()
```

ppigazzini:
- [ ] Triggered the workers update
- [ ] Backup pgns collection
- [ ] Convert pgns collection to the new gzip format

PR #1828, opened 2023-10-07
Switch to strict parsing of worker api's.

vdbergh:
Currently workers can add arbitrary fields to the api payloads, some of which may find their way into the db. This PR makes this impossible.

Note: strict parsing is possible thanks to the fix of a validation bug in #1811.

### DIFF
--- a/server/tests/test_api.py
+++ b/server/tests/test_api.py
@@ -1,4 +1,5 @@
 import base64
+import copy
 import gzip
 import io
 import sys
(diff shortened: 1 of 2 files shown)


vdbergh:
CI fails with black. But on my system black returns without errors. Slightly different versions I guess.

server/tests/test_api.py
```diff
@@ -189,6 +196,7 @@ def test_request_task(self):
         self.assertTrue(task["active"])
 
     def test_update_task(self):
+
```
ppigazzini, line 199: black dislike this empty line

ppigazzini (approved):

ppigazzini:
thank you @vdbergh :)

PR #1830, opened 2023-10-08
Don't allow non-approvers to increase priority/throughput beyond limits

XInTheDark:
Add a check to restrict a user from changing their test's priority above 0, or throughput to above 100 - now only approvers have permission to do this.


vondele:
This looks good with 200 replaced by 100.

ppigazzini:
DEV updated, modifying the Priority/Throughput as normal user the server flashes the error message, but the run is modified.
image

The PR needs some works:
1. the check introduced by the PR shuld be done before modifying the run:
https://github.com/official-stockfish/fishtest/blob/89687da7eace9cab249fd60c4ed11ea039046a3b/server/fishtest/views.py#L1124-L1132
2. the user can still create a new run with custom Priority/Throughput higher than the max values set in the PR:
https://github.com/official-stockfish/fishtest/blob/35acb37bd18d9e6299e535a5e326c91a0566cdc5/server/fishtest/templates/tests_run.mak#L523-L542

peregrineshahin:
this got superseded, it can be closed.

PR #1834, opened 2023-10-08
VLTC test presets

dav1312:
The tests are hidden, just like the PT tests, but at least this allows for some s13n

key parameters
```
"name": "VLTC",
"tc": "180+1.8",
"throughput": "50",
"threads": 1,
"options": "Hash=192",
"bounds": "standard STC",


"name": "VLTC SMP",
"tc": "60+0.6",
"throughput": "50",
"threads": 8,
"options": "Hash=512",
"bounds": "standard LTC",
```
_(not tested)_


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1835, opened 2023-10-08
Fix android bug

peregrineshahin:
fix resize events keeps triggering old error alerts if the keyboard shows or hides on  mobile.

fixes: https://discord.com/channels/435943710472011776/900771437177094156/1160537676286541824

### DIFF
--- a/server/fishtest/static/js/calc.js
+++ b/server/fishtest/static/js/calc.js
@@ -8,6 +8,8 @@ const defaultParameters = {
   "rms-bias": "191",
 };
 
+let valid_sprt = null;
+
 google.charts.load("current", { packages: ["corechart"] });
 
 let pass_chart = null;
@@ -37,10 +39,10 @@ google.charts.setOnLoadCallback(function () {
   mouse_screen.addEventListener("mousemove", handle_tooltips, true);
   mouse_screen.addEventListener("mouseleave", handle_tooltips, true);
   set_fields();
-  draw_charts();
+  draw_charts(false);
   window.onresize = function () {
     clearTimeout(resize_timeout);
-    resize_timeout = setTimeout(draw_charts, 100);
+    resize_timeout = setTimeout(() => draw_charts(true), 100);
   };
 });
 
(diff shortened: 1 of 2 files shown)


vdbergh:
I think we still want the resize to happen no? Shouldn’t we only suppress the stale error message?

vdbergh:
Hmmm this seems hard to achieve with the current code. The input validation should have been done on submission.

vdbergh:
Perhaps the script should remember the last good parameters, and rewrite the input fields to the good parameters, if the new parameters don’t validate.

peregrineshahin:
> I think we still want the resize to happen no? Shouldn’t we only suppress the stale error message?

Good point! Yes we should resize regardless

peregrineshahin:
> Perhaps the script should remember the last good parameters, and rewrite the input fields to the good parameters, if the new parameters don’t validate.

100%

peregrineshahin:
Well not really rewrite the input, Yes remember the old params to resize it again but not play with the input field

vdbergh:
That's a matter of taste. I would feel uncomfortable that the graph shown may not correspond to the input fields. But I can also see why one would think otherwise. I have no strong feelings about this.

vdbergh:
What about the URL arguments? If they are invalid then `draw_charts()` will be done with `sprt=null` it seems to me.

`valid_sprt` could be initialized with default values. But then the user would not know what they are since we are not updating the input fields.

vdbergh:
Currently the graphs are blank if the URL arguments are invalid. I guess that makes sense too.

server/fishtest/static/js/calc.js
```diff
@@ -60,35 +62,40 @@ function set_fields() {
...
-  if (val != "") {
-    alert(val);
-    return;
-  }
+  if (error) {
+    // do not show a stale alert with a resize
+    // and use last valid sprt to draw chart
+    if (!resize || !valid_sprt) {
+      alert(val);
+      return;
```
vdbergh, line 95: I guess you mean `alert(val)` here.

ppigazzini (approved):
Looks good on DEV and android

ppigazzini:

PR #1840, opened 2023-10-27
Fix bench extraction

peregrineshahin:
The Stockfish CI has picked a more suitable criteria for extracting bench in https://github.com/official-stockfish/Stockfish/commit/ca5d9a5ff0739a63f7b0a184193dfb9de3c57156 This commit adapts fishetst bench extraction to that. Fishtest Wiki instructions are outdated (as of now) by that Stockfish commit.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -650,26 +650,35 @@ def users_monthly(request):
 
 def get_master_info(url):
     try:
-        commits = requests.get(url)
-        commits.raise_for_status()
+        response = requests.get(url)
+        response.raise_for_status()
     except Exception as e:
         print(f"Exception getting commits:\n{e}")
         return None
+
     bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)")
-    for idx, commit in enumerate(commits.json()):
+    latest_bench_match = None
+
+    commits = response.json()
+    message = commits[0]["commit"]["message"].strip().split("\n")[0].strip()
+    date_str = commits[0]["commit"]["committer"]["date"]
+    date = datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%SZ")
+
+    for commit in commits:
         message_lines = commit["commit"]["message"].strip().split("\n")
-        bench = bench_search.search(message_lines[-1].strip())
-        if idx == 0:
-            message = message_lines[0].strip()
-            date_str = commit["commit"]["committer"]["date"]
-            date = datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%SZ")
-        if bench:
-            return {
-                "bench": bench.group(2),
-                "message": message,
-                "date": date.strftime("%b %d"),
-            }
-    return None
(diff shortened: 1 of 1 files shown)


vdbergh:
Not necessarily this PR, but it seems that if the current commit does not contains a bench statement (because it is non-functional), we could pick up the proper bench from the parent commits (to avoid the dreaded "This commit has no signature: please supply it manually").

peregrineshahin:
That's not actually what this PR fixes, what you say is already there in both before the PR and after it.
This PR is neccessay because fishtest have reported the wrong bench from the previous commit while the current commit had the bench in SF repo and passed the CI. There is inconsistency between CI and fishtest.

> Not necessarily this PR, but it seems that if the current commit does not contains a bench statement (because it is non-functional), we could pick up the proper bench from the parent commits (to avoid the dreaded "This commit has no signature: please supply it manually").

peregrineshahin:
Oh I see what do you mean. you mean to do the same also for new test and not only for the base test.
I don't think that's a good thing to do, if the user have forgot to provide the bench in commit message in his new patch then that solution will pick from the previous merged commit and thus the user could think they have done everything write and until the test is approved and running and the bench turns out wrong then he will fix... not sure about that

vdbergh:
No I meant the base commit. I was not aware this was already implemented. Sorry for the noise.

vdbergh:
I wonder how many people actually put bench lines in their test commits. To me this seems to be something you do when you squash while submitting a PR. But perhaps organized people do it before submitting a test.

peregrineshahin:
I think most SF devs do that. I notice that easily when seeing their commit messages after submittion under `new_tag`.
Basically more than 90% do that I would say.

> I wonder how many people actually put bench lines in their test commits. To me this seems to be something you do when you squash while submitting a PR. But perhaps organized people do it before submitting a test.

locutus2:
Yes, i do it always. You have anyway run the bench locally and so you can add it also to the commit message immediatly

vondele:
almost all devs do it, it is very convenient.

server/fishtest/views.py
```diff
@@ -785,9 +794,11 @@ def validate_form(request):
 
     def strip_message(m):
         lines = m.strip().split("\n")
-        last_line = lines[-1].strip()
-        if re.match(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)", last_line):
-            lines[-1] = ""
+        pattern = r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)"
```
ppigazzini, line 797:
```python
    bench_search = re.compile(r"(^|\s)[Bb]ench[ :]+([1-9]\d{5,7})(?!\d)")  
    for i, line in enumerate(reversed(lines)):
        new_line, n = bench_search.subn("", line)
        if n > 0:
            lines[-i-1] = new_line
            break
```

ppigazzini (approved):
Looks good on DEV

ppigazzini:

Disservin:
Thanks.

PR #1841, opened 2023-10-28
Small cleanups in the run db.

vdbergh:
- Remove the field `"results_stale"` since now the results are updated incrementally. So they should never be stale.

- Retire `get_results()` and replace it by `compute_results()` which has no side effects (`get_results()` was terrible from this point of view). `compute_results()` is called in two places: 
   - in `purge_run()` 
   - after a run is finished to verify the incrementally updated results.
- Set the `"workers"` and `"cores"` fields equal to zero when the run is finished.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -206,9 +206,10 @@ class RunDb:
                 "time_losses": 0,
                 "pentanomial": 5 * [0],
             },
-            "results_stale": False,
             "approved": False,
             "approver": "",
+            "workers": 0,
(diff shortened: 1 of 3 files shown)


vdbergh:
Drafted because I need to think this through a bit more.

vdbergh:
Undrafted.

ppigazzini (approved):
Looks good on DEV.

ppigazzini:
PROD updated, nice cleanup, thank you @vdbergh :)

PR #1842, opened 2023-10-29
Introduce generics in validation.

vdbergh:
We can now write
```
"groups" : list[str]
```
in the schema for the userdb.

Other changes:

- Put the validation code in a separate file.
- Refactor the validation code so that it is easier to work with. 
-  Introduce an email type.
- Introduce a regex type (not yet used).
- Create unit tests for the added code.


vdbergh:
My plan is to introduce a schema for the run object. One needs generics to describe the type of the task list.

vdbergh:
I drafted this again because I am making a larger PR which includes the schema for the run object.

PR #1844, opened 2023-11-02
After introduction of a itp correction based on number of running tests we have the odd situation that running more then 6 tests leads to a smaller total itp (and so total cores). Here the formula is changed that the total itp for more than 6 running tests is fixed to the same value as if tunning 6 tests.

locutus2:
…ts we have the odd situation that running more then 6 tests leads to a smaller total itp (and so total cores). Here the formula is changed that the total itp for more than 6 running tests is fixed to the same value as if tunning 6 tests.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -671,7 +671,11 @@ class RunDb:
                 itp *= bonus
 
         # Malus for too many active runs
-        itp *= 36.0 / (36.0 + count * count)
+        if count > 6:
+            # total itp for all user tests is the same as for 6 running tests
+            itp *= 3.0 / count
+        else:
+            itp *= 36.0 / (36.0 + count * count)
 
         run["args"]["itp"] = itp
 


locutus2:
It was always odd that a user which run more than 6 tests at once starts to loosing cores in the total. This fix keeps the total itp the same for 6 or more tests. 

We had discussed this also here https://github.com/official-stockfish/fishtest/pull/1831. But i factored this out for an own PR.

Maintainers please check this. I have no local fishtest instance for really testing this but i have checked the changed code part in an online python interpreter and there it works as expected.

locutus2:
I could use in this PR also my original formula from the other PR:
```
capped_count = min(count, 6)
itp *= 36.0 / (36.0 + capped_count * capped_count ) * capped_count  / count 
```

But i think this hiddes the fact that we can here two different functions combined (aka piece wise function) with the common point at 6 tests. So the if statement seems more explicit.

vondele:
The original choice was actually on purpose... it is there as an incentive to those having many tests running, to prioritize their own tests by pausing overlapping ideas.

locutus2:
@vondele 
But that is mainly archived by the behavior of the formula up to 6 tests. if i would use 12 instead of 6 my formula means that each test uses only half the resources instead of even less as in current version. And i think that should be the decision of the dev. But i think both ways are ok. Most important its contrary to my aesthetic sensibilities because its seem unlogical and counterintuitive ;-). I know in Stockfish we have several such code but at least there we gained elo!

locutus2:
As the point of view of the maintainers are now clear and no further discussion come up i close the PR.

PR #1845, opened 2023-11-02
Silence unittests by setting sys.stdout to None

vdbergh:
We also deleted a few print statement that were obviously leftover debug prints.

We changed two messages in userdb.py from stderr to stdout.


vdbergh:
I am closing this since doing `sys.stdout = None` in the global scope is too ugly. Unfortunately we cannot very well  redirect/restore `stdout` in `setup`, `tearDown` since `buffer(run, True)` seems to run in separate thread and prints to `stdout` after `tearDown` has been executed.

PR #1848, opened 2023-11-06
convert FENs in .epd to EPDs with move counters

robertnurnberg:
This PR changes downloaded (and existing) `.epd` books in the testing directory: if all lines contain full FENs (with 6 fields, including both move counters), then these are converted into the corresponding EPDs.  `cutechess-cli` only supports the latter as input from `.epd` files, and hence replaces all existing move counters with `0 1`.

With the change the playing stockfish versions will have access to the correct move number, and so even with future changes to the WDL model (where the normalization constant may become ply dependent) the correct evaluations will be displayed.

Note that this PR implements option (3) as discussed with @vondele on [discord](https://discord.com/channels/435943710472011776/1032922913499783169/1170759469903401091)

Update: [Link](https://discord.com/channels/435943710472011776/900771437177094156/1171195594765774899) to discord discussion of this PR.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -27,7 +27,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 222
+WORKER_VERSION = 223
 
 
 def validate_request(request):
--- a/worker/games.py
+++ b/worker/games.py
@@ -457,6 +457,24 @@ def unzip(blob, save_dir):
     return file_list
 
 
+def convert_book_move_counters(book_file):
+    # converts files with complete FENs, leaving others (incl. converted ones) unchanged
+    epds = []
+    with open(book_file, "r") as file:
+        for fen in file:
+            fields = fen.split()
+            if len(fields) == 6:
+                fields[4] = f"hmvc {fields[4]};"
+                fields[5] = f"fmvn {fields[5]};"
+                epds.append(" ".join(fields))
+            else:
+                return
+
+    with open(book_file, "w") as file:
+        for epd in epds:
+            file.write(epd + "\n")
+
+
 def clang_props():
     """Parse the output of clang++ -E - -march=native -### and extract the available clang properties"""
     with subprocess.Popen(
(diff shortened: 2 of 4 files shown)


peregrineshahin:
I don't think this is the place where to fix this.
The .epd is the culprit and is thus what needs to be fixed.

vondele:
I think this is a reasonable place to do this transformation, it balances between having the right move counters in the game, and not disrupting other tools out there that might rely on the books as we have it today (essentially being fens).. for example being able to feed directly in SF bench.  Doing it 'in place' at least brings us the right move counters (as verified on dev https://dfts-0.pigazzini.it/api/pgn/653569a8f2abdc6007683d75-114.pgn), without disrupting those other uses.

Eventually we could adopt fast-chess, which takes care of this internally.

vdbergh:
The conversion should only be done after downloading. If I am reading the code correctly then this PR does it always…

I also think that a valid epd can have six fields, in which case it will be ruined by the conversion.. So the filter should be tightened.

robertnurnberg:
> The conversion should only be done after downloading. If I am reading the code correctly then this PR does it always…

My thinking of doing it also independent of download was existing books in the testing directory. My understanding is that even after a worker update they will be taken as is if found, hence with potentially with the wrong format.

The worst that can happen for a book that already has the right format is that the first line of the file is read, nothing more.

> I also think that a valid epd can have six fields, in which case it will be ruined by the conversion.. So the filter should be tightened.

I wanted to keep it as simple as possible, so currently it only works for books from our repo. I don't think these will ever contain epd op codes. But happy to change the PR if there is some consensus...

vondele:
I would add the check that fields 5 and 6 are integers.

vdbergh:
@robertnurnberg 
> My understanding is that even after a worker update they will be taken as is if found, hence with potentially with the wrong format.

Is that true? It seems to me that after a worker update the workers starts with fresh testing dir and redownloads everything.

Anyway, looking at the code again, the conversion will bail out on the first line of an already converted book, so it is not a big deal that it is done always.

vondele:
The code should also work if the user would have preloaded the book in the worker dir (e.g. for those using an image with this data preloaded), so probably having the check always is needed.

vdbergh:
> The code should also work if the user would have preloaded the book in the worker dir (e.g. for those using an image with this data preloaded), so probably having the check always is needed.

That would be a rather fragile setup since it is not compatible with worker upgrades.

vdbergh:
It is strange that the designers of the EPD format did not make it an extension of the FEN format (just include the key-value pairs at the end of the FEN).

ppigazzini (approved):
Looks good on DEV with epd books.

ppigazzini:
Triggered the worker update, thank you @robertnurnberg :)

PR #1849, opened 2023-11-10
Adjusting the score threshold for draws

FauziAkram:
I suggest adjusting the score threshold for draws from 20 to 16. Here are the reasons for this proposed change:

Reducing Premature Draws: The current threshold of 20 may result in premature draws, especially in complex positions where small imbalances can still lead to decisive outcomes. Increasing the threshold to 16 would allow the game to develop further before considering it a draw.

Encouraging Dynamic Play: A higher threshold promotes more dynamic play, as sides would have to demonstrate a more significant advantage to secure a draw.

Alignment with Modern SF Assessments: SF evaluation has evolved, and now considers scores above 16 as an advantage.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@ David (dav1312)
 David Zar
 dubslow
 Fabian Fichter (ianfab)
+Fauzi Akram (fauzi2)
 Fanael Linithien (Fanael)
 FieryDragonLord
 Giacomo Lorenzetti (G-Lorenz)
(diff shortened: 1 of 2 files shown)


XInTheDark:
> SF evaluation has evolved, and now considers scores above 16 as an advantage

+0.16 is still certainly a draw?

I feel like updating the WDL model would be much more useful than adjusting this small constant here.

Disservin (commented):
I find this change quite "random" especially 
> and now considers scores above 16 as an advantage.

So I dont see any need.

vdbergh:
A change like this should be supported by evidence that it makes Fishtest more efficient. Otherwise it is just random.

Disservin:
@ppigazzini i think we can close this in the meantime. Unless the other maintainers have a different opinion.

PR #1850, opened 2023-12-02
Add initial docker support for fishtest workers

linrock:
Adds code for running fishtest workers as [docker](https://en.wikipedia.org/wiki/Docker_(software)) container instances. This enables:
- easily scaling up and managing large # of workers across many servers
- a cross-platform way of running workers locally for dev and production fishtest servers

Docker support for fishtest servers can later be added to automate setting up dev servers, which is currently a manual process.


docker/README.md
```diff
@@ -0,0 +1,29 @@
+docker containers for fishtest server (TODO) and workers
+
+
+### local worker
+
+Example of starting a single fishtest worker locally
+
+```bash
+NUM_WORKERS=1 WORKER_ARGS="username password" docker compose up --build -d
```
Disservin, line 9: This syntax isnt really os independent. I think it should use https://docs.docker.com/build/guide/build-args/.

Disservin:
@ppigazzini what do you say ? me and linrock shortly discussed this topic on discord 
https://discord.com/channels/435943710472011776/900771437177094156/1180528374335160440.

I think ultimately it would be nice if we migrate away from the currently recommended setup described here: https://github.com/official-stockfish/fishtest/wiki/Running-the-worker-on-Windows#windows-11108-instructions
to a pure docker based solution. 

This would also be a huge step if done for the fishtest server itself, at least in such a way that new contributors can just start a docker container to get a running fishtest server.

ppigazzini:
Missing instructions to send a CTRL+C to the running applications to close in proper way the workers.

docker/worker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:23.10
+
+RUN apt update -y && apt upgrade -y && apt install -y \
```
ppigazzini, line 3: `apt update` does not require the flag `-y`

vdbergh:
I played around with docker some time ago.

https://github.com/vdbergh/fishtest-docker

It still seems to work.

docker/worker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+  git vim python3 python3-distutils make g++ # clang llvm
+
+RUN useradd --create-home --shell /bin/bash worker
+
+WORKDIR /home/worker
+COPY run_workers.sh .
+RUN chmod +x run_workers.sh
+
```
vdbergh, line 11: It seems to me that with this setup it would be possible to start the same image multiple times. Am I right?

vdbergh:
When I was looking at docker I had some trouble to make sure that the same image could not be started multiple times. I solved it by giving the image access to the host's process. I am wondering if this PR does anything about this (I do not know the docker options well enough to judge).

Disservin:
> Missing instructions to send a CTRL+C to the running applications to close in proper way the workers.

Im not exactly sure what you mean by that, currently the image starts the containers in the background. So CTRL+C wont work?

ppigazzini:
> > Missing instructions to send a CTRL+C to the running applications to close in proper way the workers.
> 
> Im not exactly sure what you mean by that, currently the image starts the containers in the background. So CTRL+C wont work?

The worker deals with some signals (eg CTRL+C) to stops gracefully `cutechess-cli`.
Docker should use something like this:
```bash
docker exec -it <container_id> bash -c "kill -SIGINT \$(pidof <process_name>)"
```

Disservin:
If you think that is not enough we can increase it
https://docs.docker.com/compose/compose-file/compose-file-v3/#stop_grace_period
and we can also change the stop signal to SIGINT, if you prefer that over SIGTERM.

vdbergh:
> When I was looking at docker I had some trouble to make sure that the same image could not be started multiple times (the instances would share the same virtual disk). I solved it by giving the image access to the host's process list, so that the normal exclusion mechanism would work. I am wondering if this PR does anything about this (I do not know the docker options well enough to judge).

So is isolation between different instances of the docker image achieved via `docker compose`? I didn't know about `docker compose`.

docker/README.md
linrock: using .env resolves defining env vars independently of os. when using .env, we can skip using build args rather than taking an extra hop: `host env -> build args -> container env`. the docs also suggest to not use build args for credentials:

https://docs.docker.com/engine/reference/builder/#arg

image

linrock:
> So is isolation between different instances of the docker image achieved via docker compose? I didn't know about docker compose.

that's right, as defined in `compose.yml`, `docker compose` ensures the `fishtest-worker` image is only started once. in this PR, a single container instance runs multiple fishtest workers.

Disservin:
> ctrl c on a docker container started with compose internally uses the docker stop command, described here https://docs.docker.com/engine/reference/commandline/stop/
> 
> > The main process inside the container will receive SIGTERM, and after a grace period, SIGKILL. The first signal can be changed with the STOPSIGNAL instruction in the container's Dockerfile, or the --stop-signal option to docker run.
> 
> The worker already catches SIGTERM https://github.com/official-stockfish/fishtest/blob/master/worker/worker.py#L1534
> 
> So I think it already does the correct thing. The grace period is 10s, that should be enough?

Actually I just realized that the fishtest workers are moved into the background… so I don’t think they will get sigterm

vdbergh:
> > So is isolation between different instances of the docker image achieved via docker compose? I didn't know about docker compose.
> 
> that's right, as defined in `compose.yml`, `docker compose` ensures the `fishtest-worker` image is only started once. in this PR, a single container instance runs multiple fishtest workers.

I have another question. If you restart the worker image, will the configfile `fishtest.cfg` be preserved?

This is important since otherwise the worker will get a new uuid prefix.

vdbergh:
I tried to run it but I got this...
```sh
$ docker-compose up --build
ERROR: The Compose file './compose.yml' is invalid because:
services.fishtest-worker.restart contains an invalid type, it should be a string
```

Disservin:
But that’s not the command which is given in the readme ? 
it’s 
docker compose
not
docker-compose

vdbergh:
> But that’s not the command which is given in the readme ? it’s docker compose not docker-compose

`docker compose` did not work before but now it magically started working...

vdbergh:
I tried it and it seems to preserve the UUID which was my main concern.

I am not sure though if ctrl-C stops the workers gracefully. I had a dead task after killing docker.

vdbergh:
I got another dead task
Screenshot from 2023-12-11 15-00-33

So definitely ctrl-C does not kill the worker gracefully.

vdbergh:
It sometimes seems to hang... Not sure what is the cause. 
```
$ sudo docker compose up --build
[+] Building 0.7s (11/11) FINISHED                                                                                                                                                                          
 => [fishtest-worker internal] load build definition from Dockerfile                                                                                                                                   0.0s
 => => transferring dockerfile: 423B                                                                                                                                                                   0.0s
 => [fishtest-worker internal] load .dockerignore                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                        0.0s
 => [fishtest-worker internal] load metadata for docker.io/library/ubuntu:23.10                                                                                                                        0.6s
 => [fishtest-worker 1/6] FROM docker.io/library/ubuntu:23.10@sha256:cbc171ba52575fec0601f01abf6fdec67f8ed227658cacbc10d778ac3b218307                                                                  0.0s
 => [fishtest-worker internal] load build context                                                                                                                                                      0.0s
 => => transferring context: 36B                                                                                                                                                                       0.0s
 => CACHED [fishtest-worker 2/6] RUN apt update && apt upgrade -y && apt install -y   git vim python3 python3-distutils make g++ # clang llvm                                                          0.0s
 => CACHED [fishtest-worker 3/6] RUN useradd --create-home --shell /bin/bash worker                                                                                                                    0.0s
 => CACHED [fishtest-worker 4/6] WORKDIR /home/worker                                                                                                                                                  0.0s
 => CACHED [fishtest-worker 5/6] COPY run_workers.sh .                                                                                                                                                 0.0s
 => CACHED [fishtest-worker 6/6] RUN chmod +x run_workers.sh                                                                                                                                           0.0s
 => [fishtest-worker] exporting to image                                                                                                                                                               0.0s
 => => exporting layers                                                                                                                                                                                0.0s
 => => writing image sha256:0724bad9e7c5da47fc94b0b59d4f0c0d2672fd07e011f534a6cb705a55577b7b                                                                                                           0.0s
 => => naming to docker.io/library/fishtest-worker                                                                                                                                                     0.0s
[+] Running 1/0
 ✔ Container fishtest-worker  Created                                                                                                                                                                  0.0s 
Attaching to fishtest-worker
fishtest-worker  | fatal: destination path '/home/worker/fishtest' already exists and is not an empty directory.
fishtest-worker  | cp: cannot create regular file '/home/worker/worker1/fishtest/.git/objects/pack/pack-ac42752408b8cf43fcac8b1991ba951ca95653e2.pack': Permission denied
fishtest-worker  | cp: cannot create regular file '/home/worker/worker1/fishtest/.git/objects/pack/pack-ac42752408b8cf43fcac8b1991ba951ca95653e2.idx': Permission denied
```

Disservin:
Is the dead task issue fixed when you add
trap 'kill -SIGINT 0' INT
to before the for loop in the script?

ppigazzini:
Using `wget/curl` and `unzip` instead than `git` should help to slim down the alpine image (if that is the goal).

vdbergh:
> Using `wget/curl` and `unzip` instead than `git` should help to slim down the alpine image (if that is the goal).

Yes my own image uses this and it is only 240Mb. But  is just a proof of concept. I am indeed not sure if people are interested in a small image (it can never beat doing `python worker.py` in the console...).

vdbergh:
> Is the dead task issue fixed when you add trap 'kill -SIGINT 0' INT to before the for loop in the script?

This appears not to work (I rebuilt the image using the Dockerfile in the PR). Dead tasks are still being generated.

```
#!/bin/bash

git clone \
  https://github.com/official-stockfish/fishtest \
  ~/fishtest

trap 'kill -SIGINT 0' INT

for worker in $(seq 1 $NUM_WORKERS); do
  worker_dir=~/worker$worker
  cp -r fishtest $worker_dir
  cd $worker_dir/worker
  python3 worker.py $WORKER_ARGS &
  cd ~
done

wait
```

Disservin:
@vdbergh would you be so kind to test this?
```bash
#!/bin/bash

git clone \
  https://github.com/official-stockfish/fishtest \
  ~/fishtest

cleanup_workers() {
  echo "Cleaning up workers..."
  for pid in "${worker_pids[@]}"; do
    kill -s SIGINT "$pid"
  done
  echo "Workers cleaned up."
  exit
}

trap 'cleanup_workers' SIGINT

worker_pids=()

for worker in $(seq 1 $NUM_WORKERS); do
  worker_dir=~/worker$worker
  cp -r fishtest "$worker_dir"
  cd "$worker_dir/worker"
  python3 worker.py $WORKER_ARGS &
  worker_pids+=($!)
  cd ~
done

wait
```

vdbergh:
I tried and it seems not to work. I think the ctrl-C is simply not received by the script. Perhaps the culprit is that it is a shell script?

Even this does not seem to work
```
#!/bin/bash

git clone \
  https://github.com/official-stockfish/fishtest \
  ~/fishtest

cp -r fishtest ~/worker
cd ~/worker/worker
python3 worker.py $WORKER_ARGS 
```

vdbergh:
`docker compose` seems to be partially the culprit. When starting the image with
```
docker run -it --rm fishtest-worker
```
it clearly reacts to ctrl-C. However it doesn't really unwind things as it should.

Disservin:
btw @vdbergh it's normal that there is no more output after ctrl + c if that's how you are checking.
You will need to view the logs `docker logs fishtest-worker`

vdbergh:
> btw @vdbergh it's normal that there is no more output after ctrl + c if that's how you are checking. You will need to view the logs `docker logs fishtest-worker`

I am checking if there are dead tasks.

But what you are saying is not correct anyway. My image finishes cleanly after ctrl-C and shows all the output.

Disservin:
> But what you are saying is not correct anyway.
it only applies to the docker compose case, the run invocation should show you everything

vdbergh:
> > But what you are saying is not correct anyway.#
> 
> it only applies to the docker compose case, the run invocation should show you everything

Yes. I am using `docker run -it` ...

Disservin:
works with docker compose with these changes for me, wont instantly stop the container when the engine is currently being build probably because some signal isnt "correctly" forwarded to the make invocation, like it is for cutechess. The grace period might have to be increased in case in that case to wait for the compilation to finish.

```diff
diff --git a/docker/worker/compose.yml b/docker/worker/compose.yml
index 6ea8047..83bb576 100644
--- a/docker/worker/compose.yml
+++ b/docker/worker/compose.yml
@@ -6,6 +6,8 @@ services:
       args:
         - NUM_WORKERS=$NUM_WORKERS
         - WORKER_ARGS=$WORKER_ARGS
-
+    stop_grace_period: 1m30s
     container_name: fishtest-worker
     image: fishtest-worker
+    stdin_open: true # docker run -i
+    tty: true        # docker run -t
\ No newline at end of file
diff --git a/docker/worker/run_workers.sh b/docker/worker/run_workers.sh
index 9893533..27ee5c7 100644
--- a/docker/worker/run_workers.sh
+++ b/docker/worker/run_workers.sh
@@ -4,12 +4,30 @@ git clone \
   https://github.com/official-stockfish/fishtest \
   ~/fishtest
 
+worker_pids=()
+
+cleanup_workers() {
+  echo "Cleaning up workers..."
+  for pid in "${worker_pids[@]}"; do
+    echo "Killing worker $pid"
+    kill -s SIGINT "$pid"
+    tail --pid=$pid -f /dev/null
+  done
+  echo "Workers cleaned up."
+  exit
+}
+
+trap 'cleanup_workers' SIGTERM
+trap 'cleanup_workers' SIGINT
+
+
 for worker in $(seq 1 $NUM_WORKERS); do
   worker_dir=~/worker$worker
-  cp -r fishtest $worker_dir
-  cd $worker_dir/worker
+  cp -r fishtest "$worker_dir"
+  cd "$worker_dir/worker"
   python3 worker.py $WORKER_ARGS &
+  worker_pids+=($!)
   cd ~
 done
 ```

**Comment by vdbergh** [comment]
Did you check that no dead tasks are generated? Because if they are then it means that python did not correctly receive the signal (otherwise it would have executed a `failed_task()` api call).

**Comment by vdbergh** [comment]
> I checked over the worker output which I'd imagine is enough?
> 
> ```
> Indexing opening suite...
> Cleaning up workers...
> Killing worker 10
> 
> 
> Exception running games:
> Terminated by signal SIGINT
> Informing the server
> Post request https://tests.stockfishchess.org:443/api/failed_task handled in 130.89ms (server: 3.19ms)
> Task exited
> Waiting for the heartbeat thread to finish...
> Heartbeat stopped
> Deleting lock file /home/worker/worker1/worker/worker.lock
> Workers cleaned up.
> ```

Yes that would be enough. But I have not been able to get this. Can you post your script?

**Comment by Disservin** [comment]
> EDIT2: Now it seems to hang after ctlr-C.

hangs where ? what are the logs saying?

not sure if missed this from earlier:

> wont instantly stop the container when the engine is currently being build probably because some signal isnt "correctly" forwarded to the make invocation

**Comment by vdbergh** [comment]
> > EDIT2: Now it seems to hang after ctlr-C.
> 
> hangs where ? what are the logs saying?
> 
> not sure if you missed this from earlier:
> 
> > wont instantly stop the container when the engine is currently being build probably because some signal isnt "correctly" forwarded to the make invocation

After a while it stopped (after > 1 minute). But the worker.py script had not received the signal (I posted the logs and I checked that there is a dead task).

**Comment by vdbergh** [comment]
Maybe I should ask what OS you are using...

**Comment by Disservin** [comment]
Ubuntu 22.04.3 LTS

**Comment by Disservin** [comment]
> After a while it stopped (after > 1 minute). But the worker.py script had not received the signal (I posted the logs and I checked that there is a dead task).

`stop_grace_period: 1m30s` perhaps exactly 90s? This was always happening for me when canceled during a compilation which took longer than 90s.

**Comment by vdbergh** [comment]
> > After a while it stopped (after > 1 minute). But the worker.py script had not received the signal (I posted the logs and I checked that there is a dead task).
> 
> `stop_grace_period: 1m30s` perhaps exactly 90s? This was always happening for me when canceled during a compilation which took longer than 90s.

Sadly your patch does not work at all on my system (I just let it run till after building the engines, but the effect is still the same) ... Is their a way to force it to rebuild everything from zero?

**Comment by vdbergh** [comment]
Just to be sure I uninstalled the docker packages from the Ubuntu repositories and installed the official docker packages from the docker website.
```
docker compose up 
```
still does not handle ctrl-C (dead tasks are created) with the @Disservin patch. However 
```
docker run -it --rm fishtest-worker
```
works perfectly now (without needing any grace timeout).

Sadly with `docker run` each time you start the image, the workers will get a new UUID which is not nice (it is important to be able to track ill behaving workers across restarts). 

This could be solved by giving the workers a deterministic UUID prefix in the script (with the -u option). This would mean that one cannot start multiple copies of the image as the server will reject workers with the same UUID prefix, but this is not the intention anyway.

**Comment by vdbergh** [comment]
I found an improvement

By replacing the last line of the Dockerfile with
```
CMD ["./run_workers.sh"]
```
it is possible to kill the workers gracefully using ctrl-C, except when compiling the engines
(according to the web the [...]  avoid creating an extra shell process which eats the `ctrl-C`)

So then it is still a mystery why it does not work during the compile. Perhaps with `docker compose up` signals are not forwarded to subprocesses?

**Comment by linrock** [comment]
> CMD ["./run_workers.sh"]

nice find, i've updated the PR. i believe using an array here is equivalent to `exec ./run_workers.sh` as opposed to starting a new shell to run the script

File: docker/README.md (hunk 1 above)
Comment (linrock) [reply docker/README.md]: opting to use env vars directly instead of intermediate build args, since this avoids having to rebuild the image every time .env changes. with env vars, the container can be restarted to pick up changes. also addresses the big red warning from docker docs

**Comment by vdbergh** [comment]
> > CMD ["./run_workers.sh"]
> 
> nice find, i've updated the PR. i believe using an array here is equivalent to `exec ./run_workers.sh` as opposed to starting a new shell to run the script

This by itself seems not enough to make the worker quit gracefully. One needs Disservin's patch posted above.

**Comment by vdbergh** [comment]
Perhaps the reason that Disservin's patch cannot kill the make command, is that the make command is started in a shell (in contrast to the other subprocesses such as `cutechess`).

Thinking about a convenient way to test this theory.

**Comment by linrock** [comment]
applied Disservin's patches and confirmed ctrl-c exits `docker compose up` gracefully when the worker is playing games

**Comment by vdbergh** [comment]
I found another crucial difference between the subprocesses `make` and `cutechess-cli`. `cutechess-cli` is started with `stdout=PIPE`. This makes `Popen` non-blocking and the `Python` process keeps control so that it can kill `cutechess-cli` when necessary. 

In contrast the invocation of `make` is blocking so `Python` loses control until the `make` process finishes.

So it seems that for this PR to work well the worker needs to be adapted to handle long running processes better.

**Comment by vdbergh** [comment]
YES!!! This turned out to be the problem. I can now make the worker quit cleanly during make. I will post a patch soon.

**Comment by vdbergh** [comment]
See  #1861

One issue with `docker compose` is that it is not a drop in replacement for `python worker.py`. So extra testing is required to make sure that changes to the worker work well with `docker compose`.

**Comment by vdbergh** [comment]
Of course all problems would be solved if `docker compose` had a `-t` option (run processes in a terminal). Does anyone know if an equivalent exists?

**Comment by Disservin** [comment]
> Of course all problems would be solved if `docker compose` had a `-t` option (run processes in a terminal). Does anyone know if an equivalent exists?

My docker file had that..

```diff
+    stdin_open: true # docker run -i
+    tty: true        # docker run -t
```

vdbergh:
Yes this is now part of the PR but it is not sufficient for some reason.

vdbergh:
It appears
```
docker run -it fishtest-worker
```
does not work anymore. In principle this is a good thing since users should not user `docker run` as it generates a new UUID prefix at every invocation. So the server loses the ability to track the worker,

vdbergh:
I found a way of dealing with the `ctrl-C` problem, independently of #1861 

Replace `run_workers.sh` by
```sh
#!/bin/bash

git clone \
  https://github.com/official-stockfish/fishtest \
  ~/fishtest

cleanup_workers() {
    kill -INT -- -1
    wait
    exit
}

trap 'cleanup_workers' SIGTERM
trap 'cleanup_workers' SIGINT



for worker in $(seq 1 $NUM_WORKERS); do
  worker_dir=~/worker$worker
  cp -r fishtest "$worker_dir"
  cd "$worker_dir/worker"
  python3 worker.py $WORKER_ARGS &
  cd ~
done

wait
```
Explanation
==========

In contrast to starting the worker in the shell, docker apparently makes the workers children of `init` (`PID=1`). The init process has process group id equal to  1 (`PGID=1`). So we send an `INT` signal to the init process. This is accomplished by 
```sh
kill -INT -- -1
```
A negative argument to `kill` means a `PGID` instead of a `PID`. By convention in some shell commands (like `kill`) `--` separates the arguments from the options.

In order to catch the output of the child processes we issue a `wait` command after the `kill` command. This seems to be necessary.

vdbergh:
There is one remaining issue that needs to be solved I guess. If the container is rebuilt then a new UUID is generated. Rebuilding the container seems to happen inadvertently if one makes small configuration changes (like changing the number of workers).

peregrineshahin:
> There is one remaining issue that needs to be solved I guess. If the container is rebuilt then a new UUID is generated. Rebuilding the container seems to happen inadvertently if one makes small configuration changes (like changing the number of workers).

Hmm, that's unideal..

vdbergh:
The problem is that the algorithm for auto generating the UUID(prefix) uses some initially randomly generated state which is stored in `fishtest.cfg`.

Including some state seems unavoidable. Without it all noob's workers get the same UUID since they are launched from completely identical virtual machines. This state is destroyed if the container is rebuilt from scratch.

Of course the server will reject duplicate UUIDs, but I guess noob would not find it funny if he would have to specify himself (via the `-u` option) a different UUID for each of his workers (Technologov does this though).

In my docker image I solved this by using a virtual disk. This virtual disk is preserved, even if the container is rebuilt.

vdbergh:
One solution would be for the script that builds the container, to generate the UUID's. It could use some semi-stable information such as the machine-id of the host (assuming the host is a real machine...).

vdbergh:
The solution proposed above seems to be a good one. 

I assume we can assume that the host has Python installed (as Docker is a much heavier dependency than Python). So we can just recycle some of the worker code to autogenerate stable UUID's on the host while building the container.

vdbergh:
But now I have the impression that one can not run scripts on the host in the docker file. Not clear to me what to do about this.

vdbergh:
`docker info | grep "ID"` generates a sort of machine id.  Still not clear how we can use this.

Disservin:
Mh so I think there should be a way to change the config while avoiding the rebuild, using volumes. Additionally I talked with ppigazzini a bit about this entire topic. He proposed a second repository instead which holds the docker containers, but I think this will only make it more difficult since one would have to clone to repos and write weird reference to link to the worker code from the fishtest dir. 
Also this docker container approach is kinda abusing docker since it starts the workers multiple times inside the same container instead of having a container for each. 

The question is also what we want to achieve. Do we want to create a docker container which users can use in the future or a container which developers can use while working on the worker python code (which then would require volumes) or perhaps both. 

The entire CTRL C topic would also probably be solved if the container is used to only run one instance of the worker and the cfg file could simply be shared between the host and container. 

Ultimately imo the nice thing to have would be a complete docker setup for the server IMO ppigazzini mentioned that this would require 6 different containers to setup correctly (3 python fishtest servers, 1 net server, 1 nginx, 1 mongodb) and then it would still require some work to enable some easy way to reload the services when code has changed.

vdbergh:
> Mh so I think there should be a way to change the config while avoiding the rebuild, using volumes. Additionally I talked with ppigazzini a bit about this entire topic. He proposed a second repository instead which holds the docker containers, but I think this will only make it more difficult since one would have to clone to repos and write weird reference to link to the worker code from the fishtest dir. Also this docker container approach is kinda abusing docker since it starts the workers multiple times inside the same container instead of having a container for each.
> 
> The question is also what we want to achieve. Do we want to create a docker container which users can use in the future or a container which developers can use while working on the worker python code (which then would require volumes) or perhaps both.

IMHO this would definitely not be for developers. Working on the worker directly using git and an editor is really trivial (given that you can instantly see the effect of any changes). 

The ideal use case would be a scenario for users in which they can install a docker image on their daily OS and it just works. Alas I am not sure that it is possible to create something with docker that it is easier to use than doing `python worker.py` . Certainly not on Linux. Perhaps on Windows. 
> 
> The entire CTRL C topic would also probably be solved if the container is used to only run one instance of the worker and the cfg file could simply be shared between the host and container.
> 
> Ultimately imo the nice thing to have would be a complete docker setup for the server IMO ppigazzini mentioned that this would require 6 different containers to setup correctly (3 python fishtest servers, 1 net server, 1 nginx, 1 mongodb) and then it would still require some work to enable some easy way to reload the services when code has changed.

Yes I think having a Fishtest instance inside a docker container might offer real benefits for developers.

I am currently using an AWS instance for Fishtest. It might conceivably be replaced by a docker container.

vdbergh:
> > IMHO this would definitely not be for developers. Working on the worker [and server] directly using git and an editor is really trivial (given that you can instantly see the effect of any changes).
> 
> This is true on steroids for developing/testing/debugging the server code, that can require to change: mongodb, nginx config, python code (without checking what of the 3 python instances is affected by the change), start/stopping crond, the logs of the different services/applications etc.

As I see it, the docker container would replace the installation script (as you say there are many ingredients to this). Instead of running this script one just fires up the docker container which already contains everything needed.

The docker container would give a shell so that developers can then start hacking the Fishtest server code (mainly python, mako and javascript).

ppigazzini:
I fear to have not be clear. Docker makes more harder to develop test and debug for a skilled devop and nearly impossible for a normal developer that doesn't know Docker.

DEV and PROD have a more complex setup script that compiles a custom nginx (and a couple of nginx modules) and a script to build the updated CDIR used to set the worker flag from the IP. IMO it was very over killer for the Wiki setup script.

linrock:
> Rebuilding the container seems to happen inadvertently if one makes small configuration changes (like changing the number of workers).

@vdbergh does your `compose.yml` and `Dockerfile` match the current state of this PR? rebuilds due to config changes should've only been an issue with build args earlier in the PR.

after switching to env vars, i'm able to make config changes in `.env` and restart workers with `docker compose up` without triggering rebuilds

vdbergh:
For me this does not work

If the configuration changes new UUID's are generated. If not then the same UUID's are used.

I am using 
```
docker compose up --build
```
because that's what the readme says. Should I drop the `--build`?

Disservin:
Probably so, yes.

vondele:
chiming in a bit late, but for me the advantage of a docker container for the worker would be mostly simplicity for the user (i.e. environment known to work), but particularly added security, i.e. some level of protection against rogue patches being tested.

linrock:
i see, switching to env vars avoids rebuilding the entire image (apt update, etc.) when changes are made to `.env`. however the local state within the container does get reset upon making config changes, which changes the uuids.

> In my docker image I solved this by using a virtual disk. This virtual disk is preserved, even if the container is rebuilt.

> Mh so I think there should be a way to change the config while avoiding the rebuild, using volumes

yea i'd say either using volumes or bind-mounting a local dir inside the container is the way to go, if we need to persist container states (ie. `fishtest.cfg`) independently of image rebuilds.

linrock:
> particularly added security, i.e. some level of protection against rogue patches being tested.

definitely much safer to run arbitrary code as unpermissioned users within containers, rather than directly on host systems

vdbergh:
> i see, switching to env vars avoids rebuilding the entire image (apt update, etc.) when changes are made to `.env`. however the local state within the container does get reset upon making config changes, which changes the uuids.
> 
> > In my docker image I solved this by using a virtual disk. This virtual disk is preserved, even if the container is rebuilt.
> 
> > Mh so I think there should be a way to change the config while avoiding the rebuild, using volumes
> 
> yea i'd say either using volumes or bind-mounting a local dir inside the container is the way to go, if we need to persist container states (ie. `fishtest.cfg`) independently of image rebuilds.

I was proposing that the UUID would be generated during build from stable information on the host (e.g. the `machine-id`) and then passed to the worker as `-u` argument. However I do not see how to achieve this using docker.

linrock:
> I was proposing that the UUID would be generated during build from stable information on the host (e.g. the machine-id) and then passed to the worker as -u argument. However I do not see how to achieve this using docker.

maybe a seed for UUID generation could be added to `.env`? this way, container states won't need to persist to keep the same UUIDs

vdbergh:
> > I was proposing that the UUID would be generated during build from stable information on the host (e.g. the machine-id) and then passed to the worker as -u argument. However I do not see how to achieve this using docker.
> 
> maybe a seed for UUID generation could be added to `.env`? this way, container states won't need to persist to keep the same UUIDs

I considered this, but I thought one cannot call a script in .env (it seemed to strictly key/value to me).

linrock:
yea it's strictly key/value. can it be some constant number chosen by the user?

vdbergh:
> yea it's strictly key/value. can it be some constant number chosen by the user?

Yes. But the server will reject duplicates (from the same Fishtest user with the same number of cores). So if the idea is to run the image effortlessly on many machines then this is an issue.

linrock:
hmm i guess i don't understand why it's important to keep the same uuid across worker rebuilds across many machines.

either way, different seeds across different machines is doable either with runtime env vars or build args. something like this:

```bash
SERVERS=(
  server1
  server2
)
for server in ${SERVERS[@]}; do
  docker -H ssh://$server compose up --build -d -e SEED=$server
done
```

ppigazzini:
> chiming in a bit late, but for me the advantage of a docker container for the worker would be mostly simplicity for the user (i.e. environment known to work), but particularly added security, i.e. some level of protection against rogue patches being tested.

The way as run the worker does not belong to this repository, every user has his operating system, his use cases and his preferences.

vdbergh:
> hmm i guess i don't understand why it's important to keep the same uuid across worker rebuilds across many machines.

It is important to be able to track workers across restarts. Since it became possible to block ill behaving workers based on their UUID, the server became a much more friendly place.

So I feel that if running the worker in docker became an officially supported option then  preservation of the UUID across restarts (on a fixed machine) should be enforced to some extent. 

> 
> either way, different seeds across different machines is doable either with runtime env vars or build args. something like this:
> 
> ```shell
> SERVERS=(
>   server1
>   server2
> )
> for server in ${SERVERS[@]}; do
>   docker -H ssh://$server compose up --build -d -e SEED=$server
> done
> ```

This looks reasonable.

Disservin:
> > chiming in a bit late, but for me the advantage of a docker container for the worker would be mostly simplicity for the user (i.e. environment known to work), but particularly added security, i.e. some level of protection against rogue patches being tested.
> 
> The way as run the worker does not belong to this repository, every user has his operating system, his use cases and his preferences.

Where would server docker files then belong ?

ppigazzini:
PROD is running in a system container, DEV is running in a VM.
So docker for the server does not belong to the repo.

vdbergh:
I don't feel strongly about this, but of course nobody is suggesting running docker on production servers.

The idea would be that somebody who has an idea for a patch for the Fishtest server code (python+mako+js+html+...) can just use a docker image to test it, instead of having to install Fishtest themselves.

vdbergh:
@linrock Perhaps you may want to test https://github.com/official-stockfish/fishtest/pull/1850#issuecomment-1857339909 . It seems to completely solve the ctrl-C problem.

linrock:
> @linrock Perhaps you may want to test https://github.com/official-stockfish/fishtest/pull/1850#issuecomment-1857339909 . It seems to completely solve the ctrl-C problem.

with your updated `run_workers.sh`, I tried ctrl-c a few when running games, compiling, etc. and no dead tasks. if that's a sign of solving the ctrl-c problem, it appears solved

vdbergh:
> > @linrock Perhaps you may want to test [#1850 (comment)](https://github.com/official-stockfish/fishtest/pull/1850#issuecomment-1857339909) . It seems to completely solve the ctrl-C problem.
> 
> with your updated `run_workers.sh`, I tried ctrl-c a few when running games, compiling, etc. and no dead tasks. if that's a sign of solving the ctrl-c problem, it appears solved

I am sorry for coming back to this. If #1864 is committed then [#1850 (comment)](https://github.com/official-stockfish/fishtest/pull/1850#issuecomment-1857339909) will no longer work but Disservin's patch will work optimally.

The reason is that #1864 puts the worker in its own process group and so sending `INT` to `PGID=1` will have no effect. But the worker will forward an `INT` or `TERM` signal it receives to the whole process group. So that makes Disservin's patch work.

ppigazzini:
I am glad you are back, pointing out that a small change in fishtest code will break the simple worker docker. The application container is supposed to be immutable, which guarantees reproducible behavior. Updating a running container is a well know anti-pattern.

Disservin:
https://github.com/official-stockfish/docker-fishtest

PR #1853, opened 2023-12-03
Remove some dead code in tests_modify (views.py).

vdbergh:
This should fix https://github.com/official-stockfish/fishtest/pull/1843#issuecomment-1837540039


ppigazzini (approved):
Looks good on DEV: logs clean after modifying some tests.

ppigazzini:

PR #1854, opened 2023-12-06
Tighten up the schema in api.py.

vdbergh:
The api's in api.py are a direct connection between the workers and the db. So it makes sense to be very strict in what we accept.

Note: a serious bug was discovered in the handling of optional keys in vtjson. So vtjson should be upgraded to the latest version. Since upgrading is necessary anyway, this PR depends on a previously internal feature of vtjson, namely the possibility to pre-compile a schema.

Validating an update_task api call for an SPRT run takes 0.026ms on an AWS entry level t2.micro using Python 3.11 (without pre-compiling it takes 0.068ms).

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -14,7 +14,7 @@ from pyramid.httpexceptions import (
 )
 from pyramid.response import Response
 from pyramid.view import exception_view_config, view_config, view_defaults
-from vtjson import _validate, lax, union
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):
Looks good on DEV and some workers

ppigazzini:
PROD rebuilt and updated, thank you @vdbergh :)

PR #1856, opened 2023-12-09
Fix inconsistent <td> widths.

peregrineshahin:
in user management tables.


ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

PR #1857, opened 2023-12-09
Fix width also when table has not data

peregrineshahin:
follow up to previous commit.


ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

PR #1858, opened 2023-12-10
Generate a "pending" accept_user event on signup.

vdbergh:


vdbergh:
Hmm maybe not a good idea. People could register with an offensive username which would then be visible for everyone in the event log.

vdbergh:
Since I don't see a good solution to this issue I am closing this PR for now.

vdbergh:
It seems that the current system already has this problem. If someone registers with an offensive username then, if I understand correctly, the only way to keep that name out of the event log is to keep the user pending forever, which is not desirable.

I am reopening this to have some debate.

peregrineshahin:
That's right. I will make two buttons instead of the checkboxes, one is with Accept and one is with Reject
Not doing anything will keep the user pending
Accept will accept the user
Reject will remove them from the database.

vdbergh:
Closing this again in view of #1859

PR #1860, opened 2023-12-10
Make blocking and unblocking one button instead of a checkbox

peregrineshahin:
makes it the same button for block and unblock, 1 click is better than 2.

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -114,25 +114,22 @@
     % else:
       <%
         blocked = user['blocked'] if 'blocked' in user else False
-        blocked_checked = 'checked' if blocked else ''
       %>
-      <div class="mb-3 form-check">
-        <label class="form-check-label" for="blocked">Blocked</label>
-        <input
-          type="hidden"
-          value="0"
+      % if blocked:
+        <button
+          class="btn btn-primary w-100"
           name="blocked"
-        >
-        <input
-          type="checkbox"
-          class="form-check-input"
-          id="blocked"
+          value="0"
+          type="submit"
+        >Unblock</button>
+      % else:
+        <button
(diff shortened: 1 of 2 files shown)


peregrineshahin:
Dev is crashing, probably does not have the last push.

ppigazzini:
DEV updated.

peregrineshahin:
@vdbergh can you try it again?

ppigazzini (approved):
Looks good on DEV:
- accepted a new user
- rejected a new user
- blocked an user
- unblocked an user

vdbergh:
It seems to work.

I find it is a bit disconcerting that each time you do something in user management you are sent back to the main test page.

ppigazzini:

vdbergh:
It is possible to make very long usernames 

Screenshot from 2023-12-10 14-46-07

I tried to make even longer ones but this seemed to fail silently, which is also not nice.

peregrineshahin:
Well as you said, this particular user with abnormal long username will be rejected and thus will not be shown in the events log

vdbergh:
Well I posted a screenshot from the event log.

peregrineshahin:
> Well I posted a screenshot from the event log.

Yes that's because it was accepted, then blocked.
but normally the approver should reject him

vdbergh:
That was what I was going to say. But it is still not nice I think. There should be some reasonable limit on the length of a username.

peregrineshahin:
same goes for offensive usernames

peregrineshahin:
Yeah I think limiting username length is a good thing to have

vdbergh:
Offensive usernames are different since they cannot be rejected automatically. But usernames that are too long can.

peregrineshahin:
Agreed! my message was delayed a bit :)

PR #1861, opened 2023-12-13
Shut down the worker quickly after receiving SIGINT during make.

vdbergh:
To test this PR (on Linux, see below) one may run the worker in one terminal and during compiling send a `SIGINT` or `SIGTERM` to the python process from another terminal (e.g. via `kill <PID-python>`).  A worker running `master` will only stop when the `make` command finishes. If the worker runs this PR then it will stop immediately and clean up. 

I assume this PR will be useful if the worker is started via a service manager like `systemd` which typically stops a service by sending a `SIGTERM` to the main process. I have not tested this however.

This PR currently has no effect on `Windows` as the mechanism for terminating processes is different. Probably there is also no need for this PR on that OS.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -699,11 +699,21 @@ def setup_engine(
             cmd,
             shell=True,
             env=env,
+            stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             universal_newlines=True,
             bufsize=1,
             close_fds=not IS_WINDOWS,
         ) as p:
+            try:
+                for out in p.stdout:
+                    print(out.strip())
+            except Exception as e:
+                p.send_signal(signal.SIGINT)
+                raise FatalException(
+                    f"Executing {cmd} raised Exception: {e.__class__.__name__}: {str(e)}",
+                    e=e,
+                )
             errors = p.stderr.readlines()
         if p.returncode:
             raise WorkerException("Executing {} failed. Error: {}".format(cmd, errors))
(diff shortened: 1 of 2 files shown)


Disservin:
Btw I think in the long term we should change all process spawning to behave in the same way using a dedicated interface

vdbergh:
> Btw I think in the long term we should change all process spawning to behave in the same way using a dedicated interface

Yep! PR's welcome of course!

vdbergh:
Meanwhile a solution was found for the `ctrl-C` problem in docker (see the discussion in #1850). But I think this PR is still very useful since it also deals with other cases where the worker is started in a script.

ppigazzini:
On windows CTRL+C works fine during a build process for master (`Terminated by signal SIGINT`) but it's broken for the PR `Exception PermissionError at games.py:751 WorkerVersion: 224 ([WinError 32] The process cannot access the file because it is being used by another process: ... worker/tmp17eh296u\\ppigazzini-Stockfish-4d7f9c9\\src`

image

vdbergh:
:frowning_face: Not sure what that message even means. I guess I'll just delete the windows specific stuff.

ppigazzini:
master without `shell=true` stops correctly on Windows with CTRL+C during the build process.

In the past I tried to drop these two `shell=true` but that broken the build process (perhaps something in the Makefile, I don't recall right now). So this PR needs some test with old versions of SF as well.

vdbergh:
I removed the windows specific code. 

I assume that `shell=True/False` is actually not relevant for this application. I will see if it still works with `shell=True`.

vdbergh:
I tested it in two terminals (launching in one and killing in the other) and it still works with `shell=True`. So I changed it back (in a separate commit).

vdbergh:
In order for this PR not to get embroiled in the great docker wars I changed to a neutral title and description.

vdbergh:
#1864 is a more generic solution.

PR #1862, opened 2023-12-14
Add multiple nets support to the worker

ppigazzini:
Required by the new SF architecture with a Big + Small nets
https://github.com/official-stockfish/Stockfish/pull/4915

Raise the worker version to 224 (also server side)

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -28,7 +28,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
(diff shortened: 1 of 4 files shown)


peregrineshahin:
We probably also need a server PR
That is new net flags and base nets flags
And making sure the new small net is flaged as master and as tested

ppigazzini:
Yes, I'm doing the worker one to fix asap the crashes of some fleets.

mstembera:
@ppigazzini  @peregrineshahin  I just want to say Thank you!

PR #1864, opened 2023-12-16
Shut down the worker quickly after receiving SIGINT.

vdbergh:
To test this PR (on Linux, see below) one may run the worker in one terminal and during compiling send a SIGINT or SIGTERM to the python process from another terminal (e.g. via kill <PID-python>). A worker running master will only stop when the make command finishes. If the worker runs this PR then it will stop immediately and clean up.

I assume this PR will be useful if the worker is started via a service manager like systemd which typically stops a service by sending a SIGTERM to the main process. I have not tested this however.

This PR currently has no effect on Windows as the mechanism for terminating processes is different. Probably there is also no need for this PR on that OS.

PR #1865, opened 2023-12-16
Shut down the worker quickly after receiving SIGINT.

vdbergh:
To test this PR (on Linux, see below) one may run the worker in one terminal and during compiling send a SIGINT or SIGTERM to the python process from another terminal (e.g. via kill <PID-python>). A worker running master will only stop when the make command finishes. If the worker runs this PR then it will stop immediately and clean up.

This PR currently has no effect on Windows as the mechanism for terminating processes is different. Probably there is also no need for this PR on that OS.

This PR is less general than #1864 but also much more trivial.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -28,7 +28,7 @@ db directly. However this information may be slightly outdated, depending
 on how frequently the main instance flushes its run cache.
 """
 
-WORKER_VERSION = 224
+WORKER_VERSION = 225
 
 """
 begin api_schema
--- a/worker/games.py
+++ b/worker/games.py
@@ -703,6 +703,7 @@ def setup_engine(
             cmd,
             shell=True,
             env=env,
+            start_new_session=False if IS_WINDOWS else True,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             universal_newlines=True,
@@ -714,7 +715,7 @@ def setup_engine(
                     print(out.strip())
             except Exception as e:
                 if not IS_WINDOWS:
-                    p.send_signal(signal.SIGINT)
+                    os.killpg(p.pid, signal.SIGINT)
                 raise WorkerException(
                     f"Executing {cmd} raised Exception: {e.__class__.__name__}: {str(e)}",
                     e=e,
--- a/worker/games.py
+++ b/worker/games.py
@@ -309,6 +309,8 @@ def establish_validated_net(remote, testing_dir, net):
                         "Failed to validate the network: {}".format(net)
                     )
                 break
+            except FatalException:
+                raise
             except WorkerException:
                 if attempt > 5:
                     raise
@@ -437,12 +439,14 @@ def download_from_github(
 ):
     try:
         blob = download_from_github_raw(item, owner=owner, repo=repo, branch=branch)
-    except:
-        print("Downloading {} failed. Trying the GitHub api.".format(item))
+    except FatalException:
+        raise
+    except Exception as e:
+        print(f"Downloading {item} failed: {str(e)}. Trying the GitHub api.")
         try:
             blob = download_from_github_api(item, owner=owner, repo=repo, branch=branch)
-        except:
-            raise WorkerException("Unable to download {}".format(item))
+        except Exception as e:
+            raise WorkerException(f"Unable to download {item}", e=e)
     return blob
 
 
@@ -699,11 +703,22 @@ def setup_engine(
             cmd,
             shell=True,
             env=env,
+            stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             universal_newlines=True,
             bufsize=1,
             close_fds=not IS_WINDOWS,
         ) as p:
+            try:
+                for out in p.stdout:
+                    print(out.strip())
+            except Exception as e:
+                if not IS_WINDOWS:
+                    p.send_signal(signal.SIGINT)
+                raise WorkerException(
+                    f"Executing {cmd} raised Exception: {e.__class__.__name__}: {str(e)}",
+                    e=e,
+                )
             errors = p.stderr.readlines()
         if p.returncode:
             raise WorkerException("Executing {} failed. Error: {}".format(cmd, errors))
--- a/worker/sri.txt
+++ b/worker/sri.txt
@@ -1,1 +1,1 @@
-{"__version": 224, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "QggkAE0oS9YfF29AZS+CoHuv5Fm0iGrHsEcS8qw6+GMrbrFfQywKIvAZXCWKSX6U", "games.py": "bjiztHma22FxtqmAFeSShhOXo8qGeKgAIuao9dp63+ypHVrvsVBGNsckAWS0Glde"}
+{"__version": 224, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "QggkAE0oS9YfF29AZS+CoHuv5Fm0iGrHsEcS8qw6+GMrbrFfQywKIvAZXCWKSX6U", "games.py": "7szyHwmVhZxkiiEjEOXjCrYnys7P5aAUBef5HFa2yh8FSBXggHGKAS2G0+Qq8KsP"}
--- a/worker/sri.txt
+++ b/worker/sri.txt
@@ -1,1 +1,1 @@
-{"__version": 224, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "QggkAE0oS9YfF29AZS+CoHuv5Fm0iGrHsEcS8qw6+GMrbrFfQywKIvAZXCWKSX6U", "games.py": "m8sYwhGaxwIMl2ShvZd802KMKu+0qqpKCQmXkDHazzvkE59Y9Un1uhhGUHNnWN1u"}
+{"__version": 224, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "QggkAE0oS9YfF29AZS+CoHuv5Fm0iGrHsEcS8qw6+GMrbrFfQywKIvAZXCWKSX6U", "games.py": "bjiztHma22FxtqmAFeSShhOXo8qGeKgAIuao9dp63+ypHVrvsVBGNsckAWS0Glde"}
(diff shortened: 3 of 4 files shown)


ppigazzini:
- msys2:
```
x86_64-w64-mingw32-c++ -march=native -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-generate=profdir -pedantic -Wextra -Wshadow -Wmissing-declarations -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=0f0fdc59 -DGIT_DATE=20231216 -DARCH=x86-64-sse41-popcnt -flto -flto-partition=one   -c -o timeman.o timeman.cpp

Exception running games:
Exception ValueError at games.py:714 WorkerVersion: 224 (Unsupported signal: 2)
Informing the server
Heartbeat stopped
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 494.56ms (server: 5.92ms)
Task exited
Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder01/fishtest/worker/worker.lock
```

- WSL2 cli
```
clang++ -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-instr-generate  -pedantic -Wextra -Wshadow -Wmissing-prototypes -Wconditional-uninitialized -m64 -DUSE_PTHREADS -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=7cec72bb -DGIT_DATE=20231216 -DARCH=x86-64-sse41-popcnt -flto=full   -c -o main.o main.cpp
^C
Exception running games:
Terminated by signal SIGINT
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 122.58ms (server: 3.59ms)
Task exited
Heartbeat stopped
```

- WSL2 nohup + tail + kill
```
clang++ -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-instr-generate  -pedantic -Wextra -Wshadow -Wmissing-prototypes -Wconditional-uninitialized -m64 -DUSE_PTHREADS -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=7cec72bb -DGIT_DATE=20231216 -DARCH=x86-64-sse41-popcnt -flto=full   -c -o position.o position.cpp

Exception running games:
Terminated by signal SIGTERM
Informing the server
Heartbeat stopped
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 141.18ms (server: 3.82ms)
Task exited
```

ppigazzini:
- master on msys2 CRTL+C works fine
```
x86_64-w64-mingw32-c++ -march=native -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-generate=profdir -pedantic -Wextra -Wshadow -Wmissing-declarations -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=18fe8acd -DGIT_DATE=20231216 -DARCH=x86-64-sse41-popcnt -flto -flto-partition=one   -c -o movegen.o movegen.cpp

Exception running games:
Terminated by signal SIGINT
Informing the server
Heartbeat stopped
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 554.87ms (server: 3.78ms)
Task exited
Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder01/fishtest/worker/worker.lock 
```
image

vdbergh:
Latest commit should fix the error on msys2.

ppigazzini:
msys2 fixed
```
x86_64-w64-mingw32-c++ -march=native -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-generate=profdir -pedantic -Wextra -Wshadow -Wmissing-declarations -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=f5fd19a6 -DGIT_DATE=20231216 -DARCH=x86-64-sse41-popcnt -flto -flto-partition=one   -c -o position.o position.cpp

Exception running games:
Terminated by signal SIGINT
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 515.87ms (server: 4.87ms)
Task exited
Heartbeat stopped
Deleting lock file C:/tools/msys64/home/usr00/_git/__test_folder01/fishtest/worker/worker.lock
```

vdbergh:
In the latest push I added two lines of code to catch a `FatalException` (aka `ctrl-C`) in `establish_validated_net`. Otherwise the worker will hang for a long time if an INT/TERM signal is received while downloading the net. This is quite difficult to trigger on real hardware since downloading the net takes only a very short time. However it is different for the docker image which seems to have very slow networking (maybe it uses `slirp` on the host?).

I think this fixes the last place where the docker image will hang after receiving an INT-signal.

vdbergh:
The general principle is that when handling an exception, one should always make sure that the case of a `FatalException` is handled correctly. The  `WorkerException` class takes care of this for itself to some extent. If you do
```python
except Exception as e:
    raise WorkerException("message", e=e)
```
then if `e` is a `FatalException` it will be reraised (this is done using some `__new__` trickery in the `WorkerException` class).

vdbergh:
I added a commit with the following description

Start the make command in its own session on posix.
    
If make receives `SIGINT` then it does not kill its children but waits for them to finish. This is particularly visible during linking which takes quite a bit of time to finish.
    
By putting make into its own session and sending the signal to the corresponding process group, `make` will finish immediately, even during linking.
    
I tested that `ctrl-C` in the terminal still works, even if the worker is started from a script.
```

**Comment by ppigazzini** [comment]
> However it is different for docker containers which seem to have very slow networking (maybe they use `slirp` on the host?).

I got the same download time in my systems and 3 different container manager (docker, podman, systemd-container, all running in a VM) 

```
# system
time curl -jsSLO https://tests.stockfishchess.org/api/nn/nn-0000000000a0.nnue

real    0m5.615s
user    0m1.030s
sys     0m1.984s

# official install from docker
sudo docker buildx build --load -t time_curl_net - << EOF
FROM ubuntu:latest
RUN apt update && apt install -y curl time
CMD ["time", "curl", "-jsSLO", "https://tests.stockfishchess.org/api/nn/nn-0000000000a0.nnue"]
EOF

sudo docker run --rm time_curl_net

0.51user 0.52system 0:05.29elapsed 19%CPU (0avgtext+0avgdata 14520maxresident)k
0inputs+125960outputs (0major+1253minor)pagefaults 0swaps

# podman
sudo apt update && sudo apt install -y podman-docker
sudo docker buildx build --load -t time_curl_net - << EOF
FROM ubuntu:latest
RUN apt update && apt install -y curl time
CMD ["time", "curl", "-jsSLO", "https://tests.stockfishchess.org/api/nn/nn-0000000000a0.nnue"]
EOF

sudo docker run --rm time_curl_net

0.72user 0.30system 0:05.33elapsed 19%CPU (0avgtext+0avgdata 14232maxresident)k
10240inputs+125960outputs (88major+1196minor)pagefaults 0swaps

# systemd-container
sudo apt update && apt install -y debootstrap systemd-container
sudo debootstrap --arch=amd64 --variant=minbase jammy jammy-minbase
sudo systemd-nspawn -D jammy-minbase --machine jammy-000 /bin/sh -c 'apt install -y time curl'
sudo systemd-nspawn -D jammy-minbase --machine jammy-000 /bin/sh -c 'time curl -sSjLO https://tests.stockfishchess.org/api/nn/nn-0000000000a0.nnue'

Spawning container jammy-000 on /home/vmuser/jammy-minbase.
Press ^] three times within 1s to kill container.
0.62user 0.25system 0:05.64elapsed 15%CPU (0avgtext+0avgdata 14192maxresident)k
8928inputs+125960outputs (78major+1159minor)pagefaults 0swaps
Container jammy-000 exited successfully.
```

**Comment by vdbergh** [comment]
For me downloading the net for the worker takes about 4 seconds (sometimes more). This is after a fresh build of the image.

**Comment by vdbergh** [comment]
So you are right. Docker is not to blame for the slowness. But the issue is harder to trigger when running the worker on real hardware. In any case `FatalException` should be treated correctly (i.e. it should not be caught), which it currently isn't.

**Comment by vdbergh** [comment]
Using the docker image it is easy to start a lot of workers at the same time. The download time seems to be multiplied by the number of workers. So it becomes very significant (this should also happens on real hardware but it is harder to test).

**Comment by ppigazzini** [comment]
I tested on the office as well (1.0 Gb/s connection vs 115Mb/s at home), the container managers don't seems to introduce an overhead.

```
# system
time curl -jsSLO https://tests.stockfishchess.org/api/nn/nn-0000000000a0.nnue

real    0m1.837s
user    0m0.358s
sys     0m0.184s

# official install from docker
0.44user 0.54system 0:01.20elapsed 82%CPU (0avgtext+0avgdata 14388maxresident)k
0inputs+125960outputs (0major+1249minor)pagefaults 0swaps

# podman
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
0.32user 0.24system 0:01.44elapsed 39%CPU (0avgtext+0avgdata 14356maxresident)k
0inputs+125960outputs (0major+1249minor)pagefaults 0swaps

# systemd-container
0.42user 0.23system 0:02.00elapsed 32%CPU (0avgtext+0avgdata 14472maxresident)k
24inputs+125960outputs (0major+1227minor)pagefaults 0swaps
Container jammy-000 exited successfully.
```

vdbergh:
> I tested on the office as well (1.0 Gb/s connection vs 115Mb/s at home), the container managers don't seems to introduce an overhead.
>
You are right. I was wrong about this.

ppigazzini:
> Using the docker image it is easy to start a lot of workers at the same time. The download time seems to be multiplied by the number of workers. So it becomes very significant (this should also happens on real hardware but it is harder to test).

If the container is configured to have 4 workers, they runs in separated folders. Starting 6 containers with 4 workers requires 24 downloads. The worker will lose the control on the cores used.

To be honest, a skilled linux user is able to setup fishtest fulfilling the requirements to accomplish his use case and security needs (firejail, systemd with another user, lxd/lxc, docker, podman etc.), a not skilled linux user can have problem installing docker and managing docker. By the way the docker daemon runs indefinitely and with root privileges, podman is safer.

vdbergh:
The point is simply to make the worker quit quickly and cleanly on receiving `SIGINT`. This requires some small modifications to the worker which are contained in this PR.

ppigazzini (approved):
Tested:
- WSL2: cli, nohup and script
- Linux: cli, script, nohup, script and systemd
- msys2: cli, script

ppigazzini:
PROD updated to trigger the workers update, thank you @vdbergh , you can safely delete the branch "docker_worker" ;)

vdbergh:
Thanks! Sorry that it took so many iterations!

ppigazzini:
https://tests.stockfishchess.org/actions?page=1&user=vdbergh

You can run the docker test worker on DEV adding `--host dfts-0.pigazzini.it`

vdbergh:
> https://tests.stockfishchess.org/actions?page=1&user=vdbergh
> 
> You can run the docker test worker on DEV adding `--host dfts-0.pigazzini.it`

I know of course... I actually have a config file `fishtest_dev.cfg` for this which I copy over `fishtest.cfg` for testing. But sometimes I forget while doing a quick test.

ppigazzini:
On msys2 console, with which I usually test the PRs, sometimes/often the console hangs during the `make` of every version of SF. If I comment out the `stdout=subprocess.PIPE`, the worker runs always fine.
Running a short `build.py` program the hang is nearly deterministic when building old version of Stockfish (eg, SF 11), while never happened with master.

By the way `shell = True` can be dropped even with SF 11.

Tonight I will test what happens running the worker installed with the suggested windows installer.

vdbergh:
It is not obvious to me how to diagnose this. Does the worker really hang forever, or does it hang until the `make` process is finished?

`cutechess-cli` also runs with `stdout=subprocess.PIPE` and there are no problems there. One difference though is that in the latter case reading from the pipe goes through some extra hoops to make it non-blocking. Still a blocking read should not cause a hang. When the `make` process finishes, `stdout` is closed and the blocking read is released.

This being said, this PR has no effect on windows. So one could do the `stdout=subprocess.PIPE` only on posix.

ppigazzini:
After years from https://github.com/official-stockfish/fishtest/pull/713, on Windows, the stderr/stdout of an application started with python still have a weird behavior.
I verified that, in either the msys2 shell or the cmd used in the workers, the warnings of `make` on stderr are blocking the reads on stdout. The build process is fine using `stderr=subprocess.STDOUT` or `stdout, stderr = p.communicate()`.

PR #1866, opened 2023-12-18
Convenient link to blocked workers

peregrineshahin:
Currently, there is no convenient way to get to the particular worker page to block/unblock after misbehaving workers are spotted in test_view Tasks. You would need to write the URL manually or search for that worker on the home page.

Make the approver-only anchor redirect to the worker management rather than the user page as the user page is irrelevant after we can block only the worker.

### DIFF
--- a/server/fishtest/templates/tasks.mak
+++ b/server/fishtest/templates/tasks.mak
@@ -30,7 +30,7 @@
       <td>
     % endif
     % if approver and task['worker_info']['username'] != "Unknown_worker":
-      <a href="/user/${task['worker_info']['username']}">
+      <a href="/workers/${worker_name(task['worker_info'], short=True)}">
         ${worker_name(task['worker_info'])}
       </a>
     % elif 'worker_info' in task:


vdbergh:
There should also be a convenient link to a user profile page (to get their email address). I think this PR is good, but I don't know any other way to get to the profile page of a user, without hacking the url.

vdbergh:
I see. One can also search in the Event log for an event in which the user is source or target. Still it is a bit indirect.

Perhaps the "Contributors" or "Top month" pages could contain links to the user profile pages (for maintainers).

peregrineshahin:
Okay, that's a great suggestion.

peregrineshahin:
I will make a client js search inside this page to search for any text in a row also. so let's draft this until I come back to this

peregrineshahin:
> I will make a client js search inside this page to search for any text in a row also. so let's draft this until I come back to this

done here https://github.com/official-stockfish/fishtest/pull/1867

vdbergh:
Could you perhaps add a comment to `filterTable` about how to use it? I feel that if it is in `application.js` it should be reusable.

ppigazzini:
DEV updated with #1866 and #1867

ppigazzini (approved):
Looks good on DEV

ppigazzini:

peregrineshahin:
> Could you perhaps add a comment to `filterTable` about how to use it? I feel that if it is in `application.js` it should be reusable.

Sorry I messaged the message before the merge, It's pretty straight forward but I will add comments in next PR's probably in a cleanup

PR #1867, opened 2023-12-18
Implement client filtering.

peregrineshahin:
This adds search filtering to contributors and top-monthly client-side based on any text in a row.


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1870, opened 2023-12-19
Generalize the EvalFile net option detection

ppigazzini:
Stockfish will use more than one net with
https://github.com/official-stockfish/Stockfish/pull/4915

Get all the couples "option" and "net names" from the output of the
`stockfish uci` command.

Raise worker version to 227 (also server side)

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -231,8 +231,9 @@ def github_api(repo):
     return repo.replace("https://github.com", "https://api.github.com/repos")
 
 
-def required_net(engine):
-    net = None
+def required_nets(engine):
+    nets = {}
+    pattern = re.compile(r"(EvalFile\w*)\s+.*\s+(nn-[a-f0-9]{12}.nnue)")
     print("Obtaining EvalFile of {} ...".format(os.path.basename(engine)))
     try:
         with subprocess.Popen(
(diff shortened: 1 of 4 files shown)


worker/games.py
```diff
@@ -1341,17 +1341,14 @@ def parse_options(s):
...
             )
 
-    # Add EvalFile with full path to cutechess options, and download the networks if missimg.
-    net_base = required_net(base_engine)
-    if net_base:
-        base_options = base_options + ["option.EvalFile={}".format(net_base)]
-    net_new = required_net(new_engine)
-    if net_new:
-        new_options = new_options + ["option.EvalFile={}".format(net_new)]
+    # Add EvalFile* with full path to cutechess options, and download the networks if missimg.
```
ddobbelaere, line 1344: s/missimg/missing/ ;)

mstembera:
@ppigazzini Thank you for this.  Is this the last thing needed on the fishtest side before https://github.com/official-stockfish/Stockfish/pull/4915 can be merged?

mstembera:
@ppigazzini Can you please commit this so we can proceed with https://github.com/official-stockfish/Stockfish/pull/4915 ?  Ty!

peregrineshahin:
Maybe ppigazzini realized a big flaw in this PR and still thinking about the solution ;) ..

Disservin:
btw we should also update the memory requirements even though it's not much I guess.
The second net is 3MB, the current main net is estimated to be around 80MB in fishtest code which I think is still an old value, because the net file is around ~64MB. 

Also fishtest and the worker are out of sync regarding that. 
https://github.com/official-stockfish/fishtest/blob/master/server/fishtest/rundb.py#L1108
https://github.com/official-stockfish/fishtest/blob/master/worker/worker.py#L800

the server code is missing the 16MB ?

ppigazzini:
Tested with workers: msys2, linux (python 3.6.8) with:
 - sf_11 vs master
 - sf_classical vs master
 - sf_12 vs master
 - SF_NNUE vs master
 - DualNNUE-1100-pr vs master
 
DualNNUE-1100-pr requires `'option.EvalFile=nn-0000000000a0.nnue', 'option.EvalFileSmall=nn-c01dc0ffeede.nnue'`

image

ppigazzini:
Triggered the worker update, of course.

Disservin:
Thanks. I made a PR for the according Stockfish changes.

PR #1874, opened 2023-12-24
Modifying the "Reduce ITP with number of active tests per user"

FauziAkram:
Reasons for this change:

**Support for Active Developers:**

The current number of active developers is relatively low. Penalizing them with stringent bounds may discourage contributions. The proposed change ensures a more lenient approach, fostering a supportive environment for developers.

**Balancing Act:**

The adjustment strikes a balance between encouraging active testing and maintaining reasonable control over the number of tests. It provides a smoother reduction in itp as the number of active tests increases, offering a fair compromise.

**Community Feedback:**

I suggest seeking feedback from the fishtest community on this proposed change. Open discussions can provide valuable insights and help in refining the formula to better suit the collective needs of the community.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ David Zar
 dubslow
 Fabian Fichter (ianfab)
 Fanael Linithien (Fanael)
+Fauzi Akram Dabat (Fauzi2)
 FieryDragonLord
 Gabe (MrBrain295)
 Giacomo Lorenzetti (G-Lorenz)
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -865,7 +865,7 @@ class RunDb:
                 itp *= bonus
 
         # Malus for too many active runs
-        itp *= 36.0 / (36.0 + count * count)
+        itp *= 64.0 / (64.0 + count * count)
 
         run["args"]["itp"] = itp
 


locutus2:
To bring a bit data i created a chart which shows total itp for master and this PR. There you see that the current optimal number of 6 tests would change to 8 running tests. Personnaly i would support this PR.

Remark: i had done in the past another PR which fixed total itp at tests=6 total itp if this was reached, but this was not in line with the intensions of the maintainers: https://github.com/official-stockfish/fishtest/pull/1844

grafik

locutus2:
Further change is the ratio between the total itp if running only one test and the optimim one changes from 1:3 to 1:4. If we want maintain this ratio we can scale the formula par example like this:
`itp *= 64.0 / (64.0 + count * count + 6 * (count - 1))`
And here the chart with all three formulas:
grafik

cj5716:
I am unsure whether changing from 3 itp for 6 tests to 3 itp for 8 tests is better than 4 itp for 8 tests

vondele:
I agree with mstembera here, current setup works well.

dubslow:
I cast my vote with fauzi and locutus (a borg) that increasing the max tp to 8, and perhaps also reducing the slope before 8, would be beneficial. I agree that there is some benefit from forcing submitters to prioritize, but I think the present slope below 6 is somewhat harsh (loss of latency at maximum tp, I already have one PR from years ago about improving latency).

PR #1876, opened 2023-12-26
Make the meaning of some strings in the schemas explicit.

vdbergh:
Note: this PR conflicts (unavoidably) with #1877  . If the latter would be committed then this one has to be made into a draft, and vice versa.


ppigazzini (approved):
Looks good on DEV:
- creating, modifying, finishing, stopping, deleting tests
- blocking, unblocking users and workers
- joining, stopping workers
- query event logs and checking the DEV logs
- verified the regular expressions

ppigazzini:

PR #1877, opened 2023-12-27
Introduce log_message action.

vdbergh:
This may be used to give internal errors in the logic of Fishtest maximum visibility.

This action is currently used in two places:
    
- if a finished run does not validate;
    
- if `check_results` finds a discrepancy between the incrementally updated results and the computed results in a finished run.
    
Currently both errors are logged but otherwise they are treated silently.
    
Another application in the future might be to log certain worker errors which happen before a task is requested (e.g. failure to be able to run `cutechess-cli`).

While at it, replace all occurences of `_validate()`, which returns an error string on non-validation, by the more Pythonic `validate()`, which throws an exception instead.

Note: this PR conflicts (unavoidably) with #1876 . If the latter would be committed then this one has to be made into a draft, and vice versa.

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -2,7 +2,7 @@ from datetime import datetime, timezone
 
 from fishtest.util import hex_print, worker_name
 from pymongo import DESCENDING
-from vtjson import _validate, union
+from vtjson import ValidationError, union, validate
 
 schema = union(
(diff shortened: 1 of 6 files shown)


ppigazzini:
Drafted after merging #1876

vdbergh:
Superseded by #1906

PR #1878, opened 2023-12-28
Detect and identify an AWS instance

vdbergh:
This PR adds a suffix `(aws:<instance_type>:<instance_life_cycle>)`, for example `(aws:c4.8xlarge:spot)`, to the machine name when running on AWS (screenshots in the first comment).
    
Justification: 

- some workers may be tempted to use an under powered instance, possibly resulting in crashes and time losses. With this PR it should be possible to identify such workers. Note that the usual speed test cannot always be used reliably to eliminate such workers since some instance types (the `t-series`) have a burst mode: they will run at full speed for a while and then slow down dramatically.

- spot AWS instances may not have a stable UUID prefix. It it useful to be able to recognize this situation.

The documentation for the API used can be found here 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html


vdbergh:
Here is a screenshot:

Screenshot from 2023-12-28 16-09-13

Note that in this case the kernel version already has `aws` in it. However this is not necessarily the case.

vdbergh:
Screenshot from 2023-12-28 21-11-44

Note that here `aws` already is a substring of the kernel version. However this is a property of the image. It does not have to be the case.

vdbergh:
I am drafting this since it seem in recent instances the API has changed.

vdbergh:
The use of AWS for Fishtest seems rare. So maybe there is no need right now to clutter the worker with this PR.

PR #1880, opened 2023-12-31
Merging Non-regression STC and LTC into one

FauziAkram:
Merging the options of Non-regression STC and LTC into one.
As they have the same bounds, and therefore there are no benefits in having two separate options.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ David Zar
 dubslow
 Fabian Fichter (ianfab)
 Fanael Linithien (Fanael)
+Fauzi Akram Dabat (FauziAkram)
 FieryDragonLord
 Gabe (MrBrain295)
 Giacomo Lorenzetti (G-Lorenz)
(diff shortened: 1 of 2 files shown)


vdbergh:
They currently have the same bounds. There is no guarantee that this will always be the case.

FauziAkram:
The STC and LTC no-regression tests have shared identical bounds for over 2.5 years:
https://github.com/FauziAkram/fishtest/commit/8308edf4e1b245d3f922e4439f221094dae90a3e
With no ongoing discussions about potential divergence. If they were to diverge in the future, addressing it would involve modifying the code regardless. Therefore, maintaining a simpler version until such a divergence occurs is advisable.

peregrineshahin:
master is more informative to new users.
Apart from the PR introduces a bug. `value="something"` cannot be changed like nothing happened without looking where in the code, the old values (regression STC, regression LTC) were used.
```diff
-                   <option value="regression STC">Non-regression STC ${fb(-1.75, 0.25)}</option>
-                   <option value="regression LTC">Non-regression LTC ${fb(-1.75, 0.25)}</option>
+                  <option value="regression">Non-regression ${fb(-1.75, 0.25)}</option>
```

vdbergh:
I wonder why there is no "Test type" button for non-regression tests....

PR #1882, opened 2024-01-02
Make "This machine is too slow ..." non-fatal.

vdbergh:
This failure may be intermittent and it currently stops the worker.


ppigazzini (approved):
Looks good on DEV

ppigazzini:
Triggered the workers update.

PR #1883, opened 2024-01-04
Clear out confusion in wrong bench message.

peregrineshahin:
The previous message is confusing and vague enough to convey the reversed meaning.


dubslow:
mb `author` over `user`?

peregrineshahin:
> mb `author` over `user`?

there is the case when the user submits the test on behalf of the author :P

ppigazzini (approved):
Looks good on DEV

ppigazzini:
Triggered the workers update, thank you @peregrineshahin

PR #1885, opened 2024-01-07
Remove the fish.exit file after the worker has stopped.

vdbergh:
It seems more useful that way...


ppigazzini (approved):
Looks good with workers running linux phyton 3.6.8 and newer, and msys2 python.
I simplified a bit the code using `pathlib` as in other places.

ppigazzini:
Triggered the update fo the workers.

PR #1886, opened 2024-01-07
handle logout fail

peregrineshahin:
No redirection to home page if logout failed, clears confusion.

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -229,18 +229,30 @@ function handleApplicationLogout() {
   });
 }
 
+// Alerts errors to the UI
+function alertError(message) {
+  document.getElementById("error_div").style.display = "";
+  document.getElementById("error").textContent = message;
+}
+
 async function logout_() {
   const csrfToken = document.querySelector("meta[name='csrf-token']")[
     "content"
   ];
 
-  await fetch("/logout", {
-    method: "POST",
-    headers: {
-      "X-CSRF-Token": csrfToken,
-    },
-  });
-  window.location = "/";
+  try {
+    const response = await fetch("/logout", {
+      method: "POST",
+      headers: {
+        "X-CSRF-Token": csrfToken,
+      },
+    });
+
+    raiseForStatus(response);
+    window.location = "/";
+  } catch (error) {
+    alertError("Network error: please check your network!");
+  }
 }
 
 function logout() {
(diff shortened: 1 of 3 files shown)


peregrineshahin:
python packages failing for some reason, my change is unrelated to worker tests, but w/e

server/fishtest/static/js/application.js
```diff
@@ -225,9 +225,13 @@ function raise_for_status(response) {
...
-  const response = await fetch(url, options);
-  raise_for_status(response);
-  return response.json();
+  try {
+    const response = await fetch(url, options);
+    raise_for_status(response);
+    return response.json();
+  } catch (error) {
+    console.error("Network error: ", error);
+  }
```
vdbergh, line 234: Is there a compelling reason to change the error behavior of the wrapper functions (returning null instead of an exception)?

Also, this changes the error handling in the caller. Did you check this has no impact on the current use of the wrapper functions (e.g. in notifications.js)?

vdbergh:
Some informal Googling seems to suggest that returning an exception instead of null is the preferred error handling in most languages (it Python this is of course part of the general philosophy of the language).

peregrineshahin:
I didn't realize it's already done outside of the function elsewhere anyways.
I'm reverting this change.
```
  try {
    json = await fetch_json("/api/active_runs");
  } catch (e) {
    console_log(e, true);
    return;
  }
  ```

**Comment by ppigazzini** [comment]
DEV updated:
- login, redirected to the home
- switch in a new page (eg contributors)
- push logout -> "Logout" button grey out
- it needs a page refresh to get the "Login" button

I rebased the PR on master to fix code format and the msys2 workflow, supposed you added your GH personal repo as "origin" remote, update with:
```bash
git fetch -p origin
git switch handleLogout
git reset --hard origin/handleLogout
```

peregrineshahin:
Ofc it does not work anymore since there is a hidden merge conflict after you updated.
raise_for_status() changed to raiseForStatus()

ppigazzini:
I figured out only this way to test the PR: open a page different by home, login, change the IP to a wrong subnet, push the "Logout" button. The page doesn't change to the home for either master or PR, the error message in the console changes, though:
- fishtest master:
image
- fishtest PR:
image

peregrineshahin:
Yes, this is the expected result.. you could have turned off the internet instead of changing the IP......

peregrineshahin:
Ah... `The page doesn't change to the home for either master or PR`
Oh then the fetch returns fatal error that prevent the code to continue.. the PR is kinda useless now just that it's better to catch the error

peregrineshahin:
Never mind I add the error to the UI while at it.

ppigazzini (approved):
Looks good on DEV

image

image

ppigazzini:

PR #1887, opened 2024-01-07
Require writing old password for changing user info.

peregrineshahin:


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1888, opened 2024-01-09
Make the concurrency==cores error into a warning.

vdbergh:
The current setup is a bit inconvenient for shell scripts.

An alternative could be to test for `isatty()` but this is a bit hacky.


vdbergh:
Found a workaround.

PR #1890, opened 2024-01-15
Never delete accepted Fishtest users

peregrineshahin:
there is no need to delete users, and make them confused if they were accepted once upon a time.


peregrineshahin:
This is bugged, refactoring later.

PR #1895, opened 2024-02-18
Don't hide "non-tested" nets for logged out users

peregrineshahin:
there is no reason to hide them on the first place, and practice shows it's not the ideal thing let alone the net is still available so this is a fake frontend guard.

A follow up fix should be, website server changes after dual net, since the master net is not marked as tested for now.


vondele:
The reason for this, historically, was to hide offensive names or content that was malicious, or spam (repeated) uploads 
This might no longer be necessary (i.e. the format of the name is checked, can at most be offensive leet ;-) )

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1899, opened 2024-02-24
Fix tables overflows

peregrineshahin:
Sometimes when the text is too long, the page can expand in width, this fixes that


ppigazzini (approved):
Looks good on DEV, temporary added by hand to PROD to test it with other devices (eg iPad) before merging.

ppigazzini:

PR #1908, opened 2024-03-04
Benches both engines prior to raise a RunException

ppigazzini:
This strategy ensures that the Events Log receives detailed information,
which assists in identifying the root cause of the issue.
Possible causes could range from an erroneous bench entry on the Fishtest
new run page, a malfunctioning engine, to a worker misconfiguration,
among others.

A WorkerException is still raised immediately.

closes https://github.com/official-stockfish/fishtest/issues/1897

Thanks to @peregrineshahin and @Disservin for bringing this issue to light.

Raise worker version to 231 (also server side)


ppigazzini:
image

PR #1911, opened 2024-03-05
Add menu for blocked workers

peregrineshahin:
Also Enhances filterTable clientside function to take a predicate instead of always includes
Closes https://github.com/official-stockfish/fishtest/issues/1894

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -365,7 +365,7 @@ function padDotVersionStr(dn) {
 }
 
 // Filters tables client-side
-function filterTable(inputValue, tableId, originalRows) {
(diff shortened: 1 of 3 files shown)


ppigazzini:
The dropdown seems to be stuck on "Modified <= 5 days ago"

image

ppigazzini:
Uncaught (in promise) TypeError

image

peregrineshahin:
should be fixed

ppigazzini:
Filtering is text based on "days ago", but we should consider "seconds" and "hours" as well. 


image

peregrineshahin:
should be fixed.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1912, opened 2024-03-09
Restrict SPSA C values to be greater than 2

Viren6:
Referring to https://github.com/official-stockfish/fishtest/issues/1907#issuecomment-1974860493

I have seen multiple tunes with c < 2, some even below 0.5. There is no reason to ever tune with the same parameter value across the interval (no derivative), the evolution of a parameter can be reduced by modifying the r value instead.

Moreover, there seems to be a misinterpretation in some devs that low c values are ok because 'stochastic rounding' will handle it, without having any understanding of how the underlying algorithm is actually working. This is not surprisingly given that this formula is not mentioned in any wiki.

This PR will expose to the dev clearly that certain c values have no effect. They can combine multiplying up the value for the tune with adjusting the r value to tune more effectively if they wish. This should be controlled on the dev side rather than within fishtest.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1091,7 +1091,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], int(x["value"])
+                x["name"], math.floor(x["value"])
             )
             for x in w_params
         ]
@@ -1102,7 +1102,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], int(x["value"])
+                x["name"], math.floor(x["value"])
             )
             for x in b_params
         ]
--- a/worker/games.py
+++ b/worker/games.py
@@ -1091,7 +1091,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], x["value"]
+                x["name"], int(x["value"])
             )
             for x in w_params
         ]
@@ -1102,7 +1102,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], x["value"]
+                x["name"], int(x["value"])
             )
             for x in b_params
         ]
--- a/worker/games.py
+++ b/worker/games.py
@@ -1091,7 +1091,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], math.floor(x["value"])
+                x["name"], math.floor(x["value"] + random.uniform(0, 1))
             )
             for x in w_params
         ]
@@ -1102,7 +1102,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], math.floor(x["value"])
+                x["name"], math.floor(x["value"] + random.uniform(0, 1))
             )
             for x in b_params
         ]
--- a/worker/games.py
+++ b/worker/games.py
@@ -1091,7 +1091,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], math.floor(x["value"] + random.uniform(0, 1))
+                x["name"], x["value"]
             )
             for x in w_params
         ]
@@ -1102,7 +1102,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], math.floor(x["value"] + random.uniform(0, 1))
+                x["name"], x["value"]
             )
             for x in b_params
         ]
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -773,8 +773,8 @@ def parse_spsa_params(raw, spsa):
             continue
         if len(chunks) != 6:
             raise Exception("The line {} does not have 6 entries".format(chunks))
-        if float(chunks[4]) < 2:
-            raise Exception("The tuner requires c values greater than 2")
+        if float(chunks[4]) < 1:
+            raise Exception("The tuner requires c values greater than 1")
         param = {
             "name": chunks[0],
             "start": float(chunks[1]),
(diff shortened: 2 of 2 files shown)


worker/games.py
```diff
@@ -1091,7 +1091,7 @@ def launch_cutechess(
         cmd[:idx]
         + [
             "option.{}={}".format(
-                x["name"], math.floor(x["value"] + random.uniform(0, 1))
+                x["name"], x["value"]
```
ppigazzini, line 1094: `cutechess-cli` requires an integer value

ppigazzini:
IMO before rejecting a c value on the GUI we must write documentation and example code on the wiki to guide the devs. I don't know the SF code to write the example code.

peregrineshahin:
There is two tries in Stockfish regarding this that got rejected because of Developer's responsibility
https://github.com/official-stockfish/Stockfish/pull/4776
https://github.com/official-stockfish/Stockfish/pull/4807
 
Glad to know the reasoning about where such issue if existent should be fixed?
Is it lack of documentation?
is it better to fix in SF?
Is it better to fix in Fishtest backend?
Is it better to restrict in Fishtest frontend?
Is there any issue?
TBH this topic reoccures once in a while

Viren6:
The main issue I see is that there is no indication that c > 1 could ever have the same value across the interval, which the above 2 PRs wouldnt address, as they are on the Stockfish side. I will work on adding something to the creating my first test wiki that will explain how to handle c < 1 properly

peregrineshahin:
Can you explain why the random factor was added and needs removing now?
Also I if I remember correctly that the minimum C should be was 0.5 not 1?

Viren6:
The point is, this entire stochastic rounding concept doesnt make sense. For example, with a c value of 0.1 to make +1 and -1 10% of the time in a 100k game tune, this is effectively just a 10k game tune with added noise.

Viren6:
> IMO before rejecting a c value on the GUI we must write documentation and example code on the wiki to guide the devs. I don't know the SF code to write the example code.

image
I have added to the wiki the method for tuning effective c values less than 1

vondele:
For the wiki example, I'll make a small edit to it (in integer math this makes no big difference), but yes that is the right approach.

ppigazzini:
@Viren6 we must keep the flooring in games.py, `cutechess-cli` requires an integer value for any parameter.

peregrineshahin:
> @Viren6 we must keep the flooring in games.py, cutechess-cli requires an integer value for any parameter.
doesn't integer also trunc positive values?
I think the difference between truncing and flooring lies in the negative values from CS POV.

ppigazzini:
IMO it's weird to drop the stochastic rounding only because some people wrongly think that it fixes this particular issue. Of course, if we are able to enforce `c > 100` (at example) we could simplify away. Unfortunately, the change of scale of the parameters must be done in both fishtest GUI and SF code and it's better to minimize the work of the dev.

More perplexing it's simplifying the stochastic rounding while keeping the comment about it.

Viren6:
Hm. Maybe it's better to keep stochastic rounding and enforce a minimum c value of 2 instead? I didnt consider its interaction with flooring

Viren6:
Yeah. c>2 with stochastic rounding is the better solution, it also doesnt affect tuner behaviour from what it is currently.

ppigazzini:
A higher minimum it's better if this doesn't increased too much the work of a dev. I don't know the range of the parameters in SF code to make a judgment.

peregrineshahin:
I vaguely remember that the step is calculated as 2c
Out of curiosity can someone confirm that the wanted was 0.5 as C

ppigazzini:
@Viren6 btw the CI failed because when you change one of the worker files you need to update the `sri.txt` hash with:
```bash
python worker.py a a --only_config --no_validation
```

ppigazzini:
> I vaguely remember that the step is calculated as 2c Out of curiosity can someone confirm that the wanted was 0.5 as C

I'm not sure to have understood your question.
Anyway `c` value must be chosen to have an effect on the mini match, here is an excerpt from the wiki:

- Evaluation step: play a mini match using 2 different values of the parameter to be optimized:
   - `parameter_value - ck`
   - `parameter_value + ck`
- Update step: update the parameter according to the mini match result:
    `parameter_value = parameter_value + (ck * rk) * (wins - losses)`

Viren6:
I think 2 is a good threshold, since there are diminishing returns increasing it further for an increased dev workload. 

The threshold of 2 is affecting 15/70 params in our most recent VVLTC tune.

peregrineshahin:
How is limiting c to be minimum of 2 is any good, fauzi argued initially that he wants to tune values that can range from 1 to 4 in LMR?

peregrineshahin:
@ppigazzini my question is, is c the same as ck here?

Evaluation step: play a mini match using 2 different values of the parameter to be optimized:
parameter_value - ck
parameter_value + ck

or ck is 2*c?

ppigazzini:
`ck` is `c` at `k` iteration.

Viren6:
> How is limiting c to be minimum of 2 is any good, fauzi argued initially that he wants to tune values that can range from 1 to 4 in LMR?

It is now outlined in https://github.com/official-stockfish/fishtest/wiki/Creating-my-first-test#tuning-with-spsa how this can be achieved. He can multiply up the values within SF code and pass those to the tune. Its more work, but what he would have been doing before is a tune that is effectively c=1 with less effective games

peregrineshahin:
The goal I care about here of such PR is a minimal fix that two engine does not play with the same parameters
This is the only thing we seem to lack now
the argument should only be to limit ck to a minimum of 1
I don't get what's the randomizing for
I only care that two engines does not play the with the same paramters as minimally as possible
in such case why don't we stick to c as 1?

Viren6:
Thats the thing. If you look closely at the formula, it actually allows playing with the same params with 1 < c < 2

peregrineshahin:
So you are saying because of the random factor we have we need to up it to 2, and let anyone who wants a c of 1 to do the trick in the wiki
fair by me I guess

peregrineshahin:
Okay so another question
does C itself change that it can become < 2 not from the input but rather automatically by the algorithm?

Viren6:
We set only c_end as a minimum of 2, which is the lowest it can get in the tune

Viren6:
Closing for now since the issue seems to be resolved from the wiki update

PR #1913, opened 2024-03-12
consistent "LTC:" addition

peregrineshahin:
the addition of "LTC:" has been overwritten by the green highlighting feature which leaves one source of truth to edit it if needed.

currently this looks old way to catch LTCs, here is a conflict of green highlighting and the added text https://tests.stockfishchess.org/tests/view/65f0ca4f0ec64f0526c46a67

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -350,12 +350,14 @@ def get_tc_ratio(tc, threads=1, base="10+0.1"):
     return threads * estimate_game_duration(tc) / estimate_game_duration(base)
 
 
-def is_active_sprt_ltc(run):
+def is_sprt_ltc_data(args):
     return (
-        not run["finished"]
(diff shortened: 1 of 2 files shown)


ppigazzini:
LTC highlight seems disabled in finished tests, so the "LTC:" text add some a value. The regex is not covering all the possible templates of the TC, though.

peregrineshahin:
How about using the same function of the highlighting and since the submitted run at the time of submission will not be finished then it should match no?
```
def is_active_sprt_ltc(run):
    return (
        not run["finished"]
        and "sprt" in run["args"]
        and get_tc_ratio(run["args"]["tc"], run["args"]["threads"]) > 4
    )  # SMP-STC ratio is 4
```

ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

PR #1914, opened 2024-03-16
Make size restrictions on some field in actiondb part of the schema.

vdbergh:
Requires upgrading the vtjson package.


vdbergh:
This PR was merged in #1915

PR #1915, opened 2024-03-16
Tighten up the run schema.

vdbergh:
In particular the requirement that the final results must be the sum of the results of the individual tasks is now part of the schema.

For the PR to work vtjson should be upgraded to the latest version.


vdbergh:
This has been integrated in #1916

PR #1918, opened 2024-03-30
View actual git diff in tests_view

peregrineshahin:
This request has been pending for years, since GitHub does not provide the diffs with two dots in the API we are not able to actually see the actual diff in test view, this replaces the request to GitHub API with three dots to a local algorithm with the help of jsDiff library to find the actual diff between files.

How can we continue from there:
make tabs for the diff section such that we can see the actual diff with official master without going to GitHub.

online tool made for testing at: https://githubdifftesting.netlify.app/

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -6,6 +6,13 @@
 
 <%namespace name="base" file="base.mak"/>
 
+<script
+  src="https://cdnjs.cloudflare.com/ajax/libs/jsdiff/5.2.0/diff.js"
+  integrity="sha512-Ubw08LwzeoACkP9I1L0LqwnhiVs6Vg9LWeTFDj0leZIVp5XG28xcxK0lrlFvaL8ZqR8oMOBnFdZLo6agy5+SgQ=="
+  crossorigin="anonymous"
+  referrerpolicy="no-referrer"
+></script>
+
 % if show_task >= 0:
   <script>
     document.documentElement.style="scroll-behavior:auto; overflow:hidden;";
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV running with this PR, here is a test with a very different result using the PR:
https://tests.stockfishchess.org/tests/view/63f31bc2e74a12625bccc7e7
https://dfts-0.pigazzini.it/tests/view/63f31bc2e74a12625bccc7e7

Here are some of the previous attempts with highlighted some plus and minus about the use of the different versions of `git diff`:
https://github.com/official-stockfish/fishtest/pull/924
https://github.com/official-stockfish/fishtest/pull/959

peregrineshahin:
Looks like SPSA tunes does not work
I see that there is a conditional somewhere in the old code that compared the diff between master and the new tag, because well obviously in tunes the base and the new tags are the same

peregrineshahin:
Fixed the SPSA tunes case.

ppigazzini:
@vondele @Disservin merge conditionate to your opinions about the impact of this PR on the workflow of approvers/maintainers.

ppigazzini:
DEV updated.

vondele:
The idea of showing the actual diffs is fine with me, it is the right thing to do. I can't comment on the implementation.

For SPRT tests we basically want to know 2 things.
A) the exact difference between the base and the test code (which this patch does)
B) verify that the base branch matches official-stockfish/master at the time of submission, or be able to see these differences in exceptional cases.

For SPSA we would like to
A) Enforce that base and test are the same code (this is the enforced in fishtest AFAICT)
A) Display the exact difference between the code that runs and official-stockfish/master at the time of submission

ppigazzini (approved):

ppigazzini:

peregrineshahin:
For SPRT
A) is done
B) is doable I will do it later today

For SPSA we would like to
A) is done
B) The only commit SHA states saved in the DB are the Base and New which match for SPSA, Thus this is solvable with introducing another field for `master commit SHA` if you care about old SPSA's what we can do is use this field only for new tests.

PR #1927, opened 2024-04-05
Enhancements of tests diff

peregrineshahin:
This implements For SPRT:
verify that the base branch matches official-stockfish/master at the time of submission, or be able to see these differences in exceptional cases.
For SPSA:
Display the exact difference between the code that runs and official-stockfish/master at the time of submission and fixes some bugs.


server/fishtest/schemas.py
```diff
@@ -422,6 +422,8 @@ def final_results_must_match(run):
                 "threads": suint,
                 "resolved_base": sha,
                 "resolved_new": sha,
+                "master_sha?": sha,
+                "official_master_sha?": sha,
```
vdbergh, line 426: Will master_sha and official_master_sha always be present in new runs? In that case there is no need to make them optional. If they are only for the SPSA tests then things are different of course.

peregrineshahin: > Will master_sha and official_master_sha always be present in new runs? In that case there is no need to make them optional. If they are only for the SPSA tests then things are different of course.

Removed optional and fixed test cases.

peregrineshahin: that being said I think a type of `any` might be useful in the future.

ppigazzini:
DEV updated.

server/fishtest/schemas.py
vdbergh: The type “object” plays the role of “any”. I will put that in the readme.

peregrineshahin:
for discovering malicious code here is an example https://dfts-0.pigazzini.it/tests/view/66113a07814ea106e1c82966

vdbergh:
> DEV updated. Server logs clean after scheduling new runs (SPRT, SPSA, NumGames) and rescheduling old runs (SPRT, SPSA, NumGames).

There will be non-validation messages in the event log on PROD until all the current tests finish. I think that's ok but I wonder if there would be sensible solution to this. One thing I can think of is that the non-validation messages would only be shown for runs created after a certain date (which would be part of the commit).

vdbergh:
Or perhaps versioning the run object is better... (just like the worker has versions)?

peregrineshahin:
@vdbergh do you know how to update my vtjson package in the lxc/lxd of the wsl2 instance
It keeps saying cannot import div from vtjson even after
pip install vtjson --upgrade
fishtest@fishtest:~/fishtest$ pip install --upgrade vtjson
Requirement already satisfied: vtjson in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (1.6.4)
Requirement already satisfied: dnspython in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (2.6.1)
Requirement already satisfied: email-validator in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (2.1.1)
Requirement already satisfied: idna in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (3.6)

but I thiink this is updating in the wrong place

peregrineshahin:
> Or perhaps versioning the run object is better... (just like the worker has a version)?

Well that's why I wanted it as optional

peregrineshahin:
I was afraid that there will be a validation for the run object not just when creating it but afterwards at some point

vdbergh:
> @vdbergh do you know how to update my vtjson package in the lxc/lxd of the wsl2 instance It keeps saying cannot import div from vtjson even after pip install vtjson --upgrade fishtest@fishtest:~/fishtest$ pip install --upgrade vtjson Requirement already satisfied: vtjson in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (1.6.4) Requirement already satisfied: dnspython in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (2.6.1) Requirement already satisfied: email-validator in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (2.1.1) Requirement already satisfied: idna in /home/fishtest/.pyenv/versions/3.11.7/lib/python3.11/site-packages (from vtjson) (3.6)
> 
> but I thiink this is updating in the wrong place

You should first enter your venv and then update I think.

You can check if you are importing the right version of the package by printing `vtjson.__version__` after importing it.

peregrineshahin:
> EDIT: I noticed that it is sometimes required to do (from the fishtest home directory)
Genius it worked!!

vdbergh:
> > Or perhaps versioning the run object is better... (just like the worker has a version)?
> 
> Well that's why I wanted it as optional

Making everything optional would lead to a horribly cluttered run schema which would not be able to catch genuine errors.

Validation is done when the run is finished and it written to the db for the last time. However this is just for informational purposes. If validation fails then the run object is saved anyway (anything else would be crazy). I was talking of some way of suppressing the spurious event log messages when the run schema changes.

peregrineshahin:
> Validation is done when the run is finished and it written to the db for the last time. However this is just for informational purposes. If validation fails then the run object is saved anyway (anything else would be crazy). I was talking of some way of suppressing the spurious event log messages when the run schema changes.

Well they will be false positives but I can see how people would freak out and call it a bug when the older runs finish after the update.

ppigazzini (approved):
Looks good on DEV, when running with #1828, with finished: new runs, rescheduled old runs, modified old runs.

ppigazzini:

PR #1928, opened 2024-04-06
Versioning the run object.

vdbergh:
Currently version 0!

This is a quick go at versioning the run object to suppress spurious event log messages about validation errors for runs that were created before the schema changed.

Not yet tested but the code is fairly trivial.

This is in connection with the discussion at #1927

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -17,7 +17,7 @@ import fishtest.stats.stat_util
 from bson.binary import Binary
 from bson.objectid import ObjectId
 from fishtest.actiondb import ActionDb
-from fishtest.schemas import nn_schema, runs_schema
+from fishtest.schemas import RUN_VERSION, nn_schema, runs_schema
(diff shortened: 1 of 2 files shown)


peregrineshahin:
Would the event log supress work also if someone modified an older run and then it finished?

vdbergh:
> Would the event log supress work also if someone modified an older run and then it finished?

Not yet I guess. There is also "rescheduled_from". I'll see what I can do but these cases are rarer.

vdbergh:
> Would the event log supress work also if someone modified an older run and then it finished?

I need to look how modified and rescheduled tests are created.

vdbergh:
I think it is ok. `rescheduled_from` goes through `new_run` so it should produce runs compatible with the new schema. Modifying a run will keep the old version so event log messages will be suppressed.

ppigazzini (approved):
Looks good on DEV with finished: new runs, rescheduled old runs, modified old runs.

PR #1932, opened 2024-04-09
Fix a bug in diffs

peregrineshahin:
API limit rate shows `no diff available` then saves that in session storage, this is not intended. this prevents such text to be saved in session storage. TBH 60 requests / hour is not what I expected from GitHub, although rewriting the diff code to the server-side might help a bit it's not the best idea because the server does a lot of requests to GitHub anyways. limit rate could be easily solved once GitHub Auth is introduced (soonTM?).

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -58,7 +58,7 @@ function protectForms() {
   });
 }
 
-// Gets from the bowser the value of a saved cookie
+// Gets from the browser the value of a saved cookie
 function getCookie(cookieName) {
   return document.cookie
     .split(";")
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -652,41 +652,38 @@
     }
 
     async function fetchDiffTwoDots(rawContentUrlNew, rawContentUrlBase, apiUrlNew, apiUrlBase, diffNew, diffBase) {
-      try {
-        const [files1, files2] = await Promise.all([
-          getFilesInBranch(apiUrlBase, diffBase),
-          getFilesInBranch(apiUrlNew, diffNew),
-        ]);
+      const [files1, files2] = await Promise.all([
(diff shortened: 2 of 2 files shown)


ppigazzini:
image

ppigazzini:
I opened some run pages to hit the limit, I closed the runs pages but keeping opened the browser, I re opened that runs pages and I got this message. When closing a run page do we clear its diff from the session storage?

image

peregrineshahin:
Session storage only gets removed when the user exists the browser.

peregrineshahin:
Oh I was wrong about this, sorry not the browser the tab.

peregrineshahin:
Well this means that we need to use localStorage and expires them after a while. that would be better indeed.

ppigazzini:
I only asked if the behavior was expected.

peregrineshahin:
added a timestamp key to remove diffs of 1 day old and used localStorage, so closing the tab should persist them

> I asked only to know if the behavior was expected, after hitting the API rate limit I didn't have the possibility to check the behavior of master.

ppigazzini:
> added a timestamp key to remove diffs of 1 day old and used localStorage, so closing the tab should persist them
> 
> > I asked only to know if the behavior was expected, after hitting the API rate limit I didn't have the possibility to check the behavior of master.

On DEV the diff persists after closing the tab.

peregrineshahin:
> On DEV the diff persists after closing the tab.
looks good on local instance too

ppigazzini (approved):
The PR looks good. The only consideration here is that the optimal strategy for deleting the diff depends on the more probable (and unknown to me) user’s use case:
- if the user tends to keep tabs open for days with the interested runs, it’s better to delete the diff when closing the tab
- if users frequently close and reopen tabs for the interested runs, it’s better to delete the diff after the 1-day timeout

I'm merging this to collect the feedback from the users (if any).

ppigazzini:

PR #1934, opened 2024-04-10
Validate modify_test requests against bad requests

peregrineshahin:
this validates the post request against frontend manipulation.


ppigazzini (approved):
Lokks good on DEV

PR #1935, opened 2024-04-10
Require another approval to increase priority/throughput beyond limits

peregrineshahin:
Supersedes https://github.com/official-stockfish/fishtest/pull/1830
Dependent on  https://github.com/official-stockfish/fishtest/pull/1934

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1218,6 +1218,19 @@ def tests_modify(request):
         request.session.flash("Unable to modify another user's run!", "error")
         return home(request)
 
+    is_approver = request.has_permission("approve_run")
+    if not is_approver and (
+        (
+            run["args"]["throughput"] != int(request.POST["throughput"])
+            and int(request.POST["throughput"]) > 100
+        )
+        or (
+            run["args"]["priority"] != int(request.POST["priority"])
+            and int(request.POST["priority"]) > 0
+        )
+    ):
+        run["approved"] = False
+
     before = del_tasks(run)
     run["finished"] = False
     run["deleted"] = False
@@ -1260,7 +1273,10 @@ def tests_modify(request):
         run=before,
         message=message,
     )
-    cached_flash(request, "Run successfully modified!")
+    if run["approved"]:
+        cached_flash(request, "Run successfully modified!")
+    else:
+        cached_flash(request, "Run successfully modified!, please wait for approval.")
     return home(request)
 
 


server/fishtest/views.py
```diff
@@ -1164,82 +1207,78 @@ def tests_modify(request):
...
-        max_games = 3200000
-        if num_games > max_games:
-            request.session.flash(
-                "Number of games must be <= " + str(max_games), "error"
-            )
-            return home(request)
+    is_approver = request.has_permission("approve_run")
+    if not is_approver and (
+        (
+            run["args"]["throughput"] != int(request.POST["throughput"])
```
XInTheDark, line 1224: `!=` should probably be changed to `<` here, since we can allow e.g. decreasing throughput if it was set to above 100 previously

peregrineshahin: > `!=` should probably be changed to `<` here, since we can allow e.g. decreasing throughput if it was set to above 100 previously

Please look at the code again, != here is only the first condition to check if before != new i.e. check if changed

peregrineshahin:
Indeed that was the intended message.

peregrineshahin:
Hmm thinking about it probably it would make better sense if we allow the user to change the prio but at the same time put the test again in the pending section waiting for approval?

vdbergh:
> > DEV updated. Even when the run transitions to the approval state, the workers should continue to post results. We need to investigate whether this is causing any disruptions to the run state machine, which is currently undocumented.
> 
> What happens currently if a modified worker decided to submit a result for non approved test?

I think he can only submit results if he obtains a task. And I assume he will only be able to obtain a task for an approved test. But it would not matter to the worker if the test were suddenly unapproved (it does not care about the server state).

peregrineshahin:
Hmmm, so you are saying that if the test was unapproved suddenly the workers will still send the result of their tasks and the results will be accepted for the older tasks before approving, but it will not get new tasks after all the after the unapproval proccess?
I'm okay with that if that's the case.
i.e. no need to ditch the results/the tasks.

vdbergh:
Yes that’s what I think is true.

server/fishtest/views.py
```diff
@@ -1218,6 +1218,19 @@ def tests_modify(request):
...
+            and int(request.POST["throughput"]) > 100
+        )
+        or (
+            run["args"]["priority"] != int(request.POST["priority"])
+            and int(request.POST["priority"]) > 0
+        )
+    ):
+        run["approved"] = False
+
```
vdbergh, line 1233: Perhaps flash message here?

vdbergh:
So with this PR it seems an finished run can be unapproved.... So no rule in the schema of the form :smiling_face_with_tear: 
```python
ifthen(keys("finished"), keys("approved"))
```

ppigazzini:
DEV updated.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1936, opened 2024-04-12
Avoid timeouts for GitHub diffs

peregrineshahin:
This allows for users to use their own limited GitHub tokens to avoid exceeding the GitHub rate limit when viewing github diffs (mainly useful to approvers), which is set at 60 requests per hour for unauthenticated users. By using this token, you can surpass the standard rate limit, extending it beyond the 5000 requests per hour limit. You can find more information about GitHub's rate limits here.

It's important to note that this token will not be stored in a database but rather in your local storage. This ensures security and reduces the risk of unauthorized access.

The token should only be granted read-only permissions and limited access. This ensures that it can only be used for viewing content and not for making any modifications or accessing sensitive information.

Instructions:
1- Access the link here.
2- Generate a new token.
3- Set the Repository access to "All Repositories".
4- In the Permissions section, grant only the "Read" permission for contents, leaving all other permissions as "No Access".
5- Once generated, copy the token.
6- Save the token into the input field

the link to add/change the token https://tests.stockfishchess.org/user

Also this has many optimizations and bug fixes.

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -4,6 +4,20 @@
   document.title = "User Management | Stockfish Testing";
 </script>
 
+<script>
+  async function handleGitHubToken() {
+    await DOMContentLoaded();
+    document.getElementById("github_token").value = localStorage.getItem("github_token") || "";
+    document.getElementById("profile_form").addEventListener("submit", (e) => {
+        e.preventDefault();
+        const githubToken = document.getElementById("github_token").value;
+        localStorage.setItem("github_token", githubToken);
+        e.target.submit();
+    });
+  }
+  handleGitHubToken();
+</script>
+
 <div class="col-limited-size">
   <header class="text-md-center py-2">
     <h2>User Management</h2>
@@ -32,7 +46,7 @@
     </div>
   </header>
 
-  <form action="${request.url}" method="POST">
+  <form id="profile_form" action="${request.url}" method="POST">
     <input
       type="hidden"
       name="user"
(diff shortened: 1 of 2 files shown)


peregrineshahin:
Hopefully next step would be github auth app but it needs to be done in long steps and the final step ditching the passwords from the database.

peregrineshahin:
Slowness made me draft this since a while.
but added a big speedup (only fetch changed files according to SHA -- massive reduction in requests).
I think this is ready for review, Probably would need refactoring at some point later.

ppigazzini:
DEV updated.

peregrineshahin:
> DEV updated.

Looks good to me

server/fishtest/templates/user.mak
```diff
@@ -102,7 +116,59 @@
...
+              This ensures that it can only be used for viewing content and not for making any modifications or accessing sensitive information.
+              </p>
+              
+              <!-- Instructions on how to obtain the token -->
+              <h4>Instructions:</h4>
+              <ol>
+                <li>Access the link <a href="https://github.com/settings/tokens?type=beta">here</a>.</li>
+                <li>Generate a new token.</li>
+                <li>Set the Repository access to "All Repositories".</li>
+                <li>In the Permissions section, grant only the "Read" permission for contents, leaving all other permissions as "No Access".</li>
```
ppigazzini, line 163:
```
                <li>Press "Generate a new token".</li>
                <li>Set a "Token name".</li>
                <li>Set your preferred "Expiration" time.</li>
                <li>Set "Repository access" to "All Repositories".</li>
                <li>In "Permissions" section, grant only the "Read" permission for "Contents", leaving all other permissions as "No Access".</li>
                <li>Press "Generate token" at the bottom of the page.</li>
               <li>Copy the token and paste it into the input field above. Remember, GitHub will not show the token again for security reasons, so make sure to save it somewhere safe.</li>
```

ppigazzini: "Metadata" has mandatory Read permission.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

vondele:
Sorry for the late comment should the instructions be "All Repositories" or "Public Repositories" only? I

ppigazzini:
I tested that "Public Repositories" works fine as normal user reading some tests, but I don't know if the diff code calls the dev's repo as well and that could be private.

vondele:
the dev's repo should not be private if it is for SF development?  

Anyway, asking to generate a token that gives read access to private repos does seem a bad idea. Let's see what @peregrineshahin comments.

peregrineshahin:
Public Repo option is indeed much easier.. the reason I didn't choose it is because it gives all read-only access for all these permissions not just `contents` , I wanted it to be as minimal
Capture

vondele:
but generating a token for read access to content of private repos is a much less secure than generating a token that gives access to public stuff, right?

peregrineshahin:
That's a good point, indeed makes sense.. 
needs fixing the the instructions

PR #1937, opened 2024-04-13
When running the same binary verify signature once

peregrineshahin:
SPSA tunes and same commit vs same commit runs the same binary so verify once

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1383,16 +1383,20 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file, clear_binar
     except WorkerException as e:
         raise e
 
-    try:
-        verify_signature(
-            new_engine,
-            run["args"]["new_signature"],
-            games_concurrency * threads,
-        )
-    except RunException as e:
-        run_errors.append(str(e))
-    except WorkerException as e:
-        raise e
+    if not (
+        run["args"]["base_signature"] == run["args"]["new_signature"]
+        and new_engine == base_engine
+    ):
+        try:
+            verify_signature(
+                new_engine,
+                run["args"]["new_signature"],
+                games_concurrency * threads,
+            )
+        except RunException as e:
+            run_errors.append(str(e))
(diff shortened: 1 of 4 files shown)


worker/games.py
```diff
@@ -1372,16 +1372,17 @@ def parse_options(s):
...
-            new_engine,
-            run["args"]["new_signature"],
-            games_concurrency * threads,
-        )
-    except RunException as e:
-        run_errors.append(str(e))
-    except WorkerException as e:
-        raise e
+    if run["args"]["base_signature"] != run["args"]["new_signature"]:
+        try:
```
vdbergh, line 1376: Are you sure this is ok? What if the author of the test supplies a bogus "new_signature" equal to "base_signature"? Are these signatures validated elsewhere?

peregrineshahin: sorry I meant `new_engine != base_engine` but I your point holds nevertheless

peregrineshahin: well we can check both

peregrineshahin:
this should work
```
    if not (
        run["args"]["base_signature"] == run["args"]["new_signature"]
        and new_engine == base_engine
    ):
```

worker/games.py
vdbergh: Yes `new_engine != base_engine` seems to be a better test!

vdbergh:
Perhaps this PR will avoid the puzzling https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1714917376.066326&run_id= ?

peregrineshahin:
> Perhaps this PR will avoid the puzzling https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1714917376.066326&run_id= ?

Yes indeed.

ppigazzini (approved):
Looks good w/ DEV + workers

ppigazzini:
Triggered the workers update, thank you @peregrineshahin

PR #1941, opened 2024-04-15
Tighten the runs schema by exhibiting dependencies between flags.

vdbergh:
Note that there is some discussion in #1942 about which regex to use for usernames. In this PR, for backward compatibility, I am using a weak regex that matches all current usernames. If a consensus is reached in #1942 then in a future PR the username pattern should be updated accordingly.


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1944, opened 2024-04-17
Create one-time script to convert old usernames to new pattern (WIP)

vdbergh:
Once the conversion is done I will send an email to all the affected users.


vdbergh:
Hmm drafting this since I am now realizing that the usernames that are present in the other collections also need to be converted. This is a more substantial operation.

ppigazzini:
> Hmm drafting this since I am now realizing that the usernames that are present in the other collections also need to be converted. This is a more substantial operation.

I was waiting this Eureka moment :)

peregrineshahin:
Could you please explain what other collections means/exist?

> Hmm drafting this since I am now realizing that the usernames that are present in the other collections also need to be converted. This is a more substantial operation.

peregrineshahin:
Uh you mean the usernames will still be wrong in other connenctions like tests/runs/actions.. that's indeed annoying..

vdbergh:
> Uh you mean the usernames will still be wrong in other collections like tests/runs/actions.. that's indeed annoying..

Yes... and in principle the worker names should also be converted since they contain the username. But the use of workernames in the db is probably too recent for this to be necessary.

The conversion is not too terrible. I am working on it.

vdbergh:
I won't have time in the foreseeable future to work on this, so I am closing this for now.

PR #1945, opened 2024-04-17
Make PyMongo timezone-aware

ppigazzini:
https://pymongo.readthedocs.io/en/stable/examples/datetimes.html#reading-time

"By default all `datetime.datetime` objects returned by PyMongo will be naive but reflect UTC (i.e. the time as stored in MongoDB). By setting the `tz_aware` option on `CodecOptions`, `datetime.datetime` objects will be timezone-aware and have a `tzinfo` property that reflects the UTC timezone."

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -15,6 +15,7 @@ from datetime import datetime, timedelta, timezone
 
 import fishtest.stats.stat_util
 from bson.binary import Binary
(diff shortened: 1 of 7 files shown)


vdbergh:
I will also enforce `timezone.utc` in the schema.

ppigazzini:
Looks good on DEV, the scheduled scripts ran fine.

ppigazzini:
PROD updated.

PR #1948, opened 2024-04-21
Retrieve cached run for the tasks view

ppigazzini:
Address the issue of outdated active test data in the run retrieved from the database, which is particularly noticeable in the chi2 residuals.

Fixed the Nginx config to serve the route /tests/tasks from the main Fishtest instance.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1411,10 +1411,7 @@ def tests_stats(request):
 
 @view_config(route_name="tests_tasks", renderer="tasks.mak")
 def tests_tasks(request):
-    r_id = request.matchdict["id"]
-    run = request.rundb.runs.find_one(
-        {"_id": ObjectId(r_id)}, {"tasks": 1, "bad_tasks": 1, "results": 1, "args": 1}
-    )
+    run = request.rundb.get_run(request.matchdict["id"])
     if run is None:
         raise exception_response(404)
 


vdbergh:
I think the current setup was introduced to reduce the pressure on the main Fishtest instance under extreme conditions (e.g. when noob brings 150000 cores online).

ppigazzini:
But in this way it's impossible to debug the chi2.

We filter out the users with low frequencies (`del users[keys[idx]]`) but `keys `is not updated. In the last line of code we loop on the whole length of keys to set the residuals.

https://github.com/official-stockfish/fishtest/blob/9eb0dc9c6b7e0aa23d67a98fce02cc39c15aafe3/server/fishtest/util.py#L106-L143

I was simply improving the code and I found that possible bug.

https://github.com/official-stockfish/fishtest/compare/master...ppigazzini:fishtest:modern_python

ppigazzini:
> I think the current setup was introduced to reduce the pressure on the main Fishtest instance under extreme conditions (e.g. when noob brings 150000 cores online).

In this way it's nearly impossible to test the changes to the code about the chi2 and residuals.
(I deleted the previous comment because wrong)

vdbergh:
I understand. I am just saying that this was not really a bug but a conscious decision.

ppigazzini:
With https://github.com/official-stockfish/fishtest/pull/1663 the tasks information is loaded from the DB.
But the run loaded from the DB seems to have a residual out of sync respect to the pentanomial.
Every list of tasks shows:
- workers without a residual but with a number of games higher than the threshold of 5 (logging the process shows that the filtering code is working fine)
- wrong set of residuals for the set of pentanomials

Loading the run from the cache of the main fishtest instance fixes the issues, but the correct residuals are not saved back in the DB.

vdbergh:
> Loading the run from the cache of the main fishtest instance fixes the issues, but the correct residuals are not saved back in the DB (at least for the runs that I tested): we still got the wrong residuals switching back to loading the run from the DB.

I'll have a look. Perhaps there is a bug somewhere. Normally when a run is updated (e.g. when changing the residuals) it should be declared dirty (with `buffer(run)` and then it is saved to the disk after some time. Perhaps this has been forgotten here?

vdbergh:
A better solution would be to compute the residuals on the fly when viewing the list of tasks (not writing them to the DB).

The final residuals (which are written to the DB) will then be be computed when the run is finished (as is happening now).

vdbergh:
But now I actually think that the current code is nearly correct. The residuals are computed on the fly when the test is viewed but they are not written to the DB (which is correct for a GET request). The only issue is that the task list is computed independently from the test view (whether loaded from the db or the cache). So its residuals are not correct.

ppigazzini:
The chi2 and the residuals are computed on the fly on the cache of the main instance, only the chi2 and p-value is uded on the view. The tasks list with the residuals are retrieved from the DB, outdated and out of sync values.

ppigazzini:
So the code is bugged ad the DB has unnecessary and wrong residuals saved in the runs.

vdbergh:
I think residuals are only written to the db with a finished run no? I think those shown in tests_view are generated on the fly.

vdbergh:
Probably there should be something like compute_residuals which does not change any state. This can then be used for views.

ppigazzini:
We should simplify the functions limiting/avoiding side effects.

ppigazzini:
PROD updated to fix the bug. It's easy to revert with a different fix.

PR #1949, opened 2024-04-23
Handle failing make strip gracefully.

vdbergh:
https://tests.stockfishchess.org/actions?text=calledprocesserror


vdbergh:
I assume a failing `make strip` indicates a missing `strip` command. We could test for this before starting the worker (like we test for the presence of `make`).

PR #1950, opened 2024-04-23
More generic detection of toolchain issues.

vdbergh:


ppigazzini (approved):
Looks good on DEV with msys2 and linux workers.

ppigazzini:
Triggered the workers update, thank you @vdbergh

PR #1955, opened 2024-04-25
Implement bulk streaming of tasks PGNs for a run

ppigazzini:
This commit adds a feature that streams PGNs of a run, significantly reducing
the server's CPU/RAM usage. The API now returns a gzipped PGN, constructed by
chaining all the run's gzipped PGNs, which is streamed to the client.

Key changes:
1. update 'api_download_run_pgns' in api.py, which handles the request for
downloading the PGN of the run. It fetches the PGNs using the
'get_run_pgns' method from the RunDb class, and returns a gzipped response
    ```
    wget https://test.stockfishchess.org/api/run_pgns/<run_id>.pgn.gz
    ```
2. add 'GeneratorAsFileReader' class in util.py that provides a file-like
interface for a generator, enabling chanked reading
3. modify the 'get_run_pgns' method in rundb.py to return a
'GeneratorAsFileReader' object that yields each gzipped PGN file for the run

Note: the `gunzip on;` directive must be removed from the nginx configuration.

Improvement suggested by @peregrineshahin

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -13,7 +13,7 @@ from pyramid.httpexceptions import (
     HTTPUnauthorized,
     exception_response,
 )
-from pyramid.response import Response
+from pyramid.response import FileIter, Response
 from pyramid.view import exception_view_config, view_config, view_defaults
 from vtjson import ValidationError, validate
 
@@ -532,19 +532,18 @@ class ApiView(object):
 
     @view_config(route_name="api_download_run_pgns")
     def download_run_pgns(self):
-        tar_name = self.request.matchdict["id"]
-        match = re.match(r"^([a-zA-Z0-9]+)\.pgns\.tar$", tar_name)
+        pgns_name = self.request.matchdict["id"]
+        match = re.match(r"^([a-zA-Z0-9]+)\.pgn\.gz$", pgns_name)
         if not match:
             return Response("Invalid filename format", status=400)
         run_id = match.group(1)
-        pgns_zip = self.request.rundb.get_run_pgns(run_id)
-        if pgns_zip is None:
+        pgns_reader = self.request.rundb.get_run_pgns(run_id)
+        if pgns_reader is None:
             return Response("No data found", status=404)
-        zip_buffer = io.BytesIO(pgns_zip)
-        response = Response(content_type="application/x-tar")
-        response.app_iter = zip_buffer
-        response.content_length = zip_buffer.getbuffer().nbytes
-        response.headers["Content-Disposition"] = f'attachment; filename="{tar_name}"'
+        response = Response(content_type="application/gzip")
+        response.app_iter = FileIter(pgns_reader)
+        response.headers["Content-Disposition"] = f'attachment; filename="{pgns_name}"'
+        response.headers["Content-Encoding"] = "gzip"
         return response
 
     @view_config(route_name="api_download_nn")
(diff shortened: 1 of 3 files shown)


ppigazzini:
Opened to collect comments.

Firts implementation of the stream of the PGNs as suggested by @peregrineshahin on Discord, test on DEV with

```
curl -jLO https://dfts-0.pigazzini.it/api/run_pgns/662a442b6c6057d3bbd2978f.pgns
```

In the first tests on DEV

Plus:
- low CPU/RAM overhead
- high download bitrate

Minus:
- only one big file, losing the filename/datetimes information
- the tools must be updated

vondele:
I think having 1 file could be considered a feature, as the batch and time information is in the pgn headers. 
Tools can be updated quite easily, I think. @robertnurnberg  ?

I would prefer if there extension of the file was simply .pgn (or for the compressed stream .pgn.zip / .pgn.gz) as that is still correctly describing the content.

ppigazzini:
Extension now is ".pgn", commands updated.

robertnurnberg:
> I think having 1 file could be considered a feature, as the batch and time information is in the pgn headers. Tools can be updated quite easily, I think. @robertnurnberg ?
> 
> I would prefer if there extension of the file was simply .pgn (or for the compressed stream .pgn.zip / .pgn.gz) as that is still correctly describing the content.

Agreed about tools being easy to update. If there is a choice, I would prefer `pgn.gz` over `pgn.zip`.

ppigazzini:
The pgn files are stored in the DB as `pgn.gz`, the code simply streams those files. We are talking about the extension in the API `pgn` vs `pgn.gz`

ppigazzini:
@robertnurnberg @vondele PROD is running with the PR to test the tools change, I wait your approval to merge.

robertnurnberg:
So with https://github.com/official-stockfish/WDL_model/pull/185 the download script seems to work fine on my end.

robertnurnberg:
Btw, the  `"Content-Length"` entry seems to be no longer present in the header. (It was for the tar downloads.)

If there is an easy fix no the server side to add that, it would be great. :)

Disservin:
i merged the wdl model change, so i guess you can merge this or is the `Content-Length` a show stopper?

vondele:
LGTM.

robertnurnberg:
Ok, thanks. Having the content length is not strictly necessary. We have the number of games as a proxy for the file length for now.

peregrineshahin:
wouldn't this work?
```
# Inside the loop where you download the files
url = "https://tests.stockfishchess.org/api/run_pgns/" + test + ".pgn.gz"
try:
    response = urllib.request.urlopen(url)
    tmpName = path + test + ".tmp"
    urllib.request.urlretrieve(url, tmpName)

    # Get the file size after downloading
    file_size = os.path.getsize(tmpName)
    file_size_formatted = format_large_number(file_size) + "B"

    if args.verbose >= 2:
        print(f"Downloaded file size: {file_size_formatted}")

    # Rename the temp file
    os.rename(tmpName, path + test + ".pgn.gz")
```

peregrineshahin:
Ah you want to know the size before downloading begins, sorry about the confusion

peregrineshahin:
> We could try to update the Content-Length in the collection when uploading the worker PGN, but we must be sure that that work for active runs as well: it needs some tests and a new PR.
I think a `size` field for the database pgn entry is reasonable.

PR #1957, opened 2024-04-27
Add "Content-Length" header for run PGN

ppigazzini:
Improve https://github.com/official-stockfish/fishtest/pull/1955

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -537,13 +537,14 @@ class ApiView(object):
         if not match:
             return Response("Invalid filename format", status=400)
         run_id = match.group(1)
-        pgns_reader = self.request.rundb.get_run_pgns(run_id)
+        pgns_reader, total_size = self.request.rundb.get_run_pgns(run_id)
         if pgns_reader is None:
             return Response("No data found", status=404)
         response = Response(content_type="application/gzip")
         response.app_iter = FileIter(pgns_reader)
         response.headers["Content-Disposition"] = f'attachment; filename="{pgns_name}"'
         response.headers["Content-Encoding"] = "gzip"
+        response.headers["Content-Length"] = str(total_size)
         return response
 
     @view_config(route_name="api_download_nn")
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -257,22 +257,33 @@ class RunDb:
         return self.runs.insert_one(new_run).inserted_id
 
     def upload_pgn(self, run_id, pgn_zip):
-        self.pgndb.insert_one({"run_id": run_id, "pgn_zip": Binary(pgn_zip)})
+        self.pgndb.insert_one(
+            {"run_id": run_id, "pgn_zip": Binary(pgn_zip), "size": len(pgn_zip)}
+        )
         return {}
 
     def get_pgn(self, run_id):
         pgn = self.pgndb.find_one({"run_id": run_id})
-        if pgn:
-            return pgn["pgn_zip"]
-        return None
+        return pgn["pgn_zip"] if pgn else None
 
     def get_run_pgns(self, run_id):
-        pgns = self.pgndb.find({"run_id": {"$regex": f"^{run_id}"}})
-        if pgns is not None:
-            # Create a generator that yields each pgn.gz file
-            pgn_generator = (pgn["pgn_zip"] for pgn in pgns)
-            return GeneratorAsFileReader(pgn_generator)
-        return None
+        # Compute the total size using MongoDB's aggregation framework
+        total_size_agg = self.pgndb.aggregate(
+            [
+                {"$match": {"run_id": {"$regex": f"^{run_id}-\\d+"}}},
+                {"$group": {"_id": None, "totalSize": {"$sum": "$size"}}},
+            ]
+        )
+        total_size = total_size_agg.next()["totalSize"] if total_size_agg.alive else 0
+
+        if total_size > 0:
+            # Find the pgns and create a generator that yields each pgn.gz file
+            pgns = self.pgndb.find({"run_id": {"$regex": f"^{run_id}-\\d+"}})
+            pgns_reader = GeneratorAsFileReader(pgn["pgn_zip"] for pgn in pgns)
+        else:
+            pgns_reader = None
+
+        return pgns_reader, total_size
 
     def write_nn(self, net):
         validate(nn_schema, net, "net")


ppigazzini:
```
usr00@pp-wrk-00:~/_tests_size$ wget https://dfts-0.pigazzini.it/api/run_pgns/662cbc0ce58b87631ba593c3.pgn.gz
--2024-04-27 11:03:29--  https://dfts-0.pigazzini.it/api/run_pgns/662cbc0ce58b87631ba593c3.pgn.gz
Resolving dfts-0.pigazzini.it (dfts-0.pigazzini.it)... 94.198.98.133
Connecting to dfts-0.pigazzini.it (dfts-0.pigazzini.it)|94.198.98.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 2414993 (2.3M) [application/gzip]
Saving to: ‘662cbc0ce58b87631ba593c3.pgn.gz’

662cbc0ce58b87631ba593c3.pgn. 100%[=================================================>]   2.30M  2.71MB/s    in 0.9s

2024-04-27 11:03:30 (2.71 MB/s) - ‘662cbc0ce58b87631ba593c3.pgn.gz’ saved [2414993/2414993]

usr00@pp-wrk-00:~/_test_size$ ls -la
total 2368
drwxr-xr-x  2 usr00 usr00    4096 Apr 27 11:03 .
drwxr-x--- 40 usr00 usr00    4096 Apr 27 10:53 ..
-rw-r--r--  1 usr00 usr00 2414993 Apr 27 11:03 662cbc0ce58b87631ba593c3.pgn.gz
```

peregrineshahin:
That's exactly what I had in mind.. but I wonder how much time the aggregation takes..

peregrineshahin:
What I was thinking too, is probably it would be better schema wise.. is if we have two fields.. one refers to the pgn id and one refers to the run id. In such way we could ditch the dependence on regexes and benefit from two indexing.
But Im not sure how slow is current query.

ppigazzini:
> What I was thinking too, is probably it would be better schema wise.. is if we have two fields.. one refers to the pgn id and one refers to the run id. In such way we could ditch the dependence on regexes and benefit from two indexing. But Im not sure how slow is current query.

```
index 524f555..872d553 100644
@@ -274,12 +274,14 @@ class RunDb:
             # Create a generator that yields each pgn.gz file
             pgn_generator = (pgn["pgn_zip"] for pgn in pgns)
             # Compute the total size using MongoDB's aggregation framework
+            t0 = time.time()
             total_size = self.pgndb.aggregate(
                 [
                     {"$match": {"run_id": {"$regex": f"^{run_id}"}}},
                     {"$group": {"_id": None, "totalSize": {"$sum": "$size"}}},
                 ]
             ).next()["totalSize"]
+            print(time.time() - t0, flush=True)
             return GeneratorAsFileReader(pgn_generator), total_size
         return None, 0
```


```
wget https://dfts-0.pigazzini.it/api/run_pgns/662a442b6c6057d3bbd2978f.pgn.gz
--2024-04-27 11:27:19--  https://dfts-0.pigazzini.it/api/run_pgns/662a442b6c6057d3bbd2978f.pgn.gz
Resolving dfts-0.pigazzini.it (dfts-0.pigazzini.it)... 94.198.98.133
Connecting to dfts-0.pigazzini.it (dfts-0.pigazzini.it)|94.198.98.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 204137873 (195M) [application/gzip]
Saving to: ‘662a442b6c6057d3bbd2978f.pgn.gz’

662a442b6c6057d3bbd2978f.pgn. 100%[=================================================>] 194.68M  12.0MB/s    in 17s

2024-04-27 11:27:37 (11.2 MB/s) - ‘662a442b6c6057d3bbd2978f.pgn.gz’ saved [204137873/204137873]
```


```
Apr 27 11:27:15 dfts-0 systemd[1]: Stopped Fishtest Server port 6545.
Apr 27 11:27:15 dfts-0 systemd[1]: Started Fishtest Server port 6545.
Apr 27 11:27:16 dfts-0 pserve[4933]: Starting server in PID 4933.
Apr 27 11:27:20 dfts-0 pserve[4933]: 0.00289154052734375
```

peregrineshahin:
Makes sense, so they are naturally indexed in a way, also the aggregation seems to do a great minimalistic job! Yhanks for the reply

ppigazzini:
Companion script
```python

def add_size_key(rundb):
    pgns_without_size = rundb.pgndb.find({"size": {"$exists": False}})

    for pgn in pgns_without_size:
        size = len(pgn["pgn_zip"])
        rundb.pgndb.update_one({"_id": pgn["_id"]}, {"$set": {"size": size}})

def main():
    rundb = RunDb()
    add_size_key(rundb)

if __name__ == "__main__":
    main()
```

ppigazzini:
> Makes sense, so they are naturally indexed in a way, also the aggregation seems to do a great minimalistic job! Thanks for the reply

MMMM, the pgns collection on DEV is nearly empty, we need to test the time with the PROD collection...

ppigazzini:
Aggregation time computed on PROD, the first has 17 tasks/7k games, the second has 438 tasks/339k games
```
Apr 27 11:04:46 tests.stockfishchess.org pserve[21401]: Aggregation time: 0.03008127212524414
Apr 27 11:05:53 tests.stockfishchess.org pserve[21401]: Aggregation time: 1.187739372253418
```

ppigazzini:
@robertnurnberg PROD updated by hand, it's ready to be tested with your tool.

vondele:
Works:
```
starting download at  Sa 27 Apr 2024 13:27:03 CEST
Found 6986 downloaded tests in ./pgns/ already.
Collecting meta data for test 662574a93fe04ce4cefc3afc ...
Downloading 128MB .pgn.gz file with 123810 games (WDL = 31738 60743 31329) to ./pgns/24-04-21/662574a93fe04ce4cefc3afc/ ...
Collecting meta data for test 6627b71e3fe04ce4cefc60e0 ...
Downloading 27MB .pgn.gz file with 26484 games (WDL = 6722 13081 6681) to ./pgns/24-04-23/6627b71e3fe04ce4cefc60e0/ ...
```
matching the download size 
```
$ ls -tlr ./pgns/24-04-21/662574a93fe04ce4cefc3afc/662574a93fe04ce4cefc3afc.pgn.gz 
-rw-rw-r-- 1 vondele vondele 128237134 Apr 27 13:27 ./pgns/24-04-21/662574a93fe04ce4cefc3afc/662574a93fe04ce4cefc3afc.pgn.gz
```

robertnurnberg:
Works perfectly also here. Thanks a lot. @ppigazzini

ppigazzini:
PROD updated.

vdbergh:
I think run_id is not really the correct name for the field since it seems to also contain the task id (so it would need a different regex). Perhaps we can use two fields run_id+task_id as is done elsewhere in fishtest?

peregrineshahin:
> I think run_id is not really the correct name for the field since it seems to also contain the task id (so it would need a different regex). Perhaps we can use two fields run_id+task_id as is done elsewhere in fishtest?

Well that's what I said here, but probably it's too much work to convert old pgns
```
What I was thinking too, is probably it would be better schema wise.. is if we have two fields.. one refers to the pgn id and one refers to the run id. In such way we could ditch the dependence on regexes and benefit from two indexing.
But Im not sure how slow is current query.
```

vdbergh:
> Well that's what I said here, but probably it's too much work to convert old pgns

I guess that would be the case. Adding a new regex `run_id_pgns` is not the end of the world.

peregrineshahin:
Not sure I understood, yes I think current situation is okay.. but if you refer to your suggestion then you don't need regexes at all as the query will be strict equality.

> I guess that would be the case. Adding a new regex `run_id_pgns` is not the end of the world.

ppigazzini:
`run_id` is wrong (the first time I worked on that code it was confusing), `pgn_zip` is inaccurate as well (`pgn_gz` is better).

PR #1958, opened 2024-04-27
Update memory estimates

vondele:
after the various net updates the estimated memory need was no longer accurate also add the estimated size of cutechess-cli with the larger book


vondele:
looks like the black version I used is no longer the one used here.

ppigazzini:
> looks like the black version I used is no longer the one used here.

I will format the code when testing the PR.

ppigazzini (approved):
Looks good on DEV + workers

ppigazzini:
Triggered the workers update, thank you @vondele

PR #1959, opened 2024-04-27
Improve the queries of the run PGNs

ppigazzini:


peregrineshahin:
Yes I always thought that specifying the outfields can bring quite a nice speedup.. not sure about mongodb 100%

ppigazzini:
In this case it's a measured speedup, the code now skips the big PGNs for the total size count.

PR #1963, opened 2024-04-28
Force mime type of pgn_zip to be application/gzip in schema.

vdbergh:
Requires upgrading the` vtjson` package (which now has the `magic(mime_type)` validation method).

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -25,6 +25,7 @@ from vtjson import (
     ip_address,
     keys,
     lax,
+    magic,
     nothing,
     number,
     one_of,
@@ -56,18 +57,27 @@ epd_file = glob("*.epd", name="epd_file")
 pgn_file = glob("*.pgn", name="pgn_file")
 even = div(2, name="even")
 datetime_utc = intersect(datetime, fields({"tzinfo": timezone.utc}))
+gzip_data = magic("application/gzip", name="gzip_data")
 
 uint = intersect(int, interval(0, ...))
 suint = intersect(int, interval(1, ...))
 ufloat = intersect(float, interval(0.0, ...))
 unumber = intersect(number, interval(0, ...))
 
-pgns_schema = {
-    "_id?": ObjectId,
-    "run_id": run_id_pgns,
-    "pgn_zip": Binary,
-    "size": uint,
-}
+
+def size_is_length(x):
+    return x["size"] == len(x["pgn_zip"])
+
+
+pgns_schema = intersect(
+    {
+        "_id?": ObjectId,
+        "run_id": run_id_pgns,
+        "pgn_zip": intersect(Binary, gzip_data),
+        "size": uint,
+    },
+    size_is_length,
+)
 
 user_schema = {
     "_id?": ObjectId,
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -4,7 +4,7 @@ import io
 import re
 from datetime import datetime, timezone
 
-from fishtest.schemas import api_access_schema, api_schema
+from fishtest.schemas import api_access_schema, api_schema, gzip_data
 from fishtest.stats.stat_util import SPRT_elo, get_elo
 from fishtest.util import worker_name
 from pyramid.httpexceptions import (
@@ -509,9 +509,14 @@ class ApiView(object):
     @view_config(route_name="api_upload_pgn")
     def upload_pgn(self):
         self.validate_request("/api/upload_pgn")
+        try:
+            pgn_zip = base64.b64decode(self.pgn())
+            validate(gzip_data, pgn_zip, "pgn")
+        except Exception as e:
+            self.handle_error(str(e))
         result = self.request.rundb.upload_pgn(
             run_id="{}-{}".format(self.run_id(), self.task_id()),
-            pgn_zip=base64.b64decode(self.pgn()),
+            pgn_zip=pgn_zip,
         )
         return self.add_time(result)
 


server/fishtest/schemas.py
```diff
@@ -56,18 +57,27 @@
 pgn_file = glob("*.pgn", name="pgn_file")
 even = div(2, name="even")
 datetime_utc = intersect(datetime, fields({"tzinfo": timezone.utc}))
+zlib_data = intersect(Binary, magic("application/zlib", name="zlib_data"))
```
ppigazzini, line 60: I wonder if `application/gzip` / `gzip_data` are more appropriate than `application/zlib` / `zlib_data`
We switched from zlib to gzip with https://github.com/official-stockfish/fishtest/pull/1823

vdbergh: https://datatracker.ietf.org/doc/html/rfc6713 says application/gzip is a file format and application/zlib is a stream format. They use the same compression algorithm. Very confusing…

ppigazzini: gzip, zlib, zip are all using the same compressing algorithm. The problem with zlib is that you need some code to decompress the binary blob.
gzip has some nice propreties:
1. can be cat `cat file0.gz file1.gz > files.gz`. I'm using this to stream all the run's PGNs.gz in only one PGN.gz
2. the browser decompress a gzipped file on the fly (nginx can be configured to decompress for a client unable to decompress)

vdbergh: Yes but the data in pgnsdb is application/zlib. I cannot say that it is application/gzip when it isn’t…

ppigazzini: I wonder if that is due to the `Binary(pgn_zip)`

vdbergh: I am going to double check.

vdbergh: I double checked. Here is the code:
```
import magic
import pymongo

if __name__ == "__main__":
    client = pymongo.MongoClient()
    pgnsdb = client["fishtest_new"]["pgns"]
    pgns = pgnsdb.find({})
    for pgn in pgns:
        zip = pgn["pgn_zip"]
        mime = magic.from_buffer(zip, mime=True)
        magic_ = magic.from_buffer(zip)
        print(mime, magic_)
```

ppigazzini: https://github.com/official-stockfish/fishtest/blob/e6254159efefc88824fdee19e0f4508afd422626/worker/worker.py#L1456-L1481

vdbergh: I looked at the `pgn_zip` data and the first byte is `0x78` which according to https://datatracker.ietf.org/doc/html/rfc6713 is part of the magic for `application/zlib`. `application/gzip` starts with `0x1f`. Mysterious.

ppigazzini: `gzip` python module bugs? I was doing this to recover the datetime because the function was not implemented.
```python
                        # Extract and convert the 4 bytes starting at index 4
                        tarinfo.mtime = int.from_bytes(pgn_zip[4:8], byteorder="little")
                        tarf.addfile(tarinfo, io.BytesIO(pgn_zip))
```

vdbergh: I now saved the "pgn_zip" data to files and and inspected it with the "file" utility. It still says `application/zlib`. I think my pgn collection contains recent pgns, but just to be sure, perhaps you could also look at it on DEV or PROD?

vdbergh:
Ah wait. Perhaps it is due to the fact that for my local testing I use an old worker… Going to check again…

ppigazzini:
On DEV
```
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-623.pgn", last modified: Sat Apr 27 19:37:23 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-624.pgn", last modified: Sat Apr 27 19:48:22 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-625.pgn", last modified: Sat Apr 27 20:53:44 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-626.pgn", last modified: Sat Apr 27 21:05:36 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-627.pgn", last modified: Sat Apr 27 21:20:15 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-628.pgn", last modified: Sat Apr 27 21:25:53 2024, max compression

```

server/fishtest/schemas.py
ppigazzini: On DEV is `application/gzip` running your script:
```
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-623.pgn", last modified: Sat Apr 27 19:37:23 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-624.pgn", last modified: Sat Apr 27 19:48:22 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-625.pgn", last modified: Sat Apr 27 20:53:44 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-626.pgn", last modified: Sat Apr 27 21:05:36 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-627.pgn", last modified: Sat Apr 27 21:20:15 2024, max compression
application/gzip gzip compressed data, was "662a442b6c6057d3bbd2978f-628.pgn", last modified: Sat Apr 27 21:25:53 2024, max compression

```

vdbergh:
Yes it was the worker. So we should also enforce the correct mime type on the api level.

ppigazzini:
https://github.com/official-stockfish/fishtest/pull/1823#issuecomment-1747511591

Here you find that script that I used to convert the PROD pgns collection form zlib to gzip. You need to add the size key as well.

vdbergh:
Perhaps using similar scripts we can change run_id to run_id + task_id and pgn_zip to pgn_gzip.

ppigazzini:
Of course, it's easy. At the time I was tempted to switch to `run_task_id`

vdbergh:
In the latest push I switched to `application/gzip` and I also enforce this mime type on the api level.

ppigazzini:
Testing on DEV with some workers, away from the keyboard for one hour.

ppigazzini:
```diff
diff --git a/server/fishtest/api.py b/server/fishtest/api.py
index 65ab52b..e4584e1 100644
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -511,6 +511,7 @@ class ApiView(object):
         self.validate_request("/api/upload_pgn")
         try:
             pgn_zip = base64.b64decode(self.pgn())
+            pgn_zip = self.pgn()
             validate(gzip_data, pgn_zip, "pgn")
         except Exception as e:
             self.handle_error(str(e))
```

```
Apr 28 20:01:34 dfts-0 pserve[8436]: /api/upload_pgn: pgn (value:'H4sICP2OLmYC/zY2MmE0NDJiNmM2MDU3ZDNiYmQyOTc4Zi02NTAucGduAOy9XbPkyHEl+J6/4ppeU7cK8YmAzPSSI+0+pDGnMqW...[TRUNCATED]...') is not of type 'gzip_data': 'text/plain' is differen
```

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1964, opened 2024-04-28
Fix SPSA iteration mismatch

peregrineshahin:
after updating the number of games the iterations completion can be more than 100% 75k/60k this is because SPSA num of games should not be modifiable after creation
image (1)
See Also: https://discord.com/channels/435943710472011776/900771437177094156/1234042155346169937


ppigazzini (approved):
Priority/throughput change is still working in SPSA.

image

ppigazzini:

PR #1966, opened 2024-04-28
Drop "Content-Encoding" header for run PGNs

ppigazzini:
A client able to decompress on the fly the gzip content stops the download after extracting the first PGN (it expectes to download one file) ignoring the "Content-Length" header.

Drop the "Content-Encoding" header to complete the download of all the PGNs.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -542,7 +542,6 @@ class ApiView(object):
         response = Response(content_type="application/gzip")
         response.app_iter = FileIter(pgns_reader)
         response.headers["Content-Disposition"] = f'attachment; filename="{pgns_name}"'
-        response.headers["Content-Encoding"] = "gzip"
         response.headers["Content-Length"] = str(total_size)
         return response
 


vdbergh:
Is there a standard for this? Stopping or not after the first file may be browser dependent.

vdbergh:
Deleted my comment since I misunderstood the purpose of this PR. You want to force the client not to decompress the data on the fly.

ppigazzini:
`curl --compressed` stops as well. I figured out that with the header we communicate the download of one file, when a client is able to decompress it stop after the first task PGN. Perhaps there is a better solution.

peregrineshahin:
looks like a good solution to me.

> `curl --compressed` stops as well. I figured out that with the header we communicate the download of one file, when a client is able to decompress it stops after the first task PGN is decompressed ignoring the content length header. Perhaps there is a better solution.

ppigazzini:
With this PR we can re add the `gunzip on;` directive to have `nginx` decompressing on the fly the single PGN for a client unable to decompress.

vdbergh:
Does this PR imply that the data will no longer be compressed when the download is over?

ppigazzini:
No, simply the browser download the whole "pgn.gz"

PR #1967, opened 2024-04-28
Use JavaScript to download the run PGNs

ppigazzini:
Deal in proper way when the run has not PGNs (yet or anymore)

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -237,7 +237,7 @@
 
         % if approver:
           <div class="col-12 col-sm">
-            <a href="/api/run_pgns/${run['_id']}.pgn.gz" class="btn btn-primary text-nowrap w-100">Download games</a>
+            <button id="gamesDownloadButton" class="btn btn-primary text-nowrap w-100">Download games</button>
           </div>
         % endif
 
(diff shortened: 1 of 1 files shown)


server/fishtest/templates/tests_view.mak
```diff
@@ -860,3 +860,27 @@
     % endif
     });
 </script>
+<script>
+    document.getElementById('gamesDownloadButton').addEventListener('click', function() {
```
peregrineshahin, line 864: Nice! one issue is that this will cause a run-time if the button doesn't exist (i.e. approver is false).. in such cases we use `?.addEventListener` in the code base
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

ppigazzini:
```js
const response = await fetch('/api/run_pgns/${run['_id']}.pgn.gz');
```
I wonder if the use of the quote is consistent with our codebase.

peregrineshahin: I don't think so, prettier default is double quotes.

server/fishtest/templates/tests_view.mak
```diff
@@ -860,3 +860,27 @@
...
+        const blob = await response.blob();
+        const url = window.URL.createObjectURL(blob);
+        const a = document.createElement("a");
+        a.href = url;
+        a.download = `${run["_id"]}.pgn.gz`;
+        document.body.appendChild(a);
+        a.click();
+        document.body.removeChild(a);
+      } catch (error) {
+        alert(error.message);
```
peregrineshahin, line 881: we don't use alert, as it is non css customisable (not part of the DOM) and quite old way to alert something to the user
probably console.error is better here.

PR #1969, opened 2024-04-29
use the custom `alertError` for errors we want to show to the user

peregrineshahin:
This removes the old system-dependent `alert()` usage

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -237,7 +237,10 @@
 
         % if approver:
           <div class="col-12 col-sm">
-            <button id="gamesDownloadButton" class="btn btn-primary text-nowrap w-100">Download games</button>
+            <button 
+              id="download_games"
+              class="btn btn-primary text-nowrap w-100"
+            >Download games</button>
           </div>
         % endif
 
@@ -461,6 +464,7 @@
 ></script>
 
 <script>
+  document.title = "${page_title} | Stockfish Testing";
   let cookieTheme = getCookie("theme");
   const currentTime = new Date().getTime();
   const oneDayAgo = currentTime - (24 * 60 * 60 * 1000); // 24 hours ago in milliseconds
@@ -514,10 +518,40 @@
   document
     .getElementById("moon")
     .addEventListener("click", () => setHighlightTheme("dark"));
-</script>
 
-<script>
-  document.title = "${page_title} | Stockfish Testing";
+  async function downloadPGNs() {
+    try {
+      const response = await fetch(`/api/run_pgns/${run["_id"]}.pgn.gz`);
+      if (!response.ok) {
+        if (response.status === 404) {
+          alertError("No games found for this run");
+        } else {
+          alertError("Unable to download PGNs");
+        }
+        return;
+      }
+      const blob = await response.blob();
+
+      // Check if the blob is empty
+      if (blob.size === 0) {
+        alertError("Blob is empty!");
+        return;
+      }
+      
+      const url = window.URL.createObjectURL(blob);
+      const a = document.createElement("a");
+      a.href = url;
+      a.download = `${run["_id"]}.pgn.gz`;
+      document.body.append(a);
+      a.click();
+      document.body.removeChild(a);
+    } catch (error) {
+      alertError("Network error: please check your network!");
+    }
+  }
+
+  const downloadPGNsButton = document.getElementById("download_games");
+  downloadPGNsButton?.addEventListener("click", downloadPGNs);
 
   let fetchedTasksBefore = false;
   async function handleRenderTasks(){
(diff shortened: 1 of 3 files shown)


server/fishtest/templates/tests_view.mak
```diff
@@ -514,10 +518,41 @@
...
+        } else {
+          alertError("Unable to download PGNs");
+        }
+        return;
+      }
+      const blob = await response.blob();
+
+      / Check if the blob is empty
+      if (blob.size === 0) {
+        alertError("No games found for this run");
```
ppigazzini, line 537: IMO the message is misleading. If the total size of the PGNs is zero (no games found) the API response is 404, so the blob size zero has another cause.

https://github.com/official-stockfish/fishtest/blob/0893ce71a4999953d1976fedae1aaaf0a39ee3b7/server/fishtest/rundb.py#L291-L298

https://github.com/official-stockfish/fishtest/blob/0893ce71a4999953d1976fedae1aaaf0a39ee3b7/server/fishtest/api.py#L541-L547

server/fishtest/templates/tests_view.mak
```diff
@@ -514,10 +518,41 @@
...
+        if (response.status === 404) {
+          alertError("No games found for this run");
+        } else {
+          alertError("Unable to download PGNs");
+        }
+        return;
+      }
+      const blob = await response.blob();
+
+      / Check if the blob is empty
```
ppigazzini, line 535:
```diff
- / Check if the blob is empty
+ // Check if the blob is empty
```

server/fishtest/templates/tests_view.mak
peregrineshahin: Well I didn't get a blob size of zero, it was only error handling if the server condition was relaxed at some point.
I guess I can change the message.

server/fishtest/templates/tests_view.mak
```diff
@@ -514,10 +518,41 @@
...
+      a.click();
+      document.body.removeChild(a);
+    } catch (error) {
+      alertError("Network error: please check your network!");
+    }
+  }
+
+  document
+    .getElementById("download_games")?
+    .addEventListener("click", downloadPGNs);
```
ppigazzini, line 555: I copied the whole JS cose in a file to catch a couple of errors shown by the browser
```js
prettier --write aaaaa.js
aaaaa.js
[error] aaaaa.js: SyntaxError: Unexpected token (89:5)
[error]   87 |   document
[error]   88 |     .getElementById("download_games")?
[error] > 89 |     .addEventListener("click", downloadPGNs);
[error]      |     ^
[error]   90 |
[error]   91 |   let fetchedTasksBefore = false;
[error]   92 |   async function handleRenderTasks(){
```

prettier does not break that line
```js
document.getElementById("download_games")?.addEventListener("click", downloadPGNs);
```

I don't know the fix for this one:
```js
prettier --write aaaaa.js
aaaaa.js
[error] aaaaa.js: SyntaxError: Unexpected token, expected ")" (96:11)
[error]   94 |       await toggleTasks();
[error]   95 |     })
[error] > 96 |      if (${str(tasks_shown).lower()})
[error]      |           ^
[error]   97 |        await renderTasks();
[error]   98 |   }
[error]   99 |
```

peregrineshahin: the other one is not an error, it's only a valid syntax inside the mako, the code inside `${}` is a python code.

BTW my prettier here does break it.. I think 80 characters is default for you no?
https://prettier.io/playground/#N4Igxg9gdgLgprEAuEATCYCuBbBMB0A5nDAKIA2cusAQgJ4CSqAFADpoQDuU5EAhqgD6hPrgDO7AJQB+fANSkAbngAyASzHwocAE5tw5NWADW7ADQACdN14CACgHEAcmMkBuEGZAQADjDXQYsigfDo6XHahCEEofOScfHRBXgBGOnwmJADKonDq2sgAZnFicKnpmTBZPhlqUITIMDqYZSCl2GqNza1wAB4+umrUMHEAKrpQoWpwMcXkpV5idYSUAIqYEPBFJa0AVmK9WctrG1tIcwsgAI6ncHbhPjEgfGIAtNpwqJ+eIE18aoZ6gBhCDYbB8ZDPcjkH5LeqUACCMCaahSmHgdl0+Tg23mrQAFjBsOQAOr4tTwMQ1MBwLLRClqRQUuiQsBiZIgRQtBhQL6wLJgHRqPwI3lZGB0Si4y4+cKlEnpHyQ2UzXTKH51Uo6GD3PiEcHS1o1HRayEpPgpOAwryyuowElqVAwfHIAAcAAYvDo4Dc1N7dfqIecdl4RikHU6XUgAExeTClUYW2YhkBUS2oL6oFR8eqYPVwABiEB04ORy0hfHREBAAF8a0A

ppigazzini: I got the errors in the browser, then I copied the code to find the issue with the formatting.
The first point was an issue with the spaces/tabs perhaps.
The error message in the browser for the second point is disappeared now in the browser.

ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

ppigazzini:
BTW the javascript embedded in the mako template must be reload in the browser.
The javascript from a file in the `js` folder is automatically reloaded courtesy of the `cache_busters`.

https://github.com/official-stockfish/fishtest/blob/027a5b6590125a12278f4c1a316ac16a43bedc80/server/fishtest/__init__.py#L28-L42

https://github.com/official-stockfish/fishtest/blob/027a5b6590125a12278f4c1a316ac16a43bedc80/server/fishtest/templates/base.mak#L66-L70

ppigazzini:
Message deleted, the js in mako should be refreshed.

PR #1970, opened 2024-04-30
Do not update residuals in tests_view.

vdbergh:
tests_view is a GET request. It should not change the server state.

We now update the residuals
- after finishing a task;
- after a task failed;
- when a run is finished.

IMHO the former lack of regular updates of the residuals was also the cause of the issue #1948.

Not seriously tested (I lack the resources to generate enough tasks to have meaningful residuals).

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1306,6 +1306,8 @@ After fixing the issues you can unblock the worker at
             # done by stop_run.
             ret = {"task_alive": False}
         else:
+            if not task["active"]:
+                update_residuals(run["tasks"])
             self.buffer(run, False)
             ret = {"task_alive": task["active"]}
 
@@ -1322,6 +1324,7 @@ After fixing the issues you can unblock the worker at
         # Mark the task as inactive.
         task["active"] = False
         self.handle_crash_or_time(run, task_id)
+        update_residuals(run["tasks"])
         self.buffer(run, False)
         print(
             "Failed_task: failure for: https://tests.stockfishchess.org/tests/view/{}, "
@@ -1347,6 +1350,7 @@ After fixing the issues you can unblock the worker at
         """
         self.clear_params(run_id)  # spsa stuff
         run = self.get_run(run_id)
+        update_residuals(run["tasks"])
         for task in run["tasks"]:
             task["active"] = False
         results = run["results"]
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -20,7 +20,6 @@ from fishtest.util import (
     get_hash,
     get_tc_ratio,
     password_strength,
-    update_residuals,
 )
 from pyramid.httpexceptions import HTTPFound, exception_response
 from pyramid.security import forget, remember
@@ -1565,7 +1564,6 @@ def tests_view(request):
         task.setdefault("last_updated", datetime.min.replace(tzinfo=timezone.utc))
 
     chi2 = get_chi2(run["tasks"])
-    update_residuals(run["tasks"], cached_chi2=chi2)
 
     try:
         show_task = int(request.params.get("show_task", -1))


vondele:
Is there a risk of quadratic behavior? Say, N workers working on N tasks, leaving at the same time ...

vdbergh:
Why quadratic? Every worker has a  task. When that task finishes, the residuals are updated.

One may consider throthling updates but I don't really know if updating the residuals is very CPU intensive (the heavy lifting is done by numpy which is C).

vondele:
update_residuals goes over all tasks, so that's why I assume it is quadratic behavior.

vdbergh:
I don't think so. It goes over the tasks in a certain run. That number of tasks is not proportional to the number of workers.

vondele:
Workers are usually evenly distributed over the runs, so the number of tasks is proportional to the number of workers.

vdbergh:
The number of active tasks in a run is proportional to the number of workers. But we are iterating over all tasks in a run, which is not.

peregrineshahin:
Can't we ditch the whole residual thingy and never save colors in the database?
I never like the fact that colors are saved in the DB..
we can do stuff at frontend level, and when purging calculate the residual

vdbergh:
What is `active tasks + cst`? The number of active tasks is usually only a small fraction of the total number of tasks.

vondele:
> What is `active tasks + cst`? The number of active tasks is usually only a small fraction of the total number of tasks.

tasks = active tasks + cst

after which the cst (even if large) is simply ignored for the scaling workers argument. ... this means there is a large linear term, in addition to the quadratic term.

peregrineshahin:
> Can't we ditch the whole residual thingy and never save colors in the database? I never like the fact that colors are saved in the DB.. we can do stuff at frontend level, and when purging calculate the residual

i.e. the function does not do something that needs to be done right right now

vondele:
`cst = tasks - active tasks` (just some constant that captures the non active tasks)

vdbergh:
Why is this a constant? The total number of tasks is determined by statistics, not by the number of workers.

vondele:
the cst is the part that I consider not important, it can be any positive number. At the point of a fleet working with N workers for that run, there with be an N**2  operation (if I understand the code correctly).

vdbergh:
I think it is proportional to `N*constant` where constant is the average number of tasks in a run.

vondele:
right, but in your formula  `constant` is always larger than your `N` (i.e. number of tasks is always larger than number of active tasks?).

peregrineshahin:
If chi2 accessable in the mako, why do we have a stateful system at all?

```python
def update_residuals(tasks, cached_chi2=None):
    # If we have an up-to-date result of get_chi2() we can pass
    # it as cached_chi2 to avoid needless recomputation.
    chi2 = get_chi2(tasks) if cached_chi2 is None else cached_chi2
    residuals = chi2["residual"]

    for task in tasks:
        if "bad" in task:
            continue
        if "worker_info" not in task:
            continue
        task["residual"] = residuals.get(
            task["worker_info"]["unique_key"], float("inf")
        )

        if crash_or_time(task):
            task["residual"] = 10.0
            task["residual_color"] = "#FF6A6A"
        elif abs(task["residual"]) < chi2["z_95"]:
            task["residual_color"] = "#44EB44"
        elif abs(task["residual"]) < chi2["z_99"]:
            task["residual_color"] = "yellow"
        else:
            task["residual_color"] = "#FF6A6A"
```

vdbergh:
> right, but in your formula `constant` is always larger than your `N` (i.e. number of tasks is always larger than number of active tasks?). (My N stands for the number of workers for that run, not for the total number of workers)

Yes but the constant is much larger than N. I am not sure if stage is ever reached that N overtakes `constant`. And then probably the run will be finished so quickly that it doesn't matter.

E.g. https://tests.stockfishchess.org/tests/view/662ecb69e1ff56336c021f60 At this moment > 300 tasks and only a few active.

vdbergh:
@peregrineshahin Yes I just noticed that we actually already have a stateless computation of the residuals. I think it would be better to use that in the views. The code snippet you link to could be moved to the mako template (without of course touching actual task list).

vdbergh:
I am closing this for now.

vondele:
> Yes but the constant is much larger than N. I am not sure if stage is ever reached that N overtakes `constant`. And then probably the run will be finished so quickly that it doesn't matter.

what I try to express is that when a fleet with M workers leaves, the server will be doing O(M**2) work, which could be problematic.

peregrineshahin:
Yes but beware, that if for example we depended on residual being stateful and >= 10 then in purging the calculation needs to be redone

> @peregrineshahin Yes I just noticed that we actually already have a stateless computation of the residuals. I think it would be better to use that in the views. The code snippet you link to could be moved to the mako template (without of course touching actual task list).

vdbergh:
I think for now we need to save residuals in the db because  the purging code depends on it (I am afraid to touch that code since it is hard to test). However we only need to update the db after a run is finished (which is much less work that what this PR proposed). For the views we can use chi2.

peregrineshahin:
Probably works also..

vdbergh:
Probably will need several PR's to untangle this mess...

PR #1977, opened 2024-05-02
Do not touch task["last_updated"] in tests_view.

vdbergh:
I think that this line is actually a noop since it is only invoked if the last_updated field is not present, which cannot happen.

Not tested...


ppigazzini:
Code added in 2013 with "fixes" commit https://github.com/official-stockfish/fishtest/commit/6dc25a78f08e91e4c74b31c6f17c9a91a01353fe probably to update on the fly to a new scheme the `task` object.

ppigazzini (approved):
Looks good with DEV + workers

PR #1980, opened 2024-05-04
Enhance UX for creating tests.

peregrineshahin:
This is a suggestion by @Vizvezdenec .. sometimes the user forgets that they need to pick the right TC before editing base branch and base signature, which if they did and then pick the wanted TC then the values they wrote/copied from another place will be overwritten, this solves this by marking `LTC, STC SMP, LTC SMP, VLTC, VLTC SMP` as `don't care`.

### DIFF
--- a/server/fishtest/templates/tests_run.mak
+++ b/server/fishtest/templates/tests_run.mak
@@ -87,9 +87,7 @@
                         "options": "Hash=64",
                         "book": "${test_book}",
                         "stop_rule": "stop-rule-sprt",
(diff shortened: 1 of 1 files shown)


peregrineshahin:
On a a different note the test types should be tabs and stateful, this stateless system here making the buttons change the exact fields forgetting the older states if the user hopped between options is anti-pattern/ux and this page should be rewritten sometime in the future.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1982, opened 2024-05-05
Use "failed_task" api to flag a stale active task.

vdbergh:
This is the right thing to do in order to have an audit trail for tasks that have not played all their games.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -878,18 +878,16 @@ After fixing the issues you can unblock the worker at
                 if unique_key == task_unique_key:
                     # It seems that this worker was unable to update an old task
                     # (perhaps because of server issues).
-                    print(
-                        f'Stale active task detected for worker "{my_name_long}". Correcting...',
-                        flush=True,
-                    )
-                    self.set_inactive_task(task_id, run)
+                    # Set the task to inactive and print a message in the event log.
+                    self.failed_task(run["_id"], task_id, message="Stale active task")
                     continue
                 last_update = (now - task["last_updated"]).seconds
                 # 120 = period of heartbeat in worker.
                 if last_update <= 120:
                     error = (
                         f'Request_task: There is already a worker running with name "{task_name_long}" '
-                        f'which sent an update {last_update} seconds ago (my name is "{my_name_long}")'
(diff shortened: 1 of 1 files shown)


vdbergh:
Rebased.

vondele:
If a fleet leaves, will this lead to 1000s of messages in the event log?

vdbergh:
> If a fleet leaves, will this lead to 1000s of messages in the event log?

No because this issue only happens relatively rarely. However there are already 1000s of messages. See https://tests.stockfishchess.org/actions?action=dead_task&user=vdv&text=

Not sure if we should do something about this, and if so, what?

vondele:
Maybe we need to reconsider if such non-error events need logging. Now, it is up to you to decide if that's a logging-worthy event, i.e. if that can be helpful.

Maybe I am a bit confused between dead task and stale active task? What's the difference?

vdbergh:
> Maybe we need to reconsider if such non-error events need logging. Now, it is up to you to decide if that's a logging-worthy event, i.e. if that can be helpful.
> 
> Maybe I am a bit confused between dead task and stale active task? What's the difference?

A stale active task is a task whose attached worker is the same as the worker requesting a new tasks. This can happen if the worker was unable to update an old task because of server issues and gave up, i.e. requested a new task.

A dead task is a task which has not seen any updates for 6 minutes.

vdbergh:
> Maybe we need to reconsider if such non-error events need logging. 

Well a dead task is in some sense an error since it means that the worker did not quit cleanly (e.g. with SIGINT or SIGTERM).

This being said, in the case of a fleet, quitting the worker cleanly would just replace the dead task messages by failed task messages, so they would take up a similar amount of space.

I am not sure the dead_task messages are really a problem. They would be if the actiondb starts taking up too much space. @ppigazzini can probably comment on this.

Currently the idea is that the event log logs the full life cycle of tasks and runs.

ppigazzini (approved):
Looks good on DEV

ppigazzini:
> I am not sure the dead_task messages are really a problem. They would be if the actiondb starts taking up too much space. @ppigazzini can probably comment on this.

db.actions.totalSize() 1.031 GiB

ppigazzini:

PR #1983, opened 2024-05-05
Throttle workers on failure

vondele:
if workers fail to post a task_update, or for other reasons, they start communicating with the server more. This includes: a 5x retry to update the task, a failure message, an upload pgn of the task, a new task, etc. If the reason for failing is actually the load on the server, that load suddenly increases significantly, leading to an unstable, run-away situation in which most workers fail, and of which the server can't recover.

The attached patch tries to improve upon this, by increasing the retry time for task updates upon failure progressively. If a worker really failed, it starts with a 2min sleep before retrying.

This patch was successfully tested over the past couple of days, and made the server auto-recover under fairly large load.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -32,7 +32,7 @@ However, keep in mind that this information might be slightly outdated.
 This depends on how frequently the main instance flushes its `run_cache`.
 """
 
-WORKER_VERSION = 236
+WORKER_VERSION = 237
 
 
 def validate_request(request):
--- a/worker/games.py
+++ b/worker/games.py
@@ -1021,7 +1021,7 @@ def parse_cutechess_output(
             ):
                 # Attempt to send game results to the server. Retry a few times upon error.
                 update_succeeded = False
-                for _ in range(5):
+                for attempt in range(5):
                     try:
                         response = send_api_post_request(
                             remote + "/api/update_task", result
@@ -1048,7 +1048,7 @@ def parse_cutechess_output(
                         update_succeeded = True
                         num_games_updated = num_games_finished
                         break
-                    time.sleep(UPDATE_RETRY_TIME)
+                    time.sleep(UPDATE_RETRY_TIME * (attempt + 2))
                 if not update_succeeded:
                     raise WorkerException("Too many failed update attempts")
 
--- a/worker/sri.txt
+++ b/worker/sri.txt
@@ -1,1 +1,1 @@
-{"__version": 236, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "6+uEiLrveb452zembFH3erS4psr6m57/DXLVa7nXiO1zogUAy9AH5b9qFpsidmaJ", "games.py": "U9tidRvT37Rq3e0FByhHWLTV9p+4nfj5+c0W1wtHCseP7b58rxrlTrz6W6LQvsu1"}
+{"__version": 237, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "xIyJWylOkJngYAxSZ6wvJeBBtc0djb4njQqoz2CbeJPljOeHUOmgmxvIGQz8Jlp9", "games.py": "o8RR3ICNHS5v82eqXz+Jq8tHJsU3MTMUa+IegCjWu2UrVPveXQukqh5DEQgVhdLa"}
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -55,10 +55,10 @@ from updater import update
 # Several packages are called "expression".
 # So we make sure to use the locally installed one.
 
-WORKER_VERSION = 236
+WORKER_VERSION = 237
 FILE_LIST = ["updater.py", "worker.py", "games.py"]
 HTTP_TIMEOUT = 30.0
-INITIAL_RETRY_TIME = 15.0
+INITIAL_RETRY_TIME = 120.0
 THREAD_JOIN_TIMEOUT = 15.0
 MAX_RETRY_TIME = 900.0  # 15 minutes
 IS_COLAB = False


vdbergh:
I don't really like setting the INITIAL_RETRY_TIME to 120s since that makes casual testing of the worker more onerous (one often wants to do a quick check if the worker recovers after something goes wrong). One can however make the multiplier bigger (e.g. triple or quadruple the delay after each error).

vdbergh:
The problem with arming Fishtest against fleets of 100000 cores is that such an event very rarely happens but the resulting changes make things less and less interactive. Another example if the raise of the heartbeat period from 60s to 120s. This makes it more difficult to detect duplicate workers. Not meant as a criticism but as a general observation.

vdbergh:
Ideally the server should communitate some load metric to the worker which the worker can use to compute delays.

ppigazzini:
I merge some branches to make testing viable.
- server:
https://github.com/ppigazzini/fishtest/tree/task_duration_180
- worker:
https://github.com/ppigazzini/fishtest/tree/worker_update_freq
https://github.com/ppigazzini/fishtest/tree/worker_keep_binaries

ppigazzini:
> Ideally the server should communitate some load metric to the worker which the worker can use to compute delays.

The load on the server can be high but fine, then a run finishes and a big number of workers starts requiring a task, the answer time of the APIs increase, abruptly, from tenth of second to tens of seconds.
The missed server answer can be interpretated as a load metric.

As alternative we can check if we can add a global limit on nginx (the server is able to withstand max 170-190 calls/s).

vdbergh:
> > Ideally the server should communitate some load metric to the worker which the worker can use to compute delays.
> 
> The load on the server can be high but fine, then a run finishes and a big number of workers starts requiring a task, the answer time of the APIs increase, abruptly, from tenth of second to tens of seconds. The missed server answer can be interpretated as a load metric.

Still I would think that if the load_avg is high this should be a signal for the workers to slow down.

> 
> As alternative we can check if we can add a global limit on nginx (the server is able to withstand max 170-190 calls/s).

Throttling api calls in nginx (so that they don't reach the backend) seems like a more elegant solution than hamstringing the worker with longer and longer delays which are in fact irrelevant for every day use. What if someone decides to test a million core fleet? Should we multiply all delays by 10 to cope with that?

vdbergh:
Python has some functions relevant to the load average
```python
>>> import os
>>> os.getloadavg()
(0.50341796875, 0.3984375, 0.33544921875)
>>> os.cpu_count()
4
>>> 
```

ppigazzini:
> Still I would think that if the load_avg is high this should be a signal for the workers to slow down.

We have tested the server in these days. With 8k workers the CPU was at 45%. Joining 1k workers the CPU the CPU went to 100% in 2 seconds and the server stopped answering.

vdbergh:
> > Still I would think that if the load_avg is high this should be a signal for the workers to slow down.
> 
> We have tested the server in these days. With 8k workers the CPU was at 45%. Joining 1k workers the CPU went to 100% in 2 seconds and the server stopped answering. With this PR in place, PROD was fine with 100k cores ~14k workers.

I don't doubt that increasing worker (and other) delays makes the server more stable with high core fleets. But my point stands that this affects user friendliness in the common (=99.9%) use case.

vdbergh:
So if loadavg is not a useful metric, what about throttling api calls in nginx?

ppigazzini:
> > > Still I would think that if the load_avg is high this should be a signal for the workers to slow down.
> > 
> > 
> > We have tested the server in these days. With 8k workers the CPU was at 45%. Joining 1k workers the CPU went to 100% in 2 seconds and the server stopped answering. With this PR in place, PROD was fine with 100k cores ~14k workers.
> 
> I don't doubt that increasing worker (and other) delays makes the server more stable with high core fleets. But my point stands that this affects user friendliness in the common (=99.9%) use case.

The bottleneck is only the pserve at 6543
Your proposal can help with the friendliness during normal usage, but it is hardly useful when a fleet joins. In that case the server will stop answering to the heartbeat (served by 6543), so all workers will stay to a fast communication (barring using the lack of answer to switch the speed of communication)

ppigazzini:
> So if loadavg is not a useful metric, what about throttling api calls in nginx?
> 
> EDIT: Another useful load metric could be the number of active tasks. This should be proportional to the instant server load.

The number of cores

vdbergh:
> > So if loadavg is not a useful metric, what about throttling api calls in nginx?
> > EDIT: Another useful load metric could be the number of active tasks. This should be proportional to the instant server load.
> 
> The number of cores

I guess that's even easier...

vdbergh:
But I guess nothing will help if we connect 10000 workers instantaneously. The server will be unable to send any useful information to the workers.

vdbergh:
But I wonder if this PR would help with that. After some delay all workers will retry again and kill the server again. It might be more helpful to randomize the delays instead of making them bigger.

vdbergh:
It seems that there is plenty of stuff about nginx rate limiting on the internet. If this can be done in such a way that the backend keeps receiving the 170-190 api calls that it supports then we're fine since no amount of worker tweaking will make it possible to exceed that limit.

vdbergh:
Why was the fleet white listed? Did the number of api calls received by the backend drop below the 170-190 calls/s?

vdbergh:
> The fleet uses only one IP.

Why is that relevant? We would just want to limit the number api calls passed to the backend, independently of the ip.

vdbergh:
It seems Fishtest tries to do some rate limiting of its own
```
    # Limit concurrent request_task
    task_lock = threading.Lock()
    task_semaphore = threading.Semaphore(4)
```
But the problem I guess is that the pyramid server simply cannot handle that many connections.

ppigazzini:
> > The fleet uses only one IP.
> 
> Why is that relevant? We would just want to limit the number api calls passed to the backend, independently of the ip.

Because we want to reject a DOS

vdbergh:
> > > The fleet uses only one IP.
> > 
> > 
> > Why is that relevant? We would just want to limit the number api calls passed to the backend, independently of the ip.
> 
> Because we want to reject a DOS

Well that's an independent problem. You can whitelist the ip of the fleet and still perform rate limiting.

ppigazzini:
> > > > The fleet uses only one IP.
> > > 
> > > 
> > > Why is that relevant? We would just want to limit the number api calls passed to the backend, independently of the ip.
> > 
> > 
> > Because we want to reject a DOS
> 
> Well that's an independent problem. You can whitelist the ip of the fleet and still perform rate limiting.

That is that what I will do as soon as we stop writing here :)

vdbergh:
I was thinking that if throttling api call in nginx works then we should probably queue failed pgn uploads in the worker because we want those to succeed eventually. Or else we should white list pgn uploads.

ppigazzini:
this should work:
```nginx
map $allowlist $rate_limit_ip {
    0 $binary_remote_addr$host;
    1 "";
}

map $allowlist $rate_limit_host {
    0 $host;
    1 "";
}

map $allowlist $rate_limit_uri {
    0 $host$uri;
    1 "";
}

map $host $rate_limit_null {
    default "";
}

limit_req_zone $rate_limit_null zone=per_null:20m rate=175r/s;
limit_req zone=per_ip burst=500 delay=10;

limit_req_zone $rate_limit_ip zone=per_ip:20m rate=5r/s;
limit_req zone=per_ip burst=50 delay=10;

limit_req_status 444;


testcookie_limit_zone $rate_limit_ip zone=per_ip_cookie:20m rate=5r/s;
testcookie_limit zone=per_ip_cookie burst=100 challenge=20;

testcookie_limit_status 444;

```

vondele:
> I was thinking that if throttling api call in nginx works then we should probably queue failed pgn uploads in the worker because we want those to succeed eventually. Or else we should white list pgn uploads.

this pgn uploads is among those calls the one that seem most taxing for the

vondele:
sorry, wrong click :-)

this pgn uploads is among those calls the one that seem most taxing for the api.

I think the problem is that our server is able to handle 'just' around 100 calls per second. I a user closes a couple of test, each with a few 100 workers we are in trouble.

ppigazzini:
The upload of the PGN should be moved on another pserve, but the code of api.py has many side effects.
I started the tidy up with #1984.

vdbergh:
> I think the problem is that our server is able to handle 'just' around 100 calls per second. I a user closes a couple of test, each with a few 100 workers we are in trouble.

Do you mean the front end or the backend? nginx is supposed to be able to handle much more even on weak servers.

vondele:
nginx can handle a lot. I'm talking about our fishtest server. If you'll allow nginx to block calls at random, I predict things will get worse.

ppigazzini:
> > I think the problem is that our server is able to handle 'just' around 100 calls per second. I a user closes a couple of test, each with a few 100 workers we are in trouble.
> 
> Do you mean the front end or the backend? nginx is supposed to be able to handle much more even on weak servers.

The middleware. Our python is fast, but the conversion from python <-> json takes a lot of time.

vondele:
conversion is one aspect, but let's not forget even the db updates.

ppigazzini:
https://github.com/ijl/orjson
`fastapi` uses natively `orjson` if installed.
`pyramid` requires the definition of a custom renderer, something like this:
```python
from pyramid.renderers import JSON
import orjson

class ORJSONRenderer(JSON):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.dumps = orjson.dumps

def includeme(config):
    json_renderer = ORJSONRenderer()
    config.add_renderer('json', json_renderer)
```
The speedup in converting back and forth between json <-> python is huge, but `orjson` has different behavior wrt the standard library `json`.  It's easy to start a new project from scratch, not easy to check all the code in a big project when replacing the renderer.

vdbergh:
> nginx can handle a lot. I'm talking about our fishtest server. If you'll allow nginx to block calls at random, I predict things will get worse.

I may be wrong I don't really see why. The important api calls are `update task`, which is retried 5 times (this could be tweaked) and the pgn upload which should probably be served from a secondary instance or else could be queued.

vdbergh:
> > > I think the problem is that our server is able to handle 'just' around 100 calls per second. I a user closes a couple of test, each with a few 100 workers we are in trouble.
> > 
> > 
> > Do you mean the front end or the backend? nginx is supposed to be able to handle much more even on weak servers.
> 
> The middleware. Our python code is fast, but the conversion from python <-> json takes a lot of time.

The middleware is the pyramid server right? That's what I meant by backend (it's the thing that nginx talks to).

vondele:
Anyway, just some data, when the server is doing well under load the top profile looks like this:
```
# logging from [04/May/2024:12:44:43 +0000] to [04/May/2024:12:45:14 +0000] ########
# duration (seconds)            : 31
# calls in total                : 4000
# calls per second              : 129
# calls not reaching the backend: 
# calls handled by the backend  : 4000
                      /api/beat     2600     44.582      0.017      0.000      0.265
               /api/update_task     1182     22.960      0.019      0.001      0.427
              /api/request_task       38      4.667      0.123      0.050      0.286
                         /tests        5      4.127      0.825      0.010      1.651
                   /tests/user/        4      3.171      0.793      0.552      0.936
                /api/upload_pgn       28      1.497      0.053      0.011      0.448
                  /api/get_elo/       18      1.159      0.064      0.013      0.417
                  ```
a few minutes later adding a additional workers:
```
# logging from [04/May/2024:12:46:38 +0000] to [04/May/2024:12:47:08 +0000] ########
# duration (seconds)            : 30
# calls in total                : 4000
# calls per second              : 133
# calls not reaching the backend: 1
# calls handled by the backend  : 3999
                      /api/beat     1871   5745.340      3.071      0.004     29.993
               /api/update_task      994   2222.567      2.236      0.005     29.991
           /api/request_version      709   1104.207      1.557      0.005     29.985
              /api/request_task      243    796.810      3.279      0.113     24.285
                /api/upload_pgn       37    128.206      3.465      0.353     24.135
              /api/request_spsa      107    100.887      0.943      0.016     22.010
               /api/failed_task        7      9.125      1.304      0.375      5.003
                         /tests        9      6.597      0.733      0.010      1.549
```

In the current setup, the solution is to kill the workers as a user (who doesn't have this info).
With the patch, in this PR, things recover automatically.

**Comment by ppigazzini** [comment]
> The middleware is the pyramid server right? That's what I meant by backend (it's the thing that nginx talks to).

waitress, the WSGI used in pyramid. pserve is the script that runs waitress.

**Comment by vdbergh** [comment]
> In the current setup, the solution is to kill the workers as a user (who doesn't have this info).
With the patch, in this PR, things recover automatically.

Yes but what I find bothersome is that this is at the cost of worsening the user experience for the common use case.

> Notice that the two profiles are nearly indistinguishable in the 'calls per second'

Interesting point and not obvious to explain. Perhaps the average is not representative for the peaks that occurred when the new workers connected?

**Comment by ppigazzini** [comment]
The connections have been closed after 30 seconds
```nginx
    location /api/ {
        proxy_set_header Connection "";
        proxy_set_header X-Forwarded-Proto  $scheme;
        proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Host   $host:$server_port;
        proxy_set_header X-Forwarded-Port   $server_port;
        proxy_set_header X-Country-Code     $region;

        client_max_body_size        120m;
        client_body_buffer_size     128k;
        proxy_connect_timeout       5s;
        proxy_send_timeout          30s;
        proxy_read_timeout          30s; <-----------
        proxy_max_temp_file_size    0;
        proxy_redirect              off;
        proxy_http_version          1.1;

        proxy_pass http://$backends;
    }

```

vdbergh:
> The upload of the PGN should be moved on another pserve, but the code of api.py has many side effects. I started the tidy up with #1984.

I see. The blocking thing is "validate_request" which uses `get_run`.  We could supply it with an optional argument `use_db` which would make it load the run from the db.

ppigazzini:
Yes.

ppigazzini:
Yes, but that module is full of side effects. `validate_username_password(self)` save the body_json (used everywhere) and the timestamp, but those should be in `__init__` or in a separated function.

vdbergh:
> Yes, but that module is full of side effects. `validate_username_password(self)` at example save the body_json (used everywhere) and the timestamp, but those should be in `__init__` or in a separated function.

Perhaps. But that's not a fundamental thing.

But not being able to use `get_run` in a secondary instance is a fundamental restriction. But of course the instances of `rundb` know when they are secondary or not (since the port number is determined at start up). So we could make some behavior dependent on that.

ppigazzini:
IMO that is fundamental. We had for years not readable and bugged scripts. We need to clean up the code before adding new features.

vdbergh:
Well you have seen that code before I simplified it...

I admit that `self.request_body` and the timestamp could be set inside  __init__ instead of as in `validate_username_password`.

vdbergh:
I think you are unhappy that `validate_request(self, api)` sets some private instance variables of the `ApiView` object while validating. That kind of side effect is more like an initialization and I consider it benign (in contrast to global side effects which are dangerous since they may affect things in unexpected places). Personally I think the code is very clear but of course, having written it myself, I am biased.

vdbergh:
> Notice that the two profiles are nearly indistinguishable in the 'calls per second'

Looking again it seems that in the second profile the server suddenly got hit by a large number of `request_task` calls (compare to other calls such as the heart beat). `request_task` is quite expensive.

This makes one wonder that it perhaps only necessary to throttle `request_task` calls.

vdbergh:
I wonder though why the internal Fishtest throttling does nothing     
```
def request_task(self, worker_info):
        if self.task_semaphore.acquire(False):
            try:
                with self.task_lock:
                    return self.sync_request_task(worker_info)
            finally:
                self.task_semaphore.release()
        else:
            print("request_task too busy", flush=True)
            return {"task_waiting": False}
```
@ppigazzini Do you ever see messages on the server `"request_task too busy"`?

ppigazzini:
none in the last 180 days:
```yaml
fishtest@tests:~$ sudo journalctl -u fishtest@6543 --since "180 days ago" | grep "request_task"
Dec 08 10:34:43 tests.stockfishchess.org pserve[13971]: /api/request_task: request['worker_info']['python_version'][0] (value:-3) is not in the interval [0,...]
Dec 08 10:35:47 tests.stockfishchess.org pserve[13971]: /api/request_task: request['worker_info']['version'] (value:-222) is not in the interval [0,...]
Dec 08 10:36:18 tests.stockfishchess.org pserve[13971]: /api/request_task: request['worker_info']['version'] (value:222.0) is not of type 'int'
Dec 08 23:34:56 tests.stockfishchess.org pserve[13971]: /api/request_task: request['worker_info']['min_threads'] (value:0) is not in the interval [1,...]
Dec 08 23:35:12 tests.stockfishchess.org pserve[13971]: /api/request_task: request['worker_info']['min_threads'] (value:0) is not in the interval [1,...]
Dec 11 19:13:55 tests.stockfishchess.org pserve[30609]: /api/request_task: request is not json encoded
Dec 11 19:14:11 tests.stockfishchess.org pserve[30609]: /api/request_task: request is not json encoded
Mar 10 00:29:37 tests.stockfishchess.org pserve[17087]: /api/request_task: request is not json encoded
fishtest@tests:~$ sudo journalctl -u fishtest@6544 --since "180 days ago" | grep "request_task"
fishtest@tests:~$ sudo journalctl -u fishtest@6545 --since "180 days ago" | grep "request_task"
fishtest@tests:~$

```

vdbergh:
Thanks. So that semaphore code seems to do nothing... May as well delete it to make things clearer.

vdbergh:
I think we should set the semaphore value to 1. This means that `request_task` will never wait for `sync_request_task` to finish. I.e. it behaves as a non-blocking lock.

vdbergh:
I made the PR #1987 . It seems better to let the worker retry than to hog the server. I wonder if that semaphore code really works however...

ppigazzini:
```python
import threading
import time
import random

# Create a semaphore that allows up to N concurrent tasks
N = 4
semaphore = threading.Semaphore(N)

def worker(n):
    print(f"Worker {n} is trying to acquire the semaphore")
    if semaphore.acquire(False):
        try:
            print(f"Worker {n} has acquired the semaphore")
            sleep_time = random.randint(1, 5) * 0.1
            time.sleep(sleep_time)  # Simulate doing some work
            print(f"Worker {n} has finished its work and is releasing the semaphore")
        finally:
            semaphore.release()
    else:
        print(f"Worker {n} could not acquire the semaphore")

# Create and start M worker threads
M = 100000
for i in range(M):
    threading.Thread(target=worker, args=(i,)).start()
```

vdbergh:
Thanks! I was more wondering if different threads might not be using different semaphores or something, but looking at the code this seems impossible. 

So it seems that once the `request_task` api call reaches Fishtest, Fishtest has no difficulty handling it (otherwise the semaphore limit would have been reached easily during the fleet visits). 

So ultimately it really seems that waitress cannot handle sudden burst of connections. If these are smoothed out by nginx then perhaps the server problem would be solved?

ppigazzini:
With Joost I benched the "request_spsa" internally with 60k cores attached:

```
/api/request_spsa Total Times:
Min time: 0.0005233287811279297
Max time: 0.1949458122253418
Avg time: 0.00247439
```

This are the same min max avg seen by nginx:
```
                    route    calls       total   average    minimum    maximum
/api/request_spsa       31      7.577       0.244          0.008            1.053
```

vdbergh:
Are these in different units? To me they look quite different... E.g. max times 0.19 vs 1.053. And this is presumably when then middleware is running smoothly.

ppigazzini:
Middleware running smoothly, seconds, this is why IMO we should try with `orjson`.
A couple of years ago I tried to replace `waitress` with `gunicon`, but I got some issues.
`fastapi` is very fast, uses `uvcorn` ASGI, `gunicorn` as `uvcorn` workers manager, and switches to `orjson` if installed.

vdbergh:
```
/api/request_spsa Total Times:
Min time: 0.0005233287811279297
Max time: 0.1949458122253418
Avg time: 0.00247439
```

```
                    route    calls       total   average    minimum    maximum
/api/request_spsa       31      7.577       0.244          0.008            1.053
```
If we look at the average times (assuming as you say that both are in seconds) then almost all the time is spent in the middleware (`0.244-0.0025`).

ppigazzini:
No difference in avg time API call on PROD for `orjson` vs `json`
Same result in a test on DEV downloading a big list of task.
```bash
#!/bin/bash
url="https://dfts-0.pigazzini.it/tests/tasks/650c84f9fb151d43ae6d5342"
for i in {1..10}
do
  for j in {1..30}
  do
    curl -o /dev/null -s "$url" &
  done
  wait
done
```
- `json`
```bash
usr00@pp-wrk-00:~/_git/fishtest$ time bash parallel_curl.sh

real    0m35.129s
user    0m30.148s
sys     0m8.638s
usr00@pp-wrk-00:~/_git/fishtest$ time bash parallel_curl.sh

real    0m36.474s
user    0m31.453s
sys     0m9.037s
```
- `orjson`
```bash
usr00@pp-wrk-00:~/_git/fishtest$ time bash parallel_curl.sh

real    0m34.905s
user    0m30.843s
sys     0m7.557s

usr00@pp-wrk-00:~/_git/fishtest$ time bash parallel_curl.sh

real    0m34.521s
user    0m30.731s
sys     0m8.499s
```

vondele:
I'll close this PR, looks like some of the change and potential new developments are a better way to address this.

PR #1985, opened 2024-05-05
Make validate_request work on non-primary instances.

vdbergh:
This should make it possible to serve `/api/upload_pgn` from a non-primary instance. See https://github.com/official-stockfish/fishtest/pull/1983#issuecomment-2094916162

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -5,6 +5,7 @@ import re
 from datetime import datetime, timezone
 from urllib.parse import urlparse
 
+from bson.objectid import ObjectId
 from fishtest.schemas import api_access_schema, api_schema, gzip_data
 from fishtest.stats.stat_util import SPRT_elo, get_elo
 from fishtest.util import worker_name
@@ -24,8 +25,6 @@ Important note
 
 All APIs that rely on the `run_cache` of `rundb.get_run()`
 must be served from the main Fishtest instance.
-Note that `self.validate_request("/api/<route>")`
-uses `rundb.get_run()` under some conditions.
 
 If other Fishtest instances need information about runs,
 they should query the database directly.
@@ -122,6 +121,13 @@ class ApiView(object):
             )
 
     def validate_request(self):
+        """
+        This function will load the run from the cache or the db,
+        depending on the type of instance it runs on (primary or
+        secondary). If the request refers to a particular
+        task then one needs to make sure that it has been saved
+        to disk when invoking this function on a secondary instance.
+        """
         self.__run = None
         self.__task = None
 
@@ -137,7 +143,10 @@ class ApiView(object):
         # is a supplied run_id correct?
         if "run_id" in self.request_body:
             run_id = self.request_body["run_id"]
-            run = self.request.rundb.get_run(run_id)
+            if self.request.rundb.is_primary_instance():
+                run = self.request.rundb.get_run(run_id)
+            else:
+                run = self.request.rundb.runs.find_one({"_id": ObjectId(run_id)})
             if run is None:
                 self.handle_error("Invalid run_id: {}".format(run_id))
             self.__run = run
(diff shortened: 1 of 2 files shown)


vdbergh:
Note that validation of api calls that contain the `run_id/task_id` fields is inherently more expensive on secondary instances since it requires loading the run from the db. However for heavy calls such as `/api/upload_pgn` this may be worth it.

ppigazzini:
The PR adds also the overhead to save the run in the DB for every task added, keeping much time the run_cache_lock.

vdbergh:
I agree. I don't know how much that overhead is (it seems Fishtest is not the bottleneck). 

If one does not do this then there might be an (unlikely) race condition that the task has not been saved to disk by the time the pgn is uploaded (on a secondary instance).

vdbergh:
Actually that's not true. The pgn is only uploaded for a finished run which is saved to disk. So for pgn upload it would be ok not to save the run to disk on task creation.

ppigazzini:
> I agree. I don't know how much that overhead is (it seems Fishtest is not the bottleneck).
> 
> If one does not do this then there might be an (unlikely) race condition that the task has not been saved to disk by the time the pgn is uploaded (on a secondary instance).

Unfortunately, not so unlikely, happened some weeks ago after I moved the upload of the PGNs on a secondary instance. The PGNs failed the validation.

We could check how much code is required to write a document (w/o the PGN of course) in the pgns collection when allocating a new task, and validate the PGN upload checking the run_task_id in the pgns collection.

vdbergh:
> Unfortunately, not so unlikely, happened some weeks ago after I moved the upload of the PGNs on a secondary instance. The PGNs failed the validation.

That's a different issue. The current validation code (before this PR) uses `get_run` which does not work on a secondary instance. Then things break down rapidly!

ppigazzini:
Correct, it works only for the first request (that load in cache form the db), the next uploads get the outdated run from the cache.

vdbergh:
In the latest push I no longer do `buffer(run, True)` in `request_task` since it not necessary for pgn upload. I added however a comment to `validate_request` with some explanation.

ppigazzini (approved):
Looks good on DEV with the updated nginx config.

ppigazzini:
PROD and nginx updated, thank you @vdbergh

ppigazzini:
wiki script updated, here is the snippet of the nginx config:

```nginx
map $uri $backends {
    /upload_net/                              backend_8000;
    /tests                                    backend_6544;
    ~^/api/(actions|active_runs|calc_elo)     backend_6545;
    ~^/api/(nn/|pgn/|run_pgns/|upload_pgn)    backend_6545;
    ~^/tests/(finished|machines|user)         backend_6545;
    ~^/(actions/|contributors)                backend_6545;
    ~^/(api|tests)/                           backend_6543;
    default                                   backend_6544;
}

```

vondele:
Just for the record, current server profile under significant load look like:
```
# logging from [08/May/2024:18:50:28 +0000] to [08/May/2024:18:51:18 +0000] ########
# duration (seconds)            : 50
# calls in total                : 10000
# calls per second              : 200
# calls not reaching the backend: 1
# calls handled by the backend  : 9999
               /api/update_task     4061  23836.038      5.869      0.005     30.034
                      /api/beat     4807  22716.703      4.726      0.002     30.031
           /api/request_version      451   2507.618      5.560      0.013     30.003
              /api/request_task      352   1602.746      4.553      0.133     29.991
              /api/request_spsa       35    110.623      3.161      0.012     29.986
               /api/failed_task        9     62.856      6.984      0.100     29.984
               /tests/live_elo/        3     22.234      7.411      0.463     21.280
                /api/upload_pgn      233     21.713      0.093      0.011      0.365
                         /tests       14      8.932      0.638      0.019      1.203
                  /api/get_elo/        2      0.695      0.348      0.261      0.434
                   /tests/user/        1      0.638      0.638      0.638      0.638
                   /api/actions       17      0.507      0.030      0.006      0.084
```
looking at the 'maximum' column, upload_pgn now is doing very fine.

PR #1986, opened 2024-05-06
fall back to the three dot in case of PTs

peregrineshahin:
since official commits are always rebased

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -847,6 +847,9 @@
           dots = 3; // fall back to the three dot diff request as the diff will be rebased
       % endif
     % else:
+      % if run["args"]["new_tag"] == "master" and run["args"]["base_tag"] == pt_info["pt_branch"]:
+          dots = 3; // fall back to the three dot in case of PTs since official is always rebased
+      % endif
       const apiUrlBase = apiUrlNew;
       const diffBase = "${run["args"]["resolved_base"][:10]}";
     % endif
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1589,6 +1589,7 @@ def tests_view(request):
         "follow": follow,
         "can_modify_run": can_modify_run(request, run),
         "same_user": same_user,
+        "pt_info": request.rundb.pt_info,
     }
 
 


peregrineshahin:
> Such data is included in the rundb object. In this way it is available in mako.

Thanks for pointing that out

server/fishtest/templates/tests_view.mak
```diff
@@ -847,6 +847,12 @@
           dots = 3; // fall back to the three dot diff request as the diff will be rebased
       % endif
     % else:
+      % if (
```
ppigazzini, line 850: mako %if written in only one line.

If I'm not wrong, it does not work

image

peregrineshahin: Sorry I can't test the PR rn, I changed it to one line I'll check if that works when home.

ppigazzini:
wrong mako file

ppigazzini:
```diff
git diff
diff --git a/server/fishtest/templates/tests_view.mak b/server/fishtest/templates/tests_view.mak
index 7096f5f..435826f 100644
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -74,7 +74,12 @@
   <div class="elo-results-top">
     <%include file="elo_results.mak" args="run=run" />
   </div>
+  <div>

+  run["args"]["new_tag"]: ${run["args"]["new_tag"]} <br>
+  run["args"]["base_tag"]: ${run["args"]["base_tag"]} <br>
+  ${f"{pt_info=}"}
+  </div>
   <div class="row">
     <div class="col-12 col-lg-9">
       <div id="diff-section">
diff --git a/server/fishtest/views.py b/server/fishtest/views.py
index e0580c3..818b702 100644
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1590,6 +1590,7 @@ def tests_view(request):
         "follow": follow,
         "can_modify_run": can_modify_run(request, run),
         "same_user": same_user,
+        "pt_info": request.rundb.pt_info,
     }


@@ -1634,7 +1635,6 @@ def get_paginated_finished_runs(request):
         "num_finished_runs": num_finished_runs,
         "failed_runs": failed_runs,
         "page_idx": page_idx,
-        "pt_info": request.rundb.pt_info,
     }
```

peregrineshahin:
I don't understand how this evaluates to false
`% if run["args"]["new_tag"] == "master" and run["args"]["base_tag"] == pt_info["pt_branch"]:`
for this test
https://dfts-0.pigazzini.it/tests/view/66293d2e3fe04ce4cefc7a61

ppigazzini:
I don't know, I'm printing the variables

image

peregrineshahin:
with the printed values in that test https://dfts-0.pigazzini.it/tests/view/66293d2e3fe04ce4cefc7a61 , it should evaluate to true
I believe there is no trimming needed...
      ```
      % if run["args"]["new_tag"] == "master" and run["args"]["base_tag"] == pt_info["pt_branch"]:
          dots = 3; // fall back to the three dot in case of PTs since official is always rebased
      % endif
      ```
 the js file should have those lines....

ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

PR #1987, opened 2024-05-06
Set the request_task semaphore to one.

vdbergh:
The current value of 5 is never hit according to the server logs, even with large fleets connecting.

With value 1 the semaphore behaves as a non-blocking lock. The server will never wait for `sync_request_task` to finish but instead informs the worker that no task is available.

Not tested.


vdbergh:
Without more data one cannot know if this PR is better than just deleting the semaphore code (which currently is a noop). I am closing this for now.

PR #1988, opened 2024-05-06
Smarten Live Elo

peregrineshahin:
This avoids fetching each 20s tabs that are not focused/visible by the user, by making one tab update at a time for multiple Live Elo tabs. on the other hand updates immediately the new tab if it becomes the one in focus. Allowing for less load from the API with a lazy load approach.

### DIFF
--- a/server/fishtest/static/js/live_elo.js
+++ b/server/fishtest/static/js/live_elo.js
@@ -182,7 +182,29 @@ async function followLive(testId) {
 
   // Main worker
 
-  while (true) {
+  let isTabFocused = true;
+  let isVisibilityChange = true;
+
+  document.addEventListener("visibilitychange", function () {
+    isTabFocused = document.visibilityState === "visible";
+
+    // If the tab just becomes visible, update immediately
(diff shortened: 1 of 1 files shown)


ppigazzini:
DEV updated, please test yourself.

peregrineshahin:
I tested it on my instance before PRing.

server/fishtest/static/js/live_elo.js
```diff
@@ -182,7 +182,29 @@ async function followLive(testId) {
...
   // Main worker
 
-  while (true) {
+  let isTabFocused = true;
+  let isVisibilityChange = true;
+
+  document.addEventListener("visibilitychange", function () {
+    isTabFocused = document.visibilityState === "visible";
+
+    // If the tab have just become visible, update immediately
```
MrBrain295, line 191:
```suggestion
    // If the tab just becomes visible, update immediately
```

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #1989, opened 2024-05-07
Show up-to-date result when opening the live elo tab after delay.

vdbergh:


server/fishtest/static/js/live_elo.js
```diff
@@ -201,6 +201,7 @@ async function followLive(testId) {
         update();
       }
       await asyncSleep(20000);
+      isVisibilityChange = true;
```
peregrineshahin, line 204: Sorry about that the patch I made yesterday seems not to do what I wanted.
this line should be added when the visibility change happens
i.e. at line 190

vdbergh: I think with this line added the patch behaves as it should. However feel free to submit your own fix and then I will close this one!

ppigazzini:

PR #1990, opened 2024-05-07
Fix no immediate update when changing tab

peregrineshahin:
That's a bug in the last commit
supersedes: https://github.com/official-stockfish/fishtest/pull/1989

### DIFF
--- a/server/fishtest/static/js/live_elo.js
+++ b/server/fishtest/static/js/live_elo.js
@@ -183,14 +183,12 @@ async function followLive(testId) {
   // Main worker
 
   let isTabFocused = true;
-  let isVisibilityChange = true;
-
-  document.addEventListener("visibilitychange", function () {
+  document.addEventListener("visibilitychange", async function () {
     isTabFocused = document.visibilityState === "visible";
 
-    // If the tab just becomes visible, update immediately
-    if (isVisibilityChange && isTabFocused) {
-      isVisibilityChange = false;
+    // If the tab just becomes visible, update in 5 seconds
+    if (isTabFocused) {
+      await asyncSleep(5000);
       update();
     }
   });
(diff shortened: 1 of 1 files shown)


vdbergh:
I think you only want to show the update after some time. Otherwise you encourage people to keep changing tabs just to get quick updates.

peregrineshahin:
well, that's not a fun experience for the user.. the issue really is that people leave the tabs open..

peregrineshahin:
I mean they can already do it by refreshing.. this feels more in between, avoid the unnecessary requests when the tab is hidden but at the same time, let them see the last update only when they show the tab

vdbergh:
> well, that's not a fun experience for the user.. the issue really is that people leave the tabs open..

I am not sure. People will cycle through several tabs to see the most up to date results for each of them. I think it is strictly better to force a delay before showing the updated results (like in #1989 ).

vdbergh:
> I mean they can already do it by refreshing.. this feels more in between, avoid the unnecessary requests when the tab is hidden but at the same time, let them see the last update only when they show the tab

Not sure. Repeatedly refreshing several tabs seems like a heavier operation than just cycling through them.

peregrineshahin:
okay to avoid that misuse I added 5 second delay.

peregrineshahin:
awaiting in a loop does not delay the visibility change event to trigger, so I don't think that PR does what you wanted.

> I still think #1989 is better. It will only affect users that change tabs quickly. Users that wait longer than 20s before revisiting the tab will immediately see the latest results.

peregrineshahin:
Hmm, I see your point this is intentional, it's a bit confusing.. I have no hard opinion about this.
I will leave it to #1989, but the comments and the variable `isVisibilityChange` I think should change.

PR #1991, opened 2024-05-07
log rejecting users

peregrineshahin:
prints some information in the server log since we don't record such events publicly.
this also helps in case of a rejection by mistake, since we can contact the rejected user via email from the log.


server/fishtest/userdb.py
```diff
@@ -137,12 +137,23 @@ def save_user(self, user):
...
+
+            # Keep in logs the rejected users in case it was by mistake
+            # so we can email the user to register again
+            # and documents who rejected the user since we don't want to report it in the event log
+            if rejector:
+                current_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+                print(
+                    f"{current_time}: user: {user['username']} with the email: {user['email']} got rejected by: {rejector}",
+                    flush=True,
+                )
```
vdbergh, line 156: The server already adds a timestamp to the log entry. So there is no need to do it here.

vdbergh (commented):
Timestamp not necessary.

server/fishtest/userdb.py
peregrineshahin: Thanks, I'll remove that

vdbergh: Perhaps also 
```
user: {user['username']} with email: {user['email']} was rejected by: {rejector}
```
(nitpicking)

ppigazzini (approved):
Looks good on DEV
```yaml
May 12 15:02:37 dfts-0 systemd[1]: Started Fishtest Server port 6544.
May 12 15:02:39 dfts-0 pserve[1605]: Starting server in PID 1605.
May 12 15:02:47 dfts-0 pserve[1605]: user: new222 with email: new222new222new222new222new222new222new222new222@yahoo.com was rejected by: user00
May 12 15:04:55 dfts-0 pserve[1605]: user: test00 with email: test00test00test00test00test00test00test00test00@yahoo.com was rejected by: user00
```

ppigazzini:

PR #1992, opened 2024-05-07
Extend `get_run()` to support secondary instances

ppigazzini:
Simplify the coding of the APIs moving the logic of primary/secondary
instance in the RunDB class.
Drop old debug code.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -5,7 +5,6 @@ import re
 from datetime import datetime, timezone
 from urllib.parse import urlparse
 
-from bson.objectid import ObjectId
 from fishtest.schemas import api_access_schema, api_schema, gzip_data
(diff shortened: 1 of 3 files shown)


ppigazzini:
@vdbergh I would like to drop the DEBUG code in `rundb.py`, do you agree?

vdbergh:
This is a good start! It may be good to note that the new `get_run()` only provides read only access on secondary instances (i.e. `buffer()` will not work).

I am thinking of an improvement that would not come with the caveat that one must be careful that a task has been saved to the db.

The idea would be to create an api (hosted by the primary instance) to force the primary instance to write a run to disk, if it is dirty.

vdbergh:
> @vdbergh I would like to drop the DEBUG code in `rundb.py`, do you agree?

Yes. It hasn't been used for ages and maybe does not even work...

PR #1994, opened 2024-05-08
open some links in a new tab

MrBrain295:
This prevents the user from leaving Fishtest on accident.


peregrineshahin:
Nice, yes these should open in another tab.

ppigazzini (approved):
Looks good on DEV

ppigazzini:
PROD updated, thank you @MrBrain295

PR #1998, opened 2024-05-10
Validate run_id with ObjectId.is_valid in api_schema.

vdbergh:
This seems to be equivalent to the regex in master but in principle it does not have to be.

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -37,7 +37,7 @@ from vtjson import (
     url,
 )
 
-run_id = regex(r"[a-f0-9]{24}", name="run_id")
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
```yaml
May 10 23:05:34 dfts-0 pserve[15542]: /api/update_task: request['run_id'] (value:'QQQQQQQQQQQQQQQQ') is not of type 'valid_object_id'
May 10 23:05:35 dfts-0 pserve[15542]: Failed_task: failure for: https://tests.stockfishchess.org/tests/view/663e4ef09ea884ad4b744b3b, task_id: 0, worker: user00-3cores-86e63159-c0df*, reason: 'Too many failed update attempts'
```

ppigazzini:

PR #1999, opened 2024-05-11
Delegate the worker apis to a separate class.

vdbergh:
Worker apis are special in the sense that they need authentication.

Moreover many of the special purpose functions defined for worker apis are completely irrelevant for the other apis. This clutters the code.

Lightly tested (no spsa).

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -81,23 +81,32 @@ def authentication_failed(error, request):
     return response
 
 
-@view_defaults(renderer="json")
-class ApiView(object):
-    """All API endpoints that require authentication are used by workers"""
+def add_time(result, ts=None):
+    result["duration"] = (datetime.now(timezone.utc) - ts).total_seconds()
+    return result
 
-    def __init__(self, request):
-        self.request = request
+
+def handle_error(request, error, exception=HTTPBadRequest, ts=None):
+    if error != "":
+        full_url = request.route_url(request.matched_route.name)
+        api = urlparse(full_url).path
+        error = f"{api}: {error}"
+        print(error, flush=True)
+        raise exception(add_time({"error": error}, ts=ts))
+
+
+class worker_apis:
+    def __init__(self, apiview):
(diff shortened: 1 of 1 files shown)


vdbergh:
I am marking this as draft since I think the setup can be simplified.

server/fishtest/api.py
```diff
@@ -81,23 +81,32 @@ def authentication_failed(error, request):
...
+def handle_error(request, error, exception=HTTPBadRequest, ts=None):
+    if error != "":
+        full_url = request.route_url(request.matched_route.name)
+        api = urlparse(full_url).path
+        error = f"{api}: {error}"
+        print(error, flush=True)
+        raise exception(add_time({"error": error}, ts=ts))
+
+
+class worker_apis:
```
ppigazzini, line 98: conventional case for class 
```diff
- class worker_apis:
+ class WorkerApi
+ # class WorkerApiHandler # as alternative
```

ppigazzini:
It worked fine on DEV.

vdbergh:
> It worked fine on DEV.

That's good! In principle there is no new code but it has been reorganized.

But it should not be necessary to do the forwarding from `ApiView` to `worker_apis` (or e.g. `WorkerApi`) ourselves since Pyramid will do that for us. At least that's what I understood from reading the docs.

So something like
```python
class GenericApi:
# contains handle_error and add_time

class WorkerApi(GenenericApi):
# contains the worker apis

class UserApi(GenericApi):
# contains the other apis
```
```

ppigazzini:
Missing a class for this:
https://github.com/official-stockfish/fishtest/pull/1993

vdbergh:
Perhaps `InternalApi`?

ppigazzini:
Good name. If you are refactoring all the code, try to keep short and clean the view (that get the request, get a processed data, and send the response), with the business logic, calling a separated function (defined in this file or a new module) that processes the data.

vdbergh:
Superseded by #2000

PR #2001, opened 2024-05-11
Get port and primary_port from the config files development.ini and production.ini (RFC)

vdbergh:
This gets rid of some ugly code in rundb.py.

### DIFF
--- a/server/development.ini
+++ b/server/development.ini
@@ -17,6 +17,9 @@ debugtoolbar.hosts = 0.0.0.0/0 ::1
 
 mako.directories = fishtest:templates
 
+fishtest.port = 6542
+fishtest.primary_port = 6542
+
 ###
 # wsgi server configuration
 ###
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -41,7 +41,13 @@ def main(global_config, **settings):
         for j in static_full_path(i).glob(f"*.{i}")
     }
 
-    rundb = RunDb()
+    port = int(settings.get("fishtest.port", -1))
+    primary_port = int(settings.get("fishtest.primary_port", -1))
+    # If the port number cannot be determined like during unit tests or CI,
+    # assume the instance is primary for backward compatibility.
+    is_primary_instance = port == primary_port
+
+    rundb = RunDb(port=port, is_primary_instance=is_primary_instance)
 
     def add_rundb(event):
         event.request.rundb = rundb
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -44,46 +44,8 @@ boot_time = datetime.now(timezone.utc)
 last_rundb = None
 
 
-def get_port():
-    params = {}
-    args = sys.argv
-    if len(args) <= 1:
-        return -1
-    if os.path.basename(args[0]) != "pserve":
-        return -2
-    config = args[1]
-    for arg in args[2:]:
-        if arg[0] == "-" or "=" not in arg:
-            continue
-        arg = arg.split("=")
-        params[arg[0].strip()] = arg[1].strip()
-    c = configparser.ConfigParser(defaults=params)
-    try:
-        c.read(config)
-    except:
-        return -3
-    try:
-        section = c["server:main"]
-    except:
-        return -4
-    if "port" in section:
-        port = section["port"]
-    elif "listen" in section:
-        listen = section["listen"]
-        port = listen.split(":")[-1]
-    else:
-        return -5
-    try:
-        return int(port)
-    except:
-        return -6
-
-
-PRIMARY_INSTANCE_PORTS = [6542, 6543]
-
-
 class RunDb:
-    def __init__(self, db_name="fishtest_new"):
+    def __init__(self, db_name="fishtest_new", port=-1, is_primary_instance=True):
         # MongoDB server is assumed to be on the same machine, if not user should
         # use ssh with port forwarding to access the remote host.
         self.conn = MongoClient(os.getenv("FISHTEST_HOST") or "localhost")
@@ -96,9 +58,7 @@ class RunDb:
         self.nndb = self.db["nns"]
         self.runs = self.db["runs"]
         self.deltas = self.db["deltas"]
-        self.port = get_port()
-        if self.port < 0:
-            print(f"Unable to obtain the port number. Error: {self.port}", flush=True)
+        self.port = port
         self.task_runs = []
 
         self.task_duration = 1800  # 30 minutes
@@ -114,6 +74,8 @@ class RunDb:
 
         if self.port >= 0:
             self.actiondb.system_event(message=f"start fishtest@{self.port}")
+
+        self.__is_primary_instance = is_primary_instance
 
     def new_run(
         self,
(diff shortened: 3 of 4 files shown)


ppigazzini:
Can we drop this message? It's annoying when running a script that loads RunDB.

```python
        if self.port < 0:
            print(f"Unable to obtain the port number. Error: {self.port}", flush=True)
```

vdbergh:
I am not ať home but please feel free to edit the pr  to remove it yourself.

vdbergh:
I have squashed the PR. I have also moved the logic of recognizing a primary instance to `__init__.py`.  `rundb.py` does not need to know this logic.

ppigazzini:
DEV updated, scripts working fine
```bash
usr00@dfts-0:~/fishtest/server/utils$ ${VENV}/bin/python3 delta_update_users.py
update scan
load deltas:
len(deltas)=155220
next(iter(deltas))='663f8608976f9c59aadb72ae'
Successfully updated 2369 users
Successfully updated 103 top month users
usr00@dfts-0:~/fishtest/server/utils$ ${VENV}/bin/python3 purge_pgn.py
Finished runs:
kept  :   267 runs,  437175 tasks,       4 pgns
purged:  2427 runs,  709720 tasks,       0 pgns
Deleted runs:
kept  :    58 runs,    3723 tasks,       0 pgns
purged:   153 runs,    3665 tasks,       0 pgns
Unfinished runs:
kept  :    38 runs,   17481 tasks,     101 pgns
purged:     0 runs,       0 tasks,       0 pgns
{'bytesFreed': 1232896, 'ok': 1.0}
```

vdbergh:
We are not quite there with the config.

I am realizing that for most configurations options (e.g. `task_duration`) it makes no sense to put them in `[app:main]` as they also need to be defined if the `rundb` object is constructed directly instead of via `pserve`.

So it seems we need a config file `fishtest.ini`. It should not be part of git but perhaps `Fishstest` should write a default one if it does not exist or is incompatible. Not completely trivial. But of course the worker already does something very similar...

peregrineshahin:
I feel this is getting overthinked too much.
There is a difference between different kind of configs, enviroment/deployment related configs are something different than hardcoded constants that can be used in many places,the idea of the first can be hidden from git because it can contain sensitive data, the idea of the other config is to have a place where we can change hardcoded data without the need of looking into the code... I don't see why such config would need not to be in git.. because those should already only be development related or production related and has nothing to do with the internals of the application.
Also we shouldn't write any config runtime if there isn't one for the server bc it should be used for reading only .. and different pserves should not be a limitation at all.

vdbergh:
The reason I find it inconvenient to have a config file in git that it is more cumbersome to change things locally (e.g. `task_duration`). Before making a PR you have to change it back.

peregrineshahin:
The app depends on the config.. there is no way the app will work if you remove the hardcoded url from the config at all which should be rightful.
Suppose I have
The website url for example, and that's used in a ton of places
I would find it more convenient if I have my local instance to change it from one place..
Removing it might cause run time error or failure, but that's on me.. I would be the one who messed up the config.
How can I know what variables to change if I don't see the config at all because it's hidden from git.
Also the repo/branch used for development and PRing should not be the same one someone use..

peregrineshahin:
The worker config is different.. the code will already work without it.. because you have some default values in mind that you can suggest to the worker.
that should be not the point of the server config.
The server will not know how much task duration should be without us having such config.

vdbergh:
> The server will not know how much task duration should be without us having such config.

Internally the worker has of course defaults stored. The config file allows the user to change these defaults. Something like that is also convenient for a server. There should be a `default task_duration` for production servers, but for local testing it is convenient to make it much smaller.

vdbergh:
But it seems possible to teach git to ignore changes in tracked files. So my earlier point about not having the config file in git is moot.

peregrineshahin:
We can ofc put it in git ignore for us developers. But it should be in the official repo to know what to change...

peregrineshahin:
You can add default values but that's called error handling.. it makes no sense for example to default the weburl of the website itself to be our production url.

vdbergh:
> We can ofc put it in git ignore for us developers. But it should be in the official repo to know what to change...

.gitignore does not work for tracked files. You have to use something like this https://stackoverflow.com/questions/13630849/git-difference-between-assume-unchanged-and-skip-worktree .

But I don't understand why you keep insisting that it _should_ be in GIT. The worker config file is not in GIT and it works perfectly fine.

peregrineshahin:
Simply because there are tons of variables that I want to add to the config that are hardcoded.. making a default value is not serving my purpose at all because it simply hardcoding a value again.. you get me into infinite loop if the config is not in the repo

vdbergh:
The issue of default values is not relevant for what we are discussing.  For some values it is sensible to have a default value (e.g. task_duration). For others not (e.g. hostname). Obviously the worker does not contain a default value for the password. And still the config file works fine.

vdbergh:
I simply don't see what's wrong with having the server generate a skeleton config file, which the user can modify (this is what the worker does).

I am not saying that this is the best solution, but you seem to claim that this solution is unreasonable. That I don't accept.

vdbergh:
There are many programs that can generate their own config files. E.g. https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html .

peregrineshahin:
The skeleton config would have no bases in much case.. why are you insisting on overcomplexifying a simple thing.
Why maintain two configs when you can maintain one.. whoever wants runs the server is technically well versed.. it's not someone who runs his worker without technical knowledge.

What would I put as a skeleton for the bench regexes?
Why should be duplicate the config?

vdbergh:
I don't see why the bench regexes should under control of the user... I also don't want to put the schemas in schemas.py under control of the user.

peregrineshahin:
Do you understand that a user should not run a server to begin with.... it's for development its for us.. 
Whoever runs his own fork or instance is whoever wants to make the change... not a user 
Its an admin. Its a developer.
If we want to change a regex from 5 digits to 7 digits it should be in a config file and not hardcoded in many places.. thats the idea of a server config. That can be used thruout the app

vdbergh:
> If we want to change a regex from 5 digits to 7 digits it should be in a config file and not hardcoded in many places.. thats the idea of a server config. That can be used thruout the app

Simple: define it as a schema in `schemas.py`. I don't see why this should be in a separate config file.

vdbergh:
What I am saying is that files like `schemas.py` are in some sense already config files. So it is not necessary to duplicate their content elsewhere.

peregrineshahin:
The point of a config on the server has became an industrial standards, I deal with tons of non-developers (devOps, network people, technical support).
I always leave them a config file to change trivial settings that are sharely common to use in the app.
Even translations and labels, are configurable in modern days.

Since then, no client comes back asking about trivial changes because they don't know html, js, python.. they don't understand the code and don't want to.
Now when it comes to me... I already find this convenient also. We can't keep arguing about niche details... my whole purpose of the config is to enhance the developer experience and maintainablity.. 
Just my 2 cents..

vdbergh:
Ah sorry. You probably want to extract the bench. Vtjson cannot do that (yet). It is strictly for validation.

I would just have a file config.py containing common definitions. It is simple and it would serve your purpose.

ppigazzini:
> But it seems possible to teach git to ignore changes in tracked files. So my earlier point about not having the config file in git is moot.

I already rebuild PROD fishtest from scratch (deleting the folder and cloning the repo) with a PR that changes `setup.py`, `production.ini`, and `__init__.py`, so `.gitignore` is not a problem.
I rebuild DEV fishtest (server and worker) from scratch for any PR test.

ppigazzini (approved):
Looks good with DEV+workers and scripts.

PR #2005, opened 2024-05-14
Fix some capitalization

MrBrain295:
It looked weird not capitalizing the first word of a sentence.


ppigazzini:
Thank you @MrBrain295

PR #2006, opened 2024-05-14
remove unnecessary loop

peregrineshahin:
and change list to generator expression

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -793,12 +793,7 @@ After fixing the issues you can unblock the worker at
         # To limit db access the list is cached for
         # 60 seconds.
 
-        runs_finished = True
-        for run in self.task_runs:
-            if not run["finished"]:
-                runs_finished = False
-                break
-
+        runs_finished = all(run["finished"] for run in self.task_runs)
         if runs_finished:
             print("Request_task: no useful cached runs left", flush=True)
 
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -863,18 +863,18 @@ After fixing the issues you can unblock the worker at
 
         self.task_runs.sort(key=priority)
 
-        # get the list of active tasks
-
-        active_tasks = [
-            task for run in self.task_runs for task in run["tasks"] if task["active"]
-        ]
-
-        # We go through the list of active tasks to see if a worker with the same
-        # name is already connected.
-
         now = datetime.now(timezone.utc)
+        connections = 0
+        connections_limit = self.userdb.get_machine_limit(worker_info["username"])
+
+        active_tasks = (
+            task for run in self.task_runs for task in run["tasks"] if task["active"]
+        )
+
         for task in active_tasks:
             task_name = worker_name(task["worker_info"], short=True)
+
+            # see if a worker with the same name is already connected.
             if my_name == task_name:
                 task_name_long = worker_name(task["worker_info"])
                 my_name_long = worker_name(worker_info)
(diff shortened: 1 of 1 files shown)


ppigazzini:
The changes are not equivalent to the original code in subtle ways:
1. the first code returns the run as well (but it's not used in the following code, so the pythonic simplification is OK)
2. only one loop breaks the priority of the checks and mix the checks, this can make more difficult to debug the Event Logs. In my review of this function, I skipped this optimization because this reason, so I'm in difficulty to approve it in this PR.

vondele:
my 2cents, unrelated to an eventual merging, would it make sense to measure once the impact on the server time needed for this function under load?

ppigazzini:
Only a fast test, rn we have little space left.
```
fishtest@tests:~$ df -h
Filesystem         Size  Used Avail Use% Mounted on
/dev/ploop29670p1   99G   85G  9.7G  90% /
none               2.5G     0  2.5G   0% /sys/fs/cgroup
none               2.5G     0  2.5G   0% /dev
tmpfs              2.5G     0  2.5G   0% /dev/shm
tmpfs              2.5G  1.2M  2.5G   1% /run
tmpfs              5.0M     0  5.0M   0% /run/lock
none               2.5G     0  2.5G   0% /run/shm
tmpfs              512M     0  512M   0% /run/user/1000
```

vondele:
should probably work, some workers will need to request a task twice. I can monitor.

What about storing STC tests only for a day?

peregrineshahin:
> 2a. only one loop breaks the priority of the checks and mix the checks, this can make more difficult to debug the Event Logs.
2b. the code is a bit less readable

a) I don't think there is an inherent priority for the checks
b) in case you want another reason to return from this function with an error, it becomes unmaintainable to add yet another loop that is the same to do it, it's not like these two loops are done in a declarative way to make sense of what they are doing there are just duplicate C-like for loops which look bad.

i.e. one can argue that mapping/filtering/reducing adds to the readability and thus can be done twice, not the case here IMO.

ppigazzini:
I have no opinion on how long to retain the PGNs, as I don't know their use. Feel free to open a PR changing the numbers.

https://github.com/official-stockfish/fishtest/blob/1a3928fde2ddbf790a5d313b15fe64a3eb3ea743/server/utils/purge_pgn.py#L70-L86

robertnurnberg:
I think keeping STC just for 1 day, and LTC just for 7 days is reasonable.

Of course, if the fleet would re-appear for extended periods, even that might not be enough to keep the disk from filling up....

peregrineshahin:
This sounds reasonable feel free to override this PR with that, but it's still does not show any edge for having two loops whatsover over one, and also IMO it's confusing and a bad code not to break from loops if the code is already as imperative as a `for` when we know we are already done.

> If a worker contributor has both the issues (workers duplicated and connection limit) the new function could report the connection limit (a low priority warning) before the duplicated workers (higher priority warning, bad setup). A solution will be to complete the loop and report both the issues instead than exit as fast as possible like now with an incomplete report (with the possible wrong priority). In case of worker without issues, the loop is always completed.

vdbergh:
> it's not like these two loops are done in a declarative way to make sense of what they are doing they are just duplicate C-like for loops which look bad.

As usual we disagree... I don't think merging unrelated checks in a single loop necessarily makes the code clearer.

Basically we have two distinct checks, which both accidentally happen to involve the same loop...

peregrineshahin:
> which both accidentally happen to involve the same loop...
accidental occasions are what drive people to generalize code.

PR #2008, opened 2024-05-14
Store STC runs only for 1day

vondele:


vondele:
as suggested https://github.com/official-stockfish/fishtest/pull/2006#issuecomment-2109881731 

might need a look at the logic. Goal is to have STC for 1 day (down from 2), while keep LTC as it was (10 days)

ppigazzini:
```python
    cutoff_date_ltc = now - timedelta(days=5 * days)
    cutoff_date = now - timedelta(days=days)
...
        keep = (
            not deleted
            and finished
            and tc_regex.match(run["args"]["tc"])
            and run["last_updated"] > cutoff_date_ltc
        ) or run["last_updated"] > cutoff_date
...        
        
    # Process the runs in descending order of last_updated for the
    # last 60 days and purge the pgns collection for:
    # - runs that are finished and not deleted, and older than 2 days for STC --- first run of the function
    # - runs that are finished and not deleted, and older than 10 days for LTC --- first run of the function
    # - runs that are finished and deleted, and older than 10 days --- second run of the function
    # - runs that are not finished and not deleted, and older than 50 days --- third run of the function

    rundb = RunDb()
    out = purge_pgn(rundb=rundb, finished=True, deleted=False, days=2)
    report("Finished runs:", *out)
    out = purge_pgn(rundb=rundb, finished=True, deleted=True, days=10)
    report("Deleted runs:", *out)
    out = purge_pgn(rundb=rundb, finished=False, deleted=False, days=50)
    report("Unfinished runs:", *out)
    msg = rundb.db.command({"compact": "pgns"})
    print(msg)
```

`cutoff_date_ltc` is used only in the first run of the function, courtesy of `finished=True, deleted=False`.
Of course, it could be simplified a bit passing explicitly another parameter `ltc_days` to be used only in the first run. When `None`, falsify the first condition of `keep`.

vondele:
> `cutoff_date_ltc` is used only in the first run of the function, courtesy of `finished=True, deleted=False`. 

right, that's how I understood it as well, so that would mean the patch is correct.

ppigazzini:
Yes, the patch is correct. I wrote the function for the first run and for serendipity it worked for the other runs with a bit of convoluted logic in `keep`. Admittedly I chose the lazy way, but I was tired after reverse engineering the original, convoluted and broken, version of the script.

ppigazzini (approved):

ppigazzini:
PROD updated, thank you @vondele

PR #2011, opened 2024-05-14
two-sided resign adjudication

gahtan-syarif:
enable two-sided adjudication, meaning both engines have to agree on the resignation for it to happen. i believe it is fairer this way.

to be more specific, i think of a scenario where one of the engine might have found a line that could lead to it losing, but the other engine might not find that line, by having twosided resignation, this prevents the more "pessimistic" stockfish from being wrongly adjudicated.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1508,6 +1508,7 @@ def run_games(worker_info, password, remote, run, task_id, pgn_file, clear_binar
                     "-resign",
                     "movecount=3",
                     "score=600",
+                    "twosided=true",
                     "-draw",
                     "movenumber=34",
                     "movecount=8",


peregrineshahin:
Tbh a player forcing another to continue the game is not the most fun experience ever.
Also it looks like it hacks its way into opponent's mind and make decision based on if they see the win or not, but OTOH draw adjudication can be interpreted as a draw offer/agreement.
My 2 cent is that it doesn't really matter.. at the best case it just delays the game by some moves which is not very desired.

gahtan-syarif:
im primarily concerned on cases where the games would be drawable if the match had been kept on instead of resigning the game when one engines eval is low enough, but i guess yeah it delays games by a a ply or two in most cases.

the best way to see whether this patch would be helpful or not is to do a side by side comparison of test results at a high number of games between twosided=false and twosided=true, if the resulting elo difference between the two tests exceeds the error margin, it means setting twosided=false have been resigning a significant amount of games where it should have been a draw.

gahtan-syarif:
on a sidenote, i also wonder if the resign score of 600 can also be lowered to 400 like in OB to end matches faster, this can also be tested using the aforementioned side by side elo test

dubslow:
i thought it already was twosided, and i certainly believe that it should be important in admittedly rare edge cases. but the point of tests is that we dont know which player is better, so in principle any adjudication must result from an agreement between the players.

that said, actual data would be nice too, and i certainly think we should collect it.

vondele:
we know from recent measurements that adjudication with current settings has essentially no effect on Elo difference for the PT (agreement within error bars). Generally, one needs to agree on a draw and resigning is a one-sided decision, so the current settings make sense. Having said that, the direction should be to make adjudication more and more rare, so this would be a step in the right direction. Reducing the threshold is not a good direction (in fact I think this would revert an older change).

gahtan-syarif:
the decision to implement this i leave fully to the maintainers, and with that id like to end the note with the idea that the goal of the resign adjudication is not so much to simulate a human game where one side can quit at any given point, but more so to save resources by having the test end when it is clear that the outcome of the game would no longer change and proceeding with the match is fully a waste of time and cores. having one-sided resignation is a flaw that goes against this principle, as one engine still thinks that the match hasn't been decided, and as such i believe ending it early where the outcome still hasnt been fully decided goes against the spirit of the test

dubslow:
On further thought I quite agree with the comment that comparison with human games is not valid. Resigning in *match* play is one-sided, however this is *not* match play -- we need to be sure that the "winner" of the resignation is in fact able to convert the "won" position. Having both players in full agreement on the game is the only valid way to measure a testing result.

I vote that this PR be merged.

vondele:
> we need to be sure that the "winner" of the resignation is ... Having both players in full agreement on the game is the only valid way to measure a testing result.

both ways are an approximation, there is no certainty unless the game is played to mate. Do you have any data on wrong adjudication with the old scheme, and with the newly proposed scheme?

dubslow:
not sure anyone has any data on old or new lol

vondele:
people ran non-adjudicated on fishtest, downloaded games, and analyzed how much would have been falsely adjudicated.

PR #2013, opened 2024-05-15
Make buffer() a noop on a non-primary instance.

vdbergh:
So a non-primary instance will never accidentally write to the rundb.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -353,6 +353,13 @@ class RunDb:
         self.timer.start()
 
     def buffer(self, run, flush):
+        if not self.is_primary_instance():
+            print(
+                "Warning: attempt to use the run_cache on the",
+                f"secondary instance with port number {self.port}!",
+                flush=True,
+            )
+            return
         with self.run_cache_lock:
             if self.timer is None:
                 self.start_timer()


server/fishtest/rundb.py
```diff
@@ -353,6 +353,13 @@ def start_timer(self):
         self.timer.start()
 
     def buffer(self, run, flush):
+        if not self.is_primary_instance():
+            print(
+                "Warning: attempt to use the run_cache on the "
```
ppigazzini, line 358: I didn't know the automatic concatenation of string literals, it's surprising that this works:
```python
a ="0" "1"   "2"
```
This syntax is easier to my eyes but I'm open to other views:
```diff
            print(
-                "Warning: attempt to use the run_cache on the "
            print(
+                "Warning: attempt to use the run_cache on the",
```

vdbergh: No strong feelings about this so I changed it. Note however that the comma trick will only work in print statements...

ppigazzini (approved):
Looks good on DEV.
```diff
index 0723e34..695dc49 100644
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -602,6 +602,8 @@ class UserApi(GenericApi):
     def download_nn(self):
         nn_id = self.request.matchdict["id"]
         nn = self.request.rundb.get_nn(nn_id)
+        run = None
+        self.request.rundb.buffer(run, True)
         if nn is None:
             self.handle_error(
                 f"The network {nn_id} does not exist", exception=HTTPNotFound
```
```yaml
May 16 16:02:28 dfts-0 pserve[12422]: Warning: attempt to use the run_cache on the secondary instance with port number 6545!
```

ppigazzini:

PR #2015, opened 2024-05-16
Strip URLs before adding slashes

peregrineshahin:
this adds the practice of always stripping before adding slashes to the URLs.. one example is that the run page doesn't work with a repo with a slash at the end.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -772,7 +772,9 @@ def get_valid_books():
 
 def get_sha(branch, repo_url):
     """Resolves the git branch to sha commit"""
-    api_url = repo_url.replace("https://github.com", "https://api.github.com/repos")
+    api_url = repo_url.replace(
+        "https://github.com", "https://api.github.com/repos"
+    ).rstrip("/")
     try:
         commit = requests.get(api_url + "/commits/" + branch).json()
     except:
@@ -787,7 +789,7 @@ def get_nets(commit_sha, repo_url):
     """Get the nets from evaluate.h or ucioption.cpp in the repo"""
     api_url = repo_url.replace(
         "https://github.com", "https://raw.githubusercontent.com"
-    )
+    ).rstrip("/")
     try:
         nets = []
         pattern = re.compile("nn-[a-f0-9]{12}.nnue")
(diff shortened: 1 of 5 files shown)


dubslow:
In general, URL manipulation has specified [RFC semantics](https://datatracker.ietf.org/doc/html/rfc3986.html#section-5.4.1), and we should be using libraries to do that, e.g. https://docs.python.org/3/library/urllib.parse.html (specifically 

For example:

```
def get_sha(branch, repo_url):
    """Resolves the git branch to sha commit"""
    urlparts = urllib.urlsplit(api_url)
    urlparts.netloc = urlparts.netloc.replace(
        "github.com", "api.github.com"
    )
    api_url = urllib.urlunsplit(urlparts)
    get_url = urllib.urljoin(api_url, f"/repos/commits/{branch}")
    try:
        commit = requests.get(get_url).json()
    except:
        raise Exception("Unable to access developer repository")
    if "sha" in commit:
        return commit["sha"], commit["commit"]["message"].split("\n")[0]
    else:
        return "", ""```

that way we delegate the handling of slashes to RFC standards to people who've already done the legwork

**Comment by peregrineshahin** [comment]
that's a good suggestion, I will draft this and code that later.

**Comment by ppigazzini** [comment]
The worker (python 3.6.x) stops, wrong url, missing the repo info.
- PR, wrong:
```bash
Running master vs master
Variable task sizes used. Opening offset = 14196
Downloading https://api.github.com/zipball/6138a0fd0e43753a86e4a170a5f6e2b7b6752677
Exception in requests.get():
404 Client Error: Not Found for url: https://api.github.com/zipball/6138a0fd0e43753a86e4a170a5f6e2b7b6752677

Exception running games:
Get request to https://api.github.com/zipball/6138a0fd0e43753a86e4a170a5f6e2b7b6752677 failed
```
- master:
```
Running master vs master
Variable task sizes used. Opening offset = 14022
Downloading https://api.github.com/repos/official-stockfish/Stockfish/zipball/6138a0fd0e43753a86e4a170a5f6e2b7b6752677
Build uses default net: nn-ddcfb9224cdb.nnue
```

peregrineshahin:
it looks like this library gives 3 different combination results of joining two URLs when using `.join()`,
1) if the base URL ends with `/`
2) if the second URL begins with a slash
3) if no slashes are used in either
so stripping the slash at the end is still necessary and doesn't solve anything.

I'm not going to revert but it's a bummer.. hopefully I'm missing something

peregrineshahin:
@dubslow unfortunately your example seems to give three different results based on the slashes, I don't think unsplitting and then splitting, and then joining is reasonable even if it's done in a helper function..
```
    api_url = urllib.urlunsplit(urlparts)
    get_url = urllib.urljoin(api_url, f"/repos/commits/{branch}")
```

Issue #2003, opened 2024-05-13
Document load under fleet


vondele:
Just to keep some numbers for later reference, this is the average time in ms needed server side when the fleet is present (but not overloading) the server (50k cores), as obtained from the worker output (`Post request https://tests.stockfishchess.org:443/api/request_task handled in 1719.69ms (server: 265.24ms)`): 

```
failed_task 13.6646
request_spsa 4.04172
request_task 149.143
request_version 0.241336
update_task 4.36213
upload_pgn 92.8058
```

corresponding observation on the worker side:
```
failed_task 742.879
request_spsa 319.939
request_task 756.641
request_version 627.497
update_task 539.884
upload_pgn 1213.44
```

The sampling of the traces, for calls not leading to a lock looks like:
```
21
     File "rundb.py", line 432, in flush_buffers self.runs.replace_one(
7
     File "api.py", line 221, in request_task result = self.request.rundb.request_task(worker_info)
          File "rundb.py", line 727, in request_task return self.sync_request_task(worker_info)
               File "rundb.py", line 738, in sync_request_task w = self.workerdb.get_worker(my_name)
                    File "workerdb.py", line 17, in get_worker r = self.workers.find_one(
5
     File "api.py", line 221, in request_task result = self.request.rundb.request_task(worker_info)
          File "rundb.py", line 727, in request_task return self.sync_request_task(worker_info)
               File "rundb.py", line 902, in sync_request_task committed_games += (
4
     File "api.py", line 221, in request_task result = self.request.rundb.request_task(worker_info)
          File "rundb.py", line 727, in request_task return self.sync_request_task(worker_info)
               File "rundb.py", line 870, in sync_request_task if connections >= connections_limit:
3
     File "api.py", line 221, in request_task result = self.request.rundb.request_task(worker_info)
          File "rundb.py", line 727, in request_task return self.sync_request_task(worker_info)
               File "rundb.py", line 837, in sync_request_task task_name = worker_name(task["worker_info"], short=True)
                    File "util.py", line 47, in worker_name name = "{}-{}cores".format(username, cores)
```

Essentially all stack traces go via request_task.

vondele:
And some api times of the server under load (IIRC 75k cores) but with various stages of 'distress'
medium:
```
# logging from [08/May/2024:18:26:11 +0000] to [08/May/2024:18:27:30 +0000] ########
# duration (seconds)            : 79
# calls in total                : 10000
# calls per second              : 126
[...]

PR #2016, opened 2024-05-16
Set semaphore to 2 for "request_task" API

ppigazzini:
Tested working fine with a fleet of 75k cores 10k machines,
Waitress configured with threads=4 and connection_limit=100
HW usage: RAM 72%, primary Waitress instance CPU 85-95%

This is a good compromise for threads "privileged" for "request_task" API, and threads free for any other APIs, while keeping Waitress threads=4.

Raising the number of Waitress threads uses more RAM and can requires lowering the MongoDB cache size to avoid the OOM killer.

see #2003

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -678,8 +678,10 @@ class RunDb:
         run["workers"], run["cores"] = workers, cores
 
     # Limit concurrent request_task
+    # The semaphore must be initialized with a value
+    # less than the number of Waitress threads.
     task_lock = threading.Lock()
-    task_semaphore = threading.Semaphore(4)
+    task_semaphore = threading.Semaphore(2)
 
     task_time = 0
     task_runs = None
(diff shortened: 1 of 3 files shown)


xoto10:
Nice :)

Do we have any information on how long the request_task calls take if rejected by the semaphore?

I assume it's very fast, but it helps to theorise about what's happening if we had a rough idea of the real number.

vondele:
> Do we have any information on how long the request_task calls take if rejected by the semaphore?
```
handled in 370.54ms (server: 210.34ms)
handled in 319.99ms (server: 151.06ms)
handled in 339.38ms (server: 158.88ms)
handled in 280.25ms (server: 52.08ms)
handled in 235.64ms (server: 1.82ms)
handled in 446.53ms (server: 1.25ms)
handled in 484.77ms (server: 8.73ms)
handled in 798.18ms (server: 138.25ms)
handled in 292.24ms (server: 119.02ms)
handled in 376.86ms (server: 0.72ms)
handled in 1307.80ms (server: 5.86ms)
handled in 406.99ms (server: 1.32ms)
handled in 155.74ms (server: 6.02ms)
handled in 551.27ms (server: 278.69ms)
handled in 405.88ms (server: 0.93ms)
handled in 534.28ms (server: 253.10ms)
```
so probably 1ms on reject, 200ms on average when passing

PR #2018, opened 2024-05-16
Require confirmation before rejecting users

peregrineshahin:
this adds a confirmation dialog with the approver to confirm rejecting a new user since the decision is sensitive.

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -4,19 +4,21 @@
   document.title = "User Management | Stockfish Testing";
 </script>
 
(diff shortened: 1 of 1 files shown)


peregrineshahin:
Capture

peregrineshahin:
I gotta say that this whole right alignment for labels to fake a center alignment in the blue, is a Kindergarten level web-development, blue box looks very ugly and should be thrown away and redesigned.

ppigazzini (approved):

ppigazzini:

PR #2020, opened 2024-05-18
Use active_run_lock(run) to protect updates of aggregated data.

vdbergh:
We changed the active_run_locks to be reentrant.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -72,20 +72,21 @@ class RunDb:
 
         self.__is_primary_instance = is_primary_instance
 
-        self.task_lock = threading.Lock()
         self.request_task_lock = threading.Lock()
         self.timer = None
         self.timer_active = True
 
     def set_inactive_run(self, run):
-        with self.task_lock:
+        run_id = run["_id"]
+        with self.active_run_lock(str(run_id)):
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):
Looks good on DEV

vdbergh:
@ppigazzini Can you check in the server log if there are messages like this
https://github.com/official-stockfish/fishtest/blob/a51582a974c960ceef2695da345d07c461aa2b33/server/fishtest/rundb.py#L1202-L1209

I am trying to debug this https://tests.stockfishchess.org/actions?action=log_message&user=&text=cores

vdbergh:
Thanks. I will fix this.

vondele:
This could possibly happen if the worker sends an update_task, the rundb is updated, but the full transaction fails (e.g. nginx timeout), in which the worker sends the update_task again?

vdbergh:
> This could possibly happen if the worker sends an update_task, the rundb is updated, but the full transaction fails (e.g. nginx timeout), in which the worker sends the update_task again?

No this happens because for performance reasons `update_task()` and `request_task()` are not synchronized. But with PR #2020 they are at least partially synchronized so that we can now do something in `request_task()`. See #2021 .

PR #2021, opened 2024-05-18
Do not assign a task for a finished run.

vdbergh:


ppigazzini:

PR #2022, opened 2024-05-19
Danger alert for mismatching options

peregrineshahin:
this gives a warning for mismatching options
a patch with hash=512 vs hash=64 has made its place in Stockfish master
the yellow highlight is for demonstration purposes (done with paint, I don't think we need any).
Capture

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -249,6 +249,12 @@
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


ppigazzini:
Spaces before/after/between options trigger the Danger.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

peregrineshahin:
> Spaces before/after/between options trigger the Danger. EDIT: fixed.

this should be checked in saving the run when creating too, stripping those spaces

PR #2024, opened 2024-05-19
Set days_ltc in clear way and rename script

ppigazzini:
The new name reflects that we are purging the "pgns" collection. The scheduled task must be updated with the new name.


ppigazzini:
- [x] PROD scheduled job updated with the new script name 
 - [x] wiki server setup script updated

PR #2027, opened 2024-05-19
Enhance ugly design

peregrineshahin:
old
Capture2
vs new
Capture

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -25,24 +25,18 @@
       <h4 class="alert-heading">
         <a href="/tests/user/${user['username']}" class="alert-link col-6 text-break">${user['username']}</a>
       </h4>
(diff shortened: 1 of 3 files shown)


vondele:
while at it, maybe we don't need the registration time with microsecond resolution displayed?

peregrineshahin:
> while at it, maybe we don't need the registration time with microsecond resolution displayed?

image

ppigazzini:
image

peregrineshahin:
I cannot reproduce that at all, on firefox, edge, and chrome

peregrineshahin:
nvm I replecated that.

peregrineshahin:
should be fixed now

ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

PR #2028, opened 2024-05-19
This adds a 404 page

peregrineshahin:
I personally prefer a wide banner just like GitHub in https://github.com/ghost/dummy , but this is good enough to have as a placeholder, feel free to overwrite me with a design.
Capture


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #2032, opened 2024-05-21
Show more info about user and in profile

peregrineshahin:
shows tests repo and the groups the user belong to

### DIFF
--- a/server/fishtest/templates/signup.mak
+++ b/server/fishtest/templates/signup.mak
@@ -86,6 +86,32 @@
       <label for="email" class="d-flex align-items-end">Email</label>
     </div>
 
+    <div class="input-group mb-3">
+      <div class="form-floating">
+        <input
+          class="form-control"
+          id="tests_repo"
+          name="tests_repo"
+          placeholder="GitHub Stockfish fork URL"
+        >
+        <label for="tests_repo" class="d-flex align-items-end">Tests Repository</label>
+      </div>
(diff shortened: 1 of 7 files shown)


peregrineshahin:
drafting this, since I think registered users should input their GitHub test repo and should be able to change it as discussed before with @vdbergh. will implement that probably today or tomorrow.
it gives also a better anti spam user-regestriation..

vondele:
A bit related, having a separate page (visible only to approvers) with a list of all approvers and their last active date would be useful at times.

peregrineshahin:
> A bit related, having a separate page (visible only to approvers) with a list of all approvers and their last active date would be useful at times.

Yes this would be a nice feature, I think we should know who are other approvers.
that's easily doable, but I think the 'last active' is not available at first glance that, I think we only track 'last active' for users by their worker activity, we probably should also do that for (test submision/test approval/modification) now not sure if we do it.

vondele:
you are probably right we don't have that, other than via a search in the actions database, at least now approvals are being tracked.

vondele:
note that people that contribute resources (cpu time) might not have a github repo, and might be confused about why this is needed? If we add that, consider clarifying that this is only useful for contributors that want to code software.

peregrineshahin:
undrafted, looks good to me on local instance.

Capture3
Capture
Capture2

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #2036, opened 2024-05-27
Implement time control based estimation improvements

MinetaS:
TODO: Disable worker TC scaling if nodestime option exists


MinetaS:
Now workers should handle nodestime correctly as well.

PR #2038, opened 2024-05-27
Warn about too low hash values/tc_ratio

peregrineshahin:
code and idea by @Dubslow, math reviewed by Viz

### DIFF
--- a/server/fishtest/helpers.py
+++ b/server/fishtest/helpers.py
@@ -1,3 +1,6 @@
+from fishtest.util import get_hash, get_tc_ratio
+
+
 def tests_repo(run):
     return run["args"].get(
         "tests_repo", "https://github.com/official-stockfish/Stockfish"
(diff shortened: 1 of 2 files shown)


server/fishtest/helpers.py
```diff
@@ -19,3 +22,23 @@ def diff_url(run):
...
+    target_hash = 64 * tc_ratio / 6
+    # if hash is outside of 5% of the target, emit warning upon submission.
+    return 0.95 <= hash / target_hash <= 1.05
+
+
+def reasonable_run_hashes(run):
+    # if this func returns false, then emit warning to user to verify hashes
+    base_hash = get_hash(run["args"]["base_options"])
+    new_hash = get_hash(run["args"]["new_options"])
+    tc_ratio = get_tc_ratio(run["args"]["tc"], run["args"]["threads"])
```
dubslow, line 43: we should probably add a comment that this doesn't work on asymetric-time tests (which are exceedingly rare)

and ofc i have no idea how nodestime works either

server/fishtest/templates/tests_view.mak
```diff
@@ -249,6 +249,12 @@
         </div>
       </div>
 
+      % if not h.reasonable_run_hashes(run):
+        <div class="alert alert-danger mb-2">
+          Hash is too low for this TC ratio
```
dubslow, line 254: in principle the hash could be too large as well but this is probably fine

server/fishtest/helpers.py
peregrineshahin: > we should probably add a comment that this doesn't work on asymetric-time tests (which are exceedingly rare)
> 
> and ofc i have no idea how nodestime works either

We can disable all these cases in the worst case scenario.
Which IMO should be the way to go since we don't know what to do about nodes_time atm, but time odds could be trivially disabled. (semi-drafting this)

peregrineshahin: Disabled from the warning*

server/fishtest/templates/tests_view.mak
```diff
@@ -249,6 +249,12 @@
         </div>
       </div>
 
+      % if not h.reasonable_run_hashes(run):
+        <div class="alert alert-danger mb-2">
+          Hash options are too low or too high for this TC ratio
```
ppigazzini, line 254: IMO is better to drop "ratio" in the message

ppigazzini (approved):
Looks good on DEV

image

ppigazzini:

Issue #2041, opened 2024-05-29
hanging server


vondele:
overnight the server went in a hanging state. The signature as we have observed previously. pserve 6545 at 100% load. 

Using the new facility to get traces, I sampled both 6543 and 6545. Interestingly, 6545 doesn't show any thread active in a part of the fishtest python code. 6543 however seems to be deadlocked, or at least in fixed patterns in which all threads in fishtest code are waiting on locks.

```
===============================================================
Collecting 600 samples for pserve 6543 spaced 0.5 seconds apart
Total time needed 300 seconds 
Found pid 16765 for pserve 6543
Showing 10 most frequent samples traces: 
127 
     File "rundb.py", line 465, in flush_buffers self.scavenge(self.run_cache[oldest]["run"]) 
          File "rundb.py", line 491, in scavenge self.set_inactive_task(task_id, run) 
               File "rundb.py", line 86, in set_inactive_task with self.active_run_lock(str(run_id)): 
127 
     File "api.py", line 309, in beat self.validate_request() 
          File "api.py", line 129, in validate_request run = self.request.rundb.get_run(run_id) 
               File "rundb.py", line 363, in get_run with self.run_cache_lock: 
127 
     File "api.py", line 238, in update_task result = self.request.rundb.update_task( 
          File "rundb.py", line 1173, in update_task return self.sync_update_task(worker_info, run_id, task_id, stats, spsa) 
               File "rundb.py", line 1303, in sync_update_task self.buffer(run, False) 
                    File "rundb.py", line 398, in buffer with self.run_cache_lock: 
127 
     File "api.py", line 237, in update_task self.validate_request() 
          File "api.py", line 129, in validate_request run = self.request.rundb.get_run(run_id) 
               File "rundb.py", line 363, in get_run with self.run_cache_lock: 
127 
     File "api.py", line 221, in request_task result = self.request.rundb.request_task(worker_info) 
          File "rundb.py", line 758, in request_task return self.sync_request_task(worker_info) 
               File "rundb.py", line 1063, in sync_request_task self.buffer(run, False) 
                    File "rundb.py", line 398, in buffer with self.run_cache_lock: 
```
I have the feeling 6543 is in a deadlock situation.

dubslow:
[...]

PR #2029, opened 2024-05-20
Every three minutes validate a random run from the cache.

vdbergh:
The reason for this PR is that before introducing more dynamic updates we would like to verify that the dynamic updates of run["workers"] and run["cores"] introduced in
https://github.com/official-stockfish/fishtest/pull/2010 are bug free.

In order to facilitate the use of periodic timers in Fishtest we have introduced a new scheduler in utils.py which may be interesting in its own right.

Currently the timers defined are the cache flush timer and the timer introduced in this PR.

PR #2042, opened 2024-05-29
Presumably a fix for deadlock (see #2041).

vdbergh:
We use a separate lock to update aggregates. To this end we extend self.active_run_lock() with an optional argument `name` to be able to have different locks associated with the same `run_id`.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -76,14 +76,14 @@ class RunDb:
 
     def set_inactive_run(self, run):
         run_id = run["_id"]
-        with self.active_run_lock(str(run_id)):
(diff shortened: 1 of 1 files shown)


vdbergh:
If #2029 would be merged then a better fix would be to do the scavenging of dead tasks as a separate scheduled task (as mentioned in https://github.com/official-stockfish/fishtest/pull/2029#issuecomment-2121582490). This does not need the `run_cache_lock` so the cycle would be broken in another way.

vdbergh:
I have drafted this since #2029 is now being tested.

vdbergh:
I am closing this now in view of the merging of #2029

PR #2044, opened 2024-05-29
Refactor minimum requirement version for compiler and add logic to clang

Disservin:
Same behaviour as before except that we now reject clang in case it is older than 8.0 (released 2019), 7.0 wasn't able to compile latest master for me.


ppigazzini (approved):
Looks good with workers and DEV

ppigazzini:
Triggered the workers update, thank you @Disservin

PR #2045, opened 2024-06-01
Add `run["finished"]=True` to `set_inactive_run()`.

vdbergh:
This eliminates a race condition between `stop_run()` and `request_task()`.


ppigazzini:
It's running fine on DEV, I will add it to PROD in 20 minutes, as soon as the MongoDB backup finishes.

ppigazzini:
PROD running with #2045 and #2046

ppigazzini (approved):
Looks good on PROD

PR #2049, opened 2024-06-03
More dynamically updated aggregated data.

vdbergh:
We introduce two new aggregated variables: committed_games and total_games. Both are used in "request_task", the former being a bit more important since it may be computed several times.

Also: some code refactoring. The functions with name "compute_..." that compute the various types of aggregated data are now grouped together in schemas.py.

Also: we have updated the cache_schema as the field "last_scavenge_time" is no longer used. We use the cache_schema to validate the cache during initialization, the idea being that an unused schema will quickly suffer from bitrot and will hence also lose its documentation value.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -101,7 +101,7 @@ def main(global_config, **settings):
             signal.signal(signal.SIGINT, rundb.exit_run)
             signal.signal(signal.SIGTERM, rundb.exit_run)
             rundb.schedule_tasks()
-            rundb.update_workers_cores()
+            rundb.update_aggregated_data()
 
     config.add_subscriber(add_rundb, NewRequest)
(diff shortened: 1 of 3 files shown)


vdbergh:
I will make further PRs involving request_task on top of this one.

vdbergh:
Converted to draft since I may have introduced a bug.

vdbergh:
Bug fixed.

vondele:
will post that here, looks like we have a validation error 

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1717598694.553362&run_id=

The run object 66605cb2c340c8eed7756a00 does not validate: run['workers'] (value:-71) is not greater than or equal to 0

vdbergh:
Since #2052 is already being tested, we can close this.

PR #2056, opened 2024-06-07
Enhancement of validation.

vdbergh:
Many data structure in Fishtest contain references to a run objects. Naively validating such data structures means that we are also validating the embedded run objects, which is wasteful as these are already validated elsewhere.

The latest version of vtjson makes it possible to change subschemas during validation.

In this PR we define the cache_schema as
```python
cache_schema = {
    run_id: {
        "run": set_label(runs_schema, "runs_schema"),
	"is_changed": bool,
        "last_sync_time": ufloat,
        "last_access_time": ufloat,
    },
}
```
and we perform validation via
```
validate(cache_schema, self.run_cache, name="run_cache", subs={"runs_schema": dict})
```
Requires upgrade of vtjson.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -17,7 +17,7 @@ from bson.objectid import ObjectId
 from fishtest.actiondb import ActionDb
 from fishtest.schemas import (
     RUN_VERSION,
-    cache_schema_fast,
+    cache_schema,
     compute_committed_games,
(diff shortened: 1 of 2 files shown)


ppigazzini (approved):
Looks good on DEV

ppigazzini:
PROD rebuilt from scratch, thank you @vdbergh

PR #2057, opened 2024-06-07
Update task_runs list via a scheduled task with period 1 min.

vdbergh:
Now it is computed (and cached) at the beginning of request_task. This could possibly lead to variation in the duration of request_task.

Also: general cleanup of request_task.

Also: we now maintain a set of unfinished runs so that it is no longer necessary to access the data base to get this list.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -68,7 +68,8 @@ class RunDb:
         self.runs = self.db["runs"]
         self.deltas = self.db["deltas"]
         self.port = port
-        self.task_runs = []
+        self.unfinished_runs = []
+        self.unfinished_runs_lock = threading.Lock()
         self.wtt_map = {}
         self.wtt_lock = threading.RLock()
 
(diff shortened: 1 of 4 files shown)


vdbergh:
I am drafting this I need to think how to make CI work with this.

vdbergh:
Undrafted.

vdbergh:
In a follow up PR I will update the list of unfinished runs dynamically on the main instance, so that it does not require db access. Then only the itp values need to be updated in a schedule tasks. This will reduce side effects.

ppigazzini:
There is a conflict, perhaps after merging #2056

vdbergh:
I am working on this PR. Will rebase later.

vdbergh:
Undrafted and rebased.

ppigazzini (approved):
Looks good on DEV and in 2 hours on PROD,.

ppigazzini:
Thank you @vdbergh

PR #2058, opened 2024-06-09
Retire `self.update_itp_task.schedule_now()`.

vdbergh:
`update_itp()` is a successor to code that also updated the list of unfinished runs but the latter is now managed dynamically.

Itp values move slowly on the Fishtest time scale and there is no need to micro-manage them.


ppigazzini (approved):
Looks good on DEV, PROD updated, thank you @vdbergh

PR #2059, opened 2024-06-10
Update comments; validate whole runs.

vdbergh:
Update comments about administrative flags.
    
Also: update schemas to more rigorously reflect this.

Validate whole run in "validate_random_run".
    
There are some race conditions in Fishtest so this will trigger some (event) log messages. This will help us fix the race conditions.


ppigazzini (approved):
Looks good on PROD

PR #2065, opened 2024-06-12
Replace print by the Python logging framework (RFC)

vdbergh:
Log messages with level >= ERROR give an event log message.

Also: move scheduler to a separate file scheduler.py.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 import math
 import os
 import random
@@ -15,6 +16,7 @@ from bson.codec_options import CodecOptions
 from bson.errors import InvalidId
 from bson.objectid import ObjectId
 from fishtest.actiondb import ActionDb
+from fishtest.scheduler import Scheduler
 from fishtest.schemas import (
     RUN_VERSION,
     cache_schema,
@@ -30,8 +32,8 @@ from fishtest.schemas import (
 from fishtest.stats.stat_util import SPRT_elo
 from fishtest.userdb import UserDb
 from fishtest.util import (
+    EventHandler,
     GeneratorAsFileReader,
-    Scheduler,
     crash_or_time,
     estimate_game_duration,
     format_bounds,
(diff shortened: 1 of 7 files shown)


server/fishtest/rundb.py
```diff
@@ -723,13 +725,12 @@ def scavenge_dead_tasks(self):
...
             task = run["tasks"][task_id]
-            print(
+            self.logger.info(
                 "dead task: run: https://tests.stockfishchess.org/tests/view/{} task_id: {} worker: {}".format(
                     run["_id"], task_id, worker_name(task["worker_info"])
                 ),
-                flush=True,
             )
-            self.handle_crash_or_time(run, task_id)
+            self.handle_clrash_or_time(run, task_id)
```
vondele, line 733: typo

ppigazzini:
If I'm not wrong you are using the root logger, the best practice is to log at the module level, see at example
<https://betterstack.com/community/guides/logging/python/python-logging-best-practices/>

```python
class ActionDb:
    def __init__(self, db):
        self.db = db
        self.actions = self.db["actions"]
        self.logger = logging.getLogger(__name__)
```

vdbergh:
I am not using the root logger but a child logger with name "fishtest". I am not using `__name__` since I am using the same logger everwhere.

I had quite a bit of trouble with this PR since the tutorials did not work. The reason seems to be that pserve already does some configuration. Switching to a child logger fixed things.

server/fishtest/rundb.py
vdbergh: Thanks! Fixed.

ppigazzini:
Pyramid example logs at module level, <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/logging.html>

The logger configuration in Pyramid is in the usual ini files:

https://github.com/official-stockfish/fishtest/blob/322eea2e377023df3393d577587b9100666949f0/server/production.ini#L37-L67

https://github.com/official-stockfish/fishtest/blob/322eea2e377023df3393d577587b9100666949f0/server/development.ini#L32-L62

vdbergh:
Hmm that's nice! It seems pyramid already does some of the heavy lifting.

It is not clear to me however how one declares a project specific handler. I am using a handler that posts to the event log.

vdbergh:
https://stackoverflow.com/questions/30301824/add-custom-log-handler-to-pyramid

vdbergh:
@ppigazzini The default logging configuration precedes every line with a time stamp. But your logs already seem to have timestamps (provide by systemd?). So I assume you don't want timestamps in the pyramid logs. Is this correct?

ppigazzini:
Systemd already add the time stamp.

vdbergh:
Another problem is that currently the constructor for EventHandler takes rundb.actiondb as argument (which itself is constructed inside rundb.__init__(). It is not obvious to me how handle this without resorting to global variables. 

I guess the EvenHandler will have to open a new connection to the db in the constructor. Not a big deal.

ppigazzini:
If the best practice is to log at module level, `log` is a global variable that somewhat replace the `print` function:

```python
def foo():
    message = "Error!"
    print(message)
```
```python
import logging
log = logging.getLogger(__name__)

def bar():
    message = "Error!"
    log.debug(message)
```

vdbergh:
I want to define a custom logger that also logs to the event log for `level >= ERROR`. 

Unfortunately I don't know how to do this in Pyramid framework (I tried the last two hours or so but there are errors I don't understand).

vdbergh:
I think I am getting to work now.

vdbergh:
Closing in favor of #2067

vdbergh:
This the code for the custom handler...
```python
class EventHandler(logging.Handler):
    def __init__(self):
        super().__init__(level=logging.ERROR)
        logging.disable(level=logging.CRITICAL)
	rundb = RunDb(is_primary_instance=False)
        logging.disable(level=logging.NOTSET)
        self.actiondb = rundb.actiondb

    def emit(self, record):
        message=self.format(record)
	self.actiondb.log_message(
            username="fishtest.system",
            message=message,
        )
```

PR #2067, opened 2024-06-12
Replace print by the Python logging framework

vdbergh:
We use the logger "fishtest" as configured in production.ini.

For messages to the event log we use the child logger "fishtest.event", also configured in production.ini.

Log messages are formatted as follows:
```
INFO [fishtest] [waitress-0:rundb.py:failed_task:1446] <...message...>
```
Also: move scheduler to a separate file scheduler.py.

### DIFF
--- a/server/development.ini
+++ b/server/development.ini
@@ -35,10 +35,10 @@ port = 6542
 ###
 
 [loggers]
-keys = root, fishtest
+keys = root, fishtest, fishtest.event
 
 [handlers]
-keys = console
+keys = console, events
 
 [formatters]
 keys = generic
@@ -52,11 +52,20 @@ level = DEBUG
 handlers =
 qualname = fishtest
 
+[logger_fishtest.event]
+level = DEBUG
+handlers = events
+qualname = fishtest.event
+
 [handler_console]
 class = StreamHandler
 args = (sys.stderr,)
 level = NOTSET
 formatter = generic
 
+[handler_events]
+class = fishtest.utils.EventHandler
+formatter = generic
+
 [formatter_generic]
-format = %(asctime)s %(levelname)-5.5s [%(name)s][%(threadName)s] %(message)s
+format = %(levelname)s [%(name)s] [%(threadName)s:%(filename)s:%(funcName)s:%(lineno)s]  %(message)s
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -1,7 +1,7 @@
 from datetime import datetime, timezone
 
 from fishtest.schemas import action_schema
-from fishtest.util import hex_print, worker_name
+from fishtest.util import event_log, hex_print, logger, worker_name
 from pymongo import DESCENDING
 from vtjson import ValidationError, validate
 
@@ -219,13 +219,7 @@ class ActionDb:
         try:
             validate(action_schema, action, "action")
         except ValidationError as e:
-            message = (
-                f"Internal Error. Request {str(action)} does not validate: {str(e)}"
-            )
-            print(message, flush=True)
-            self.log_message(
-                username="fishtest.system",
-                message=message,
-            )
+            message = f"Request {str(action)} does not validate: {str(e)}"
+            event_log.error(message)
             return
         self.actions.insert_one(action)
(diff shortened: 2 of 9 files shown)


vdbergh:
This the code for the custom handler...
```python
class EventHandler(logging.Handler):
    def __init__(self):
        super().__init__(level=logging.ERROR)
        logging.disable(level=logging.CRITICAL)
	rundb = RunDb(is_primary_instance=False)
        logging.disable(level=logging.NOTSET)
        self.actiondb = rundb.actiondb

    def emit(self, record):
        message=self.format(record)
	self.actiondb.log_message(
            username="fishtest.system",
            message=message,
        )
```

vdbergh:
The logging section of `production.ini`
```ini
[loggers]
keys = root, fishtest

[handlers]
keys = console, events

[formatters]
keys = generic

[logger_root]
level = DEBUG
handlers = console

[logger_fishtest]
level = DEBUG
handlers = events
qualname = fishtest

[handler_console]
class = StreamHandler
args = (sys.stderr,)
level = NOTSET
formatter = generic

[handler_events]
class = fishtest.rundb.EventHandler
formatter = generic

[formatter_generic]
format = %(levelname)-5.5s [%(name)s][%(threadName)s] %(message)s
```

vdbergh:
Event log output after injection of a schema error in an  event log message.
```
ERROR [fishtest.actiondb][waitress-0] Request {'action': 'failed_task', 'username': 'user01', 'worker': 'user01-1cores-39761dde-4b0d', 'run_id': '666691306a6aedb9a0923317blah', 'run': 'master-c84ada9', 'task_id': 22, 'message': 'Terminated by signal SIGINT', 'time': 1718224271.578238} does not validate: action['run_id'] (value:'666691306a6aedb9a0923317blah') is not of type 'valid_object_id'
```

ppigazzini:
With DEBUG the systemd log is flooded

```ini
[logger_fishtest]
level = INFO
handlers = events
qualname = fishtest
```

I wonder if this is a simpler way to do the intended action...

```python
class EventHandler(logging.Handler):
    def __init__(self):
        super().__init__(level=logging.ERROR)
        rundb = RunDb(is_primary_instance=False)
        self.actiondb = rundb.actiondb

    def emit(self, record):
        if record.levelno >= self.level:  # Check if message level is ERROR or higher
            message = self.format(record)
            self.actiondb.log_message(
                username="fishtest.system",
                message=message,
            )
```

vdbergh:
I am not sure I understand. The handler is already created with level equal to level.ERROR. I don’t understand the need for a further test in emit().

I am AFK but if you want to raise the log levels from DEBUG to INFO then please do!

vdbergh:
I changed the log levels from DEBUG to INFO and fixed a typo.

vdbergh:
I now reverted the root level to ERROR and set the fishtest level to DEBUG so that we can use `logger.debug(message)` inside Fishtest (for example to instrument code). According the specs, this is how it should be.

vdbergh:
Rebased.

vdbergh:
I switched back to using a single main logger (this pattern is also used by other packages). The main logger is the logger “fishtest” which is a child of the root logger which is preconfigured in production.ini. Using a different child logger of “fishtest” per source file did not seem to offer any benefits and moreover it complicates some things as we are also using a logger which logs to the event log. In this PR this logger is called `fishtest.event`. With module level logging it should be called `__name__.event` but then AFAICS it is no longer possible to configure it in production.ini (wildcards in keys appear no to work).

vdbergh:
Rebased.

vdbergh:
Rebased

vdbergh:
@ppigazzini  Could you maybe have a look at this PR? It seems like a nobrainer (replacing print by logger.info) and tying into the Python logger framework has many benefits for interoperability (I plan to do the same for the worker).

But the real reason I am mentioning this is the following: I would like to do some work on the scheduler. This PR moves the scheduler to a separate file `scheduler.py`. That makes this PR difficult to rebase when there are changes to the scheduler in master as they have to be copied manually to the new file which is very error prone.

ppigazzini:
I'm working in stealth mode in this weekend: trying new nginx/systemd settings, instrumenting on the fly the code, documenting in the setup script the new setting successfully tested.

PR #2068, opened 2024-06-13
Introduce a worker global cache

vondele:
introduce a global_cache to reduce downloads from github and the net server if multiple workers having a shared filesystem are run by one user. This should allow a fleet to run without overloading github.

If the user specifies the path that corresponds to the global cache, the network and the source zipball are cached.  They are used if available, and only downloaded (and stored) if not.

Storing the data in the cache has been done in such a way that it becomes available atomically on most file systems.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 239
+WORKER_VERSION = 240
 
 
 @exception_view_config(HTTPException)
--- a/worker/games.py
+++ b/worker/games.py
@@ -135,6 +135,41 @@ def send_sigint(p):
         p.send_signal(signal.SIGINT)
 
 
+def cache_read(cache, name):
+    """Read a binary blob of data from a global cache on disk, None if not available"""
+    if cache == "":
+        return None
+
+    try:
+        return (Path(cache) / name).read_bytes()
+    except Exception as e:
+        return None
+
+
+def cache_write(cache, name, data):
+    """Write a binary blob of data to a global cache on disk in an atomic way, skip if not available"""
+    if cache == "":
+        return
+
+    try:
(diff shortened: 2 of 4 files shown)


vondele:
tested now under load, seems to do its job very well.

ppigazzini:
The PR always adds the `global_cache` field in `fishtest.cfg`.
So, if I'm not wrong:
- `cache` variable is never `None`
- ways to stop using the global cache:
  - running the worker with the introduced flag and a not existent path, at example an empty string (`""`)
  - delete the field `global_cache` from `fishtest.cfg`

vondele:
I can indeed remove the comparison with `None`, just felt it was nicer to have inside the code as a more 'pythonic' why to switch it off for specific downloads, but it is a feature that is not used.

let me know if you prefer that.

vondele:
I can't really comment on this, I always start with a config file from scratch. I guess it should be treated like the other options we have? Which I assume is 1 ?

vdbergh:
> Options:
> 
>     1. adding the instruction to drop the cache (flag with empty string "") and dropping `None` in the conditionals
> 
>     2. make the flag mandatory not adding the `global_cache` field in the config file (in this case the `cache` variable is `None` without the flag)

I would prefer option 1. So `global_cache` being equal to `""` (the default) by convention means: no global cache.

vondele:
I've tried to implement that.

ppigazzini (approved):
Looks good on worker:
- linux
- wsl2
- msys2 bash
- msys2 cmd (wiki installer)

ppigazzini:
I will change a bit the help message when merging because the default is the string written in the config file.
```bash
  -g GLOBAL_CACHE, --global_cache GLOBAL_CACHE
                        Useful only when running multiple workers
                        concurrently: an existing absolute path to be used to
                        globally cache on disk certain downloads, reducing
                        load on github or net server. The default empty string
                        ("") disables using a cache. (default:
                        /home/usr00/_git/__test_folder00/fishtest/cache)
```

ppigazzini:
Triggered the workers update.

PR #2069, opened 2024-06-14
Introduce compute_flags() and validate them.

vdbergh:
Previously the flags `is_green` and `is_yellow` were derived from formatting information in the result of `format_results(run)`. Obviously this is not nice.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -20,6 +20,7 @@ from fishtest.schemas import (
     cache_schema,
     compute_committed_games,
     compute_cores,
+    compute_flags,
     compute_results,
     compute_total_games,
     compute_workers,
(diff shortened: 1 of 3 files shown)


ppigazzini:
The PR upgrades the run version. I wonder if the new version is compatible with the old one or I need to reload the DB after testing the PR on DEV.

vdbergh:
Sorry. I went for a walk.

No it is not necessary to change the db. The reason the run version has been raised in this PR is that I made a small change in the algorithm for the color assignment of fixed length runs. But this will not be noticed in practice.

ppigazzini:
On DEV the PR seems fine, the 2 tests are finished green and red, logs clean.
PROD running with the PR.

ppigazzini (approved):
Looks good on DEV and PROD

ppigazzini:
Thnak you @vdbergh

PR #2070, opened 2024-06-14
Periodically validate Fishtest's internal data structures.

vdbergh:
In this way we guarantee that the schemas in schemas.py are kept up-to-date and hence preserve their documentary value.

Also: switch to string keys for the worker_runs dict. This is what we do elsewhere and it makes the schema more readable:
```python
worker_runs_schema = {
    uuid: {
        run_id: True,
        "last_run": run_id,
    }
}
```
Also: retire the separate variable "purge_count" and include it in the "active_run" dict.

Requires upgrade of vtjson for the set datatype used in "unfinished_runs_schema".

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -17,15 +17,20 @@ from bson.objectid import ObjectId
 from fishtest.actiondb import ActionDb
 from fishtest.schemas import (
     RUN_VERSION,
+    active_runs_schema,
     cache_schema,
     compute_committed_games,
     compute_cores,
     compute_results,
     compute_total_games,
     compute_workers,
+    connections_counter_schema,
     nn_schema,
     pgns_schema,
     runs_schema,
+    unfinished_runs_schema,
+    worker_runs_schema,
+    wtt_map_schema,
 )
 from fishtest.stats.stat_util import SPRT_elo
 from fishtest.userdb import UserDb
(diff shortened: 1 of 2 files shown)


ppigazzini:
I wonder if this type of check should be blocking and done once in CI. In this way we are suggesting that it's fine to run the server with some outdated schemas.

vdbergh:
This is to catch for example a developer using `run["_id"]` instead of `str(run["_id"])` as  a key in new code. It may take a long time before this is noticed. I have been burnt by this myself.

It cannot be done only at startup since many of the data structures start out empty and they are filled while the server is running.

vdbergh:
In other words I see no other way to enforce that the data structures used in Fishtest correspond to the schemas in schemas.py. Sure, currently they do, but as the data structures in Fishtest evolve, people will "forget" to update the corresponding schemas so that they become outdated. 

These are the "non-db" schemas verified in this PR:
```python
cache_schema = {
    run_id: {
        "run": runs_schema,
        "is_changed": bool,  # Indicates if the run has changed since last_sync_time.
        "last_sync_time": ufloat,  # Last sync time (reading from or writing to db). If never synced then creation time.
        "last_access_time": ufloat,  # Last time the cache entry was touched (via buffer() or get_run()).
    },
}

wtt_map_schema = {
    short_worker_name: (runs_schema, task_id),
}

connections_counter_schema = {
    ip_address: suint,
}

unfinished_runs_schema = {
    run_id,
}

active_runs_schema = {
    "purge_count?": suint,
    run_id: {
        "time": ufloat,
        "lock": threading.RLock,
    },
}

worker_runs_schema = {
    uuid: {
        run_id: True,
        "last_run": run_id,
    }
}
```

ppigazzini (approved):
Looks good on DEV and PROD

ppigazzini:
PROD already running with the PR, thank you @vdbergh

Issue #2064, opened 2024-06-12
The 404 page masks error messages


vdbergh:
The api's in api.py sometimes return a HTTPNotFound exception (404) with a body explaining the error. This error message is no longer displayed.

Can we add the error message to the 404 page, or is this not done?

peregrineshahin:
Well, this is unintended. it's a clear bug
but it's not obvious to me how to keep this custom renderer while at the same time return the error as json as before the indrocution of this new page.
I tried 
```
@notfound_view_config(renderer="notfound.mak")
def notfound_view(request):
    if request.exception:
        return request.exception
    else:
        return {}
```
but it adds some additional text like (page not found) and returns the response as text rather than json..

peregrineshahin:
Well I forgot to answer your question, no I don't think it's a good idea to return HTML as response for non view APIs as error.

vdbergh:
I think one would only have to modify `notfound.mak` using the content of `request.exception` to return json. The string value of the exception seems to be a string encoding of the json error.

PR #2073, opened 2024-06-15
Partial fix for #2064.

vdbergh:
I had to sacrifice including `base.mak` since I do not know how to do it conditionally.

### DIFF
--- a/server/fishtest/templates/notfound.mak
+++ b/server/fishtest/templates/notfound.mak
@@ -1,51 +1,60 @@
-<%inherit file="base.mak"/>
+% if json:
+  ${json_body|n}
+% else:
+  <!doctype html>
+  <html lang="en">
+    <head>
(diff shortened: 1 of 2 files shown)


peregrineshahin:
This can be done much cleaner, let me try a fix now that you inspired me how to actually get back the json and make the content type json conditionally.

vdbergh:
The issue is not the content type but the fact that inheriting from `base.mak` happens unconditionally.

vdbergh:
Anyway I happy to see a better fix.

peregrineshahin:
returning a new Response already escapes the renderer as per how pyramid works

vdbergh:
> returning a new Response already escapes the renderer as per how pyramid works

Ah ok! That's much better.

vdbergh:
I will close this, waiting for your PR.

PR #2075, opened 2024-06-15
Show approvers to user management

peregrineshahin:
and enhance the page with count statistics

image

### DIFF
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -1,5 +1,9 @@
 <%inherit file="base.mak"/>
 
+<%!
+  from fishtest.util import format_group
+%>
+
 % if profile:
   <script>
     document.title = "Profile | Stockfish Testing";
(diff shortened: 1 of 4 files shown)


peregrineshahin:
now one thing to add to the schema is `last_action`, but it requires getting the users and updating it on each action as well.
which I will do in a different PR based on @vondele suggestion.

peregrineshahin:
well, now I realize we have already an actions collection, can benefit from that instead no?

peregrineshahin:
thing is the last_updated in the contributors section is not even related to the user, it's part of the workers schema..

ppigazzini (approved):
Looks good on DEV

vondele:
interestingly, one approver ([protonspring](https://tests.stockfishchess.org/user/protonspring)) seems to have a double group 'approvers, approvers'

ppigazzini:
fixed

PR #2076, opened 2024-06-16
Bugfix in clean_cache().

vdbergh:
Old runs without any tasks don't have "cores".


vdbergh:
Actually now I realize that this may no longer be a problem since the db has been updated recently. I need to fix my local db.

vdbergh:
The script to fix the db was posted here https://github.com/official-stockfish/fishtest/pull/2047#issuecomment-2143376747 .

PR #2078, opened 2024-06-17
sprt calculator: allow for specifying alpha and beta parameters

gahtan-syarif:


peregrineshahin:
If the math holds, worth changing it in elo calc api as well
https://github.com/official-stockfish/fishtest/blob/12981ffad22f5e44420d40ea9298b6ff83c0bfc2/server/fishtest/api.py#L563

peregrineshahin:
changing it to be optional* and not hard-coded.

gahtan-syarif:
> If the math holds, worth changing it in elo calc api as well
> 
> https://github.com/official-stockfish/fishtest/blob/12981ffad22f5e44420d40ea9298b6ff83c0bfc2/server/fishtest/api.py#L563

That would depend on if the calculation for the SPRT-adjusted elo is generalized to all alpha/beta values or only 0.05. I think vdbergh would know the answer to that?

PR #2084, opened 2024-06-25
Refactored locking implementation.

vdbergh:
We have refactored the locking implementation in the worker as a self-contained package "packages/openlock". The package is also available on github

https://github.com/vdbergh/openlock

and on PyPi

https://pypi.org/search/?q=openlock

This PR cleans up the worker code considerably.


vdbergh:
Rebased after #2068

ppigazzini:
https://stackoverflow.com/questions/48571212/why-is-sys-exit-causing-a-traceback
A note if some user should report this behavior: the wiki setup for Windows runs worker.py in an interactive session, so it shows a traceback when exiting:
```bash
*** Another worker (with PID=3424) is already running in this directory ***
  File "C:\Users\WDAGUtilityAccount\Desktop\worker\worker.py", line 1625, in <module>
    sys.exit(worker())
SystemExit: 1
```

vdbergh:
Is there any particular reason why on Windows Python is invoked as  `python.exe -i`?

ppigazzini:
The worker is started with a double click on a launcher cmd, with the -i the user views the traceback, otherwise the windows simply close.

ppigazzini (approved):
Tested successfully on:
- linux
- windows wsl2
- windows 10/11 msys2 bash
- windows 10/11 msys2 cmd

ppigazzini:
Triggered the worker updated, thank you @vdbergh

PR #2085, opened 2024-06-25
Fix flake8 errors/warnings

ppigazzini:


ppigazzini:
Looks good on DEV and PROD.

PR #2089, opened 2024-06-30
Fix red color `pre` before approving

peregrineshahin:
currently fixed games test_view results_pre shows a red color if the test isn't even approved, fix that since the state isn't really rejected.

e.g. https://dfts-0.pigazzini.it/tests/view/66818436b72c7bf0c0059a07

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -288,7 +288,7 @@ def format_results(run_results, run):
 
         if los >= 0.95:
             state = "accepted"
-        else:
+        elif run["approved"]:
             state = "rejected"
 
     result["info"].append(


vdbergh:
This due to me. I changed the coloring algorithm for fixed length test to be in line with SPRT tests: `red = rejected` (i.e. the opposite of passed). However this means that no colors should be shown while the fixed games test is running, which is not what people expect.

In retrospect, fixed length tests have a different function from SPRT tests: their primary use case is measuring ELO. So it is not surprising that the coloring algorithm is different.

I will revert.

PS. In the future Fishtest will no longer store the results widget in the db. This means that it will be possible to change the coloring algorithm retroactively.

peregrineshahin:
>I will revert.

Okay there is also the case of yellow appearing that's novel I think. I will close this in the meantime

vdbergh:
I would suggest to leave this issue open until the revert has actually happened.

peregrineshahin:
Well this is a PR not issue..

vdbergh:
Ah sorry I had noticed this.

dubslow:
> However this means that no colors should be shown while the fixed games test is running, which is not what people expect.

This is exactly what people expect, and how it's always been. Or, more precisely, if elo is within error bar of 0, then no color; if error bar is above and excludes 0, green, and otherwise, red; by the default has always been no color.

vdbergh:
> > However this means that no colors should be shown while the fixed games test is running, which is not what people expect.
> 
> This is exactly what people expect, and how it's always been. Or, more precisely, if elo is within error bar of 0, then no color; if error bar is above and excludes 0, green, and otherwise, red; by the default has always been no color.

You are contradicting yourself. People do expect colors to be shown while the test is running as you state in the last sentence (note that such colors are statistically meaningless).

But I already said I would revert so we should not discuss this further.

dubslow:
Yes, my bad, I committed a sign error while reading the original sentence :facepalm:

PR #1993, opened 2024-05-08
Proof of concept implementation of get_run via api (RFC).

vdbergh:
With this PR `/api/serialize_run` hosted by the main instance can be invoked by a secondary instance to receive a copy of the run from the cache.

In this way there is never an issue that the run might not be up to date.

Sadly I don't know how feasible this is. A typical serialized run is 5.5k. Is this too much?

Note: `/api/serialize_run` restricts access to `localhost` over `http`. Some form of access control is necessary but it can also be done externally.

I have tested that the PR works. The deserialized runs pass validation (for this I needed to supply the `tz_aware` codec option to `bson.decode()`).

PR #2090, opened 2024-07-01
Make new_run return a string instead of an ObjectId.

vdbergh:
The convention throughout the code base is that `run_id` is a string.

Since a `run_id` and the corresponding ObjectId print in exactly the same way, violating this convention may lead to hard to understand bugs such as this one

https://tests.stockfishchess.org/actions?action=log_message&user=&text=internal

which was introduced in #1993.


ppigazzini (approved):
Looks good on DEV.
Already running on PROD.
Thank you @vdbergh

PR #2091, opened 2024-07-01
Revert back to the coloring algorithm before #2069

vdbergh:
The aim of the change was to treat `num_games` and SPRT tests uniformly. This was not implemented correctly and users complained.

But independently of this, `num_games` tests and SPRT tests serve a fundamentally different purpose. So it makes sense that they use different coloring algorithms, which are adapted to their respective use cases.


ppigazzini:

PR #2092, opened 2024-07-01
Do not hold on to the run_cache_lock while writing to the db.

vdbergh:
Use the per-run active_run_lock instead.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -659,7 +659,6 @@ class RunDb:
     # Cache runs
     run_cache = {}
     run_cache_lock = threading.Lock()
-    run_cache_write_lock = threading.Lock()
 
     # handle termination
     def exit_run(self, signum, frame):
@@ -702,8 +701,8 @@ class RunDb:
                 flush=True,
             )
             return
+        r_id = str(run["_id"])
         with self.run_cache_lock:
-            r_id = str(run["_id"])
             if flush:
                 self.run_cache[r_id] = {
                     "is_changed": False,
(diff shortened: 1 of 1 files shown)


vondele:
I'm trying to understand this a bit (not knowing this code well).. so with this, somebody can own the run_cache_lock, and modify `run` while at the same time the owner of `active_run_lock` might be writing that run to disk?

vdbergh:
Here is how I see it.

The holder of the `run_cache_lock` should not change the actual runs, just the metadata about those runs in the cache (I do not exclude that there are bugs of course).

All modifications to actual run objects (or access to more than one property) should be done by grabbing the corresponding `active_run_lock` (again I do not exclude that there might be bugs).

When we need to write a run to the db we also need to grab the `active_run_lock`, otherwise we risk the run being modified midflight, potentially leading to a run on disk with incorrect aggregated data.

vdbergh:
Actually it seems that in mongod single document writes are atomic https://www.mongodb.com/docs/manual/tutorial/model-data-for-atomic-operations/ . Not sure if pymongo honors that.

vondele:
> Here is how I see it.
> 
> The holder of the `run_cache_lock` should not change the actual runs, just the metadata about those runs in the cache (I do not exclude that there are bugs of course).
> 
> All modifications to actual run objects (or access to more than one property) should be done by grabbing the corresponding `active_run_lock` (again I do not exclude that there might be bugs).
> 
> When we need to write a run to the db we also need to grab the `active_run_lock`, otherwise we risk the run being modified midflight, potentially leading to a run on disk with incorrect aggregated data.

OK, I think I understand what the approach is.

ppigazzini:
Looks good on DEV, testing on PROD with many workers.

ppigazzini (approved):
Already running on PROD, it looks good.
Thank you @vdbergh

PR #2093, opened 2024-07-01
Introduce `min_delay` parameter for `create_task`.

vdbergh:
Set it to 1.0 for the `flush_buffers` task.


ppigazzini:
Looks good on DEV, testing on PROD with many workers.

ppigazzini (approved):
Already running on PROD, it looks good.
Thank you @vdbergh

PR #2096, opened 2024-07-02
Protect internal data by lock before validation

vdbergh:
Deals with https://github.com/official-stockfish/fishtest/issues/2094#issuecomment-2201970849


ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #2097, opened 2024-07-02
return 404 if a non-sprt run_id is abused to be used in a live_elo link

peregrineshahin:
e.g. https://tests.stockfishchess.org/tests/live_elo/66833b9895b0d1e881e81489 https://tests.stockfishchess.org/tests/live_elo/6683a096c4f539faa03268e5


vdbergh:
A slightly more elaborate project would be to make the live elo page work for num_games tests (with an Elo dial and a LOS dial).

peregrineshahin:
> elaborate

We still would need to guard it against tunes, but yeah it might be reasonable at some point.

ppigazzini (approved):
Looks good on DEV.

PR #2099, opened 2024-07-03
Explicitly declare some schema fields as posix timestamps.

vdbergh:
This may yield clearer non-validation messages.


ppigazzini (approved):
Looks good on DEV.

ppigazzini:

PR #2100, opened 2024-07-03
Fully initialize the rundb object before starting tasks.

vdbergh:
From the "better to be safe than sorry" department.


ppigazzini (approved):
Looks good on DEV.

ppigazzini:

PR #2105, opened 2024-07-04
Add first name and last name to signup and profile

peregrineshahin:
not tested yet, needs adding the fields to the user collection (converting db), but still optional for those who prefer to be anonymous.

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -90,6 +90,8 @@ pgns_schema = intersect(
 user_schema = {
     "_id?": ObjectId,
     "username": username,
+    "firstname": str,
+    "lastname": str,
     "password": str,
     "registration_time": datetime_utc,
     "pending": bool,
--- a/server/fishtest/templates/contributors.mak
+++ b/server/fishtest/templates/contributors.mak
@@ -2,6 +2,7 @@
 
 <%!
   import urllib
+  from fishtest.util import format_fullname
 %>
 
 <script>
@@ -127,6 +128,7 @@
       <tr>
         <th></th>
         <th>Username</th>
+        <th>Full Name</th>
         <th class="text-end">Last active</th>
         <th class="text-end">Games/Hour</th>
         <th class="text-end">CPU Hours</th>
@@ -148,6 +150,7 @@
           ${user['username']}
           % endif
           </td>
+          <td>${format_fullname(user['firstname'], user['lastname'])}</td>
           <td data-diff="${user['diff']}" class="text-end">${user['str_last_updated']}</td>
           <td class="text-end">${int(user['games_per_hour'])}</td>
           <td class="text-end">${int(user['cpu_hours'])}</td>
--- a/server/fishtest/templates/signup.mak
+++ b/server/fishtest/templates/signup.mak
@@ -38,6 +38,30 @@
       <label for="username" class="d-flex align-items-end">Username</label>
     </div>
 
+    <div class="form-floating mb-3">
+      <input
+        type="text"
+        class="form-control mb-3"
+        id="firstname"
+        name="firstname"
+        placeholder="First Name"
+        maxlength="30"
+      >
+      <label for="firstname" class="d-flex align-items-end">First Name</label>
+    </div>
+
+    <div class="form-floating mb-3">
+      <input
+        type="text"
+        class="form-control mb-3"
+        id="lastname"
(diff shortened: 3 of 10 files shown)


server/fishtest/schemas.py
```diff
@@ -90,6 +90,8 @@ def size_is_length(x):
 user_schema = {
     "_id?": ObjectId,
     "username": username,
+    "firstname": union("", str),
```
peregrineshahin, line 93: I wonder if this union is redundant, btw I didn't include any regex, only length limit, because from experience no matter how smart you would be about a regex for names, you will either allow everything or miss the most obscure legal name ever.

vdbergh: Yes it is redundant since “” is a string.

The union means that the object should either be equal to “” or else be of type string. This is equivalent to being of type string.

ppigazzini:
DEV updated.

ppigazzini:
```yaml
Jul 04 23:30:22 dfts-0 pserve[102834]: Traceback (most recent call last):
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/env/lib/python3.12/site-packages/pyramid_mako/__init__.py", line 148, in __call__
Jul 04 23:30:22 dfts-0 pserve[102834]:     result = template.render_unicode(**system)
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/env/lib/python3.12/site-packages/mako/template.py", line 443, in render_unicode
Jul 04 23:30:22 dfts-0 pserve[102834]:     return runtime._render(
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/env/lib/python3.12/site-packages/mako/runtime.py", line 874, in _render
Jul 04 23:30:22 dfts-0 pserve[102834]:     _render_context(
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/env/lib/python3.12/site-packages/mako/runtime.py", line 916, in _render_context
Jul 04 23:30:22 dfts-0 pserve[102834]:     _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/env/lib/python3.12/site-packages/mako/runtime.py", line 943, in _exec_template
Jul 04 23:30:22 dfts-0 pserve[102834]:     callable_(context, *args, **kwargs)
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/fishtest/templates/base.mak", line 561, in render_body
Jul 04 23:30:22 dfts-0 pserve[102834]:     <div>${self.body()}</div>
Jul 04 23:30:22 dfts-0 pserve[102834]:   File "/home/usr00/fishtest/server/fishtest/templates/user.mak", line 90, in render_body
Jul 04 23:30:22 dfts-0 pserve[102834]:     value="${user['firstname']}"
Jul 04 23:30:22 dfts-0 pserve[102834]: KeyError: 'firstname'

```

peregrineshahin:
I think I mentioned above that we need to add such keys to every old user before this can be tested.. now while the current way of perfectly making the schema predictable in the mako rendering side, I agree that it might feel shaky that thereis one point of failure. But at the same time not using `run.get("firstname","")` can work as a fallback to catch errors since the tendering errors happen before shipping that to production.. so I'm not entirely sure if when using such a fullstack technology as we do, should convert all code base to accept wrong schema in the renderer.

maximmasiutin:
Please consider using this version which I updated to fulfill this request: https://github.com/maximmasiutin/fishtest-contributors/blob/main/fishtest_contributors.py

You may copy this file to your own repository, be it a separate repository for this script that you may create, or just add it to an existing repository such as the main Stockfish repository, so that you would not need to maintain this file separately.

server/fishtest/schemas.py
peregrineshahin: Well the limit is more to avoid spam rather than an intrinsic property of the names.. but noted for the future.. I guess this schema is possible because of vtjson, thanks for the package!

MinetaS:
I prefer one single display name field to two first/last name fields because 1) middle name does exist and 2) some people in Indonesia have only one name afaik.

Disservin:
I like the idea of a display name as well.

PR #2108, opened 2024-07-06
Raise pserve threads and semaphores

ppigazzini:


ppigazzini:
In this way every `pserve` should use: 100 * 10 = 1000 connections, requiring 2000 files.
PROD is running with 6 pserve, so we must be sure that `pserve` and `nginx`, both managed by `systemd`, are not capped by the limit of number of files.

ppigazzini:
PROD updated.

ppigazzini:
@vondele found the way to set the number of file limit for a service managed by systemd
https://access.redhat.com/solutions/1257953

vdbergh:
I have been reading about JAVA virtual threads https://openjdk.org/jeps/425 . In short: virtual threads are thread like object in which all code runs asynchronously. So there is no need for an async/await framework. Async/await is nice but it is hard to mix with legacy synchronous code (like Fishtest).

Virtual threads are so cheap that there is no need to do any pooling. You can essentially create an infinite number of them! Incidentally, the JEP recommends using semaphores to limit access to constrained resources as we are now doing in Fishtest!

If Python had virtual threads then there would be no need to put a limit on the number of waitress threads. I did some googling, but sadly there do not seem to be any projects to implement virtual threads in Python.

PR #2109, opened 2024-07-06
Only scavenge 500 dead tasks per turn.

vdbergh:
When a fleet quits the server is suddenly confronted with 1000s of dead tasks which need to be handled. With the current PR we throttle this.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -104,6 +104,8 @@ class RunDb:
         self.request_task_lock = threading.Lock()
         self.scheduler = None
 
+        self.dead_task_throttle = 500
+
     def schedule_tasks(self):
         if self.scheduler is None:
             self.scheduler = Scheduler(jitter=0.05)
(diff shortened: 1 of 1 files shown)


ppigazzini:
The PR code works as expected, but pserve's CPU load halved immediately when the cores stopped, so it looks like we are saving CPU when we have a lot of free CPU.
It's not clear to me the negative effects of the PR.
Unfortunately, the fleet was schedule to do other tasks and we have not the time to monitoring the cores leaving with master in place.

vdbergh:
I agree it is not clear cut.

I am worried about the run locks being acquired during the handling of dead tasks (potentially interfering with the handling of the worker apis). But the locks are not held continuously (they are acquired and released for each dead task that is handled). So perhaps that gives the other threads an opportunity run.

Another effect is that if the handling of dead tasks takes a long time then other tasks are delayed - in particular the flush buffers task (since all tasks run in one thread). However this could be solved using a thread pool as I wrote above.

vdbergh:
Do you have  a log of a fleet leaving before this PR? How long did it it take?

vdbergh:
I am closing this for now. There is a defect this approach. If a fleet disconnects and reconnects again you want to clear the old dead tasks as quickly as possible so that they can be assigned to the newly connected workers.

I will submit a new PR in which the scavenging of dead tasks is done in a background thread.

PR #2111, opened 2024-07-06
Soften log message on expected non-validation of old run object.

vdbergh:


ppigazzini:

PR #2118, opened 2024-07-12
Increasing the maximum number of tests users can follow from 15 to 20

FauziAkram:
Increasing the maximum number of tests users can follow from 15 to 20.
So users can keep track of more tests simultaneously, allowing for a more comprehensive overview of ongoing tests.
Also, the previous limit of 15 was arbitrary, and increasing the limit poses no risks. This adjustment aligns with user needs for better monitoring and management of tests.

P.s: The new limit of 20 is also an arbitrary number, but it is better than before, and from my point of view is enough to satisfy users activity. But the limit can be increased further if needed.

### DIFF
--- a/server/fishtest/static/js/notifications.js
+++ b/server/fishtest/static/js/notifications.js
@@ -160,7 +160,7 @@ function notifyElo(entryStart, entryFinished) {
   });
 }
 
-const designCapacity = 15;
+const designCapacity = 20;
 
 function getNotifications() {
   let notifications;


FauziAkram:
@ppigazzini are there any drawbacks to increasing the limit?

FauziAkram:
Why is it taking so long to resolve this PR?
Currently, around 35-60 active tests are running in fishtest, being able to follow only 15 at a time is very limiting, as a user I would prefer to be able to follow more than that.

Disservin:
well if you take a look at the commits the last commit was on `Jul 6, 2024`... so it's no surprise there's no feedback on this + this is just a temporary solution to a permanent problem..
there's no good reason why this should be 20 especially when you say 35-60 tests are active... the proper way would be just to rewrite this bit and not use a capped container

ppigazzini:
PROD updated, thank you @FauziAkram

PR #2119, opened 2024-07-13
Switch from cutechess-cli to fastchess

vondele:
fixes https://github.com/official-stockfish/fishtest/issues/2106

This PR switches from cutechess-cli to fast-chess.

cutechess-cli has been serving us well in the past years, however, some issues have accumulated, namely the difficulty of compiling cutechess-cli, the observed timeouts at high concurrency and short TC, and e.g. slowness when indexing larger books.  fast-chess https://github.com/Disservin/fast-chess has addressed these issues, and has now probably become mature enough to serve as the game manager for fishtest.

As an example of its ability to deal with short TC and high concurrency:
https://github.com/official-stockfish/fishtest/pull/2119#issuecomment-2571563064
with concurrency 48, and TC 1+0.01s, no timeouts are observed.

fast-chess is built from sources, with the zip download as well as the binary cached as needed.  There is fine-grained control over which version of fast-chess is used, so we can easily upgrade for new features.

In this PR, fast-chess was built in cutechess compatibility to facilitate integration, and to benefit from the existing fishtest checks. After validation the full switch to 'native' fast-chess output, with its pentanomial and wld output for every game pair, has significantly simplified the worker's book-keeping.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -67,7 +67,7 @@ def is_64bit():
 
 
 HTTP_TIMEOUT = 30.0
-CUTECHESS_KILL_TIMEOUT = 15.0
+FASTCHESS_KILL_TIMEOUT = 15.0
 UPDATE_RETRY_TIME = 15.0
 
 RAWCONTENT_HOST = "https://raw.githubusercontent.com"
@@ -504,24 +504,6 @@ def unzip(blob, save_dir):
     return file_list
 
 
-def convert_book_move_counters(book_file):
-    # converts files with complete FENs, leaving others (incl. converted ones) unchanged
-    epds = []
-    with open(book_file, "r") as file:
-        for fen in file:
-            fields = fen.split()
-            if len(fields) == 6 and fields[4].isdigit() and fields[5].isdigit():
-                fields[4] = f"hmvc {fields[4]};"
-                fields[5] = f"fmvn {fields[5]};"
-                epds.append(" ".join(fields))
-            else:
-                return
-
-    with open(book_file, "w") as file:
-        for epd in epds:
-            file.write(epd + "\n")
-
-
 def clang_props():
     """Parse the output of clang++ -E - -march=native -### and extract the available clang properties"""
     with subprocess.Popen(
@@ -876,89 +858,7 @@ def enqueue_output(stream, queue):
         queue.put(line)
 
 
-def update_pentanomial(line, rounds):
-    saved_rounds = copy.deepcopy(rounds)
-    saved_line = line
-
-    def result_to_score(_result):
-        if _result == "1-0":
-            return 2
-        elif _result == "0-1":
-            return 0
-        elif _result == "1/2-1/2":
-            return 1
-        else:
-            return -1
-
-    if "pentanomial" not in rounds.keys():
-        rounds["pentanomial"] = 5 * [0]
-    if "trinomial" not in rounds.keys():
-        rounds["trinomial"] = 3 * [0]
-
-    saved_sum_trinomial = sum(rounds["trinomial"])
-    current = {}
-
-    # Parse line like this:
-    # Finished game 4 (Base-SHA vs New-SHA): 1/2-1/2 {Draw by adjudication}
-    line = line.split()
-    if line[0] == "Finished" and line[1] == "game" and len(line) >= 7:
-        round_ = int(line[2])
-        rounds[round_] = current
-        current["white"] = line[3][1:]
-        current["black"] = line[5][:-2]
-        i = current["result"] = result_to_score(line[6])
-        if round_ % 2 == 0:
-            if i != -1:
-                rounds["trinomial"][2 - i] += 1  # reversed colors
-            odd = round_ - 1
-            even = round_
-        else:
-            if i != -1:
-                rounds["trinomial"][i] += 1
-            odd = round_
-            even = round_ + 1
-        if odd in rounds.keys() and even in rounds.keys():
-            assert rounds[odd]["white"][0:3] == "New"
-            assert rounds[odd]["white"] == rounds[even]["black"]
-            assert rounds[odd]["black"] == rounds[even]["white"]
-            i = rounds[odd]["result"]
-            j = rounds[even]["result"]  # even is reversed colors
-            if i != -1 and j != -1:
-                rounds["pentanomial"][i + 2 - j] += 1
-                del rounds[odd]
-                del rounds[even]
-                rounds["trinomial"][i] -= 1
-                rounds["trinomial"][2 - j] -= 1
-                assert rounds["trinomial"][i] >= 0
-                assert rounds["trinomial"][2 - j] >= 0
-
-    # make sure something happened, but not too much
-    # this sometimes fails: we want to understand why
-    assertion = (
-        current.get("result", -1000) == -1
-        or abs(sum(rounds["trinomial"]) - saved_sum_trinomial) == 1
-    )
-    if not assertion:
-        raise WorkerException(
-            "update_pentanomial() failed. line={}; rounds before={}; rounds after={}".format(
-                saved_line, saved_rounds, rounds
-            )
-        )
-
-
-def validate_pentanomial(wld, rounds):
-    def results_to_score(results):
-        return sum([results[i] * (i / 2.0) for i in range(len(results))])
-
-    LDW = [wld[1], wld[2], wld[0]]
-    s3 = results_to_score(LDW)
-    s5 = results_to_score(rounds["pentanomial"]) + results_to_score(rounds["trinomial"])
-    assert sum(LDW) == 2 * sum(rounds["pentanomial"]) + sum(rounds["trinomial"])
-    epsilon = 1e-4
-    assert abs(s5 - s3) < epsilon
-
-
-def parse_cutechess_output(
+def parse_fastchess_output(
     p, current_state, remote, result, spsa_tuning, games_to_play, batch_size, tc_limit
 ):
     hash_pattern = re.compile(r"(Base|New)-[a-f0-9]+")
@@ -968,7 +868,23 @@ def parse_cutechess_output(
         return "-".join([word[0], word[1][:10]])
 
     saved_stats = copy.deepcopy(result["stats"])
-    rounds = {}
+
+    # patterns used to obtain fastchess WLD and ptnml results from the following block of info:
+    # --------------------------------------------------
+    # Results of New-e443b2459e vs Base-e443b2459e (0.601+0.006, 1t, 16MB, UHO_Lichess_4852_v1.epd):
+    # Elo: -9.20 +/- 20.93, nElo: -11.50 +/- 26.11
+    # LOS: 19.41 %, DrawRatio: 42.35 %, PairsRatio: 0.88
+    # Games: 680, Wins: 248, Losses: 266, Draws: 166, Points: 331.0 (48.68 %)
+    # Ptnml(0-2): [43, 61, 144, 55, 37], WL/DD Ratio: 4.76
+    # --------------------------------------------------
+    pattern_WLD = re.compile(
+        r"Games: ([0-9]+), Wins: ([0-9]+), Losses: ([0-9]+), Draws: ([0-9]+), Points: ([0-9.]+) \("
+    )
+    pattern_ptnml = re.compile(
+        r"Ptnml\(0-2\): \[([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+), ([0-9]+)\]"
+    )
+    fastchess_WLD_results = None
+    fastchess_ptnml_results = None
 
     q = Queue()
     t_output = threading.Thread(target=enqueue_output, args=(p.stdout, q), daemon=True)
@@ -986,7 +902,7 @@ def parse_cutechess_output(
         except Empty:
             if p.poll() is not None:
                 break
-            time.sleep(1)
+            time.sleep(0.1)
             continue
 
         line = hash_pattern.sub(shorten_hash, line)
@@ -997,70 +913,83 @@ def parse_cutechess_output(
             if num_games_updated == games_to_play:
                 print("Finished match cleanly")
             else:
-                raise WorkerException("Finished match uncleanly")
+                raise WorkerException(
+                    "Finished match uncleanly {} vs. required {}".format(
+                        num_games_updated, games_to_play
+                    )
+                )
 
         # Parse line like this:
         # Warning: New-SHA doesn't have option ThreatBySafePawn
         if "Warning:" in line and "doesn't have option" in line:
-            message = r'Cutechess-cli says: "{}"'.format(line)
+            message = r'fastchess says: "{}"'.format(line)
             raise RunException(message)
 
         # Parse line like this:
         # Warning: Invalid value for option P: -354
         if "Warning:" in line and "Invalid value" in line:
-            message = r'Cutechess-cli says: "{}"'.format(line)
+            message = r'fastchess says: "{}"'.format(line)
             raise RunException(message)
 
         # Parse line like this:
         # Finished game 1 (stockfish vs base): 0-1 {White disconnects}
-        if "disconnects" in line or "connection stalls" in line:
+        if "disconnect" in line or "stall" in line:
             result["stats"]["crashes"] += 1
 
-        if "on time" in line:
+        if "on time" in line or "timeout" in line:
             result["stats"]["time_losses"] += 1
 
-        # Parse line like this:
-        # Score of stockfish vs base: 0 - 0 - 1  [0.500] 1
-        if "Score" in line:
-            # Parsing sometimes fails. We want to understand why.
+        # fastchess WLD and pentanomial output parsing
+        m = pattern_WLD.search(line)
+        if m:
             try:
-                chunks = line.split(":")
-                chunks = chunks[1].split()
-                wld = [int(chunks[0]), int(chunks[2]), int(chunks[4])]
-            except:
-                raise WorkerException("Failed to parse score line: {}".format(line))
+                fastchess_WLD_results = {
+                    "games": int(m.group(1)),
+                    "wins": int(m.group(2)),
+                    "losses": int(m.group(3)),
+                    "draws": int(m.group(4)),
+                    "points": float(m.group(5)),
+                }
+            except Exception as e:
+                raise WorkerException(
+                    "Failed to parse WLD line: {} leading to: {}".format(line, str(e))
+                )
 
-            validate_pentanomial(
-                wld, rounds
-            )  # check if cutechess-cli result is compatible with
-            # our own bookkeeping
+        m = pattern_ptnml.search(line)
+        if m:
+            try:
+                fastchess_ptnml_results = [int(m.group(i)) for i in range(1, 6)]
+            except Exception as e:
+                raise WorkerException(
+                    "Failed to parse ptnml line: {} leading to: {}".format(line, str(e))
+                )
 
-            pentanomial = [
-                rounds["pentanomial"][i] + saved_stats["pentanomial"][i]
+        # if we have parsed the block properly let's update results
+        if (fastchess_ptnml_results is not None) and (
+            fastchess_WLD_results is not None
+        ):
+            result["stats"]["pentanomial"] = [
+                fastchess_ptnml_results[i] + saved_stats["pentanomial"][i]
                 for i in range(5)
             ]
-            result["stats"]["pentanomial"] = pentanomial
 
-            wld_pairs = {}  # trinomial frequencies of completed game pairs
-
-            # rounds['trinomial'] is ordered ldw
-            wld_pairs["wins"] = wld[0] - rounds["trinomial"][2]
-            wld_pairs["losses"] = wld[1] - rounds["trinomial"][0]
-            wld_pairs["draws"] = wld[2] - rounds["trinomial"][1]
-
-            result["stats"]["wins"] = wld_pairs["wins"] + saved_stats["wins"]
-            result["stats"]["losses"] = wld_pairs["losses"] + saved_stats["losses"]
-            result["stats"]["draws"] = wld_pairs["draws"] + saved_stats["draws"]
+            result["stats"]["wins"] = (
+                fastchess_WLD_results["wins"] + saved_stats["wins"]
+            )
+            result["stats"]["losses"] = (
+                fastchess_WLD_results["losses"] + saved_stats["losses"]
+            )
+            result["stats"]["draws"] = (
+                fastchess_WLD_results["draws"] + saved_stats["draws"]
+            )
 
             if spsa_tuning:
                 spsa = result["spsa"]
-                spsa["wins"] = wld_pairs["wins"]
-                spsa["losses"] = wld_pairs["losses"]
-                spsa["draws"] = wld_pairs["draws"]
+                spsa["wins"] = fastchess_WLD_results["wins"]
+                spsa["losses"] = fastchess_WLD_results["losses"]
+                spsa["draws"] = fastchess_WLD_results["draws"]
 
-            num_games_finished = (
-                wld_pairs["wins"] + wld_pairs["losses"] + wld_pairs["draws"]
-            )
+            num_games_finished = fastchess_WLD_results["games"]
 
             assert (
                 2 * sum(result["stats"]["pentanomial"])
@@ -1068,9 +997,12 @@ def parse_cutechess_output(
                 + result["stats"]["losses"]
                 + result["stats"]["draws"]
             )
-            assert num_games_finished == 2 * sum(rounds["pentanomial"])
+            assert num_games_finished == 2 * sum(fastchess_ptnml_results)
             assert num_games_finished <= num_games_updated + batch_size
             assert num_games_finished <= games_to_play
+
+            fastchess_ptnml_results = None
+            fastchess_WLD_results = None
 
             # Send an update_task request after a batch is full or if we have played all games.
             if (num_games_finished == num_games_updated + batch_size) or (
@@ -1111,10 +1043,6 @@ def parse_cutechess_output(
                 else:
                     current_state["last_updated"] = datetime.now(timezone.utc)
 
-        # Act on line like this:
-        # Finished game 4 (Base-SHA vs New-SHA): 1/2-1/2 {Draw by adjudication}
-        if line.startswith("Finished game"):
-            update_pentanomial(line, rounds)
     else:
         raise WorkerException(
             "{} is past end time {}".format(datetime.now(timezone.utc), end_time)
@@ -1123,7 +1051,7 @@ def parse_cutechess_output(
     return True
 
 
-def launch_cutechess(
+def launch_fastchess(
     cmd, current_state, remote, result, spsa_tuning, games_to_play, batch_size, tc_limit
 ):
     if spsa_tuning:
@@ -1154,7 +1082,7 @@ def launch_cutechess(
         w_params = []
         b_params = []
 
-    # Run cutechess-cli binary.
+    # Run fastchess binary.
     # Stochastic rounding and probability for float N.p: (N, 1-p); (N+1, p)
     idx = cmd.index("_spsa_")
     cmd = (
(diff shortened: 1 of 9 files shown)


vondele:
The PR is ready for review from the fishtest side, still in a draft state to allow for a couple of fixes and updates to be made to fast-chess.

vondele:
current pushed version runs correctly for fishtest AFAICT. Still not master fast-chess, but nearly there.

I think the most important thing now verifying that fast-chess is building with all the version of the compilers we need, and in all environments. 

Current fishtest version check is gcc >= 7.3 and clang >= 8.0 (current online workers are gcc >= 9.3 and clang >= 13.0).

vondele:
updated to master now.

vondele:
updated to current fast-chess, verified to compile with gcc 7.3.0 and clang 8.0.0, the fishtest minimum required versions.

worker/games.py
```diff
@@ -999,16 +981,16 @@ def shorten_hash(match):
             else:
                 raise WorkerException("Finished match uncleanly")
 
-        # Parse line like this:
+        # Parse line like this:                                          # TODO TODO https://github.com/Disservin/fast-chess/issues/535
```
Disservin, line 984: done

worker/worker.py
```diff
@@ -442,89 +425,115 @@ def verify_required_cutechess(cutechess_path):
...
         return False
 
     if major + minor + patch == 0:
-        print("Unable to find the version of cutechess-cli.")
+        print("Unable to find the version of fast-chess.")
         return False
 
-    if (major, minor) < (1, 2):
-        print("Requires cutechess 1.2 or higher, found version doesn't match")
+    if (major, minor) < (0, 9):
```
Disservin, line 445: lets also have these as constants at the top next around line 69

vondele: actually, I think this check can be removed, now that we explicit check the sha.

Disservin:
there was some code somewhere which made high concurrency worker only pickup ltc tasks, arguably we don't need that bit anymore with fc, but can be done later

vondele:
> there was some code somewhere which made high concurrency worker only pickup ltc tasks, arguably we don't need that bit anymore with fc, but can be done later

OK, will remove that, it is a server change.

Disservin:
Should we also first compile the binary with tests, to make sure that everything works on the host ? `make -j tests && ./fast-chess-tests`
Not quite sure

vondele:
To be considered... will the tests work if fast-chess is compiled with 'USE_CUTE' and how long does it take? 
Are there additional dependencies if done (gtest or similar) ?

Disservin:
> will the tests work if fast-chess is compiled with 'USE_CUTE' 

it is a separate binary which can only run the tests

> how long does it take?

the tests are rather quick, ~3s.. compiling them takes considerably more time... i need to check if that can be improved

> Are there additional dependencies if done (gtest or similar) ?

there's a depedency on doctest, but the repo has source code for that

vondele:
testing worked locally with gcc 7.3 and clang 8.0, and is simple to add. Trying in CI.

ppigazzini:
https://github.com/official-stockfish/fishtest/blob/12981ffad22f5e44420d40ea9298b6ff83c0bfc2/worker/games.py#L1157-L1180

if I recall correctly `fast-chess` does not need stochastic rounding.

vondele:
> if I recall correctly `fast-chess` does not need stochastic rounding.

Isn't that more a property of SF? I think parameters are integers in SF, or at least only that is well tested.

ppigazzini:
`cutechess` accepts only integers, but it's a pleasure to start blaming SF devs for the convoluted SPSA parameters rounding.

vondele:
They consulted their legal department, which points to the UCI spec
```
		* spin
			a spin wheel that can be an integer in a certain range
```

ppigazzini:
No way a legal department will take `can be` as a specification.

Disservin:
> working. Timing added is not so important, this happens just once and the binary is reused.

just for documentation, we get a one time addition of 2min - 4min

image

vondele:
ah, you mean in fishtest CI. For workers, I think it is acceptable. 
Also, this is CI, where I hardcoded concurrency to 1. Let me change that to 4, I recall that's the concurrency that is available to actions.

Disservin:
> ah, you mean in fishtest CI. For workers, I think it is acceptable.

actually meant workers, but fishtest ci can be used to measure the increased time quite nicely :D

vondele:
workers have typically even larger concurrency so will be even faster, and also this binary is preserved over worker restarts, so will be compiled only rarely.

vondele:
https://github.com/Disservin/fast-chess/pull/601#issuecomment-2234219897

worker/games.py
```diff
@@ -1317,7 +1299,7 @@ def run_games(
     start_game_index = opening_offset + input_total_games
     run_seed = int(hashlib.sha1(run["_id"].encode("utf-8")).hexdigest(), 16) % (2**30)
 
-    # Format options according to cutechess syntax.
+    # Format options according to fastchess syntax.
```
robertnurnberg, line 1302: fast-chess?

worker/games.py
```diff
@@ -1554,15 +1531,17 @@ def make_player(arg):
         if any(substring in book.upper() for substring in ["FRC", "960"]):
             variant = "fischerandom"
 
-        # Run cutechess binary.
-        cutechess = "cutechess-cli" + EXE_SUFFIX
+        # Run fastchess binary.
```
robertnurnberg, line 1534: fast-chess? I guess variable/function/directory names can stay without hyphen.

worker/worker.py
```diff
@@ -392,139 +396,143 @@ def get_credentials(config, options, args):
...
-            "Exception downloading or extracting {}:\n".format(zipball),
-            e,
-            sep="",
-            file=sys.stderr,
-        )
-    else:
-        print("Finished downloading {}".format(cutechess))
-
+def verify_required_fastchess(fastchess_path, fastchess_sha):
+    # Verify that fastchess is working and has the required minimum version.
```
robertnurnberg, line 400: fast-chess

worker/worker.py
```diff
@@ -392,139 +396,143 @@ def get_credentials(config, options, args):
...
+        item_url = (
+            "https://api.github.com/repos/"
+            + username
+            + "/fast-chess/zipball/"
+            + fastchess_sha
+        )
 
-    cutechess = "cutechess-cli" + EXE_SUFFIX
-    cutechess_path = testing_dir / cutechess
+        print("Building fast chess from sources at {}".format(item_url))
```
robertnurnberg, line 476: fast-chess

vondele:
integrate https://github.com/official-stockfish/fishtest/pull/2121 directly in this PR, as it makes sense to do the transition in 1 step.

gahtan-syarif:
github.com/official-stockfish/fishtest/blob/dc51a478d2051051b63902020d9b4efa25d1af43/worker/games.py#L1247
this can be increased to `2**64`

https://github.com/official-stockfish/fishtest/blob/dc51a478d2051051b63902020d9b4efa25d1af43/worker/games.py#L1449
can be modified to include nodes tracking for debugging, format is `-pgnout file=NAME nodes=true`

vdbergh:
I have been running this PR for a while (with a one core worker). I have observed no issues.

Disservin:
> I think the culprit is that `fastchess_WLD_results["games"]` is not computed as W+L+D (which would include the `saved_stats`).

Mh fastchess updates calculates this as w+l+d.. now what comes to my mind is that maybe some race could lead to flawed updates but this shouldn't happend because all relevant parts are behind a lock. Can you give more information perhaps i.e. was the assertion wrong by 1 or 2? meaning a pair was missed or something was not updated in pairs.. also are you on the latest version of this pr? I think there were issues in the past not sure if they were ever included here or fixed before

vdbergh:
> > I think the culprit is that `fastchess_WLD_results["games"]` is not computed as W+L+D (which would include the `saved_stats`).
> 
> Mh fastchess updates calculates this as w+l+d.. now what comes to my mind is that maybe some race could lead to flawed updates but this shouldn't happend because all relevant parts are behind a lock. Can you give more information perhaps i.e. was the assertion wrong by 1 or 2? meaning a pair was missed or something was not updated in pairs.. also are you on the latest version of this pr? I think there were issues in the past not sure if they were ever included here or fixed before

There is nothing wrong with fastchess. In the case of SPSA the worker computes WLD of a task as the sum of the WLD for the "mini tasks" (corresponding to a choice of parameters). The same should be done for the total number of games.

vdbergh:
One may wonder however what the point is of this summing of WLD data corresponding to different sets of parameters...

Disservin:
> There is nothing wrong with fastchess. In the case of SPSA the worker computes WLD of a task as the sum of the WLD for the "mini tasks" (corresponding to a choice of parameters). The same should be done for the total number of games.

oh! i missed the fact that both assertion errors came from a spsa test, thanks

vondele:
> I got two assertion errors https://tests.stockfishchess.org/actions?max_actions=2&action=&user=&text=AssertionError&before=1722328311.350552&run_id=
> 
> I think the culprit is that `fastchess_WLD_results["games"]` is not computed as W+L+D (which would include the `saved_stats`).
> 
> I am bit surprised however that this issue (if present) does not cause more problems.

@vdbergh is this the proper fix (quick local test suggest it is):

```diff
diff --git a/worker/games.py b/worker/games.py
index 88a26c2..cc82c02 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -990,7 +990,12 @@ def parse_fastchess_output(
                 spsa["losses"] = fastchess_WLD_results["losses"]
                 spsa["draws"] = fastchess_WLD_results["draws"]
 
-            num_games_finished = fastchess_WLD_results["games"]
+            num_games_finished = (
+                fastchess_WLD_results["games"]
+                + saved_stats["wins"]
+                + saved_stats["losses"]
+                + saved_stats["draws"]
+            )
 
             fastchess_ptnml_results = None
             fastchess_WLD_results = None
```

vdbergh:
Yes I think so.

vdbergh:
For reference: there are some more assertion errors https://tests.stockfishchess.org/actions?max_actions=2&action=&user=&text=AssertionError&before=1723033368.540148&run_id=

I have not debugged it yet (it is again for SPSA tests).

vdbergh:
Ok this is master
```
            assert num_games_finished == 2 * sum(rounds["pentanomial"])
            assert num_games_finished <= num_games_updated + batch_size
            assert num_games_finished <= games_to_play
```
This is this PR
```
            assert num_games_finished == 2 * sum(result["stats"]["pentanomial"])
            assert num_games_finished <= num_games_updated + batch_size
            assert num_games_finished <= games_to_play
```
`num_games_finished` is the number of games finished in the current mini task (so the fix in https://github.com/official-stockfish/fishtest/pull/2119#issuecomment-2264774549 is actually wrong, sorry about this).

The difference between `master` and this PR is the right hand side of the first assert.
```
2 * sum(rounds["pentanomial"])
```
in master vs
```
2 * sum(result["stats"]["pentanomial"])
```
in this PR.

The difference it that the first expression only refers to the current mini task and the second expression refers to the aggregated results.

So I think the second expression should be replaced by
```
2*sum(fastchess_ptnml_results)
```

vdbergh:
This would be the patch wrt the current PR. I am now testing it
```diff
diff --git a/worker/games.py b/worker/games.py
index cc82c02..9ef2796 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -990,15 +990,7 @@ def parse_fastchess_output(
                 spsa["losses"] = fastchess_WLD_results["losses"]
                 spsa["draws"] = fastchess_WLD_results["draws"]
 
-            num_games_finished = (
-                fastchess_WLD_results["games"]
-                + saved_stats["wins"]
-                + saved_stats["losses"]
-                + saved_stats["draws"]
-            )
-
-            fastchess_ptnml_results = None
-            fastchess_WLD_results = None
+            num_games_finished = fastchess_WLD_results["games"]
 
             assert (
                 2 * sum(result["stats"]["pentanomial"])
@@ -1006,10 +998,13 @@ def parse_fastchess_output(
                 + result["stats"]["losses"]
                 + result["stats"]["draws"]
             )
-            assert num_games_finished == 2 * sum(result["stats"]["pentanomial"])
+            assert num_games_finished == 2 * sum(fastchess_ptnml_results)
             assert num_games_finished <= num_games_updated + batch_size
             assert num_games_finished <= games_to_play
 
+            fastchess_ptnml_results = None
+            fastchess_WLD_results = None
+
             # Send an update_task request after a batch is full or if we have played all games.
             if (num_games_finished == num_games_updated + batch_size) or (
                 num_games_finished == games_to_play
```

worker/games.py
```diff
@@ -970,6 +870,23 @@ def shorten_hash(match):
     saved_stats = copy.deepcopy(result["stats"])
     rounds = {}
```
Viren6, line 871: I dont think rounds is accessed anywhere

Viren6:
Seems time losses can be a lot higher than games played:

image

Checking the pgn shows the time losses is exactly twice as high as it should be, so there is probably an error between pairs vs games somewhere.

gahtan-syarif:
> Seems time losses can be a lot higher than games played:
> 
> image
> 
> Checking the pgn shows the time losses is exactly twice as high as it should be, so there is probably an error between pairs vs games somewhere.

it seems i found out what went wrong, fastchess when encountering time losses displays a warning to the user like below:
```
Warning; Engine Engine2 loses on time
Finished game 3 (Engine1 vs Engine2): 1-0 {Black loses on time}
```
this warning does not exit on cutechess.
fishtest counts the occurrences of the string "loses on time" which means it counts twice for every timeloss on fastchess output

Disservin:
> > Seems time losses can be a lot higher than games played:
> > image
> > Checking the pgn shows the time losses is exactly twice as high as it should be, so there is probably an error between pairs vs games somewhere.
> 
> it seems i found out what went wrong, fastchess when encountering time losses displays a warning to the user like below:
> 
> ```
> Warning; Engine Engine2 loses on time
> Finished game 3 (Engine1 vs Engine2): 1-0 {Black loses on time}
> ```
> 
> this warning does not exist on cutechess. fishtest counts the occurrences of the string "loses on time" which means it counts twice for every timeloss on fastchess output. the same applies to crashes as fc also sends a warning for disconnects

Lest just get rid of that I don’t see any value in the warning report

gahtan-syarif:
> > > Seems time losses can be a lot higher than games played:
> > > image
> > > Checking the pgn shows the time losses is exactly twice as high as it should be, so there is probably an error between pairs vs games somewhere.
> > 
> > 
> > it seems i found out what went wrong, fastchess when encountering time losses displays a warning to the user like below:
> > ```
> > Warning; Engine Engine2 loses on time
> > Finished game 3 (Engine1 vs Engine2): 1-0 {Black loses on time}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this warning does not exist on cutechess. fishtest counts the occurrences of the string "loses on time" which means it counts twice for every timeloss on fastchess output. the same applies to crashes as fc also sends a warning for disconnects
> 
> Lest just get rid of that I don’t see any value in the warning report

i suggest making it logger::trace so it could still show up on the log file

vondele:
Updated fast-chess after the fix, thanks for analysis and fixes.

gahtan-syarif:
```python
        # Parse line like this:
        # Finished game 1 (stockfish vs base): 0-1 {White disconnects}
        if "disconnects" in line or "connection stalls" in line:
            result["stats"]["crashes"] += 1

        if "on time" in line:
            result["stats"]["time_losses"] += 1
```
i suggest changing to:
```
        # Parse line like this:
        # Finished game 1 (stockfish vs base): 0-1 {White disconnects}
        if "disconnect" in line or "stall" in line:
            result["stats"]["crashes"] += 1

        if "on time" in line or "timeout" in line:
            result["stats"]["time_losses"] += 1
```

based on: https://github.com/cutechess/cutechess/blob/780065637f9936bc29cc592c6f0b2007ccbf66de/projects/lib/src/board/result.cpp#L109-L132

vondele:
converting to draft, awaiting a change in convention for mate in plies / moves in the pgn output:

https://github.com/Disservin/fastchess/pull/696

gahtan-syarif:
```cpp
                "-tournament",
                "gauntlet",
```
better to get rid of this since its unecessary and fastchess currently supports only roundrobin format

ppigazzini:
Tested fine on DEV with workers with several combinations of:
- Windows/Linux/WSL2
- gcc/clang
- python versions 3.6-3.13
- RAM 1.5MiB - 64MiB
- concurrency 1 - 48

I found no issue with a low RAM worker (1.5 Mib and 4 cores).
But `cutechess` and `fast-chess` have different RAM usages, so I share my data to get feedback before merging the PR.

**cutechess with 64 GiB RAM**
- concurrency 1
```
top - 12:22:39 up  1:02,  1 user,  load average: 1.78, 20.00, 21.97
Tasks:  29 total,   1 running,  28 sleeping,   0 stopped,   0 zombie
%Cpu(s):  2.1 us,  0.2 sy,  0.0 ni, 97.8 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem :  2.8/64251.3  [                                                                                                    ]      MiB Swap:  0.0/16384.0  [                                                                                                    ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  19733 usr00     20   0  253.4m 186.1m   4.1m S  49.7   0.3   0:25.76 stockfish_f656f
  19732 usr00     20   0  253.6m 186.1m   4.1m S  45.8   0.3   0:25.82 stockfish_f656f
  19728 usr00     20   0   70.8m  46.4m   3.2m S  11.8   0.1   0:40.88 cutechess-cli
```
- concurrency 48
```
top - 12:14:54 up 54 min,  1 user,  load average: 54.16, 30.42, 20.34
Tasks: 124 total,   4 running, 120 sleeping,   0 stopped,   0 zombie
%Cpu(s): 50.1 us, 43.0 sy,  0.0 ni,  6.9 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem : 28.6/64251.3  [|||||||||||||||||||||||||||||                                                                       ]
MiB Swap:  0.0/16384.0  [                                                                                                    ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  18870 usr00     20   0   81.4m  51.1m   3.2m S  2449   0.1  86:20.35 cutechess-cli
  19085 usr00     20   0  246.6m 179.2m   4.0m S  35.5   0.3   0:46.86 stockfish_f656f
  18937 usr00     20   0  247.2m 179.8m   4.1m S  34.6   0.3   0:47.39 stockfish_f656f
  19039 usr00     20   0  247.2m 179.6m   4.0m S  32.2   0.3   0:44.01 stockfish_f656f
  18935 usr00     20   0  246.6m 181.3m   4.2m S  30.6   0.3   0:48.48 stockfish_f656f
```
**cutechess with 1.5 GiB RAM**
- concurrency 1
```
top - 12:20:28 up  2:29,  2 users,  load average: 1.09, 1.17, 1.54
Tasks: 128 total,   1 running, 127 sleeping,   0 stopped,   0 zombie
%Cpu(s): 22.8 us,  1.8 sy,  0.0 ni, 75.2 id,  0.2 wa,  0.0 hi,  0.1 si,  0.0 st
MiB Mem : 48.6/1519.7   [||||||||||||||||||||||||||||||||||||||||||||                                              ]
MiB Swap:  2.5/2048.0   [||                                                                                        ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   7632 vmuser    20   0  248.0m 180.8m   4.0m S  50.0  11.9   0:10.72 stockfish_f656f
   7633 vmuser    20   0  247.9m 180.7m   4.0m S  46.7  11.9   0:10.35 stockfish_f656f
   7628 vmuser    20   0   70.7m  44.3m   2.8m S  10.7   2.9   1:21.70 cutechess-cli

```
- concurrency 4
```
top - 12:07:43 up  2:16,  2 users,  load average: 2.47, 2.11, 2.03
Tasks: 133 total,   2 running, 131 sleeping,   0 stopped,   0 zombie
%Cpu(s): 65.3 us,  5.0 sy,  0.0 ni,  7.2 id, 19.6 wa,  0.0 hi,  2.9 si,  0.0 st
MiB Mem : 98.0/1519.7   [||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||  ]
MiB Swap: 18.7/2048.0   [|||||||||||||||||||                                                                                 ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   6628 vmuser    20   0  247.6m 172.0m   3.8m S  34.0  11.3   0:10.82 stockfish_f8771
   6625 vmuser    20   0  244.8m 175.6m   3.6m S  32.7  11.6   0:03.96 stockfish_f8771
   6620 vmuser    20   0   71.5m   2.6m   1.9m R  23.4   0.2   2:07.66 cutechess-cli
```


**fast-chess with 64 GiB RAM**
- concurrency 1
```
top - 14:47:58 up  3:30,  1 user,  load average: 0.98, 3.53, 9.37
Tasks:  29 total,   1 running,  28 sleeping,   0 stopped,   0 zombie
%Cpu(s):  2.0 us,  0.1 sy,  0.0 ni, 97.9 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem :  3.3/64251.3  [|||                                                                                                 ]
MiB Swap:  0.0/16384.0  [                                                                                                    ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  27132 usr00     20   0  250.0m 182.9m   4.2m S  52.6   0.3   1:13.41 stockfish_f8771
  27130 usr00     20   0  249.9m 182.7m   4.1m S  42.5   0.3   1:14.17 stockfish_f8771
  27126 usr00     20   0  482.3m 407.9m   4.5m S   4.2   0.6   0:07.96 fastchess
```
- concurrency 48
```
top - 14:33:00 up  3:15,  1 user,  load average: 53.87, 37.49, 18.49
Tasks: 124 total,   2 running, 122 sleeping,   0 stopped,   0 zombie
%Cpu(s): 96.6 us,  1.3 sy,  0.0 ni,  2.1 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem : 30.3/64251.3  [||||||||||||||||||||||||||||||                                                                      ]
MiB Swap:  0.0/16384.0  [                                                                                                    ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  22504 usr00     20   0 3866.5m 413.1m   4.7m S  63.7   0.6   2:48.84 fastchess
  22704 usr00     20   0  250.6m 184.7m   4.0m S  62.0   0.3   2:05.07 stockfish_f656f
  22719 usr00     20   0  251.9m 184.7m   4.0m S  60.7   0.3   2:03.61 stockfish_f656f
  22564 usr00     20   0  250.3m 182.8m   4.1m S  59.4   0.3   2:05.92 stockfish_f656f
  22765 usr00     20   0  251.0m 183.8m   4.1m S  59.4   0.3   2:03.78 stockfish_f656f
  22666 usr00     20   0  249.7m 184.3m   4.1m S  58.7   0.3   2:05.35 stockfish_f656f
```

**fast-chess with 1.5 GiB RAM**
- concurrency 1
```
top - 12:49:47 up  2:58,  2 users,  load average: 2.04, 2.95, 2.85
Tasks: 127 total,   1 running, 126 sleeping,   0 stopped,   0 zombie
%Cpu(s): 24.4 us,  0.8 sy,  0.0 ni, 74.7 id,  0.0 wa,  0.0 hi,  0.2 si,  0.0 st
MiB Mem : 70.5/1519.7   [||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||                          ]
MiB Swap:  2.5/2048.0   [||                                                                                        ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   9046 vmuser    20   0  249.1m 182.0m   4.1m S  50.5  12.0   0:15.23 stockfish_f656f
   9048 vmuser    20   0  248.9m 181.6m   4.0m S  48.2  11.9   0:14.50 stockfish_f656f
   9042 vmuser    20   0  482.3m 408.0m   4.6m S   2.3  26.8   0:02.59 fastchess

```
- concurrency 4
```
top - 12:47:16 up  2:56,  2 users,  load average: 4.05, 3.10, 2.83
Tasks: 134 total,   1 running, 133 sleeping,   0 stopped,   0 zombie
%Cpu(s): 90.5 us,  1.8 sy,  0.0 ni,  1.0 id,  6.1 wa,  0.0 hi,  0.6 si,  0.0 st
MiB Mem : 98.9/1519.7   [||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||| ]
MiB Swap: 50.0/2048.0   [|||||||||||||||||||||||||||||||||||||||||||||                                             ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   8617 vmuser    20   0  250.0m 151.1m   2.9m S  45.8   9.9   1:06.56 stockfish_f656f
   8622 vmuser    20   0  249.5m 161.7m   2.9m S  43.2  10.6   1:08.29 stockfish_f656f
   8615 vmuser    20   0  249.5m 155.6m   3.0m S  36.5  10.2   1:06.21 stockfish_f656f
   8606 vmuser    20   0  698.3m   4.9m   3.1m S   6.0   0.3   0:10.89 fastchess
```

Disservin:
A small difference to cutechess is that fastchess loads the book into memory but only the openings which will be played, it will remove openings which it will never reach with the configuration it was launched. 

This makes it use some more memory, I think the book is also copied twice currently with both references being alive, but the other copy should be released relatively soon afterwards. 

It seems that if the system is under more load (high concurrency) the memory usage goes down

64GiB RAM
1th -> 64th => 407M -> 184M

1.5 GiB RAM
1th -> 4th => 408M -> 4.9M

This is probably due to the fact that the memory is still reserved for the program for future allocations from the OS, if the OS is under more load it will reclaim the free memory from fastchess.

ppigazzini:
The new version fast-chess https://github.com/Disservin/fastchess/commit/3022d2a6dc0d13e70ac1a4016e9a5595c34618e2 is using half the memory with 48 threads now.

```
top - 20:30:32 up  9:19,  1 user,  load average: 53.53, 46.69, 48.05
Tasks: 125 total,   2 running, 123 sleeping,   0 stopped,   0 zombie
%Cpu(s): 96.3 us,  1.3 sy,  0.0 ni,  2.4 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem : 30.1/64251.3  [||||||||||||||||||||||||||||||                                                                      ]      MiB Swap:  0.0/16384.0  [                                                                                                    ]

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 178911 usr00     20   0 3665.1m 212.0m   4.5m S  70.1   0.3   2:27.81 fastchess
 179194 usr00     20   0  250.7m 183.1m   4.2m S  63.1   0.3   1:49.55 stockfish_f656f
 178999 usr00     20   0  251.5m 183.9m   4.2m S  62.8   0.3   1:49.57 stockfish_f656f
 178977 usr00     20   0  249.7m 195.0m   4.2m S  61.5   0.3   1:49.80 stockfish_f656f
 179156 usr00     20   0  248.9m 193.8m   4.1m S  58.5   0.3   1:48.92 stockfish_f656f
 179017 usr00     20   0  250.1m 182.7m   4.4m S  57.1   0.3   1:50.36 stockfish_f656f
```

Disservin:
fyi, pondering is currently not implemented in fastchess

ppigazzini:
On Windows with 4 threads uses 6.5 MiB

image

ppigazzini:
Workers at concurrency 48 in a test with TC 1+0.01, screenshots for reference.
- msys2
image
- WSL2
image

ppigazzini (approved):
Tested fine on DEV.

ppigazzini:
Worker update triggered
image

ppigazzini:
Thank you to @vondele @Disservin @gahtan-syarif

PR #2121, opened 2024-07-18
Switch from cutechess-cli to fast-chess (II)

vondele:
Alternative to (or continuation on top of) https://github.com/official-stockfish/fishtest/pull/2119

This takes the transition a step further and uses the capability of fast-chess to output the pentanomials. As a result some of the worker code can be simplified.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,6 +14,7 @@ Fanael Linithien (Fanael)
 Fauzi Akram Dabat (FauziAkram)
 FieryDragonLord
 Gabe (MrBrain295)
+Gahtan Nahdi (gahtan-syarif)
 Giacomo Lorenzetti (G-Lorenz)
 Gian-Carlo Pascutto (gcp)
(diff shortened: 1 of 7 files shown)


vondele:
@vdbergh you might want to have a careful look at this.

vdbergh:
I had a look and everything seems fine. It is more robust that the previous code which - IIRC - relied on the fact that cutechess schedules games in a game pair consecutively to obtain the pentanomial frequencies. 

The server will reject results with incompatible WLD and pentanomial frequencies (this amounts to checking some equalities and some inequalities). So that is also a good check.

vondele:
OK, good to know that the server does that check. 

Yes, in that case, it almost certainly is fine, and I would tend to prefer this PR directly over https://github.com/official-stockfish/fishtest/pull/2119 ... no need to go in 2 steps.

vondele:
closing this PR, integrating it directly in #2119

PR #2122, opened 2024-07-20
Allow a user to delete their own account

peregrineshahin:
image

### DIFF
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -58,6 +58,11 @@ class ActionDb:
 
         return actions_list, count
 
+    def action_related_to_user(self, username):
+        q = {"username": username}
+        action_exists = self.actions.find_one(q)
+        return action_exists is not None
+
     def failed_task(self, username=None, run=None, task_id=None, message=None):
         task = run["tasks"][task_id]
         self.insert_action(
--- a/server/fishtest/templates/user.mak
+++ b/server/fishtest/templates/user.mak
@@ -17,7 +17,42 @@
           e.target.submit();
       });
     }
+    
+    async function handleDeleteAccount() {
+      await DOMContentLoaded();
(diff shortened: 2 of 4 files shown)


vdbergh:
How about flagging the account as deleted, instead of deleting it from the database?

peregrineshahin:
> How about flagging the account as deleted, instead of deleting it from the database?

Tests will not be deleted and shall be accessible with links as they are not user-url-encoded, although the test page of the user will not show their tests as it will give 404 after clicking their tests link from a test which is fine I think.
Removing the user from the database in case they wanted so, keeps the integrety of the website and is by today's standard not just security but also related to legal ramifications from GDPR and similar laws.

ddobbelaere:
After the account is deleted, is it still possible (e.g. by somebody else) to reregister the same account name? 

If so, this certainly does not feel right to me and means that there should be at least a list if past account names kept. But you might as well "flag" the account as deleted then (along with removing all possible personal data except the user name).

peregrineshahin:
> After the account is deleted, is it still possible (e.g. by somebody else) to reregister the same account name?
> 
> If so, this certainly does not feel right to me and means that there should be at least a list of past account names kept. But you might as well "flag" the account as deleted then (along with removing all possible personal data except the user name), as @vdbergh alludes to.

That's not the only way to prevent impersonation, we can check for registeration on demand if there was any action related such username before.

vondele:
some comments here that come to my mind... nothing definitive.

* I do think we need to keep a trace of that user in the DB. The user handle is part of the logs / tests etc.
* There need to be a mechanism to prevent misuse, e.g. create account, spam, delete account is probably not ideal. So, like we do for creation, I think it makes sense to require approval for deletion.
* I think rather than deleting the user, I think removing user data associated with that field is the way to go (which I think is just the email, maybe the real name is provided as well).
* We do need to make sure the user handle can't be recycled.

PR #2127, opened 2024-08-07
Improve eAccessibility

peregrineshahin:


peregrineshahin:
Undrafted this, probably a suffecient first step towards a some sort of full Web-Accessibility,
I tested most of the changes with some screen-readers, and introduced some changes that should help newbies in general.
We can improve on it later on.

PR #2131, opened 2024-08-17
Fix worker updates

### LINKED ISSUE #2130: Worker updates broken after refactored locking implementation
After an update is pushed, workers repeatedly print:

*** Another worker (with PID=7) is already running in this directory ***

This hasn't been discovered yet on the main instance since there have been no worker updates since its introduction.
The reason this is occuring is because the worker lock is not being released before the restart.

Viren6:
Releases the worker lock before restarting for an update. Fixes #2130

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 241
+WORKER_VERSION = 242
 
 
 @exception_view_config(HTTPException)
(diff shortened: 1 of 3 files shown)


vondele:
So that's potentially an important thing to fix first.... will it be possible to apply this fix automatically, or will users need to do something by hand (like restart the worker). If the latter is the case, we might need to go in two steps (e.g. skip lock check in a first update, and only apply the full fix in a second step).

Viren6:
I believe they would need to manually restart the worker. My workers are dockerised and I had to hard redeploy them to get them working.

vondele:
so, that probably means we will have to apply the fix in two steps if that's possible.

vdbergh:
I guess I messed up the updating process... I neglected to test it since I thought that it would work in the same way as before. But now I realize that there is a fundamental distinction between the original adhoc locking implementation and the refactored one. The former one was sort of reentrant and so it had no issues with upgrading.

vdbergh:
> so, that probably means we will have to apply the fix in two steps if that's possible.

It is possible I guess. 

PR1: this PR+comment out locking statement in the worker. 
PR2: comment in locking statements.

ppigazzini (approved):
Tested working on DEV, commenting out all the worker lock code.
Finger crossed for the workers update...

ppigazzini:
The workers are updating to 242.
Thank you @Viren6 (and @vdbergh for the 2 PRs trick)

peregrineshahin:
it looks like the two commits/PRs trick has proved failing to get most of the workers on track.
@vdbergh would there be any other solution that doesn't require the client to redeploy clean workers?

vdbergh:
It seems to be mainly @linrock 's machines that are offline. Perhaps someone can contact him on Discord to see what happened?

Disservin:
I wrote him a message

vdbergh:
I am realizing that the double upgrade trick still has the issue that if a worker was offline during the upgrade 241->242 then it can now again not upgrade without reinstalling. There is not much to be done about this I am afraid.

xu-shawn:
sounds a little bit hacky but maybe we can try renaming the lockfile to something else?

vdbergh:
> sounds a little bit hacky but maybe we can try renaming the lockfile to something else?

Great idea! It seems to me this might actually work!

xu-shawn:
Made PR #2155

vdbergh:
> I am realizing that the double upgrade trick still has the issue that if a worker was offline during the upgrade 241->242 then it can now again not upgrade without reinstalling. There is not much to be done about this I am afraid.

Sorry I haven't thought about this for a while. Reinstalling isn't actually necessary. Restarting is enough (I just tested it) so it is not  such a big deal. Still renaming the lock file seems like a nice additional safety net.

ppigazzini:
To be frank, I was well aware that a stopped worker on 241 has the issue on the first start after updating to 243. But with the second start the worker runs fine.
The real goal was to keep online the already running worker. I waited that all the running workers were on 242 before triggering the update to 243.

PR #2135, opened 2024-09-15
Fix benign schema error.

vdbergh:
See

https://tests.stockfishchess.org/actions?max_actions=1&action=log_message&user=&text=&before=1725836603.523324&run_id=

Probably, in a future PR, we will delegate the purging of the "active_runs" data structure to a scheduled task so that there is no more need for the "purge_count" field.


ppigazzini (approved):

ppigazzini:

PR #2136, opened 2024-09-21
Update base settings for PT tests

vondele:
as needed since SF 17 release.


ppigazzini:
Server updated, thank you @vondele

PR #2155, opened 2025-01-03
Rename lock file

xu-shawn:
Rename lock file from `worker.lock` to `fishtest_worker.lock`. Might fix the issue with some workers being unable to update.


vdbergh:
I do not quite know how to test this PR but the essential point is that the worker and the upgraded version have different lock file name, and this is something that can be tested:

- Check out 12981ffad22f5e44420d40ea9 (worker version 241).
- Rename the lock file in `worker.py` to something else.
- Start worker.

The upgrade works!

ppigazzini:
> I do not quite know how to test this PR but the essential point is that the worker and the upgraded version have different lock file names, and this is something that can be tested:

The trick to test the PR is replacing:
https://github.com/official-stockfish/fishtest/blob/1598bdd664ea6d3b28d1a0e89e5799e755c04e8b/worker/updater.py#L13-L13
with:
```python
WORKER_URL = "https://github.com/xu-shawn/fishtest/archive/refs/heads/rename_lock.zip"
```

ppigazzini:
@xu-shawn yesterday I triggered the update to 242 (that skips the check of the lock), all the running worker updated fine. Then I triggered the update to 243 and some workers didn't recover. It seems that for some machines the old `worker.lock` stayed in place. The PR to update to 242 should have deleted the lock.

@vdbergh I wonder if, during the update, the worker pre-OpenLock maintains the PID and with OpenLock changes the PID.

vdbergh:
> It seems that for some machines the old worker.lock stayed in place. The PR to update to 242 should have some code to delete the lock file as well.

The point of 242 was to release the lock before upgrading. So I do not see how the lock file could have remained in place.

> I wonder if, during the update, the worker pre-OpenLock maintains the PID and with OpenLock changes the PID. I tested the PRs but I haven't investigated the inners.

I am not sure I understand the question. The lock file contains the PID of the locking process. But during updating the new worker gets the same PID (basically it is replacing the original worker in memory). So it is refusing to start. But 242 should have solved this...

vdbergh:
Ah now I understand the question. You are wondering why upgrading worked before openlock. 

The previous adhock locking implementation allowed the lock to be reacquired. So no problem with upgrading.

It wasn't a truly reentrant lock however as unlocking just amounted to deleting the lock file which needed to be done only once.

I didn't like this behaviour (which was just a side effect of the implementation). Either the lock should be truly reentrant (which is cumbersome to implement) or it should be as it is now.

xu-shawn:
> > It seems that for some machines the old worker.lock stayed in place. The PR to update to 242 should have some code to delete the lock file as well.
> 
> The point of 242 was to release the lock before upgrading. So I do not see how the lock file could have remained in place.

But would that not rely on 241 releasing the lock **before** calling `os.execv()` (`updater.py` line 16)? I don't think the program would continue after that point, and there's no lock release before it.

vdbergh:
> > > It seems that for some machines the old worker.lock stayed in place. The PR to update to 242 should have some code to delete the lock file as well.
> > 
> > 
> > The point of 242 was to release the lock before upgrading. So I do not see how the lock file could have remained in place.
> 
> But would that not rely on 241 releasing the lock **before** calling `os.execv()` (`updater.py` line 16)? I don't think the program would continue after that point, and there's no lock release before it.

Here is what I think should have happened: 242 disabled the locking code (i.e. it would not try to acquire a lock). So 241 could start 242. 242 also released the lock (created by 241 in this case) before upgrading to 243 which reenabled the locking code (and will now release the lock before upgrading further).

vdbergh:
Another fleet that seems to have disappeared is olafm. I don't know if it was online during the upgrade.

xu-shawn:
rebased

ppigazzini (approved):

ppigazzini:
Triggered the worker update. Thank you @xu-shawn :)
Before the update we had 222 workers from 47 contributors:


worker | contributor
-- | --
1 | 0x539
1 | Calis007
37 | ChessDBCN
1 | CounterFlow
2 | DanielMiao1
1 | Fifis
1 | Goatminola
1 | Hyphrat
1 | Lorenz
1 | Mineta
1 | MooTheCow
1 | OliverClarke
1 | RickGroszkiewicz
1 | Torom
1 | YvesKn
1 | borinot
1 | fishtester
2 | fishtrawler
1 | gentoothebest
1 | gopeto
1 | gvreuls
1 | jibarbosa
3 | leszek
32 | linrock
1 | malfoy
1 | maposora
2 | markkulix
1 | mecevdimitar
2 | mgrabiak
8 | micpilar
2 | naclosagc
2 | okrout
73 | olafm
4 | oz
3 | romangol
5 | rpngn
1 | skyfish
1 | somethingintheshadows
3 | sschnee
2 | t3hf1sht3ster
7 | technologov
1 | vaskoul
2 | vidar808
1 | wizardassassin
1 | xuhdev
3 | zeeblefritz
1 | zhujianzhao

PR #2156, opened 2025-01-06
Update the maintenance page

ppigazzini:


server/fishtest/static/html/maintenance.html
```diff
@@ -91,16 +91,16 @@ <h1>We&rsquo;ll be back soon!</h1>
...
       allowtransparency="true"
       frameborder="0"
     ></iframe>
     <script>
       const countDown = () => {
-        // write a stop date in ISO8601-compliant format (UTC)
+        // Write a stop date in ISO8601-compliant format (UTC)
         // "YYYY-MM-DDTHH:mm:ssZ"
-        const countDay = new Date("2023-01-01T00:00:00Z");
+        const countDay = new Date("2025-01-01T00:00:00Z");
```
Disservin, line 103: maybe just `const countDay = new Date(new Date().getFullYear() + "-01-01T00:00:00Z");`

PR #2158, opened 2025-01-06
Fix race condition.

vdbergh:
See
https://tests.stockfishchess.org/api/get_run/677a6ec46b38f4f9629fe598

This run is finished and yet it has nonzero nps and games_per_minute. This is possible if the scheduled task runs at the same time as the finishing of the run. In this PR we fix this and we enforce nps=0.0 and games_per_minute=0.0 in the schema.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -126,7 +126,8 @@ class RunDb:
             nps = 0.0
             games_per_minute = 0.0
             # excess of caution
-            with self.active_run_lock(str(run["_id"])):
+            run_id = str(run["_id"])
+            with self.active_run_lock(run_id):
                 tasks = copy.copy(run["tasks"])
             for task in tasks:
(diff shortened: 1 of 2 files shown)


ppigazzini:
Testing on PROD.

vdbergh:
Sorry I made an error in the schema

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1736192486.153756&run_id=

(inverted condition).

vdbergh:
No actually the schema is correct. So I need to think where the message 

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1736192486.153756&run_id=

comes from....

vdbergh:
I fixed the error in the schema.

ppigazzini:
Testing on PROD.

ppigazzini (approved):
log_messages look good on PROD

Issue #2124, opened 2024-07-28
Notify also when followed tests gets approved


peregrineshahin:
not an issue but nice to have. maybe can look to add at it later

peregrineshahin:
worth mentioning that a submitter follows his own tests automatically as of now

peregrineshahin:
An issue is that the folow=1 should be a post request instead of get, it makes no sense that if I send someone a link with `?follow=1` they follow the test???
need to be based on user actions not tricking them to follow

vdbergh:
I can explain why it is that way. Normally one would want to follow a test after creating it. The problem however is that the server may reject a new run so that in that case one would be following a non-existing test.

Hence the `?follow=1` hack when the test page is loaded after creating the test (the fact that the test page is loaded means that the run is valid). I had not anticipated the people would be sharing these links.

Currently I think that the problem mentioned in the first paragraph is not really serious. Following a non-existing test is pretty harmless since after 4.5 hours it will be purged. Moreover, in most cases people will edit the new run form until a valid run is created anyway.

vdbergh:
I will keep this issue in mind, but I am a bit hesitant to create new PRs at this time since there is already a sizable queue. I'd rather wait until the queue becomes a bit shorter.

vdbergh:
Actually now I remember that there is a more fundamental reason for the `?follow=1` hack. When we submit a new run form, the run id isn't known yet. So we cannot start following the test when we submit. We really have to do it when the server sends us back  the test page.

So the main question would be: what is a good alternative for `?follow=1`? The server sending a cookie? Including some data in a meta tag?

peregrineshahin:
Better to keep the follow hack but change it to submited_test=1 and check here https://github.com/official-stockfish/fishtest/blob/12981ffad22f5e44420d40ea9298b6ff83c0bfc2/server/fishtest/templates/tests_view.mak#L29
if the user is the same as the submitter, that would be the easiest.

peregrineshahin:
Or you can completely remove it and each time a user views his own test they start following it if they didn't.. and since you redirect to the test view immediately, the follow=1 is already redundant

peregrineshahin:
Ine trick also so that this works on all devices is most likely thee user will open the home page which if they did we can loop with js if they have any active tests or you can pass it from python.. then you can make it work on all devices by following them on this device

vdbergh:
That's not a bad idea (I am not too worried about the slight difference in functionality).

[...]

PR #2160, opened 2025-01-09
Implement the solution to #2124 suggested by peregrineshahin.

vdbergh:


ppigazzini:
If I understood correctly, with this PR the link
https://dfts-0.pigazzini.it/tests/view/67814962d2cc5840d6658793?follow=1
enable the following of the test only for the submitter (if logged in).

In my tests the link does nothing for a different user logged in, or a user not logged.

vdbergh:
> If I understood correctly, with this PR the link https://dfts-0.pigazzini.it/tests/view/67814962d2cc5840d6658793?follow=1 enable the following of the test only for the submitter (if logged in).
> 
> In my tests the link does nothing for a different user logged in, or a user not logged, and enable the following for the submitter.

Yes that is the intention of this PR. The "follow=1" url was created as a trick to subscribe the author of a test (this is explained in the issue linked to) but apparently people (unintentionally) share this follow=1 url, after creating a test, and then everyone who opens it is subscribed.

ppigazzini (approved):
Looks good on DEV

ppigazzini:

PR #2161, opened 2025-01-09
Kill worker quickly with "killall python" while building fastchess.

vdbergh:
We use the same mechanism which is used for compiling stockfish.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 247
+WORKER_VERSION = 248
 
 
 @exception_view_config(HTTPException)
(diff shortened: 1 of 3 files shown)


Disservin:
btw I just realized we are missing some warning checks, 
https://github.com/official-stockfish/fishtest/blob/0dfb3e79bee7ca9bbe43c9ac1df043b45b2d7fd8/worker/games.py#L956

fastchess has lines `Warning:` and `Warning;` (semicolon) because iirc openbench had a parser which required me to use ; for some stuff otherwise it would parse it as result lines..

Ideally the games.py would need to be changed to check for both cases and all kinds of `Warnings`

vdbergh:
Warnings are currently checked rather sparingly since there was some issue with one of the windows mechanisms for running the worker (IIRC stderr output was delayed for some reason).

vdbergh:
> Mh my comment wasn't really about your pr, I just realized that fastchess stdout warnings might not be checked entirely.. (the warnings after invoking fastchess are all stdout)

Good to know that warnings are stdout! Then we can make more use of them.

ppigazzini:
I tried unsuccessfully to have a slow exit after an error building fastchess: wrong cmd, deleting source code during the compilation etc. I'm missing what is the improvement.

By the way a sure improvement is to add a `finally:` clause to delete the temporary folder.

```diff
diff --git a/worker/worker.py b/worker/worker.py
index ee399a68..6f995630 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -518,8 +518,6 @@ def setup_fastchess(worker_dir, compiler, concurrency, global_cache):
                 )

         shutil.copy("fastchess" + EXE_SUFFIX, testing_dir)
-        os.chdir(cd)
-        shutil.rmtree(tmp_dir)

     except Exception as e:
         print(
@@ -528,6 +526,10 @@ def setup_fastchess(worker_dir, compiler, concurrency, global_cache):
             sep="",
             file=sys.stderr,
         )
+    finally:
+        os.chdir(cd)
+        shutil.rmtree(tmp_dir)
+

     return verify_required_fastchess(testing_dir / fastchess, fastchess_sha)
```

vdbergh:
> I'm missing in what case there is the improvement.

Consider the following sequence
```sh
nohup python -u worker.py
```
Check in a different terminal with `tail -f nohup.out` that fastchess starts building
```sh
killall python
```
With master fastchess will continue building. With this PR the building will stop.

ppigazzini:
Mmmm, there is an issue with my suggestion. The 2 variables are defined in the middle of the try except, they could be undefined in the finally clause.

vdbergh:
I'll have a look...

ppigazzini:
The try except should have only the critical code, right now has nearly all the code of the function.

vdbergh:
I made a bad aid fix for the issue you mention. Rewriting the function is a bit out of the scope of this PR I think (I just inherited the code).

vdbergh:
I can attempt later (I need to study the code).

ppigazzini (approved):

ppigazzini:

PR #2162, opened 2025-01-09
Backport a bug fix in openlock.

vdbergh:


ppigazzini:
https://github.com/official-stockfish/fishtest/blob/8c15e79f6a8c7cc656b031eb5e8db538fa3fc285/worker/worker.py#L63

I suggest to add also this change to the local package to avoid warning with Pylance.
```diff
diff --git a/worker/packages/openlock/openlock.py b/worker/packages/openlock/openlock.py
index 9252ee06..5dc05bb3 100644
--- a/worker/packages/openlock/openlock.py
+++ b/worker/packages/openlock/openlock.py
@@ -115,7 +115,7 @@ def set_defaults(**kw):
 class FileLock:
     def __init__(
         self,
-        lock_file="openlock.lock",
+        lock_file=Path("openlock.lock"),
         timeout=None,
     ):
         self.lock_file = Path(lock_file)
```

Upstream it should be something like this (not tested):

```python
class FileLock:
    """
    The lock constructor. An :py:class:`openlock.FileLock` object
    supports the context manager protocol.
    """

    lock_file: str | Path
    """
    The `Path` object representing the lock file.
    """
    timeout: float | None
    """
    The value of the timeout parameter.
    """
    __lock: threading.Lock
    __acquired: bool
    __retry_period: float
    __race_delay: float
    __tries: int

    def __init__(
        self,
        lock_file: str | Path = "openlock.lock",
        timeout: float | None = None,
    ) -> None:
```

vdbergh:
Thanks for the type correction! I will incorporate it upstream. Is it relevant for the worker version which has no type annotations?

ppigazzini:
image

Studio Code shows the warning: "Path" is not assignable to a "str"

vdbergh:
I will think about it tomorrow. 

This seems to be a case of type inference. Since there are no type annotations, the type checker infers from the signature of `__init__`   that `lockfile` is a string (first assignment) and then complains if it is assigned a `Path` object. 

However this is not a Python error since Python variables do not have a type.

I am not sure if it makes sense to use a type checker on a Python library that does not contain type annotations (like  the `openlock` version bundled with the worker) ...

I hope that the worker can be upgraded some day to a more recent Python version so that we can properly annotate everything...

vdbergh:
Well `pyright` is a static type checker. So it should preferably be used with type annotated libraries (otherwise its type inference will make mistakes).

This being said, while `mypy` (frequently mentioned in the Python documentation) runs cleanly on `vtsjon` and the upstream version of `openlock` (which are both fully type annotated), `pyright` has quite a few issues with them. 

I need to check if these issues are real. It could be that mypy is simply smarted than pyright....

vdbergh:
I corrected the type annotation for `lock_file` (it passes `mypy`). I don't think it makes sense to do it for the bundled copy of `openlock` since this is not annotated.

```python
def __init__(
        self,
-       lock_file: str = "openlock.lock",
+       lock_file: str | Path = "openlock.lock",
        timeout: float | None = None,
) -> None:
```

ppigazzini:
I use `pyright` in my editors and `mypy` as standalone tool installed with pipx
`mypy  --python-executable env/bin/python3 *.py`

`pyrigth` is designed to work in python language server, so it performs a lazy evaluation with speed as requirement.
`mypy` on the other hand is more accurate (but slower), it has the big defect to integrates with some delay the last python features.

In April 2024 `mypy` was not able to assess correctly the generics, se the snippet of code.

https://docs.python.org/3/library/typing.html#generics

```python
class TripleBuffer[T: (bytes, ObjectsContainer)]:
    def __init__(self) -> None:
        self.lock = threading.Lock()
        # (data, etag)
        self.read_buffer: tuple[T, str] | None = None
        self.write_buffer: tuple[T, str] | None = None
        self.swap_buffer: tuple[T, str] | None = None

    def write(self, data: T) -> None:
        pass

    def read(self) -> tuple[T, str] | None:
        pass
        
class AppState:
    def __init__(self) -> None:
        self.objects_container = TripleBuffer[ObjectsContainer]()
        self.image = TripleBuffer[bytes]()
```

ppigazzini:
script to setup on Ubuntu `neovim` with the language server.

```bash
#!/bin/bash
# 250111
# Setup Neovim with node, LSP

# Install Neovim
sudo apt update
sudo apt install -y neovim

# setup nvm and install latest node.js
# https://github.com/nvm-sh/nvm
# update with:
# cd "$NVM_DIR" && git fetch --tags origin && git switch --detach $(git describe --tags --match "v[0-9]*" `git rev-list --tags --max-count=1`) && . "$NVM_DIR/nvm.sh"
# nvm install node --reinstall-packages-from=node
# nvm install-latest-npm
# npm install -g prettier pyright
if [ ! -d "${HOME}/.nvm" ]; then
  nvm_dir="${HOME}/.nvm"
  git clone https://github.com/nvm-sh/nvm.git "$nvm_dir"
  cd "$nvm_dir"
  git checkout `git describe --abbrev=0 --tags --match "v[0-9]*" $(git rev-list --tags --max-count=1)`
  . "${nvm_dir}/nvm.sh"
  nvm install node

  cat << EOF >> ${HOME}/.profile

# nvm
export NVM_DIR="\$HOME/.nvm"
# [ -s "\$NVM_DIR/nvm.sh" ] && . "\$NVM_DIR/nvm.sh" --no-use
[ -s "\$NVM_DIR/nvm.sh" ] && . "\$NVM_DIR/nvm.sh"
EOF

  # load nvm
  . "${nvm_dir}/nvm.sh"
  # install some packages
  npm install -g prettier pyright
fi

# Clone nvim-lspconfig plugin
git clone https://github.com/neovim/nvim-lspconfig "$HOME/.config/nvim/pack/nvim/start/nvim-lspconfig"

# Write init.lua configuration file
cat > "$HOME/.config/nvim/init.lua" <<'EOF'
-- Disable mouse
vim.o.mouse = ""
-- Move cursor to last position when opening a file
vim.api.nvim_exec([[
  autocmd BufReadPost *
    \ if line("'\"") >= 1 && line("'\"") <= line("\$") && &ft !~# 'commit'
    \ |   exe "normal! g`\""
    \ | endif
]], false)

require'lspconfig'.pyright.setup{}
-- Setup language servers.
local lspconfig = require('lspconfig')
lspconfig.pyright.setup {}
lspconfig.ts_ls.setup {}
lspconfig.rust_analyzer.setup {
  -- Server-specific settings. See `:help lspconfig-setup`
  settings = {
    ['rust-analyzer'] = {},
  },
}

-- Global mappings.
-- See `:help vim.diagnostic.*` for documentation on any of the below functions
vim.keymap.set('n', '<space>e', vim.diagnostic.open_float)
vim.keymap.set('n', '[d', vim.diagnostic.goto_prev)
vim.keymap.set('n', ']d', vim.diagnostic.goto_next)
vim.keymap.set('n', '<space>q', vim.diagnostic.setloclist)

-- Use LspAttach autocommand to only map the following keys
-- after the language server attaches to the current buffer
vim.api.nvim_create_autocmd('LspAttach', {
  group = vim.api.nvim_create_augroup('UserLspConfig', {}),
  callback = function(ev)
    -- Enable completion triggered by <c-x><c-o>
    vim.bo[ev.buf].omnifunc = 'v:lua.vim.lsp.omnifunc'

    -- Buffer local mappings.
    -- See `:help vim.lsp.*` for documentation on any of the below functions
    local opts = { buffer = ev.buf }
    vim.keymap.set('n', 'gD', vim.lsp.buf.declaration, opts)
    vim.keymap.set('n', 'gd', vim.lsp.buf.definition, opts)
    vim.keymap.set('n', 'K', vim.lsp.buf.hover, opts)
    vim.keymap.set('n', 'gi', vim.lsp.buf.implementation, opts)
    vim.keymap.set('n', '<C-k>', vim.lsp.buf.signature_help, opts)
    vim.keymap.set('n', '<space>wa', vim.lsp.buf.add_workspace_folder, opts)
    vim.keymap.set('n', '<space>wr', vim.lsp.buf.remove_workspace_folder, opts)
    vim.keymap.set('n', '<space>wl', function()
      print(vim.inspect(vim.lsp.buf.list_workspace_folders()))
    end, opts)
    vim.keymap.set('n', '<space>D', vim.lsp.buf.type_definition, opts)
    vim.keymap.set('n', '<space>rn', vim.lsp.buf.rename, opts)
    vim.keymap.set('n', '<space>ca', vim.lsp.buf.code_action, opts)
    vim.keymap.set('n', 'gr', vim.lsp.buf.references, opts)
    vim.keymap.set('n', '<space>f', function()
      vim.lsp.buf.format { async = true }
    end, opts)
  end,
})
EOF
```

ppigazzini (approved):

ppigazzini:
Thank you @vdbergh :)

ppigazzini:
@vdbergh I played a bit with `ruff`
- lint like `flake8`
```
ruff check --output-format=concise --extend-exclude packages .
ruff check --statistics --extend-exclude packages .
```
- lint extreme edition
```
ruff check --select ALL --output-format=github --extend-exclude packages .
ruff check --select ALL --statistics --extend-exclude packages .
```
- replacement of `black` and `isort`
```bash
ruff check --select I --fix --extend-exclude packages .
ruff format --exclude packages .
```

PR #2163, opened 2025-01-09
Make fish.exit quit the worker after the next update_task.

vdbergh:
In other words: do not wait for the whole task to finish.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 249
+WORKER_VERSION = 250
 
 
 @exception_view_config(HTTPException)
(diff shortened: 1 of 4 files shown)


ppigazzini:
I got the duplicated worker error when restarting the worker asap after `fish.exit` (linux and msys2):
```
Remaining number of GitHub api calls = 5000
Fetching task...
Error from remote: Request_task: There is already a worker running with name "user00-3cores-f0420551-3c27" which 3 seconds ago sent an update for task 678196166e9276046ac73f35/88 (my name is "user00-3cores-f0420551-87e1")
Post request https://dfts-0.pigazzini.it:443/api/request_task handled in 114.84ms (server: 2.13ms)
Waiting 15.0 seconds before retrying
```
It seems that the exit is not totally clean like with or CTRL+C or `fish.exit` on master.

vdbergh:
Good catch! 

We need to invoke the failed task api to stop the task. 

I am going to draft this PR and think about it.

vdbergh:
Undrafted.

ppigazzini (approved):

ppigazzini:

PR #2164, opened 2025-01-10
Add priority field to cache.

vdbergh:
If we create a new task then we buffer the run with priority 1. In that way it will be the first in line for saving to the db.

There are several reasons for this PR.

- `upload_pgn` is served from a secondary instance. So we want to write the corresponding task to the db as quickly as possible, otherwise the api will fail.

-  If the tasks on the test page are served from a secondary instance then there may be entries in the event log that refer to tasks that have not yet been written to disk, probably baffling the users. So we want to make this interval as short as possible.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -729,6 +729,7 @@ class RunDb:
                     self.run_cache[r_id] = {
                         "last_access_time": time.time(),
                         "last_sync_time": time.time(),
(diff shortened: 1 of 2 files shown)


ppigazzini:
Looks good on DEV, loaded the PR on PROD.

ppigazzini (approved):
Looks good on PROD

ppigazzini:

PR #2165, opened 2025-01-11
Update fastchess

vondele:
use release alpha 1.2.0

Fixes an issue with the build (test) process on large core workers (>200)


ppigazzini:

PR #2166, opened 2025-01-11
Check for semicolon in fastchess Warning

Disservin:
Now matches https://github.com/Disservin/fastchess/blob/master/app/src/engine/uci_engine.cpp#L258


ppigazzini (approved):

ppigazzini:
Thank you @Disservin :)
I will trigger the worker update when merging #2165

PR #2167, opened 2025-01-12
Increase max core count for SPSA tasks

vondele:
it appears this limit is now less important than it was previously. Gently increase it.


ppigazzini (approved):

ppigazzini:
PROD updated.

PR #2169, opened 2025-01-13
Include compiler and env in engine name. Retire clear_binaries.

vdbergh:
Currently the worker deletes cached binaries on restart. The reason is that the user may have changed compilers, so that new engine binaries would be matched against old engine binaries, built with a different compiler.
    
In this PR we add compiler+version and a hash of selected environment variables to the engine name so that there is no risk of matching engine binaries built with different compilers, or with  different environment variables.
    
A possible enhancement (not in this PR) would be to also add the full build command to the engine name so that engine binaries could even be preserved across worker updates (a worker update may change the build command).

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -721,6 +721,7 @@ def setup_engine(
     concurrency,
     compiler,
     global_cache,
+    env,
 ):
     """Download and build sources in a temporary directory then move exe to destination"""
     tmp_dir = Path(tempfile.mkdtemp(dir=worker_dir))
@@ -767,10 +768,6 @@ def setup_engine(
             f"ARCH={arch}",
             f"COMP={comp}",
         ]
-
-        # append -DNNUE_EMBEDDING_OFF to existing CXXFLAGS environment variable, if any
-        cxx = os.environ.get("CXXFLAGS", "") + " -DNNUE_EMBEDDING_OFF"
-        env = dict(os.environ, CXXFLAGS=cxx.strip())
 
         with subprocess.Popen(
             cmd,
@@ -1208,7 +1205,6 @@ def run_games(
     run,
     task_id,
     pgn_file,
-    clear_binaries,
     global_cache,
 ):
     # This is the main fastchess driver.
@@ -1298,7 +1294,7 @@ def run_games(
     # Clean up old engines (keeping the num_bkps most recent).
     worker_dir = Path(__file__).resolve().parent
     testing_dir = worker_dir / "testing"
-    num_bkps = 0 if clear_binaries else 50
+    num_bkps = 50
     try:
         engines = sorted(
             testing_dir.glob("stockfish_*" + EXE_SUFFIX),
@@ -1323,11 +1319,36 @@ def run_games(
                     sep="",
                     file=sys.stderr,
                 )
+
+    def create_environment():
+        # OS and TEMP are necessary for msys2
+        white_set = {"PATH", "OS", "TEMP"}
+        env = {k: v for k, v in os.environ.items() if k in white_set}
+        env["CXXFLAGS"] = "-DNNUE_EMBEDDING_OFF"
+
+        # Do not hash directories such as PATH and TEMP
+        hash_set = {"OS", "CXXFLAGS"}
+        hashed_env = {k: v for k, v in env.items() if k in hash_set}
+
+        env_hash = hashlib.sha256(str(hashed_env).encode()).hexdigest()[0:10]
+        return env, env_hash
+
+    env, env_hash = create_environment()
+
+    # build signature
+    bs = (
+        worker_info["compiler"]
+        + "_"
+        + str(".".join([str(s) for s in worker_info["gcc_version"]]))
+        + "_env_"
+        + env_hash
+    )
     # Create new engines.
     sha_new = run["args"]["resolved_new"]
     sha_base = run["args"]["resolved_base"]
-    new_engine_name = "stockfish_" + sha_new
-    base_engine_name = "stockfish_" + sha_base
+
+    new_engine_name = "stockfish_" + sha_new + "_" + bs
+    base_engine_name = "stockfish_" + sha_base + "_" + bs
 
     new_engine = testing_dir / (new_engine_name + EXE_SUFFIX)
     base_engine = testing_dir / (base_engine_name + EXE_SUFFIX)
@@ -1344,6 +1365,7 @@ def run_games(
             worker_info["concurrency"],
             worker_info["compiler"],
             global_cache,
+            env,
         )
     if not base_engine.exists():
         setup_engine(
@@ -1356,6 +1378,7 @@ def run_games(
             worker_info["concurrency"],
             worker_info["compiler"],
             global_cache,
+            env,
         )
 
     os.chdir(testing_dir)
@@ -1565,7 +1588,7 @@ def run_games(
                 "-engine",
                 "name=New-" + run["args"]["resolved_new"],
                 "tc={}".format(scaled_new_tc),
-                "cmd=./{}".format(new_engine_name),
+                "cmd={}".format(new_engine),
                 "dir=.",
             ]
             + new_options
@@ -1574,7 +1597,7 @@ def run_games(
                 "-engine",
                 "name=Base-" + run["args"]["resolved_base"],
                 "tc={}".format(scaled_tc),
-                "cmd=./{}".format(base_engine_name),
+                "cmd={}".format(base_engine),
                 "dir=.",
             ]
             + base_options
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 250
+WORKER_VERSION = 251
 
 
 @exception_view_config(HTTPException)
--- a/worker/sri.txt
+++ b/worker/sri.txt
@@ -1,1 +1,1 @@
-{"__version": 250, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "rRWsXlFkv+whgfQS9IIJVEWnbquhlDa0DYRkkJ43/L7NTRR9pt2+sA2hqHH0x9M/", "games.py": "iYN33XDHqgfXnQheUtghu9GsuYhL0XU85/wUijSu87QtuwbyeAzJf2/Roqfatwgf"}
+{"__version": 251, "updater.py": "Mg+pWOgGA0gSo2TuXuuLCWLzwGwH91rsW1W3ixg3jYauHQpRMtNdGnCfuD1GqOhV", "worker.py": "mfLxjM3Pq6448qSPwbLq5lTqX1nxMVBZbz8iv4Pkn9jPzhwHULsNCtFWw7ThcPAy", "games.py": "IZmJeczQV7IDMOvWS1NqR/tPe+Zg3rI8D1QDYIz/dvRE8auUU66PEjyfhWf9nbNe"}
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -69,7 +69,7 @@ MIN_GCC_MINOR = 3
 MIN_CLANG_MAJOR = 8
 MIN_CLANG_MINOR = 0
 
-WORKER_VERSION = 250
+WORKER_VERSION = 251
 FILE_LIST = ["updater.py", "worker.py", "games.py"]
 HTTP_TIMEOUT = 30.0
 INITIAL_RETRY_TIME = 15.0
@@ -1298,7 +1298,6 @@ def fetch_and_handle_task(
     password,
     remote,
     current_state,
-    clear_binaries,
     global_cache,
     worker_lock,
 ):
@@ -1391,7 +1390,6 @@ def fetch_and_handle_task(
             run,
             task_id,
             pgn_file,
-            clear_binaries,
             global_cache,
         )
         success = True
@@ -1609,14 +1607,12 @@ def worker():
     # Start the main loop.
     delay = INITIAL_RETRY_TIME
     fish_exit = False
-    clear_binaries = True
     while current_state["alive"]:
         success = fetch_and_handle_task(
             worker_info,
             options.password,
             remote,
             current_state,
-            clear_binaries,
             options.global_cache,
             worker_lock,
         )
@@ -1637,7 +1633,6 @@ def worker():
                 safe_sleep(delay)
                 delay = min(MAX_RETRY_TIME, delay * 2)
         else:
-            clear_binaries = False
             delay = INITIAL_RETRY_TIME
 
     if fish_exit:


worker/games.py
```diff
@@ -1323,11 +1322,17 @@ def parse_options(s):
                     sep="",
                     file=sys.stderr,
                 )
+    # build signature
+    bs = (
+        worker_info["compiler"]
+        + "_"
+        + str(".".join([str(s) for s in worker_info["gcc_version"]]))
```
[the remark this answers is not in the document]
vdbergh: The field name `gcc_version` is a historical artifact. It means "compiler_version". It should probably be changed. It needs to be changed both in the worker and on the server.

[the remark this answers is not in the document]
Disservin: ah i see :D

ppigazzini:
> different comiler.

Typo in the commit message.

Nitpicking:
- after merging this, all the contributors will start randomly the worker with or without prepending CXXFLAGS="-march=native and other flags"

vdbergh:
> > different comiler.
> 
> Typo in the commit message.

Thanks. Will fix.
> 
> Nitpicking:
> 
>     * after merging this, all the contributors will start randomly the worker with or without prepending CXXFLAGS="-march=native and other flags"

I don't understand. The build procedure has not changed. The only change is that the name of the binary includes the compiler name and the version. I.e. in master
```
stockfish_c76c1793615b17b97dd504f8c81c1ff2c63c232a
```
and in the PR
```
stockfish_c76c1793615b17b97dd504f8c81c1ff2c63c232a_g++_13.3.0
```
Note that the engine name in the pgn does not change in this PR. This is set separately.

ppigazzini:
```
python3 worker.py
CXXFLAGS="-march=native" python3 worker.py
```
If I'm not wrong both the commands will produce different binaries but with the same name.

vdbergh:
Hmm... I was not aware of this possibility.

It seems bad that the users are allowed to mess with the build procedure....

ppigazzini:
It's a feature of Stockfish Makefile, that reads some environment variables.

vdbergh:
I don't know this make feature very well.

Do we know which environment variables affect the build of stockfish? In principle we could hash them and include the hash in the engine name.

vdbergh:
If the goal of this PR is to simplify the life of a fishtest dev/tester,

Partially. But it also for people that frequently restart their worker (like me...). Recompiling all binaries each time takes a significant amount of time.

Disservin:
every variable in stockfish's makefile can be overwritten from the environment

> Do we know which environment variables affect the build of stockfish? In principle we could hash them and include the hash in the engine name.

I wouldn't really want to have fishtest rely on this, it should ideally now rather little about stockfishs build process.. but to answer your question basically those here,
https://github.com/official-stockfish/Stockfish/blob/master/src/Makefile#L418-L422

read more about environment variables here:
https://www.gnu.org/software/make/manual/make.html#Environment

basically everything gets passed and used, unless overwritten by the Makefile (which well just overwrites the env stuff)

vdbergh:
> I wouldn't really want to have fishtest rely on this, it should ideally know rather little about stockfish's build process.. but to answer your question basically those here,
https://github.com/official-stockfish/Stockfish/blob/master/src/Makefile#L418-L422

It's only 3 variables. That isn't too bad. I think it might be better to clear these variables in the worker before we start the build. I really think that choosing appropriate flags should be the responsibility of the worker and not the user (otherwise we are going back to the time of `compile.txt`).

vdbergh:
Ok I now see what should be done. Whitelisting a few environment variables, useful for developers, and these should be hashed.

vdbergh:
Undrafted. We use this.
```python
    def create_environment():
        white_list = ["COMPCXX"]
        env = {}
        for k, v in os.environ.items():
            if k in white_list:
                env[k] = v
        env["CXXFLAGS"] = "-DNNUE_EMBEDDING_OFF"
        env_hash = hashlib.sha256(str(env).encode()).hexdigest()[0:10]
        env["PATH"] = os.environ["PATH"]
        return env, env_hash
```
The engine name now looks like this
```
stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_13.3.0_env_hash_25b32ad706
```

worker/games.py
```diff
@@ -1323,11 +1319,34 @@ def parse_options(s):
                     sep="",
                     file=sys.stderr,
                 )
+
+    def create_environment():
+        white_list = ["COMPCXX"]
```
ppigazzini, line 1324: I was joking in keeping `COMPCXX`, it's the dangerous one. :(
Perhaps it's better to keep the white_list for a while in case of issue.

vdbergh: Ah ok. Do you want to keep the white list as it is now. Or just use 
```
white_list = []
```

vdbergh: What I write in the OP would be an enhancement on top on this PR. The build signature would also
have a hash of the build command besides a hash of the environment. It does not replace this PR. It also requires more work. Currently the binaries are in the testing folder, which is simply replaced during an upgrade. This procedure would need to be enhanced.

vdbergh: I now set the white list to `[]`.

ppigazzini:
I got this error on msys2 worker, I think that the Makefile is expecting some variables for Windows.

```
x86_64-w64-mingw32-c++ -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-generate=profdir -pedantic -Wextra -Wshadow -Wmissing-declarations -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=b92b54ca -DGIT_DATE=20250113 -DARCH=x86-64-sse41-popcnt -flto -flto-partition=one   -c -o movepick.o movepick.cpp
x86_64-w64-mingw32-c++ -DNNUE_EMBEDDING_OFF -Wall -Wcast-qual -fno-exceptions -std=c++17 -fprofile-generate=profdir -pedantic -Wextra -Wshadow -Wmissing-declarations -DNDEBUG -O3 -funroll-loops -DIS_64BIT -msse -msse3 -mpopcnt -DUSE_POPCNT -DUSE_SSE41 -msse4.1 -DUSE_SSSE3 -mssse3 -DUSE_SSE2 -msse2 -DGIT_SHA=b92b54ca -DGIT_DATE=20250113 -DARCH=x86-64-sse41-popcnt -flto -flto-partition=one   -c -o position.o position.cpp
make[2]: Leaving directory '/home/usr00/_git/__test_folder01/fishtest/worker/tmpgifw07zi/official-stockfish-Stockfish-c085670/src'
make[1]: Leaving directory '/home/usr00/_git/__test_folder01/fishtest/worker/tmpgifw07zi/official-stockfish-Stockfish-c085670/src'

Exception running games:
Executing ['make', '-j8', 'profile-build', 'ARCH=native', 'COMP=mingw'] failed. Error: ['which: no wine in (/ucrt64/bin:/usr/local/bin:/usr/bin:/usr/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)\n', 'which: no wine in (/ucrt64/bin:/usr/local/bin:/usr/bin:/usr/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)\n', 'which: no wine in (/ucrt64/bin:/usr/local/bin:/usr/bin:/usr/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)\n', 'which: no wine in (/ucrt64/bin:/usr/local/bin:/usr/bin:/usr/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'make[2]: *** [<builtin>: misc.o] Error 127\n', 'Cannot create temporary file inmake[2]: *** Waiting for unfinished jobs....\n', ' C:\\WINDOWS\\: Permission denied\n', 'make[2]: *** [<builtin>: bitboard.o] Error 127\n', 'make[2]: *** [<builtin>: evaluate.o] Error 127\n', 'make[2]: *** [<builtin>: movegen.o] Error 127\n', 'make[2]: *** [<builtin>: benchmark.o] Error 127\n', 'make[2]: *** [<builtin>: main.o] Error 127\n', 'make[2]: *** [<builtin>: movepick.o] Error 127\n', 'make[2]: *** [<builtin>: position.o] Error 127\n', 'make[1]: *** [Makefile:1091: gcc-profile-make] Error 2\n', 'make: *** [Makefile:945: profile-build] Error 2\n']
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 493.69ms (server: 13.49ms)
Task exited
Waiting 15.0 seconds before retrying
```

ppigazzini:
https://github.com/official-stockfish/Stockfish/blob/c085670b8474dd2137446ff278f6b73f4374cc68/src/Makefile#L22-L38

If I'm not wrong the issue is about `$(OS)`

```
### Establish the operating system name
KERNEL := $(shell uname -s)
ifeq ($(KERNEL),Linux)
	OS := $(shell uname -o)
endif

### Target Windows OS
ifeq ($(OS),Windows_NT)
	ifneq ($(COMP),ndk)
		target_windows = yes
	endif
else ifeq ($(COMP),mingw)
	target_windows = yes
	ifeq ($(WINE_PATH),)
		WINE_PATH := $(shell which wine)
	endif
endif
```

vdbergh:
Could it be WINE_PATH? Is that in the environment of msys?

ppigazzini:
WINE_PATH is passed to cross_compile un linux targetting windows.
It should not enter in else, so I suspect $(OS)

ppigazzini:
```bash
$ env
...
OS=Windows_NT
...
```

vdbergh:
> WINE_PATH is passed on linux to cross_compile targetting windows. It should not enter in that else, so I suspect $(OS) is degraded.

Ok. Is OS in the environment in msys2?

You can add it to the white list to test...

vdbergh:
Ah OK. I missed your reply. I'll white list OS.

ppigazzini:
That is fixed but it seems that another variable is missing (it seems something for the temporary folder).

vdbergh:
> That is fixed but it seems that another variable is missing (it seems something for the temporary folder).

Hmm what is the error message?

ppigazzini:
That was easy:
```diff
diff --git a/worker/games.py b/worker/games.py
index d62f175b..d54229b4 100644
--- a/worker/games.py
+++ b/worker/games.py
@@ -1321,7 +1321,7 @@ def run_games(
                 )

     def create_environment():
-        white_list = []
+        white_list = ["OS", "TEMP"]
         env = {}
         for k, v in os.environ.items():
             if k in white_list:
```

This is the hard one:
```
TC limit 736.1928080689958 End time: 2025-01-13 20:57:47.784134+00:00
Failed to get console mode. Error code: 6
Engine not found at: .\./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94
```

vdbergh:
> Engine not found at: .\./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94

One would have to know which (shell?) commands generates this message...

ppigazzini:
It should be `fastchess`, but I don't understand what is wrong. I'm adding some variables to the white_list.

vdbergh:
Could it be that fastchess on windows cannot handle the engine name for some reason @Disservin ?

vdbergh:
The mix of `/` and `\` is a bit strange.

ppigazzini:
> Could it be that fastchess on windows cannot handle the engine name for some reason @Disservin ?

...pathname too long..., I will try in another folder.

ppigazzini:
still got error also using a shorter folder tree
```
Step 4/4. Deleting profile data ...
make ARCH=x86-64-sse41-popcnt COMP=mingw profileclean
make[1]: Entering directory '/home/worker/tmpjkcy8hba/official-stockfish-Stockfish-c085670/src'
make[1]: Leaving directory '/home/worker/tmpjkcy8hba/official-stockfish-Stockfish-c085670/src'
strip stockfish.exe
Obtaining EvalFile of stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_11d2464bbd.exe ...
Obtaining EvalFile of stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_11d2464bbd.exe ...
Verifying signature of stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_11d2464bbd.exe ...
CPU factor : 2.408834622470337 - tc adjusted to 144.530+1.445
CPU factor : 2.408834622470337 - tc adjusted to 2.409+0.024
CPU factor : 2.408834622470337 - tc adjusted to 2.409+0.024
TC limit 701.5730837944857 End time: 2025-01-13 21:26:15.154455+00:00
Failed to get console mode. Error code: 6
Engine not found at: .\./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_11d2464bbd

Waiting for fastchess to finish ... done
Task exited
```

vdbergh:
I tried to use Pathlib to set the engine argument to fastchess. Can you check if it makes a difference?

ppigazzini:
Tried, no difference.

I ran in testing dir fastchess by hand, but the command line on the wiki seems to be wrong on NNUE option.
```
./fastchess.exe -repeat -rounds 10000 -games 2 \
-resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20 \
-concurrency 8 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random plies=16 \
-engine name=base cmd=stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe option.EvalFile=nn-37f18f62d772.nnue option."Use NNUE=true" \
-engine name=test cmd=stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe option.EvalFile=nn-37f18f62d772.nnue option."Use NNUE=true" \
-ratinginterval 1 -each tc=10.0+0.1 proto=uci option.Threads=1 -pgnout result.pgn
Failed to open log file.
Indexing opening suite...
Started game 1 of 20000 (base vs test)
Started game 5 of 20000 (base vs test)
Started game 4 of 20000 (test vs base)
Started game 6 of 20000 (test vs base)
Started game 3 of 20000 (base vs test)
Started game 8 of 20000 (test vs base)
Started game 7 of 20000 (base vs test)
Started game 2 of 20000 (test vs base)
Warning; test doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; test doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Warning; base doesn't have option Use NNUE
Finished match
```

vdbergh:
> Tried, no difference.

Well I would hope that we would at least not get `.\./stockfish...` in the message (mix of `/` and `\` separators). Correct?

vdbergh:
> ```
> ./fastchess.exe -repeat -rounds 10000 -games 2 \
> -resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20 \
> -concurrency 8 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random plies=16 \
> -engine name=base cmd=stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe option.EvalFile=nn-37f18f62d772.nnue option."Use NNUE=true" \
> -engine name=test cmd=stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe option.EvalFile=nn-37f18f62d772.nnue option."Use NNUE=true" \
> -ratinginterval 1 -each tc=10.0+0.1 proto=uci option.Threads=1 -pgnout result.pgn
> Failed to open log file.
> Indexing opening suite...
> Started game 1 of 20000 (base vs test)
> Started game 5 of 20000 (base vs test)
> Started game 4 of 20000 (test vs base)
> Started game 6 of 20000 (test vs base)
> Started game 3 of 20000 (base vs test)
> Started game 8 of 20000 (test vs base)
> Started game 7 of 20000 (base vs test)
> Started game 2 of 20000 (test vs base)

But is seems that at least FC is able to parse the `cmd` argument correctly here.

vdbergh:
Perhaps the working directory of `fastchess` is incorrect? Maybe there is some environment variable in msys2 that specifies the working directory?

ppigazzini:
The command line on the wiki is outdated, I am writing a new one from the output of the worker.

ppigazzini:
This worked, match finished
```
./fastchess.exe -recover -repeat -games 2 -rounds 10 -ratinginterval 1 -scoreinterval 1 -autosaveinterval 0 -report penta=true -pgnout results-aa61a7e4-e423-48c7-bfb3-05ffd73054c5.pgn -site https://tests.stockfishchess.org/tests/view/678435a84fe0ba5930ec15c6 -event Batch 153: master vs master -srand 5895699939700649196 -resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20 -variant standard -concurrency 3 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random plies=16 start=15619 -engine name=New-c085670b8474dd2137446ff278f6b73f4374cc68 tc=1.0+0.01 cmd=./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe dir=. option.Hash=16 option.EvalFile=nn-1c0000000000.nnue option.EvalFileSmall=nn-37f18f62d772.nnue -engine name=Base-c085670b8474dd2137446ff278f6b73f4374cc68 tc=1.377+0.014 cmd=./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe dir=. option.Hash=16 option.EvalFile=nn-1c0000000000.nnue option.EvalFileSmall=nn-37f18f62d772.nnue -each proto=uci option.Threads=1
./fastchess.exe -recover -repeat -games 2 -rounds 10 -ratinginterval 1 -scoreinterval 1 -autosaveinterval 0 -report penta=true -pgnout results-aa61a7e4-e423-48c7-bfb3-05ffd73054c5.pgn -site https://tests.stockfishchess.org/tests/view/678435a84fe0ba5930ec15c6 -event Batch 153: master vs master -srand 5895699939700649196 -resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20 -variant standard -concurrency 3 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random plies=16 start=15619 -engine name=New-c085670b8474dd2137446ff278f6b73f4374cc68 tc=1.0+0.01 cmd=./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe dir=. option.Hash=16 option.EvalFile=nn-1c0000000000.nnue option.EvalFileSmall=nn-37f18f62d772.nnue -engine name=Base-c085670b8474dd2137446ff278f6b73f4374cc68 tc=1.377+0.014 cmd=./stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94.exe dir=. option.Hash=16 option.EvalFile=nn-1c0000000000.nnue option.EvalFileSmall=nn-37f18f62d772.nnue -each proto=uci option.Threads=1
```

vdbergh:
But it is strange that same command does not work in the worker.... 

The only explanation I can think of is that there is some environment variable in msys2 that specifies the current working directory.

vdbergh:
I made the paths absolute. Can you perhaps try again?

ppigazzini:
Ok, I'm trying. I updated the wiki with the correct fastchess command line.

ppigazzini:
> But it is strange that same command does not work in the worker....
> 
> The only explanation I can think of is that there is some environment variable in msys2 that specifies the current working directory.

I have already added PWD and OLDPWD.

ppigazzini:
Same error...
```
Failed to get console mode. Error code: 6
Engine not found at: C:/msys64/home/usr00/_git/__test_folder01/fishtest/worker/testing/stockfish_c085670b8474dd2137446ff278f6b73f4374cc68_g++_14.2.0_env_hash_a112752c94

Waiting for fastchess to finish ... done
```

vdbergh:
Ok this is an absolute well formed path and yet fastchess does not find it. 

I think we cannot solve this without @Disservin 's help...

vdbergh:
This is the `fastchess` code that seems to generate the error. There seems to be nothing wrong with it.
The call `std::filesystem::exists(p)` appears to give an incorrect answer.
```
#ifndef NO_STD_FILESYSTEM


        // convert path to a filesystem path
        auto p = std::filesystem::path(configs[i].dir) / std::filesystem::path(configs[i].cmd);


#    ifdef _WIN64
        // add .exe if . is not present
        if (p.extension().empty()) {
            p = std::filesystem::path(configs[i].dir) / std::filesystem::path(configs[i].cmd + ".exe");
        }
#    endif


        if (!std::filesystem::exists(p)) {
            throw std::runtime_error("Engine not found at: " + p.string());
        }
#endif
```

ppigazzini:
The .exe is missing in all the error messages.

Disservin:
Windows needs the .exe extension if you check the error message from fastchess you can see it's missing, 
but the filesystem only has the file with .exe, normally this would be caught by `if (p.extension().empty())`.
However in your new filename you already have an extension because of the gcc version, mainly the extension in your case is `.1.0_env_hash_1a7687b56b`, as such no `.exe` will be added by fastchess and fastchess is unable to find the engine.

vdbergh:
Ok. Perhaps I fixed it.

ppigazzini:
Worker started, finger crossed...

ppigazzini:
It works.

ppigazzini:
@Disservin if I'm not worng the 2 failed warnings are from `fastchess`

```
Failed to get console mode. Error code: 6
Failed to open log file.
Indexing opening suite...
```

vdbergh:
I am going to call it a day! Thanks for helping out!

ppigazzini:
Good night!

Disservin:
> @Disservin if I'm not wrong the 2 failed warnings are from `fastchess`
> 
> ```
> Failed to get console mode. Error code: 6
> Failed to open log file.
> Indexing opening suite...
> ```

yeah I noticed that too.. the console mode error is for nicer printing of different colored text on windows, which is not important as it is only used for `--help` and `--compliance`. I already moved that bit out of the normal code path https://github.com/Disservin/fastchess/commit/9043663fa34b422b5dd0f9ebc49d9ad8eabf5a17

and just added a commit to drop the warning about the log file which can't be opened

https://github.com/Disservin/fastchess/commit/2068a8286351b0ea2cc47b95cc020fb0cb4162f2

worker/games.py
```diff
@@ -1323,11 +1319,35 @@ def parse_options(s):
                     sep="",
                     file=sys.stderr,
                 )
+
+    def create_environment():
+        # This is necessary for msys2.
+        white_list = ["OS", "TEMP"]
```
vdbergh, line 1325: After the dust has settled... I wonder if it is really necessary to white list `TEMP`. The Stockfish makefile appears not to use it,

ppigazzini: Yes, it's necessary.

```
make[2]: Leaving directory '/home/Pasquale Pigazzini/_git/__test_folder01/fishtest/worker/tmpvvpa7ie_/official-stockfish-Stockfish-c085670/src'
make[1]: Leaving directory '/home/Pasquale Pigazzini/_git/__test_folder01/fishtest/worker/tmpvvpa7ie_/official-stockfish-Stockfish-c085670/src'

Exception running games:
Executing ['make', '-j8', 'profile-build', 'ARCH=native', 'COMP=mingw'] failed. Error: ['Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'Cannot create temporary file in C:\\WINDOWS\\: Permission denied\n', 'make[2]: *** [<builtin>: misc.o] Error 127\n', 'make[2]: *** Waiting for unfinished jobs....\n', 'make[2]: *** [<builtin>: movegen.o] Error 127\n', 'make[2]: *** [<builtin>: movepick.o] Error 127\n', 'make[2]: *** [<builtin>: benchmark.o] Error 127\n', 'make[2]: *** [<builtin>: evaluate.o] Error 127\n', 'make[2]: *** [<builtin>: bitboard.o] Error 127\n', 'make[2]: *** [<builtin>: main.o] Error 127\n', 'make[2]: *** [<builtin>: position.o] Error 127\n', 'make[1]: *** [Makefile:1091: gcc-profile-make] Error 2\n', 'make: *** [Makefile:945: profile-build] Error 2\n']
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 497.02ms (server: 14.02ms)
Task exited
Waiting 15.0 seconds before retrying
```

ppigazzini: This is when running the worker in a msys2 shell. Today, when I will get some spare time, I will test also the windows worker when run in a cmd shell (like the official windows setup does)

vdbergh: Thank! I wanted to make sure!

vdbergh:
I wonder if it is possible that if we use Fishtest to match SF against an older version that the two versions are actually built for different architectures (since the worker extracts the optimal architecture from the makefile and the newer makefile may support newer architectures).

ppigazzini:
It should be the case. But in some sense, that should be an improvement of Stockfish (intended as whole framework, not strictly the cpp code) to be tested and accepted/rejected. At example in my small sample size adding `-march=native` on some ARCH is a little consistent speedup while on the other ARCHs is neutral. That change, if added to the Makefile, could be tested.

vdbergh:
I don't actually understand why the worker uses a complicated procedure to determine the build command. Why not just
```
make profile-build
```
and let the compiler figure it out?

I checked that `make profile-build` has been supported since SF12.

ppigazzini:
If you intend the ARCH detection, it was added in the Makefile 18 months ago with SF 16.1
https://github.com/official-stockfish/Stockfish/commit/708319a433951ee5d5d74e0bf1cda218c14dd18e

Dropping the fishtest legacy code to detect the ARCH simplifies the code but stops the test vs versions before September 2023.

vdbergh:
> If you intend the ARCH detection, it was added in the Makefile 18 months ago with SF 16.1 [official-stockfish/Stockfish@708319a](https://github.com/official-stockfish/Stockfish/commit/708319a433951ee5d5d74e0bf1cda218c14dd18e)
> 
> Dropping the fishtest legacy code to detect the ARCH simplifies the code but stops the test vs versions before September 2023.




> If you intend the ARCH detection, it was added in the Makefile 18 months ago with SF 16.1 [official-stockfish/Stockfish@708319a](https://github.com/official-stockfish/Stockfish/commit/708319a433951ee5d5d74e0bf1cda218c14dd18e)
> 
> Dropping the fishtest legacy code to detect the ARCH simplifies the code but stops the test vs versions before September 2023.

Yes I just checked this as well. I guess it is to early to drop the legacy code. But I wonder if we could use the script ./get_native_properties.sh included with SF17 (I don't know how it works) to get a sensible ARCH.

ppigazzini:
> > If you intend the ARCH detection, it was added in the Makefile 18 months ago with SF 16.1 [official-stockfish/Stockfish@708319a](https://github.com/official-stockfish/Stockfish/commit/708319a433951ee5d5d74e0bf1cda218c14dd18e)
> > Dropping the fishtest legacy code to detect the ARCH simplifies the code but stops the test vs versions before September 2023.
> 
> Yes I just checked this as well. I guess it is to early to drop legacy ARCH detection in the worker. But I wonder if we could use the script `scripts/get_native_properties.sh` included with SF17 (I don't know how it works) to get a sensible ARCH.

I wrote that script with strict POSIX compliance, tested on linux/windows/macos/android x86/arm etc.
On Windows it works only with msys2/wsl2 (or other system that have /cat/proc).

On a native Windows I use `coreinfo.exe` to get the CPU feature.
https://github.com/[ppigazzini/stockfish-downloader](https://github.com/ppigazzini/stockfish-downloader)

ppigazzini:
There is the issue that the script is updated to support new ARCH added in the codebase, so we should use the right version of the script otherwise `make` fails.

We should decide if the test is about the framewrk as well (Makefile) or we test only the code. In the first case we should use the best setting (eg ARCH) for any of the two engines tested, in the second case we should use the least common ARCH.

vdbergh:
Perhaps we should just keep the legacy code for a while (`build_hash=legacy`) but for recent SF's switch to `make COMP=... profile-build`, leaving it to the SF makefile to select the optimal ARCH.

worker/games.py
```diff
@@ -1323,11 +1319,42 @@ def parse_options(s):
                     sep="",
                     file=sys.stderr,
                 )
+
+    def create_environment():
+        env = {}
+        # OS and TEMP are necessary for msys2
+        white_list = ["PATH", "OS", "TEMP"]
```
ppigazzini, line 1326: tests look good, the only way to annoy you a bit is nitpicking now:
- set assures immutability, less memory, no duplication, search time O(1) vs O(n) and set operations
- comprehension is faster/shorter/readable than loop
```diff
-        env = {}
        # OS and TEMP are necessary for msys2
-        white_list = ["PATH", "OS", "TEMP"]
-        for k, v in os.environ.items():
-            if k in white_list:
-                env[k] = v
+        white_set = {"PATH", "OS", "TEMP"}
+        env = {k: v for k, v in os.environ.items() if k in white_set}
        env["CXXFLAGS"] = "-DNNUE_EMBEDDING_OFF"
        
        # Do not hash directories such as PATH and TEMP
-        hashed_env = {}
-        hash_list = ["OS", "CXXFLAGS"]
-        for k, v in env.items():
-            if k in hash_list:
-                hashed_env[k] = v
+        hash_set = {"OS", "CXXFLAGS"}
+        hashed_env = {k: v for k, v in env.items() if k in hash_set}
```

vdbergh: I will change it (I did not know comprehension works for dictionaries).

vdbergh: Ok done!

ppigazzini (approved):
Looks good with: linux worker, msys2 worker in msys2 terminal; msys2 worker in cmd terminal.

ppigazzini:

PR #2170, opened 2025-01-15
Index worker_runs data structure by short_worker_name.

vdbergh:
If the worker runs out of GitHub api call then the server consults the `worker_runs` data structure to send the worker only runs it has seen before.

However currently this data structure is indexed by `uuid` which does not persist across worker restarts. In this PR we change the index to the worker name, which is persistent.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -1121,7 +1121,7 @@ After fixing the issues you can unblock the worker at
         near_github_api_limit = worker_info["near_github_api_limit"]
 
         # Now we sort the list of unfinished runs according to priority.
-        last_run_id = self.worker_runs.get(unique_key, {}).get("last_run", None)
+        last_run_id = self.worker_runs.get(my_name, {}).get("last_run", None)
 
         def priority(run):  # lower is better
             return (
@@ -1191,8 +1191,7 @@ After fixing the issues you can unblock the worker at
             # GitHub API limit...
             if near_github_api_limit:
                 have_binary = (
-                    unique_key in self.worker_runs
-                    and run_id in self.worker_runs[unique_key]
+                    my_name in self.worker_runs and run_id in self.worker_runs[my_name]
                 )
                 if not have_binary:
                     continue
@@ -1286,10 +1285,10 @@ After fixing the issues you can unblock the worker at
         # the worker has seen, as well as the last id that was seen.
         # Note that "worker_runs" is empty after a server restart.
         with self.worker_runs_lock:
-            if unique_key not in self.worker_runs:
-                self.worker_runs[unique_key] = {}
-            self.worker_runs[unique_key][run_id] = True
-            self.worker_runs[unique_key]["last_run"] = run_id
+            if my_name not in self.worker_runs:
+                self.worker_runs[my_name] = {}
+            self.worker_runs[my_name][run_id] = True
+            self.worker_runs[my_name]["last_run"] = run_id
 
         return {"run": run, "task_id": task_id}
 
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -849,7 +849,7 @@ active_runs_schema = {
 }
 
 worker_runs_schema = {
-    uuid: {
+    short_worker_name: {
         run_id: True,
         "last_run": run_id,
     }


ppigazzini:
I don't understand the behavior of my test.
I ran a worker only with one SPSA test, then I exhaust all the GitHub api calls.
A second worker is not able to get a test.
When I restart the first worker, I got the message that it can run only already ran tests.
The worker starts downloading new source code form GitHub?!? (Different test? Download works with exhausted API calls?)

image

vdbergh:
Hmm this really crazy. This was supposed to be a completely trivial PR.

> I don't understand the behavior of my test. I ran a worker only with one SPSA test (set to higher priority), then I exhaust all the GitHub api calls. A second worker is not able to get a test because the exhausted api calls. 

This is expected since the second worker has a different name.

> I set another test to the higher priority. When I restart the first worker, I got the message that it can run only already seen tests. The worker starts downloading form GitHub the source code of the never saw test 

This should not have happened... It must be a bug although the code is really trivial.

> (the download works with exhausted API calls?)

This is also weird. Maybe, these days, downloading something does not count towards your api calls since you can get the files in any case from raw.github.com? I noticed that my api calls are almost equal to 60.

If that's true then there would not be much point anymore in keeping track of github api calls...


> 
> ```
> Post request https://dfts-0.pigazzini.it:443/api/request_version handled in 103.01ms (server: 0.46ms)
> {'resources': {'core': {'limit': 60, 'remaining': 0, 'reset': 1737107777, 'used': 60, 'resource': 'core'}, 'graphql': {'limit': 0, 'remaining': 0, 'reset': 1737109795, 'used': 0, 'resource': 'graphql'}, 'integration_manifest': {'limit': 5000, 'remaining': 5000, 'reset': 1737109795, 'used': 0, 'resource': 'integration_manifest'}, 'search': {'limit': 10, 'remaining': 10, 'reset': 1737106255, 'used': 0, 'resource': 'search'}}, 'rate': {'limit': 60, 'remaining': 0, 'reset': 1737107777, 'used': 60, 'resource': 'core'}}
> Remaining number of GitHub api calls = 0
> 
>   We have almost exhausted our GitHub api calls.
>   The server will only give us tasks for tests we have seen before.
> ```
>

vdbergh:
Did you perhaps interrupt the compilation earler with ^C? That would explain the behaviour.

- server hands worker a run
- worker downloads the .zip
- users interrupts the compile
- server hands the same run again, thinking the worker has the binary
- the worker does not actually have the binary and redownloads the .zip (perhaps this does not cost any api calls...)

ppigazzini:
Yes CTRL+C.
I redo the test from scratch, loading a clean DB on DEV, rebuilding the workers, avoiding the CTRL+C.

ppigazzini:
Withe the setup built from scratch the worker is getting only tasks form the already seen run.

vdbergh:
> With the setup built from scratch the worker is getting only tasks form the already seen run.
> 
That's good right?

> Command to exhaust the API calls.
> 
> ```shell
> for i in {1..60}; do wget https://api.github.com/users/official-stockfish/repos; done
> ```

I was wondering how to do that.

ppigazzini:
The worker seems to not use any API call, I'm restarting 2 workers after deleting the content of "testing" folder (but fastchess) and the API count is still at 60.

The download from raw does not use API call.

`curl -JLO https://raw.githubusercontent.com/official-stockfish/books/master/UHO_Lichess_4852_v1.epd.zip && curl https://api.github.com/rate_limit`

We download from the API in case of error.
https://github.com/official-stockfish/fishtest/blob/620979af9c4c0945f25193bb42ea998613c4d503/worker/games.py#L483-L515

https://github.com/official-stockfish/fishtest/pull/1456

vdbergh:
Yes but the engine binaries do not use the `download_from_github` api. Perhaps they should (but files appear on `raw.github...` with a slight delay).

I am thinking that perhaps simply downloading files from `api.github.com` does not consume api calls.

ppigazzini:
Downloading raw files form GitHub seem to have an unofficial rate limit of 5000/hour/IP

https://stackoverflow.com/questions/66522261/does-github-rate-limit-access-to-public-raw-files
https://stackoverflow.com/questions/76645249/github-rate-limits-for-raw-githubusercontent-com-for-private-repos-orgs

ppigazzini (approved):
Looks good on DEV.

ppigazzini:

ppigazzini:
> Yes but the engine binaries do not use the `download_from_github` api. Perhaps they should (but files appear on `raw.github...` with a slight delay).
> 
> I am thinking that perhaps simply downloading files from `api.github.com` does not consume api calls.

In effect no API calls used running this:
```
rm official-stock* && curl -JLO https://api.github.com/repos/official-stockfish/Stockfish/zipball/c085670b8474dd2137446ff278f6b73f4374cc68 && curl https://api.github.com/rate_limit
```

vdbergh:
By viewing a few tests, api calls can go down very quickly however :) (I know about the token...).

PR #2173, opened 2025-01-16
Revert "Disable caching the main page."

ppigazzini:
Reverts official-stockfish/fishtest#2172


vdbergh:
Yep sorry. The global variable were used somewhere else.

PR #2175, opened 2025-01-18
Save a run with finished tasks with higher priority.

vdbergh:
Sometimes the worker list on the main page shows a worker seemingly working on two different tasks - which is impossible. This illusion occurs when one of the tasks is a new one and the other task is a finished task by the same worker which has not been written to disk yet. So from the point of view of the db it is still unfinished.

This PR mitigates this issue by buffering a run with finished tasks with priority 1 (note: a run with new tasks is buffered with priority 2, other runs are buffered with priority 0).

Another reason for this PR is that we want to reduce the number of direct buffer calls. This PR moves some buffer calls deeper in the call chain.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -325,6 +325,7 @@ class RunDb:
                             del self.connections_counter[remote_addr]
                     except Exception as e:
                         print(f"Error while deleting connection: {str(e)}", flush=True)
+        self.buffer(run, False, priority=1)
 
     def set_bad_task(self, task_id, run, residual=None, residual_color=None):
         zero_stats = {
@@ -365,6 +366,7 @@ class RunDb:
             # to zero.
             task["bad"] = True
             task["stats"] = copy.deepcopy(zero_stats)
+            self.buffer(run, False, priority=1)
 
     # Do not run two copies of this function in parallel!
     def update_aggregated_data(self):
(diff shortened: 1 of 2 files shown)


server/fishtest/rundb.py
```diff
@@ -738,6 +740,12 @@ def get_run(self, r_id):
             return self.runs.find_one({"_id": r_id_obj})
 
     def buffer(self, run, flush, priority=0, create=False):
+        """
+        Guidelines
+        ==========
+        priority=1 : finished task
+        priority=2 : new taak
```
vondele, line 747: task

vdbergh: Thanks! Fixed.

vdbergh:
Note concerning deadlock: I believe that it is currently always safe to call `buffer`, unless you call `buffer(run, True)` while holding an `active_run_lock` for a different run. Such a situation should be considered a bug.

ppigazzini (approved):
Looks good on DEV with many time odds tests (to finish in short time) and workers.

ppigazzini:

vdbergh:
Thanks! It appears to be working. I have a little program that screen scrapes the worker list and there are no more duplicate workers.

ppigazzini:
With the fleet running (real max 400k cores), sometimes the bookkeeping goes up. I wonder if this worker duplication was one of the suspects.

image

vdbergh:
I think that when a fleet disconnects brutally (the equivalent of `kill -9`) and then reconnects with new uuid-prefixes then for a few minutes the old and the new fleet will be added together. This PR will not help with that since the server has no way of knowing that the new workers are the same as the old once (since the names have changed).

Still that would not be enough to reach 2552825 cores. For this it would be necessary to repeat this scenario several times.

vondele:
vdbergh's description is accurate, this does happen. 

I believe these screenshots are actually user modifications of the webpage output, just for the lulz

PR #2176, opened 2025-01-19
Move some more buffer calls deeper in the call chain

vdbergh:
In the second commit in this PR we convert the `priority` argument to `buffer` to use an `Enum` and we retire the positional `flush` argument in favor of `priority=Prio.SAVE_NOW`.


ppigazzini (approved):
Looks good on DEV with several running/finished tests and workers.

ppigazzini:

PR #2185, opened 2025-01-25
Eliminate benign race condition.

vdbergh:


ppigazzini (approved):
Looks good on DEV, thank you @vdbergh :)

PR #2186, opened 2025-01-25
Eliminate some more unsynchronized changes to the run object.

vdbergh:


ppigazzini (approved):
Looks good on DEV, scheduled PROD update, thank you @vdbergh :)

PR #2189, opened 2025-01-26
Move the scheduler code to a separate file.

vdbergh:
In this way it is easier to maintain.


ppigazzini (approved):
Looks good on DEV.

ppigazzini:

PR #2191, opened 2025-01-30
Synchronize flush_all().

vdbergh:
The former comment about deadlock seems outdated to me. Signal handlers run in the main thread and Fishtest executes no code in the main thread once initialization is finished. So signal handlers behave like asynchronous events in the same way as api calls do, and the latter appear currently to be deadlock free.

In a second commit we make a small tweak in the procedure to stop the scheduler.


ppigazzini (approved):
Looks good on DEV, I will update PROD after the merge, thank you @vdbergh :)

PR #2192, opened 2025-02-03
Leave the run_cache in a good state after flush_all.

vdbergh:
Currently this is not important since flush_all is only called when quiting Fishtest, but it is a good programming practice to make it self contained.


ppigazzini (approved):
Quit tested on DEV.

ppigazzini:

PR #2193, opened 2025-02-03
Refresh theme cookie expiration date

MinetaS:


ppigazzini:
https://dfts-0.pigazzini.it/tests

DEV updated.
Randomly clicking to the sun/moon icons and to the home page link, the sun/moon icons seem to stop working.

MinetaS:
It seems that the first line adds dark theme CSS link first and then the duplicate one is added afterwards.

MinetaS:
I make the script set theme and removed according lines in the template.
Should be good now, but have no idea whether this is a good practice or not.

server/fishtest/static/js/application.js
```diff
@@ -77,9 +77,7 @@ function formatBytes(bytes) {
 }
 
 function handleApplicationThemes() {
-  if (!getCookie("theme")) {
-    setTheme(mediaTheme());
-  }
+  setTheme(getCookie("theme") ?? mediaTheme());
```
peregrineshahin, line 80: This passes `null` in case the left side is false, which the function doesn't handle obviously because it only checks for dark and consider anything else as light, you need to make a separate refresh function to call in the else part of the original code if statement, looks so wrong and bugged.

ppigazzini:
I confirm that it's bugged.
Select the dark theme, open a slow page, the page is loaded with the light theme, when the load is complete the page switches to the dark theme.
<https://dfts-0.pigazzini.it/contributors>

MinetaS:
I don't have any expertise in web development so I'll close this and open an issue instead.

PR #2194, opened 2025-02-04
Remove old workaround

vondele:
no longer grow the number of games per task,
it is no longer needed for fishtest performance
and limits the number of workers that can be assigned e.g. to a progression test.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -873,10 +873,7 @@ class RunDb:
         game_time = estimate_game_duration(run["args"]["tc"])
         concurrency = worker_info["concurrency"] // run["args"]["threads"]
         assert concurrency >= 1
-        # as we have more tasks done (>250), make them longer to avoid
-        # having many tasks in long running tests
-        scale_duration = 1 + min(4, len(run["tasks"]) // 250)
-        games = self.task_duration * scale_duration / game_time * concurrency
+        games = self.task_duration / game_time * concurrency
         if "sprt" in run["args"]:
             batch_size = 2 * run["args"]["sprt"].get("batch_size", 1)
             games = max(batch_size, batch_size * int(games / batch_size + 1 / 2))


ppigazzini (approved):
Looks good on DEV, PROD updating in minutes, thank you @vondele

vdbergh:
Wasn’t this introduced to avoid hitting the document size limit in mongodb?

vondele:
hmm, I didn't recall that aspect, and also that will have improved with the ability to run larger core counts in the worker and the increased stability of the workers, but it nevertheless has become more likely. The limit is around 16MiB, but it would be good to know to what that corresponds in number of tasks.

ppigazzini:
My fault, too tired to check the blame before merging.
However, better to add a clear comment about the MongoDB limit when reverting or adding another policy.

Commit and issue about hitting the 16 MiB limit in MongoDB
https://github.com/official-stockfish/fishtest/commit/11cc76e39cbabcfa5aa91f7da34a56b8f9d15adf
https://github.com/official-stockfish/fishtest/issues/1299

The bigger Viz run restored from an old backup has 20523 tasks
https://tests.stockfishchess.org/tests/view/620b4522170fcff54b33a7bf

vondele:
ok, so for reference the trigger for this was https://tests.stockfishchess.org/tests/view/67a27b0ceb183d11c659458b where there were just 118 tasks (for 60k games) and https://tests.stockfishchess.org/tests/view/67a27b0eeb183d11c659458d with 526 tasks, so despite prio 1 could not take all workers. That's quite far from the 20k that we did hit before. Will ponder about an improved solution.

vdbergh:
I guess this issue arises because these runs were small (only 60000 games). One could let the task size depend on the size of the run, to achieve more granularity.

vondele:
so, some pondering, I think it is just fine as is now. First, we're far away from that limiting number, and second, it is no good protection for reaching that limit anyway. To get to 20k tasks, what needs to be happening is mostly failure of workers to start on it, and zero sized tasks... the task size is not limiting in that case (e.g. for SPRT tests). So, I think we're fine, and actually could reduce a bit the task_duration to reduce from 30 to 10min the expected time a worker spends on a task (which would be the time for a progression test in the presence of the fleet9.

PR #2196, opened 2025-02-06
Save network file w/ Pyramid

ppigazzini:
The VPS now hosts both the fishtest server and the neural networks repository.
Save directly the network file via Pyramid retiring the FastAPI server.
Add helper script to verify and to backup to AWS S3 the network files.
The helper is scheduled to run on cron with a limited environment, it reads
the necessary environment variables directly by the .profile.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1,4 +1,5 @@
 import copy
+import gzip
 import hashlib
 import html
 import json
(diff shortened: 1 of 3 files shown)


ppigazzini:
Changes on PROD:
- [ ] create the folder `/var/www/fishtest/nn` with user permission
- [ ] symbolic link `/var/www/fishtest/static` to `${HOME}/fishtest/server/fishtest/static`
- [ ] nginx: move the root of the net to `/var/www/fishtest/nn`
- [ ] nginx: move the root of static content to `/var/www/fishtest/static`
- [ ] cron: replace the scheduled scripts with the new one and retire old variables
- [ ] systemd: delete net-server unit file
- [ ] delete net-server folder

PR #2201, opened 2025-02-08
show approver on test details

Disservin:
easier than digging through the event log


ppigazzini (approved):
Looks good on DEV, PROD update in minutes, thank you @Disservin :)

PR #2202, opened 2025-02-08
make approver name clickable in details

Disservin:
cleanup #2201


ppigazzini:
shortcut to block the approvers :)

ppigazzini (approved):

PR #2203, opened 2025-02-09
Revert "feat: capture all warnings from fastchess"

Disservin:
Reverts official-stockfish/fishtest#2200


Disservin:
okay my initial patch was not ideal since sometimes when an engine loses on time it might not report a final bestmove and instead get "interrupted" by fc... in which case fc will report a warning that no bestmove was found

this is actually simply a timeloss and shouldn't be a fatal event to trigger a stop_run, so revert it for now

https://tests.stockfishchess.org/actions?action=&user=&text=%22fastchess+says%3A+Warning%3B+No+bestmove+found%22

Disservin:
@vdbergh is there currently a way to report something to the event log but not stop the worker?

vdbergh:
> @vdbergh since I think you added this initially (the runexception logic) is there currently a way to report something to the event log but not stop the worker?

Currently there are three exceptions:

- WorkerException (stop the task and create a `failed_task` event).
- RunException (stop the run, if authorized, and create a `stop_run` even).
- FatalException (stop the worker).

Not sure what you want. Do you want to log something on the server without doing anything else?

Disservin:
> Do you want to log something on the server without doing anything else?

basically yes

Disservin:
though not entirely sure what the best behavior is.. these are some warnings which fastchess reports.. though i think most of them are useful to "watch out" for others can be ignored since they likely happen together with timeouts, i.e. "No output from", "No bestmove found in the last line from", "Last info string with score not found from" others are more "critical"

```
Warning; Engine {} is not responsive.
Warning; Engine {} didn't respond to uci.
Warning; {} doesn't have option {}
Warning; Invalid value for option {}; {}
Warning; Cannot start engine {};
Warning; No output from {}
Warning; No bestmove found in the last line from {}
Warning; Last info string with score not found from {}
Warning: No opening book specified! Consider using one, otherwise all games will be
Warning: Unknown opening format, {}. All games will be played from the starting position.
Warning: Stats will be dropped for more than 2 engines.
Warning: Bayesian SPRT model not available with pentanomial statistics. Disabling..
Warning; Move does not match uci move format, lowercase and 4/5 chars.
Warning; Illegal move {} played by {}
Warning; Illegal pv move {} pv; {}
Warning; localtime_s failed
Warning: No opening book specified! Consider using one, otherwise all games will be played from the starting position.
Warning: Unknown opening format, 2. All games will be played from the starting position.
```

how illegal moves currently handled? I don't see anything related to that in fishtest?

robertnurnberg:
Best practice might be to have a list of fc warnings that should trigger some action by fishtest. As it was in master for `doesn't have option`.

vdbergh:
Not unrelated: it might be time to figure out where the "finished match uncleanly" come from. See https://tests.stockfishchess.org/actions?action=&user=&text=uncleanly .

Disservin:
> > how are illegal moves currently handled? I don't see anything related to that in fishtest?
> 
> No there is nothing. I am guessing that cutechess considered an illegal move as a loss for the engine that makes the illegal move, but I don't know.

this is not really ideal at all.. rare illegal moves don't really happen often enough to cause a significant impact on the playing strength

vdbergh:
> > > how are illegal moves currently handled? I don't see anything related to that in fishtest?
> > 
> > 
> > No there is nothing. I am guessing that cutechess considered an illegal move as a loss for the engine that makes the illegal move, but I don't know.
> 
> this is not really ideal at all.. rare illegal moves don't really happen often enough to cause a significant impact on the playing strength and thus might not be detected

Ok. That would perhaps be an example of something that should be logged but not stop the run?

vdbergh:
> Illegal moves should stop the run imo.

They are similar to crashes and these also do not stop the run.

ppigazzini:
@Disservin I tested the PR, let me know if it's ok to merge it.

robertnurnberg:
> > Illegal moves should stop the run imo.
> 
> They are similar to crashes and these also do not stop the run.

Right. I assume that is because crashes could be not the fault of the engine?

vdbergh:
> > > Illegal moves should stop the run imo.
> > 
> > 
> > They are similar to crashes and these also do not stop the run.
> 
> Right. I assume that is because crashes could be not the fault of the engine?

That's true... In the case of SF they likely aren't.

Disservin:
> @Disservin I tested the PR, let me know if it's ok to merge it.

yeah please do, the other issues can be addressed later

ppigazzini (approved):
Looks good on test workers.

ppigazzini:
Triggered the workers update.

Disservin:
> > Illegal moves should stop the run imo.
> 
> They are similar to crashes and these also do not stop the run.

mh but currently as you said there isn't a way to report those without stopping the run, so currently the easiest solution is stop the run..?

vdbergh:
> > > Illegal moves should stop the run imo.
> > 
> > 
> > They are similar to crashes and these also do not stop the run.
> 
> mh but currently, as you said there isn't a way to report those without stopping the run, so currently the easiest solution is stop the run..?

Yes.

PR #2206, opened 2025-02-09
add log run endpoint for illegal moves and pv moves

Disservin:
a prototype for the illegal move issues discussed in #2203 based on a separate log message

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -165,6 +165,7 @@ def main(global_config, **settings):
     config.add_route("api_update_task", "/api/update_task")
     config.add_route("api_failed_task", "/api/failed_task")
     config.add_route("api_stop_run", "/api/stop_run")
+    config.add_route("api_log_run", "/api/log_run")
     config.add_route("api_request_version", "/api/request_version")
     config.add_route("api_beat", "/api/beat")
     config.add_route("api_request_spsa", "/api/request_spsa")
--- a/server/fishtest/actiondb.py
+++ b/server/fishtest/actiondb.py
@@ -213,6 +213,18 @@ class ActionDb:
             message=message,
         )
 
+    def log_run(self, username, run, task_id, message):
+        task = run["tasks"][task_id]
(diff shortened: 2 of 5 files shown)


Disservin:
tbh i'm not too sure if this is really a good approach.. i'd like to hear some ideas

for this approach i had to implement a log cache to avoid spamming the server when tests produce a lot of illegal moves, the only advantage i see with this approach is maybe being able to play against old bugged versions

vondele:
not 100% sure I understand. I think for illegal moves the run can simply be stopped? At least for fishtest?

I think timeouts are the one of the few things that might reasonably happen with SF.

Disservin:
Fishtest currently doesn't do anything when it comes to illegal moves.

@vdbergh argued that illegal moves shouldn't stop the run

vdbergh:
> Fishtest currently doesn't do anything when it comes to illegal moves.
> 
> @vdbergh argued that illegal moves shouldn't stop the run

Well I am really rather neutral here.  I agree with @robertnurnberg that the analogy with crashes (which don't stop the run) is slightly broken.

robertnurnberg:
We could add to this pr other run stopping warning messages from fastchess. Basically things that indicate something seriously broken in the engine (SF patch most likely).

vondele:
I agree this could be handled / logged as crashes, i.e. illegal real move is handled as a crash.
Illegal PV move, I don't know. For SF they are anyway not likely.

robertnurnberg:
The thing is that crashes at present are not run stopping iiuc. Probably because they may not be the fault of the engine, and more likely hint at issues with the worker (like time losses).

So a possibility would be to deal with illegal moves like we deal with missing options, for example. A run-stopping incident.

Disservin:
Problem with reporting it as a crashes is that we don't get the stdout line. 

I'd prefer if both illegal moves and illegal pv's just stop the run or how does openbench for example do it ?

vdbergh:
Let's go for stop_run then. We can see how it goes.

PR #2215, opened 2025-02-11
refactor(server): use not ambiguous naming

ppigazzini:


ppigazzini:
I always wrongly supposed that that "l" ambiguous variables highlighted by the linter were for losses...

vdbergh:
> I always wrongly supposed that that "l" ambiguous variables highlighted by the linter were for losses...

Sometimes l was used for losses. But not always I guess.

PR #2218, opened 2025-02-12
refactor(worker): improve import from "packages"

ppigazzini:
- Update `__init__.py` to lazily load submodules using a lightweight wrapper.
- When a module (e.g. "requests") is not available from the current
  environment, the fallback in the "packages" folder is used.
- Adjust `sys.path` only when "requests" is loaded from the fallback location,
  ensuring that the package path is appended only when necessary.
- The lazy loading mechanism is compatible with Python 3.6

Raise worker version to 260 (also server side)

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -34,7 +34,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 259
+WORKER_VERSION = 260
 
 
 @exception_view_config(HTTPException)
--- a/worker/games.py
+++ b/worker/games.py
@@ -23,7 +23,10 @@ from pathlib import Path
 from queue import Empty, Queue
 from zipfile import ZipFile
 
-import requests
+try:
+    import requests
+except ImportError:
+    from packages import requests
 
 IS_WINDOWS = "windows" in platform.system().lower()
 IS_MACOS = "darwin" in platform.system().lower()
--- a/worker/packages/__init__.py
+++ b/worker/packages/__init__.py
@@ -1,0 +1,28 @@
+import importlib
+import sys
+import types
+from pathlib import Path
+
+
+class LazyModule(types.ModuleType):
+    def __init__(self, name):
+        super().__init__(name)
+        self._module_path = f".{name}"
+        self._module = None
+
+    def _load(self):
+        if self._module is None:
+            pkg_path = str(Path(__file__).resolve().parent)
+            if pkg_path not in sys.path:
+                sys.path.append(pkg_path)
+            self._module = importlib.import_module(self._module_path, __name__)
+        return self._module
+
+    def __getattr__(self, item):
+        module = self._load()
+        return getattr(module, item)
+
+
+requests = LazyModule("requests")
+
+__all__ = ["requests"]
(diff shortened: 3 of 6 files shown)


vdbergh:
I was hit once by the fact that there is more than one package called expression. E.g. https://pypi.org/project/Expression/3.1.0/

ppigazzini:
One of the purposes of the PR is to revert to having only an optional dependency on the `packages` folder. Previously, the Windows worker installer deleted that folder and set up a proper virtual environment, using the last version of the dependencies.

By the with at least python 3.7 the code will be simpler:
https://peps.python.org/pep-0562/
```python
# packages/__init__.py
import importlib
import sys
from pathlib import Path


def __getattr__(name):
    if name == "requests":
        pkg_path = str(Path(__file__).resolve().parent)
        if pkg_path not in sys.path:
            sys.path.append(pkg_path)
        return importlib.import_module(f".{name}", __name__)
    raise AttributeError(f"Module {__name__} has no attribute {name}")


__all__ = ["requests"]
```

vdbergh:
I wonder if it would be possible/desirable to install the needed packages on the fly (in a venv). After all we install fastchess if it is missing. Why not Python packages?

vdbergh:
In principle we can get the wheel (which is just a zip file) directly from PyPi . E.g. https://files.pythonhosted.org/packages/6b/23/4399ec694e43c500f881dc357e69cd3adc0e0db217f7eba628a5afbb97c9/openlock-1.2.1-py3-none-any.whl.

ppigazzini:
https://docs.astral.sh/uv/concepts/projects/dependencies/#managing-dependencies
`uv` allow to set dependencies in many ways, `uv` is a single binary that can be installed everywhere (eg our "testing" folder).
uv install a standalone python if missing during the `uv sync`.

ppigazzini:
I tested this PR on linux, msys2, wsl2, with python 3.6.15 - 3.13.2
I hope that this will (not) bring down the fleet of @noobpwnftw :)

vdbergh:
> https://docs.astral.sh/uv/concepts/projects/dependencies/#managing-dependencies `uv` is a single binary that can be installed everywhere (eg our "testing" folder), it allows to set dependencies in many ways, `uv sync` installs a standalone python of a given version if missing during the `uv sync`.

Are you proposing that the worker installs uv first and then uses it to download other package?

ppigazzini:
> > https://docs.astral.sh/uv/concepts/projects/dependencies/#managing-dependencies `uv` is a single binary that can be installed everywhere (eg our "testing" folder), it allows to set dependencies in many ways, `uv sync` installs a standalone python of a given version if missing during the `uv sync`.
> 
> Are you proposing that the worker installs uv first and then uses it to download other packages?

`uv sync` installs the virtual env withe the required python version as well, searching first on the available pythons locally and then downloading a standalone python as fallback.

Disservin:
i mean if we go as far as downloading python itself through uv we might as well get an up to date compiler ourselfs lol (though i don't think a tool like uv exists for c/c++ compilers, maybe an opportunity to write one? jk )

ppigazzini:
> i mean if we go as far as downloading python itself through uv we might as well get an up to date compiler ourselfs lol (though i don't think a tool like uv exists for c/c++ compilers, maybe an opportunity to write one? jk )

The multiplatform standalone compiler exists, fast native build (all the libraries are built using `-march=native` and LTO), no PGO yet.
https://ziglang.org/download/

Bug fixed upstream: https://github.com/ziglang/zig/pull/22156 , https://github.com/ziglang/zig/pull/22241

zig is able now to build the target for mingw as well (disabling LTO in Makefile, though).

```
make -j build ARCH=x86-64-sse41-popcnt COMP=clang CXX="zig c++ -target x86_64-windows-gnu"

PS D:\> .\stockfish.exe compiler
Stockfish dev-20241208-cf10644d by the Stockfish developers (see AUTHORS file)

Compiled by                : clang++ 19.1.0 on MinGW64
Compilation architecture   : x86-64-sse41-popcnt
Compilation settings       : 64bit SSE41 SSSE3 SSE2 POPCNT
Compiler __VERSION__ macro : Clang 19.1.0 (https://github.com/ziglang/zig-bootstrap 46b9e66db90230fe62404b27b85a378ccf2c82c2)
```

Disservin:
Well that is great but not what I was talking about :D My comment was about switching compilers so that we can make use of newer c++ features in the language and the std. That's the same with zig :D

PR #2223, opened 2025-02-16
make download games button available for all

Disservin:


ppigazzini (approved):
Looks good on DEV, PROD update in minutes, thank you @Disservin :)

PR #2224, opened 2025-02-16
chore(tooling): bump to latest version

ppigazzini:


ppigazzini:
A little experiment to check that a future tooling bump will go smoothly.

PR #2227, opened 2025-02-18
Do not revive a failed run.

vdbergh:
This failed run was revived
https://tests.stockfishchess.org/actions?run_id=67b2f36338b351d1ce6405e1 and then predictably it failed again after some time. With this PR we disallow reviving a failed run.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -568,7 +568,9 @@ class RunDb:
 
         # "failed"
         # set in /api/stop_run
-        # cleared in set_active_run()
+        # in principle cleared in set_active_run() to comply with the schema
+        # but this method should never be invoked on a failed run (if it is then
+        # it is a bug)
         new_run["failed"] = False
 
         # "is_green"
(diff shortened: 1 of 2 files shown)


ppigazzini (approved):
Got on DEV the "You cannot modify a failed run" with a failed run.

ppigazzini:

ppigazzini:
@vdbergh on Discord at least one approver is unhappy with this PR.
Today a test was stopped as failed by a worker, the pgn shows wrong evaluation by master for all the positions, tomorrow the owner will check the RAM of the pc.
https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1740077574.468933&run_id=
It seems that it's not a rare case that a "bad" worker has stopped a run.

vdbergh:
As far as I am concerned you can revert his patch. Maybe it would be better to require that at two (three?) workers reject the patch before stopping the run. That is harder to implement though.

ppigazzini:
If technically not complicated, we could allow only to an approver to revive a failed run.

PR #2228, opened 2025-02-19
Overhaul cachebuster logic

Disservin:
Use the pyramid builtin way to write a cache buster by extending QueryStringCacheBuster.
While at it move the routes to a separate file for easier access.

The four individual 
```
config.add_static_view("css", "static/css", cache_max_age=3600)
config.add_static_view("js", "static/js", cache_max_age=3600)
config.add_static_view("img", "static/img", cache_max_age=3600)
config.add_static_view("html", "static/html", cache_max_age=3600)
```

were combined to just

```
config.add_static_view("static", "static", cache_max_age=3600)
```

I don't think this makes a big difference?

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -1,5 +1,3 @@
-import base64
-import hashlib
 import signal
 import sys
 import threading
@@ -7,6 +5,7 @@ import traceback
 from pathlib import Path
 
 from fishtest import helpers
+from fishtest.routes import setup_routes
 from fishtest.rundb import RunDb
 from pyramid.authentication import AuthTktAuthenticationPolicy
 from pyramid.authorization import ACLAuthorizationPolicy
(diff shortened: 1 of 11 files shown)


ppigazzini:
It works fine on DEV, if I'm not wrong the commits should be squashed.

ppigazzini (approved):
Looks good, on DEV, PROD update in minutes, very nice cleanup, thank you @Disservin

Disservin:
I forgot to adapt the js logic for the dark theme, https://github.com/official-stockfish/fishtest/blob/master/server/fishtest/static/js/application.js#L118

switching to dark theme currently doesnt work without a page refresh, i'll try to fix that later today

ppigazzini:
> I forgot to adapt the js logic for the dark theme, https://github.com/official-stockfish/fishtest/blob/master/server/fishtest/static/js/application.js#L118
> 
> switching to dark theme currently doesnt work without a page refresh, i'll try to fix that later today

I tested the original PR changing the dark theme, see below. I didn't find a different behavior in master vs the PR, but I can have missed some corner cases.

```diff
--- a/server/fishtest/static/css/theme.dark.css
+++ b/server/fishtest/static/css/theme.dark.css
@@ -2,7 +2,7 @@

 :root {
   color-scheme: dark;
-  --bg-dark-1: #111;
+  --bg-dark-1: #fff;
   --bg-dark-2: #222;
   --bg-dark-3: #333;
```

Disservin:
mh works fine in chrome but not in firefox

PR #2229, opened 2025-02-19
use webp images

Disservin:
quality is indistinguishable for me, 91% size reduction for stockfish and 72% for flags

serving the png variants doesn't hurt i guess


server/fishtest/static/html/maintenance.html
```diff
@@ -58,7 +58,7 @@
   <body>
     <article>
       <div class="header">
-        <img src="/img/stockfish.png" alt="logo" />
+        <img src="/img/stockfish.webp" alt="logo" />
```
ppigazzini, line 61: Weird, the maintenance page loads the pgn but does not find the webp.

Disservin: mh.. works for me?
grafik

Disservin: but yeah not on dev.. https://dfts-0.pigazzini.it/, though not even the any of the (static) assets https://dfts-0.pigazzini.it/img/stockfish.webp

ppigazzini: More weirdness, not it works, and the mime type is correct, before it was html/txt ...
image

ppigazzini: In a sandbox the webp is not loaded and the mime type is wrong.

image

Disservin: Do the other static files load ? 
Does it work with master ?

ppigazzini: I fixed a bug in the maintenance nginx conf. The pgn was loaded for serendipity because already cached.

ppigazzini: IMO you can delete the png, we don't have a fallback in place.

ppigazzini (approved):
Looks good on DEV, PROD update in minutes, thank you @Disservin

PR #2230, opened 2025-02-20
fix darkmode switch

Disservin:
fixes https://github.com/official-stockfish/fishtest/pull/2228#issuecomment-2670725836


ppigazzini (approved):
Looks good on DEV with chrome and firefox.

ppigazzini:
PROD updated, thank you @Disservin

PR #2232, opened 2025-02-21
Do not use raw references to run objects in wtt_map.

vdbergh:
It is dangerous to keep long running references to run objects outside the cache since such a run object may be evicted from the cache and then reinserted. This would result in a new run object with is equal to the original one, but not the same.

Currently there is one such instance (in wtt_map) which we fix in this PR.


ppigazzini (approved):

PR #2235, opened 2025-02-22
Cleanup: retire `cached_flash`.

vdbergh:
Since the main page is no longer cached, `chached_flash` has become superfluous.


ppigazzini (approved):

PR #2236, opened 2025-02-22
Create a "warning" queue for flash messages.

vdbergh:
This is used in the view `tests_modify` to alert the user that the run will have to be reapproved.


ppigazzini (approved):

PR #2239, opened 2025-02-23
Simplify handling of SPSA.

vdbergh:
Rather than maintaining a separate data structure per worker, we store the SPSA data of the worker (c,R and flip) in the corresponding task. To save space in the db we delete this data when the task becomes non-active.

We also make sure that the api "request_spsa" is properly synchronized.


ppigazzini (approved):

PR #2240, opened 2025-02-23
feat: verify crc32 checksum before pgn upload

Disservin:
Verification that the checksum of the file on the filesystem and the checksum of the data from fastchess match.


worker/worker.py
```diff
@@ -1430,40 +1432,50 @@ def fetch_and_handle_task(
...
-            except Exception as e:
-                print("Exception deleting PGN file:\n", e, sep="", file=sys.stderr)
+            # Check that the file is not corrupted
+            if crc_actual != crc_expected:
+                print(
+                    f"Checksum of file ({crc_actual}) does not match expected value ({crc_expected}).\nSkipping upload."
+                )
+            else:
+                # Ignore non UTF-8 characters in PGN file
+                data = pgn_file.read_text(encoding="utf-8", errors="ignore")
```
vondele, line 1468: as only minor nit, can we reuse the pgn_file.read_bytes() data, which we have just verified, and convert those bytes directly, instead of reading the file a second time?

Disservin:
mh.. while testing i got a mismatch after a ctrl+c event.. I'm not too sure about the order of events in the worker but I suppose this could happen

vondele:
I think that can happen. I believe what's going on is that the worker stops following fastchess output on ctrl+c, but fastchess still writes a game to the file before it is terminated. Not 100% sure what the best solution to this is... only stop watching fastchess output if the process has terminated? Idk, that might be a bit tricky.

Disservin:
yeah probably something like that..

are we btw sure that fishtest starts the upload once fastchess has completely finished? 
I had a look at the bogus pgn from the WDL model repo and the batch which had the pgn without the new lines didn't finish all it's games, so could it be that a ctrl+c in the worker currently might result in a not correct upload?
https://tests.stockfishchess.org/tests/view/678cd339f4dc0a8b4ae8ddaa

```
1311  okrout-22cores-0a0cde5b-d78f  os: Linux 6.12.9-200.fc41.x86_64; ram: 7992MiB; compiler: g++ 14.2.1; python: 3.13.1; worker: 254; arch: 64bit AVX2 SSE41 SSSE3 SSE2 POPCNT  2025-01-23 03:00:17  008 / 098  [0, 2, 1, 1, 0]  0  0  -
```

worker/worker.py
```diff
@@ -1430,40 +1432,52 @@ def fetch_and_handle_task(
...
+                mode="wb",
+                fileobj=gz_buffer,
+            ) as gz:
+                gz.write(pgn_data.encode())
+            payload["pgn"] = base64.b64encode(gz_buffer.getvalue()).decode()
+
+        print("Uploading compressed PGN of {} bytes".format(len(payload["pgn"])))
+        return send_api_post_request(remote + "/api/upload_pgn", payload)
+
+    if not pgn_file["name"] or not pgn_file["name"].exists():
```
vondele, line 1449: Better also check for zero-sized files, which otherwise leads to a crc check error:

```diff
-    if not pgn_file["name"] or not pgn_file["name"].exists():
+    if not pgn_file["name"] or not pgn_file["name"].exists() or pgn_file["name"].stat().st_size == 0:
```

vdbergh:
I am not totally sure how the control flow goes after SIGINT. The structure is 
```python
try:
    try:
      <parse fastchess output>
    finally:
      <kill fastchess>
except FatalException:
    <send pgn>
```
Looking at this it seems conceivable that the final pgn is sent before fastchess is killed when a ^C is received.

vdbergh:
It seems OK. I made the following test program
````
import signal

def on_sigint(signal, frame):
    raise Exception()

signal.signal(signal.SIGINT, on_sigint)

try:
    try:
        print(1)
        while True:
            pass
        print(2)
    finally:
        print(3)
except:
    print(4)
```
The output is 
```text
1
^C3
4
```
which is what we want!

vondele:
I think the issue is not that fastchess is still writing when we upload the pgn, but rather that fastchess can have written more to the pgn that we have not yet acknowledged, i.e. we have not parsed that output crc but the file has already more data.

vondele:
It is unrelated to this PR though, but I think one could do it slightly differently, when fastchess is running, an interrupt kills fastchess first. Once it is dead, we parse and handle all its output, and continue tearing down the worker.

vdbergh:
> It is unrelated to this PR though, but I think one could do it slightly differently, when fastchess is running, an interrupt kills fastchess first. Once it is dead, we parse and handle all its output, and continue tearing down the worker.

Well this is exactly what is happening according to my experiment.

worker/worker.py
```diff
@@ -1430,40 +1432,56 @@ def fetch_and_handle_task(
...
+            with gzip.GzipFile(
+                filename=f"{str(run_id)}-{task_id}.pgn.gz",
+                mode="wb",
+                fileobj=gz_buffer,
+            ) as gz:
+                gz.write(pgn_data.encode())
+            payload["pgn"] = base64.b64encode(gz_buffer.getvalue()).decode()
+
+        print("Uploading compressed PGN of {} bytes".format(len(payload["pgn"])))
+        return send_api_post_request(remote + "/api/upload_pgn", payload)
```
ppigazzini, line 1447: I wonder why we are returning this info `{'duration': 0.006097}`
the return value is not catched/used in the calling line.

Disservin: correct but if at any point someone wants to parse that return value they can :D

vondele:
> > It is unrelated to this PR though, but I think one could do it slightly differently, when fastchess is running, an interrupt kills fastchess first. Once it is dead, we parse and handle all its output, and continue tearing down the worker.
> 
> Well this is exactly what is happening according to my experiment.

I don't think this is really what happens in the code. What if the worker is in 
https://github.com/official-stockfish/fishtest/blob/e54829ffa53f90aa5d77111914b79fc3b1bd450c/worker/games.py#L980
when it receives the interrupt, and the queue contains more (unhandled) lines.

vdbergh:
> > > It is unrelated to this PR though, but I think one could do it slightly differently, when fastchess is running, an interrupt kills fastchess first. Once it is dead, we parse and handle all its output, and continue tearing down the worker.
> > 
> > 
> > Well this is exactly what is happening according to my experiment.
> 
> I don't think this is really what happens in the code. What if the worker is in
> 
> https://github.com/official-stockfish/fishtest/blob/e54829ffa53f90aa5d77111914b79fc3b1bd450c/worker/games.py#L980
> 
> when it receives the interrupt, and the queue contains more (unhandled) lines.

Ok I see what you mean. The content of the pgn does not always correspond exactly to the results of the task. But I wonder if fixing this is really worth the added complexity of the worker. The control flow is already quite tricky to follow.

worker/worker.py
ppigazzini: No code is in place to parse the response, someone needs in any case to add the required code.

vondele:
> Ok I see what you mean. The content of the pgn does not always correspond exactly to the results of the task. But I wonder if fixing this is really worth the added complexity of the worker. The control flow is already quite tricky to follow.

I agree with the sentiment, but maybe the insight allows for coming up with a scheme that is not more complex. 
I don't think we should do it as part of this PR though.

ppigazzini:
On windows/msys2 with `CTRL+C ` I got in all my 5 attempts a wrong crc
```
Games: 6, Wins: 1, Losses: 2, Draws: 3, Points: 2.5 (41.67 %)
Ptnml(0-2): [0, 2, 0, 1, 0], WL/DD Ratio: -nan
--------------------------------------------------
Started game 15 of 464 (New-fa6c30af81 vs Base-fa6c30af81)
Heartbeat stopped

Waiting for fastchess to finish ... done

Exception running games:
Terminated by signal SIGINT
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 86.63ms (server: 12.00ms)
Checksum of file (0x51f11001) does not match expected value (0xd487998f).
Skipping upload.
Task exited
Releasing the worker lock
```

ppigazzini:
The crc check fails every single time on my msys2 worker, the pgn is never uploaded.

vondele:
> On windows/msys2 with `CTRL+C ` I got in all my 5 attempts a wrong crc

do the checksums match if you wait for the task to finish (just to make sure the crc calculation is actually correct on windows, e.g. line endings ...).

ppigazzini:
Cross post, the crc fails always, not only with CTRL+C.

ppigazzini:
the msys2 worker crc32 matches the 7zip crc32, fastchess crc32 seems to be wrong on windows.

Disservin:
The EOL of what ?

worker/worker.py
```diff
@@ -1430,40 +1432,56 @@ def fetch_and_handle_task(
...
-                    )
+    if "spsa" not in run["args"]:
+        try:
+            file_content = pgn_file.read_bytes()
 
-            try:
-                pgn_file.unlink()
-            except Exception as e:
-                print("Exception deleting PGN file:\n", e, sep="", file=sys.stderr)
+            crc_actual = hex(zlib.crc32(file_content))
```
ppigazzini, line 1465: this should work on linux, windows, mac
```diff
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1461,10 +1461,12 @@ def fetch_and_handle_task(
     if "spsa" not in run["args"]:
         try:
             file_content = pgn_file.read_bytes()
+            normalized_content = file_content.replace(b'\r\n', b'\n').replace(b'\r', b'\n')
+            crc_actual = hex(zlib.crc32(normalized_content) & 0xffffffff)

-            crc_actual = hex(zlib.crc32(file_content))
```

ppigazzini:
> The EOL of what ?

pgn text file:
linux `\n`
windows `\r\n`
mac `\r`

worker/worker.py
ppigazzini: https://docs.python.org/3/library/zlib.html#zlib.crc32

I dropped `& 0xffffffff` because starting python 3.0 `zlib.crc32()` return an unsigned on all the platforms

Disservin:
Mh I'm a bit surprised that fastchess is writing is dependent new lines. That's why I was asking.. I will check that tomorrow

Disservin:
Also Mac nowadays is also \n, \r is from pre osx times

robertnurnberg:
> I think the issue is not that fastchess is still writing when we upload the pgn, but rather that fastchess can have written more to the pgn that we have not yet acknowledged, i.e. we have not parsed that output crc but the file has already more data.

A simple solution would be to look at the pgn data until the last empty line, and compare that with crc. Possibly iterating further.

That is, if we think having these pgn would be useful. Otherwise we could simply forgo any pgn's from interrupted tasks.

Disservin:
i added the newline normalization but dropped the replace for the old mac newlines, those macs don't even have python 3 support

ppigazzini:
I wonder if we should upload the normalized content as well to have all the pgns with the same EOL convention.
@robertnurnberg this could help with scripting?

robertnurnberg:
> I wonder if we should upload the normalized content as well to have all the pgns with the same EOL convention. @robertnurnberg this could help with scripting?

I am neutral on this. Whatever is easier for fishtest/this PR is fine.

ppigazzini:
> > I wonder if we should upload the normalized content as well to have all the pgns with the same EOL convention. @robertnurnberg this could help with scripting?
> 
> I am neutral on this. Whatever is easier for fishtest/this PR is fine.

From a fast check it seems that all the uploaded pgns have linux EOL, sorry for the noise.

Disservin:
mh is this because cutechess always writes in LF format or because the server does some compression?

robertnurnberg:
I think 
```python
data = pgn_file.read_text(encoding="utf-8", errors="ignore")
```
here potentially normalizes any newline encoding to the linux one.

ppigazzini:
> mh is this because cutechess always writes in LF format or because the server does some compression?

I checked a running test, with pgns form fastchess:
https://tests.stockfishchess.org/tests/view/67b746c8cfa7b0de4134f493

robertnurnberg:
See also https://github.com/python/cpython/pull/110880. So there is an option to specify the desired newline character. If the option is not given, the linux one is used iiuc.

Disservin:
> > mh is this because cutechess always writes in LF format or because the server does some compression?
> 
> I checked a running test, with pgns from fastchess: https://tests.stockfishchess.org/tests/view/67b746c8cfa7b0de4134f493

i don't know what this is supposed to answer ? :D

Disservin:
running this cutechess command on my windows machine
```
G:\chess\H2H-Engines\cutechess.exe -engine cmd=stockfish.exe name=stockfish -engine cmd=stockfish.exe name=stockfish2 -each proto=uci tc=1+0.01 dir=G:\Github\Stockfish\src -concurrency 24 -rounds 10 -games 2 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random -srand 888215712035837424 -pgnout mytest.pgn
```

gives me a pgn with unix line endings.. so i'll guess i'll adapt fastchess

robertnurnberg:
> gives me a pgn with unix line endings.. so i'll guess i'll adapt fastchess

I think that is a good idea. It would mean we no longer need to convert newline encodings, and external tools would report the same crc as fastchess and worker.

ppigazzini:
> > > mh is this because cutechess always writes in LF format or because the server does some compression?
> > 
> > 
> > I checked a running test, with pgns from fastchess: https://tests.stockfishchess.org/tests/view/67b746c8cfa7b0de4134f493
> 
> i don't know what this is supposed to answer ? :D

Question: "Has the bugged fastchess affected the PGNs uploaded so far?"
Answer: "No, courtesy of fishtest bug resilience feature (up and running since 2013 ignoring tons of bugs)"

ppigazzini:
check successful, pgn uploaded.
```
--------------------------------------------------
Started game 24 of 464 (Base-fa6c30af81 vs New-fa6c30af81)
Heartbeat stopped

Waiting for fastchess to finish... done

Exception running games:
Terminated by signal SIGINT
Informing the server
Post request https://dfts-0.pigazzini.it:443/api/failed_task handled in 70.00ms (server: 12.91ms)
Uploading compressed PGN of 22692 bytes
Post request https://dfts-0.pigazzini.it:443/api/upload_pgn handled in 94.22ms (server: 15.29ms)
Task exited
Releasing the worker lock
```

ppigazzini (approved):
Looks good on DEV with windows and linux workers uploading PGNs. PROD update in seconds, thank you @Disservin @vondele :)

PR #2241, opened 2025-02-23
Fix schema error introduced in #2239

vdbergh:
See

Screenshot from 2025-02-23 17-27-33


ppigazzini:
Added by hand to PROD.

vdbergh:
It is a "non-functional bug" :)

ppigazzini:
Added to DEV as well.

ppigazzini (approved):
Logs clean on DEV with a running SPSA.

PR #2242, opened 2025-02-23
Add a print statement before benchmarking

vondele:
As benchmarking takes some time, give some info to the user before this starts.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1459,6 +1459,7 @@ def run_games(
         pass
 
     # Verify that the signatures are correct.
+    print("Benchmarking worker and verifying signature", flush=True)
     run_errors = []
     try:
         base_nps, cpu_features = verify_signature(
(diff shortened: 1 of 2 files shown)


ppigazzini (approved):
Looks good on DEV, with workers (linux, msys2) and python versions (3.6 - 3.13)

ppigazzini:
Triggered the workers update, thank you @vondele

worker/worker.py
```diff
@@ -529,7 +521,7 @@ def setup_fastchess(worker_dir, compiler, concurrency, global_cache, tests=False
                     "Executing {} failed. Error: {}".format(cmd, errors)
                 )
 
-        shutil.copy("fastchess" + EXE_SUFFIX, testing_dir)
+        (build_dir / "fastchess").with_suffix(EXE_SUFFIX).rename(fastchess_path)
```
Disservin, line 524: I think this broke something..

yesterday fauzi got an error after a worker update which required a new fastchess build

grafik

"Winerror cannot create a file when that file already exists"

ppigazzini: This night I will trigger the fastchess rebuild to check if I'm able to replicate the issue.

ppigazzini: fixed in #2248

PR #2243, opened 2025-02-24
Make sure to buffer the run after request_spsa.

vdbergh:
If we don't do this then the spsa_params for the workers may not survive a server restart.

This fixes another issue with #2239.


vdbergh:
You can add a bug label to this.

ppigazzini (approved):
Looks good on DEV with some server restarts and VPS reboots. PROD update in seconds, thank you @vdbergh for the bug fix :)

PR #2244, opened 2025-02-24
Make the new spsa code robust against a server crash

vdbergh:
In a task we store not only the spsa data but also the number of games in the task when the spsa data was generated.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -37,6 +37,7 @@ from fishtest.stats.stat_util import SPRT_elo
 from fishtest.userdb import UserDb
 from fishtest.util import (
     GeneratorAsFileReader,
+    count_games,
     crash_or_time,
     estimate_game_duration,
     get_bad_workers,
@@ -50,8 +51,6 @@ from fishtest.util import (
 from fishtest.workerdb import WorkerDb
 from pymongo import DESCENDING, MongoClient
 from vtjson import ValidationError, validate
-
-boot_time = datetime.now(timezone.utc)
 
 
 class RunDb:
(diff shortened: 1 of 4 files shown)


server/fishtest/rundb.py
```diff
@@ -1254,12 +1253,10 @@ def sync_update_task(self, worker_info, run_id, task_id, stats, spsa):
 
         error = ""
 
-        def count_games(d):
-            return d["wins"] + d["losses"] + d["draws"]
-
         num_games = count_games(stats)
         old_num_games = count_games(task["stats"]) if "stats" in task else 0
         spsa_games = count_games(spsa) if "spsa" in run["args"] else 0
+        spsa_iter = (num_games - spsa_games) // 2
```
ppigazzini, line 1259: Ruff is highlighting that `spsa_iter` is never used.

ppigazzini: image

I don't know if the widget is visible only upon a condition.

vdbergh: I forgot to delete this variable. I have fixed it.

ppigazzini:
I wonder if you are using `pre-commit` locally, the wrong format should never happen in theory...

ppigazzini:
> I still need to do this. My local fishtest is a bit messed up right now.

I can format for you, let me know.

ppigazzini:
Your last attempt is good on my fishtest.

ppigazzini (approved):
Looks good on DEV, with a couple of SPSA, few server restarts and VPS reboots.
Events log and journalctl are clean, workers fine, SPSA charts without hiccups.
PROD update in seconds, thank you @vdbergh :)

EDIT: DEV is impossible to crash, I trust your local crashes :)

PR #2246, opened 2025-02-25
refactor(server): util.py

ppigazzini:
- use UHO_Lichess_4852_v1
- improve code and variable naming

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -32,8 +32,8 @@ class GeneratorAsFileReader:
         pass  # No cleanup needed, but method is required
 
 
-def hex_print(s):
-    return hashlib.md5(str(s).encode("utf-8")).digest().hex()
+def hex_print(run_id):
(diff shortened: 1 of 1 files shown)


server/fishtest/util.py
```diff
@@ -444,18 +443,19 @@ def remaining_hours(run):
         if llr >= llr_bound:
             return 0
 
-        # Assume all tests use default book (UHO_4060_v3).
-        book_positions = 242201
-        t = scipy.stats.beta(1, 15).cdf(min(N / book_positions, 1.0))
-        expected_games_llr = int(2 * N * llr_bound / llr)
+        # Assume all tests use default book (UHO_Lichess_4852_v1).
```
ppigazzini, line 446: @vdbergh I updated the book/positions according to the default book for test, but I noticed that the regression test still uses the previous book. If I'm not wrong the update is still correct.

PR #2247, opened 2025-02-26
Add document size to tests page.

vdbergh:
The main aim is to track the impact on document size of the recents spsa changes.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -177,6 +177,10 @@
                 % endif
               % endfor
               <tr>
+                <td>document size</td>
+                <td>${document_size} bytes</td>
+              </tr>
+              <tr>
                 <td>events</td>
                 <td><a href="/actions?run_id=${str(run['_id'])}">/actions?run_id=${run['_id']}</a></td>
               </tr>
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -8,6 +8,7 @@ import re
 from datetime import datetime, timezone
 from pathlib import Path
 
+import bson
 import fishtest.stats.stat_util
 import requests
 from fishtest.run_cache import Prio
@@ -1611,6 +1612,7 @@ def tests_view(request):
         "can_modify_run": can_modify_run(request, run),
         "same_user": same_user,
         "pt_info": request.rundb.pt_info,
+        "document_size": len(bson.BSON.encode(run)),
     }
 
 


vdbergh:
Currently the way information in the task is stored is very redundant. We can reduce it to storing just the flips as booleans
(`1=True`, `-1=False`). A boolean is stored as a byte.

vdbergh:
A similar comment holds for the history. We could store `iters` (4 bytes) instead of `R` and `c` (2 x 8 = 16 bytes).

ppigazzini:
> A similar comment holds for the history. We could store `iters` (4 bytes) instead of `R` and `c` (2 x 8 = 16 bytes).

This requires more work, `c` is currently used for the `% c` chart (to show the effect of the variable `c` and to have 2 slightly different shapes when switching between absolute-percentage with 1 parameter selected). Or the JS deals with 2 different types of info in SPSA, or we convert the SPSA in the DB, or we reduce the parameters used in the chart.

ppigazzini (approved):
Looks good on DEV, PROD updated in seconds, thank you @vdbergh

vdbergh:
> > A similar comment holds for the history. We could store `iters` (4 bytes) instead of `R` and `c` (2 x 8 = 16 bytes).
> 
> This requires more work, `c` is currently used for the `% c` chart (to show the effect of the variable `c` and to have 2 slightly different shapes when switching between absolute-percentage with 1 parameter selected). Or the JS deals with 2 different types of info in SPSA, or we convert the SPSA in the DB, or we reduce the parameters used in the chart.

Yes I know that this requires more work. I was just thinking out loud.

vdbergh:
So for now there do not seem to be real issues, even with the current somewhat wasteful implementation (I will fix this). The real test will be when the fleet visits during a large tune.

ppigazzini:
```
fishtest_new> db.runs.stats().size
24401115484
fishtest_new> db.pgns.stats().size
22891789426
fishtest_new> db.actions.stats().size
514442282
fishtest_new> db.deltas.stats().size
4426558
fishtest_new> db.nns.stats().size
680002
fishtest_new> db.users.stats().size
540844
fishtest_new> db.user_cache.stats().size
516399
fishtest_new> db.top_month.stats().size
31301
fishtest_new> db.workers.stats().size
17859
```

vdbergh:
The new spsa implementation has no effect on the size of finished runs. But it increases the size of unfinished runs since some data is stored in active tasks (and the number of active tasks increases with the number of workers present). 

We need to see how close we get to the 16Mb document size limit.  With the currently ongoing tunes https://tests.stockfishchess.org/tests/view/67be601fd8d5c2c657c53123 and https://tests.stockfishchess.org/tests/view/67be6a52d8d5c2c657c53201 we are still very far from this limit.

ppigazzini:
`param_history` is permanent, though.

vdbergh:
Yes but nothing was changed related to `param_history`.

ppigazzini:
> A similar comment holds for the history. We could store `iters` (4 bytes) instead of `R` and `c` (2 x 8 = 16 bytes).

PR #2249, opened 2025-02-26
Greatly reduce the amount of spsa data stored in active tasks.

vdbergh:


ppigazzini (approved):
Looks good on DEV, with some SPSA, logs clean. PROD update in seconds, thank you @vdbergh

PR #2250, opened 2025-02-28
Use bit array for storing flips in active spsa tasks

vdbergh:
Also: do not allow multiple update_spsa for a single request_spsa.

### DIFF
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -592,3 +592,28 @@ def strip_run(run):
 
 def count_games(stats):
     return stats["wins"] + stats["losses"] + stats["draws"]
+
+
+def pack_flips(flips):
+    """
+    This transforms a list of +-1 into a sequence of bytes
+    with the meaning of the indivual bits being 1:1, 0:-1.
+    """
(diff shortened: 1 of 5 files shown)


ppigazzini:
ot: suddenly my local `ruff` started sorting the import in all files in weird way, issue fixed cleaning the cache with `ruff clean`
I added a note in https://github.com/official-stockfish/fishtest/issues/634#issue-605333511

server/fishtest/util.py
```diff
@@ -592,3 +592,21 @@ def strip_run(run):
 
 def count_games(stats):
     return stats["wins"] + stats["losses"] + stats["draws"]
+
+
+def pack_flips(a):
```
ppigazzini, line 597: I wonder if you could use variable names more explicative, otherwise I need to reverse engineering the function with the risk to set a bad/wrong name :)

vdbergh:
I changed the names of the argument.  Have a look if it is ok...

ppigazzini (approved):
Looks good on DEV with few SPSA, logs clean. PROD update in seconds, thank you @vdbergh :)

PR #2252, opened 2025-02-28
Reduce the amount of statistics printed on worker start.

vdbergh:
Perhaps this is controversial... This is how it looks

```
Benchmarking worker and verifying signature...
Concurrency    :               4
Threads        :               1
Statistics for stockfish_93b966829bd7d2d9d9dce49f11e20125f48d0cfd_g++_13.3:
Mean nps       :       583686.87
Std (%)        :            1.13
Statistics for stockfish_f551f1b5a204b728c3c3842192db51e1c1032be1_g++_13.3:
Mean nps       :       561182.67
Std (%)        :            2.00
```

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -405,20 +405,11 @@ def compute_bench_statistics(results, threads, depth, *, bench_nps=False):
     median_nps = statistics.median(bench_nps_values)
     std_nps = statistics.stdev(bench_nps_values) if len(bench_nps_values) > 1 else 0
 
-    print(
-        f"Bench for {'nps' if bench_nps else 'nodes'}\n"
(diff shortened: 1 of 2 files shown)


ppigazzini:
For a while we are keeping all the whole output to collect information in case of issue with a worker and to check if the bench for nps is correct for SMP test.
Some of those statistics debunked some no-data-based hypothesis: min/max that there is not a deterministic outlier, the mean time that `bench 13` is not too short on the latest CPUs.

In these days the output has helped some users on Discord to understand why their workers had a different nps wrt the past in normal test and SMP test: a SMP test has different concurrency/threads values in the bench for nodes and bench for nps.

PR #2253, opened 2025-03-01
perf(util.py): use numpy functions

ppigazzini:


ppigazzini:
Looks good on DEV with SPSA and SPRT time odds (with tainted workers to trigger the purge).

PR #2254, opened 2025-03-02
Delete SPSA glucode that is no longer needed

vdbergh:


ppigazzini (approved):
Looks good on DEV with a SPSA, PROD update in seconds, thank you @vdbergh

PR #2257, opened 2025-03-03
Do not wrap byte arrays with Binary.

vdbergh:
The Binary wrapper (with subtype 0) seems to stem from the days that Python made no distinction between strings and byte arrays. I checked that wrapping a byte array with Binary makes no difference in the resulting bson encoding. Moreover a Binary wrapper is directly decoded as bytes and not as a Binary wrapper.

Solves:

https://tests.stockfishchess.org/actions?max_actions=1&action=log_message&user=&text=&before=1740938754.68217&run_id=

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -11,7 +11,6 @@ from datetime import UTC, datetime
 
 import fishtest.run_cache
 import fishtest.stats.stat_util
-from bson.binary import Binary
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):
Looks good on cli tests and with SPSA/SPRT tests on DEV (pgns downloaded are fine). PROD update in seconds, thank you @vdbergh

PR #2258, opened 2025-03-04
feat(server): configuration with env variables

ppigazzini:


ppigazzini:
We can retire the secret files. The environment variables must be provided according to how the developer starts his server:
- systemd unit file
- docker compose config file
- `.profile`
- launcher sh script

vdbergh:
Will just executing pserve on the command line still work with this PR?

ppigazzini:
> Will just executing pserve on the command line still work with this PR?

The goal of this PR is to ensure the server runs under any condition, without embedding fixed configurations for PROD in the code. However, I still need to decide whether to use the PROD configuration as a fallback.
I was surprised that the changes are minimal:
- the only real change is the link to the NN repo (data.stockfishchess.org)
- the FISHTEST_URL is used only for printing in 2 scheduled server tasks, so its absence has no impact
- the secrets were already empty in the wiki server setup script, they are required only on PROD
- mongodb remote connection has no use
I'm considering replacing the authentication secret used with PROD with one with a better entropy and without symbols that require escaping in some contexts.

ppigazzini:
> Will just executing pserve on the command line still work with this PR?

DEV looks good running with the 3 instances started on the cli, without setting the environment variables and the secrets.
This PR fix the user registration without the captcha secret.

ppigazzini:
PR tested:
- on PROD (running fine since this morning)
- DEV
- clean install in a VM using http and IPv4

vdbergh:
Nice!
```
pserve ~/fishtest/server/production.ini http_port=6543
FISHTEST_AUTHENTICATION_SECRET is missing, using an insecure default for authentication.
Starting server in PID 1078.
```

ppigazzini:
Updated the wiki script to setup the server, the secret is generated without symbols to avoid issues (in parametric unit file, `%` must be escaped to `%%` at example)
```bash
openssl rand -base64 $(( 40 * 5 ))  | tr -dc '[:alnum:]' |  head -c 40
```
Here is an example of the systemd unit file written by the setup.
```ini
[Unit]
Description=Fishtest Server port %i
After=network.target mongod.service

[Service]
Type=simple
Environment="FISHTEST_URL="
Environment="FISHTEST_NN_URL="
Environment="FISHTEST_AUTHENTICATION_SECRET=N6Nsf3N8UfbCskr2xidzRM90HP30f8XcDPjyn6CQ"
Environment="FISHTEST_CAPTCHA_SECRET="
ExecStart=/home/user000/fishtest/server/.venv/bin/pserve production.ini http_port=%i
Restart=on-failure
RestartSec=3
User=user000
WorkingDirectory=/home/user000/fishtest/server

[Install]
WantedBy=multi-user.target
```

ppigazzini:
Since a couple of weeks, I have setup DEV and PROD only with the "server" folder using sparse checkout.
```bash
rm -rf fishtest
git init --initial-branch=master fishtest
cd fishtest
git remote add origin https://github.com/official-stockfish/fishtest.git
git config user.email 'you@example.com'
git config user.name 'your_name'
git config core.sparseCheckout true
echo "server/" >> .git/info/sparse-checkout
git pull origin master
```
The autocompletion of the filename with tab gets a little speedup, `git grep` returns results only for the "server" files.
Sparse checkout for my test workers as well.

PR #2260, opened 2025-03-07
Block all requests after SIGINT or SIGTERM is received.

vdbergh:
In this way it is possible to do a very clean shutdown of the server.


vdbergh:
To test this PR you can for example remove `sys.exit(0)` from `rundb.py::exit_run()`.

ppigazzini:
> To test this PR you can for example remove `sys.exit(0)` from `rundb.py::exit_run()`.

What is wrong with a bucket of soaped water on DEV?

ppigazzini (approved):

PR #2261, opened 2025-03-07
feat(server): sanitize run options

ppigazzini:
Also:
refactor(worker): format fastchess options
feat(worker): use dynamic url


ppigazzini:
Stop creativity in options

image

vdbergh:
I agree this was long overdue.

PR #2265, opened 2025-03-09
feat(worker): bench setting the hash

ppigazzini:
Raise worker version to 268 (also server side)


ppigazzini:
The PR:
- benches for signature with only one process, using 1/5 of the previous time for a worker with 48 threads
- warmup the worker benching at depth 10
- benches using also the hash set in the test
- prints statistics only for the bench for nps
- prints the time used for the signature and the warmup 

The nps bench for new_engine could be dropped to speedup further the process.

Note: alternate attempts to get a bench with lower Stdev (%), running more benches to warmup and/or cooldown directly in `run_parallel_benches`, have failed with concurrency 48 at 1 thread. The CPU load was always at 100% as expected during the bench, courtesy of the warmup/cooldown bench. But max_nps / min_nps ~ 2, with 2x Stdev (%). Same result using only a cooldown bench.

```python
def run_parallel_benches(engine, concurrency, threads, hash_size, depth):

    engine_list = [engine] * concurrency *3,
    hash_list =[hash_size] * concurrency * 3,
    threads_list = [threads] * concurrency * 3,
    depth_list = [depth - 3] * concurrency + [depth] * concurrency + [depth - 3] * concurrency
    try:
        with ProcessPoolExecutor(max_workers=concurrency) as executor:
            results = list(
                executor.map(
                    run_single_bench,
                    engine_list,
                    hash_list,
                    threads_list,
                    depth_list,
                )
            )
    except Exception as e:
        message = f"Failed to run engine bench: {str(e)}"
        raise WorkerException(message, e=e)

    return results[concurrency: -concurrency]
```

vdbergh:
This code appears to be doing the wrong thing
```python
    new_engine = (testing_dir / new_engine_name).with_suffix(EXE_SUFFIX)
    base_engine = (testing_dir / base_engine_name).with_suffix(EXE_SUFFIX)
```
It seems to replace the part after the last dot by `EXE_SUFFIX`. So in particular it deletes the environment part of the engine name. E.g. a test print shows:
```python
new_engine_name = stockfish_f3bfce353168b03e4fedce515de1898c691f81ec_g++_13.3.0_env_25b32ad706
new_engine = /home/vdbergh/fishtest/worker/testing/stockfish_f3bfce353168b03e4fedce515de1898c691f81ec_g++_13.3
```

vdbergh:
`with_suffix` is used at a few other places. It seems dangerous since it does something unexpected for path names containing a dot.

ppigazzini:
the `with_suffix` does the right thing with a proper filename (hyphen as token separator, underscore as in-token separator, at max one dot for the extension). 
Search tools and other tools consider hyphen a token separator and underscore as a part of the token (try with a `cw` in vim).

PR #2266, opened 2025-03-10
feat: add new -testEnv option for fastchess

Disservin:
With this option fastchess prints all necessary debug information for illegal pv moves in one line, seperated by ` :: ` instead of `\n`. Which means fishtest is able to properly propagate this back to the user viewing the Events log. Hopefully adding more debug information to help users fix their tests and maybe https://github.com/official-stockfish/Stockfish/issues/5890.

See https://github.com/Disservin/fastchess/pull/812.

Also removed some dead code.


ppigazzini (approved):
Looks good on DEV with linux/windows workers. PROD update in seconds, thank you @Disservin :)

PR #2267, opened 2025-03-10
More intrinsic way of protecting spsa state against server crashes

vdbergh:
We include a signature in the data sent to the worker which the workers returns with its spsa_results.


ppigazzini:
Note: a not updated worker has issue with the PR, to be investigated if the failure will trigger the worker update. 
```
Ptnml(0-2): [0, 1, 2, 1, 0], WL/DD Ratio: 1.00
Error from remote: /api/update_task: request['spsa']['sig'] is missing
Post request [https://dfts-0.pigazzini.it:443/api/update_task](https://dfts-0.pigazzini.it/api/update_task) handled in 355.35ms
```

vdbergh:
The error is intentional. It should trigger a worker upgrade.

vdbergh:
Latest push is just correcting a typo in the commit message.

ppigazzini (approved):
Looks good on DEV, PROD update in seconds (finger crossed), thank you @vdbergh

PR #2270, opened 2025-03-16
Replace dot by _ in engine name.

vdbergh:
Dots in filenames appear to be bug magnets. See

https://github.com/official-stockfish/fishtest/pull/2265#issuecomment-2727259209


ppigazzini:
Bets practices for a talking filename is to have at max 1 dot for the extension and to use or underscore or hyphen (best for searching the single token, at least on Windows) as token separator.

vdbergh:
Well currently there is a bug which should be fixed. But if you prefer to fix it in a different way then I don't mind! You can just edit this PR.

ppigazzini (approved):
Looks good on DEV and linux/msys2 workers. PROD update in seconds to trigger the workers update. Thank you @vdbergh

PR #2271, opened 2025-03-17
Move base diff link to "base branch not same as master" card

vdbergh:
Here is how it looks now:

Screenshot from 2025-03-17 18-12-29

Before

image


ppigazzini (approved):

PR #2273, opened 2025-03-21
Look for curl and/or wget during toolchain check in worker.

vdbergh:
See https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1742552611.384284&run_id=

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -35,7 +35,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 271
+WORKER_VERSION = 272
 
 
 @exception_view_config(HTTPException)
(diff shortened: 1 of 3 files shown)


ppigazzini (approved):
Looks good on windows/linux workers.
While testing I refactored a bit the function.
PROD update in seconds, thank you @vdbergh

vdbergh:
Actually I am having second thoughts about this PR. Investigating further it seems that normally `curl` or `wget` are not needed since the worker downloads the nets itself using the requests package before starting the compile. But what is then the cause of these failed tasks https://tests.stockfishchess.org/actions?action=&user=&text=wget ?

ppigazzini:
Reverting branch tested https://github.com/ppigazzini/fishtest/tree/drop_curl_wget

vdbergh:
We should probably wait with reverting until the fate of my PR to fix this on the SF side is known. https://github.com/official-stockfish/Stockfish/pull/5938

ppigazzini:
> We should probably wait with reverting until the fate of my PR to fix this on the SF side is known. [official-stockfish/Stockfish#5938](https://github.com/official-stockfish/Stockfish/pull/5938)

I simply pushed a branch on my repo, no PR open.

PR #2274, opened 2025-03-24
Salvage recent network files on worker update

vdbergh:


vdbergh:
Drafted to refine things a little.

ppigazzini:
I would like to have other opinions on this feature. I don't know:
- frequency of new net testing
- success rate for a new net
- the connection speed of the contributors

Plus:
- avoid downloading nets used in active tests

Minus:
- the master nets could be older than the 10 preserved nets
- CloudFlare caches the nets, the download takes a couple of seconds on a 100 Mbps connection
- without any recent net testing, we are saving useless nets

vdbergh:
> 
> Minus:
> 
>     * the master nets could be older than the 10 preserved nets
>
The PR checks access time, not creation time. Most tests involve master so the master net should normally survive since it is accessed all the time (if it doesn't then nothing is really lost compared to the current situation).
 
>     * CloudFlare caches the nets, the download takes a couple of seconds on a 100 Mbps connection

This seems to depend on where you are. At university we have of course a very fast connection but downloading the net can easily take several seconds, if not more. Also people running a private fishtest should not be expected to set up cloudflare caching.
 > 
>     * without any recent net testing, we are saving useless nets

We could delete nets that have not been accessed for a certain period.

vdbergh:
I am going to close this PR in favor of a refactored version.

PR #2275, opened 2025-03-24
Salvage recently accessed network and book files on worker update

vdbergh:
This is a more generic version of the closed PR #2274


vdbergh:
In the latest push I added an expiration time for files
```python
        # Preserve some old files
        backup_pattern = (
            # (pattern, num_bkps, expiration_in_days)
            ("fastchess", 1, 9999),
            ("fastchess.exe", 1, 9999),
            ("nn-*.nnue", 10, 30),
            ("*.epd", 4, 30),
            ("*.pgn", 4, 30),
        )
```

vdbergh:
@ppigazzini 

Are you still having second thoughts about this? This is my testing directory
```
-rwxrwxr-x 1 vdbergh vdbergh   1476176 Mar 25 10:56 fastchess
-rw-rw-r-- 1 vdbergh vdbergh  74874478 Mar 25 23:07 nn-1c0000000000.nnue
-rw-rw-r-- 1 vdbergh vdbergh   3519630 Mar 25 23:07 nn-37f18f62d772.nnue
-rw-rw-r-- 1 vdbergh vdbergh 174889110 Mar 25 22:57 UHO_Lichess_4852_v1.epd
```
`nn-1c0000000000.nnue`  and `UHO_Lichess_4852_v1.epd` together are 250Mb which this PR avoids redownloading after a worker update. Moreover the code is also more generic than the current code which salvages only the fastchess binary (to avoid a lengthy recompile).

ppigazzini:
I will test the PR as soon as I find some spare time, as I need to focus on work during weekdays (and my laptop SSD died this Monday, I'm waiting the replacement from HP). I do like the simplification of the code.

However, the math is off for 130 MB, the book is zipped :)
```
$ ls -lah UHO_Lichess_4852_v1.epd.zip
-rwxrwxrwx 1 usr00 usr00 41M Mar 26 22:22 UHO_Lichess_4852_v1.epd.zip
```
https://github.com/official-stockfish/fishtest/blob/b97a93292555728da30c25342bb459b41ee2ef65/worker/games.py#L1439-L1443

The nets are stored gzipped (with only a limited gain in space, though). Browser and curl/wget with proper flags decompress locally the gzip file, requests is able to download the gzipped file if setting the header. The Cloudflare cache should be double checked, though.
```
$ ls -lah nn-1c0000000000.nnue.gz
-rw-r--r-- 1 fishtest fishtest 59M Nov 10 17:49 nn-1c0000000000.nnue.gz

ls -lah nn-37f18f62d772.nnue.gz
-rw-r--r-- 1 fishtest fishtest 2.8M May 27  2024 nn-37f18f62d772.nnue.gz
```

vdbergh:
Ok I see. I forgot about compression :( Still saving a 100Mb (41+59) download is not nothing. During testing - or active development - many updates may happen in a row....

I also think that by building on this PR I found an elegant _and_ foolproof way to preserve most engine binaries across worker updates, saving more on compile time....

ppigazzini:
Error with python 3.6.x (used by noob's workers)
```
Updating worker version to 272
Moving logfile /home/usr00/_git/__test_folder00/fishtest/worker/api.log to /home/usr00/_git/__test_folder00/fishtest/worker/api.log.previous
The file /home/usr00/_git/__test_folder00/fishtest/worker/tmpdpg0chk0/fishtest-master/worker/sri.txt matches the worker files!
Failed to preserve/delete the file /home/usr00/_git/__test_folder00/fishtest/worker/_testing_1743072548.257949/fastchess:
'PosixPath' object has no attribute 'rstrip'
Failed to preserve/delete the file /home/usr00/_git/__test_folder00/fishtest/worker/_testing_1743072548.257949/nn-37f18f62d772.nnue:
'PosixPath' object has no attribute 'rstrip'
Failed to preserve/delete the file /home/usr00/_git/__test_folder00/fishtest/worker/_testing_1743072548.257949/nn-1c0000000000.nnue:
'PosixPath' object has no attribute 'rstrip'
Failed to preserve/delete the file /home/usr00/_git/__test_folder00/fishtest/worker/_testing_1743072548.257949/UHO_Lichess_4852_v1.epd:
'PosixPath' object has no attribute 'rstrip'
```

vdbergh:
Hmm `rstrip` isn't used anywhere. This seems like a Python bug.  Perhaps because of the dots in the temporary filenames?

I can't really check now (at work).

ppigazzini:
> Unfortunately it seems you cannot install python 3.6 using uv...

I installed 3.6.15 with `pyenv` (3.6.8 does not compile anymore on `pyenv`)

vdbergh:
Thanks I fixed it (and added a comment).

vdbergh:
If you have time can you check that the function below in `games.py` does not print an error message on Python 3.6 (it uses pathlib)? If it fails then it will mess up the logic of the PR a little bit.

```python
def update_atime(path):
    # atime is used by the updater to check which files should be preserved
    # but on a modern Linux system it is updated very lazily. Therefore we
    # update it manually when required.
    atime = time.time()
    try:
        mtime = os.stat(path).st_mtime
        os.utime(path, times=(atime, mtime))
    except OSError as e:
        print(
            f"Failed to update the atime of {path}:\n",
            e,
            sep="",
            file=sys.stderr,
        )
```

ppigazzini (approved):
Looks good in local test, workers update in minutes, finger crossed...
Thank you @vdbergh

PR #2276, opened 2025-03-28
Salvage engine binaries compiled with ARCH=native

vdbergh:
(PR on top of #2275)
    
Salvaging such binaries is completely safe. To mark an engine as natively compiled (all recent stockfishes) we append `-native` to the name of the binary.
    
To achieve this we refactor `setup_engine` and simplify its use.
    
Also: we delete the left over files with names `results-xx...xx.pgn` on worker update.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -765,19 +765,43 @@ def find_arch(compiler):
     return arch
 
 
+def create_environment():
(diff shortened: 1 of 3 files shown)


vdbergh:
There is something wrong with this PR. For some reason I cannot even convert it to draft.

PR #2277, opened 2025-03-28
Salvage engine binaries compiled with ARCH=native

vdbergh:
To achieve this we refactor "setup_engine" and simplify its use.
    
Also: we delete the left over files with names results-xx...xx.pgn on worker update.

This PR is a natural continuation of #2275


ppigazzini:
Stockfish defaults to `ARCH=native` since September 2023, so fishtest uses only native SF binaries. If there is not a really important reason, I will prefer to avoid the `-native` suffix, considering the 260 chars length limit for the path on Windows.

vdbergh:
Well we can replace it by just "-n" for example. 

Some indication is necessary to be able to do tests with old stockfishes which do not support `ARCH=native` (in that case the binary is not preserved across worker update).

BTW: If the length of the filename is a concern we can use only part of the sha.

vdbergh:
We could also invert the logic and use a suffix "-old" for non-native binaries. Let me do that.

ppigazzini:
I'm missing the reason why only ARCH=native is safe.
Beware that I wrote that bash code.

vdbergh:
The worker algorithm for selecting the ARCH may change during a worker update. In principle this may affect the ARCH chosen for an old SF which does not understand ARCH=native.

vdbergh:
In any case I now dropped `-native` and flagged old binaries with `-old`. It is very little code to make that distinction, but of course we could decide to recycle all binaries, since the risk is very small.

vdbergh:
The backup pattern is now
```python
        # Preserve/delete some old files
        backup_pattern = (
            # (pattern, num_bkps, expiration_in_days)
            ("fastchess" + EXE_SUFFIX, 1, math.inf),
            ("stockfish-*-old" + EXE_SUFFIX, 0, -1),
            ("stockfish-*" + EXE_SUFFIX, 50, 30),
            ("nn-*.nnue", 10, 30),
            ("results-*.pgn", 0, -1),
            ("*.epd", 4, 30),
            ("*.pgn", 4, 30),
        )
```

ppigazzini:
> The worker algorithm for selecting the ARCH may change during a worker update. 

The problem is worse than that.
In the past fishtest selected the building ARCH and we dropped all the old SF binaries to be sure to adapt after a change of the ARCH selection.
Since 2023 the `ARCH=native` algorithm can change in Stockfish in any moment. My Zen4 could have master built with `vnni256` and test built with `vnni512` at example.

vdbergh:
> > The worker algorithm for selecting the ARCH may change during a worker update.
> 
> The problem is worse than that. In the past fishtest selected the building ARCH and we dropped all the old SF binaries to be sure to adapt after a change of the ARCH selection. Since 2023 the `ARCH=native` algorithm can change in Stockfish in any moment. My Zen4 could have master built with `vnni256` and test built with `vnni512` at example.

IUC the difference is that when ARCH=native it is the stockfish Makefile that selects the actual ARCH and not the worker. So the actual ARCH is determined by the sha, which is part of the name of the binary.

ppigazzini:
> IIUC the difference is that when ARCH=native it is the stockfish Makefile that selects the actual ARCH and not the worker. So the actual ARCH is determined by the sha, which is part of the name of the binary.

We should name the binary after the build, not before.
```
$ ./stockfish-6ceaca4c7b2cc1fa87617b1b9e83d38d8e880924-g++_14_2_0-b1537a0cca.exe compiler
Stockfish dev-20250327-328dc0c6 by the Stockfish developers (see AUTHORS file)

Compiled by                : g++ (GNUC) 14.2.0 on MinGW64
Compilation architecture   : x86-64-sse41-popcnt
Compilation settings       : 64bit SSE41 SSSE3 SSE2 POPCNT
Compiler __VERSION__ macro : 14.2.0
```

vdbergh:
> > IIUC the difference is that when ARCH=native it is the stockfish Makefile that selects the actual ARCH and not the worker. So the actual ARCH is determined by the sha, which is part of the name of the binary.
> 
> We should name the binary after the build, not before.
> 
I think this is unnecessary for native binaries and it does not not help for non-native binaries since the worker algorithm to select the ARCH depends on having the makefile available. 

In other words even if we include the ARCH in the filename of the binary, to see if we can reuse that binary after a worker update we would have to download the source code anyway...

vondele:
I think these changes add quite some logic to optimize for a corner case (which unfortunately happens to be more common for developers). Just rebuilding binaries as needed is a few seconds overhead, but we won't need to worry about it being right, or covering all cases.

vdbergh:
> I think these changes add quite some logic to optimize for a corner case (which unfortunately happens to be more common for developers).

Actually if you look at this PR then you see that the logic of setup_engines had become simpler. Also the update logic has become simpler, since it is now table driven.

>  Just rebuilding binaries as needed is a few seconds overhead, 

Building the binaries perhaps takes a couple of seconds on a 70 core worker. But we should not forget lower core workers. I believe the entry level for Fishtest testing should be very low. For developers it is a plus that constant recompiling is avoided.
  
> but we won't need to worry about it being right, or covering all cases.

 This PR does not attempt to cover all cases. It only handles ARCH=native, which is the default these days. In all other cases it falls back to the current behavior.

vdbergh:
> I think these changes add quite some logic to optimize for a corner case (which unfortunately happens to be more common for developers). Just rebuilding binaries as needed is a few seconds overhead, but we won't need to worry about it being right, or covering all cases.

I think it is incorrect to say that this PR optimizes a corner case. It optimizes the safe case ARCH=native which has been the default for almost two years (and the logic is actually simpler than before). 

Compile time may be negligible on a 70 core worker, but it is not on a low core worker.

ppigazzini:
> IIUC the difference is that when ARCH=native it is the stockfish Makefile that selects the actual ARCH and not the worker. So the actual ARCH is determined by the sha, which is part of the name of the binary.

In case of a change in the `get_native_properties.sh`, the two binaries may be built with different ARCHs on certain CPUs.
The ARCH selection now is embedded in the SF project. We can choose to accept this side effect, but it’s important to be aware that it occurs.

vdbergh:
> In case of a change in the get_native_properties.sh, the two binaries may be built with different ARCHs on certain CPUs ina regression test.
> The ARCH selection now is embedded in the SF project. We can choose to accept this side effect, but it’s important to be aware that it occurs.

Yes one should be aware of this, but this is how it is now anyway. So this comment is orthogonal to this PR.

This PR is about the observation that it is wasteful for the updater to delete a binary when the worker likely has to build that exact same binary again a bit later.

robertnurnberg:
So running `make clean && time make -j4 profile-build` on my laptop gives
```
95.601u 7.216s 0:49.55 207.4%   0+0k 42760+754264io 335pf+0w
```
so about 50s wall time. I would think most workers on fishtest run with better hardware (and more cores).

vdbergh:
But it is not about a single binary. The worker keeps 50 binaries. A fair number of  these need to be recompiled after a worker update. 

Note: I run a 1 core worker...

ppigazzini (approved):
Looks good in local test with msys2 and linux (python 3.6) workers. Workers update in minutes. Thank you @vdbergh

PR #2280, opened 2025-03-31
Further tweak warnings/notes in tests_view.mak

vdbergh:
Here is how it looks (not all warnings):

Screenshot from 2025-03-31 18-44-02


ppigazzini (approved):

PR #2282, opened 2025-04-13
Update min GCC to 9.3 and Clang to 10.0

Disservin:
I propose we update the GCC and Clang minimum required version to 9.3 and 10.0 to slowly force users with very old systems to update their compilers.

gcc 9.3 has been chosen as the minimum version for gcc since it is the at the time of writing the oldest used version (by noob).

clang 10.0 has been chosen since it almost matches the release date of gcc 9.3 (gcc March 12, 2020 vs clang 24 Mar 2020).

At the time of writing there are no active clang contributors.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -35,7 +35,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 274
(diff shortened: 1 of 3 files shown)


Disservin:
cc @vondele

vondele:
sounds good to me.

ppigazzini (approved):
Looks good in local test, PROD update in seconds, thank you @Disservin

PR #2287, opened 2025-04-23
update book used for PTs

robertnurnberg:
The previous version `UHO_4060_v3.epd` contained some duplicates, which have now been removed. See https://github.com/official-stockfish/books/pull/50.

@dav1312 FYI


ppigazzini (approved):
PT looks good on DEV (and using the new UHO_4060_v4.epd), PROD update in seconds, thank you @robertnurnberg

PR #2288, opened 2025-04-25
Refactor removing old files.

vdbergh:
Removing old files is currently done at three places. We refactor this via calling a single function.


robertnurnberg:
I think your latest force-push overwrote the change from 30 to 365 days for local books.

robertnurnberg:
Also, is there any chance to change the name of `num_bkps` to something more intuitive? ;) I believe it stands for `num_backups`, right?

vdbergh:
> I think your latest force-push overwrote the change from 30 to 365 days for local books.

I did not notice this (I was wondering why there was a merge conflict in `updater.py`).

Currently it is 30 days after the last use. So actively used books will never be deleted. If we make it 365 then we may as well make it `math.inf`. What do you think?

vdbergh:
> Also, is there any chance to change the name of `num_bkps` to something more intuitive? ;) I believe it stands for `num_backups`, right?

I am not sure who uses `num_bkps`? @ppigazzini ?

robertnurnberg:
> > I think your latest force-push overwrote the change from 30 to 365 days for local books.
> 
> I did not notice this (I was wondering why there was a merge conflict in `updater.py`).
> 
> Currently it is 30 days after the last use. So actively used books will never be deleted. If we make it 365 then we may as well make it `math.inf`. What do you think?

I was thinking about some fleet machines which might come back after several months. I think a year is fine. Some books might be retired on the server for example.

ppigazzini:
> I am not sure who uses `num_bkps`? @ppigazzini ?

Introduced 5 years ago by https://github.com/official-stockfish/fishtest/commit/06126a6bd75cc989f778bce6247c7500d743d971 to keep only the 3 most recent backup copies of the `testing` folder.

ppigazzini:
> Also, is there any chance to change the name of `num_bkps` to something more intuitive? ;) I believe it stands for `num_backups`, right?

```yaml
$ git grep -n num_bkps
games.py:1422:    # Clean up old engines (keeping the num_bkps most recent).
games.py:1425:    num_bkps = 50
games.py:1440:        for old_engine in engines[num_bkps:]:
games.py:1517:    # Clean up the old networks (keeping the num_bkps most recent)
games.py:1518:    num_bkps = 10
games.py:1521:    )[num_bkps:]:
updater.py:103:            # (pattern, num_bkps, expiration_in_days)
updater.py:112:        for pattern, num_bkps, expiration_days in backup_pattern:
updater.py:122:                    if idx >= num_bkps:
updater.py:137:        # Clean up old folder backups (keeping the num_bkps most recent).
updater.py:138:        num_bkps = 3
updater.py:141:        )[num_bkps:]:
```

vdbergh:
> > > I think your latest force-push overwrote the change from 30 to 365 days for local books.
> > 
> > 
> > I did not notice this (I was wondering why there was a merge conflict in `updater.py`).
> > Currently it is 30 days after the last use. So actively used books will never be deleted. If we make it 365 then we may as well make it `math.inf`. What do you think?
> 
> I was thinking about some fleet machines which might come back after several months. I think a year is fine. Some books might be retired on the server for example.

I think machines in a fleet are usually started from an image, so they would redownload the books anyway @vondele ? Anyway I have no strong feelings about this.

ppigazzini (approved):
Looks good in some tests with windows/linux workers.
Workers upgrade in minutes, thank you @vdbergh

PR #2290, opened 2025-04-28
RFC: Replace print by the Python logging framework in worker.

vdbergh:
Some advantages:

(1) The Python logging framework is is thread safe so we do not have to worry about printing from different threads.

(2) The Python logging framework is very configurable which may be useful in the future.

This implementation is very rudimentary since I want to make sure that other people also agree that this is a good thing to do.

I have drafted this since it is a RFC.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -3,6 +3,7 @@ import ctypes
 import hashlib
 import io
 import json
+import logging
 import math
 import multiprocessing
 import os
@@ -23,6 +24,8 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from queue import Empty, Queue
 from zipfile import ZipFile
+
+logger = logging.getLogger(__name__)
 
 try:
     import requests
@@ -91,16 +94,11 @@ def backup_log():
         logfile = Path(__file__).resolve().parent / LOGFILE
         logfile_previous = logfile.with_suffix(logfile.suffix + ".previous")
         if logfile.exists():
-            print("Moving logfile {} to {}".format(logfile, logfile_previous))
+            logger.info("Moving logfile {} to {}".format(logfile, logfile_previous))
             with LOG_LOCK:
                 logfile.replace(logfile_previous)
     except Exception as e:
-        print(
-            "Exception moving log:\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error("Exception moving log: " + str(e))
 
 
 def str_signal(signal_):
@@ -148,12 +146,7 @@ def update_atime(path):
         mtime = os.stat(path).st_mtime
         os.utime(path, times=(atime, mtime))
     except OSError as e:
-        print(
-            f"Failed to update the atime of {path}:\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error(f"Failed to update the atime of {path}: " + str(e))
 
 
 def cache_read(cache, name):
@@ -213,12 +206,7 @@ def requests_get(remote, *args, **kw):
         result = requests.get(remote, *args, **kw)
         result.raise_for_status()  # also catch return codes >= 400
     except Exception as e:
-        print(
-            "Exception in requests.get():\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error("Exception in requests.get(): " + str(e))
         raise WorkerException("Get request to {} failed".format(remote), e=e)
 
     return result
@@ -229,12 +217,7 @@ def requests_post(remote, *args, **kw):
     try:
         result = requests.post(remote, *args, **kw)
     except Exception as e:
-        print(
-            "Exception in requests.post():\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error("Exception in requests.post(): " + str(e))
         raise WorkerException("Post request to {} failed".format(remote), e=e)
 
     return result
@@ -261,15 +244,10 @@ def send_api_post_request(api_url, payload, quiet=False):
                 api_url
             )
         )
-        print(
-            "Exception in send_api_post_request():\n",
-            message,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error("Exception in send_api_post_request(): " + message)
         raise WorkerException(message)
     if "error" in response:
-        print("Error from remote: {}".format(response["error"]))
+        logger.error("Error from remote: {}".format(response["error"]))
 
     t1 = datetime.now(timezone.utc)
     w = 1000 * (t1 - t0).total_seconds()
@@ -283,8 +261,8 @@ def send_api_post_request(api_url, payload, quiet=False):
     )
     if not quiet:
         if "info" in response:
-            print("Info from remote: {}".format(response["info"]))
-        print(
+            logger.info("Info from remote: {}".format(response["info"]))
+        logger.info(
             "Post request {} handled in {:.2f}ms (server: {:.2f}ms)".format(
                 api_url, w, s
             )
@@ -301,7 +279,7 @@ def github_api(repo):
 def required_nets(engine):
     nets = {}
     pattern = re.compile(r"(EvalFile\w*)\s+.*\s+(nn-[a-f0-9]{12}.nnue)")
-    print("Obtaining EvalFile of {}...".format(engine.name))
+    logger.info("Obtaining EvalFile of {}...".format(engine.name))
     try:
         with subprocess.Popen(
             [engine, "uci"],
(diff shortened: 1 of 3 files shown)


robertnurnberg:
I may be useful to see a comparison between old and new output? We could also take the chance to consistently change to usage of f-strings?

By the way, completely unrelated: since https://github.com/official-stockfish/books/pull/52 we have SRI checksums for all the books. So we can check the integrity of local books when deciding whether to download books from the repo. I think you know that part of the worker code best, in case you want to give it a try.

vdbergh:
> I may be useful to see a comparison between old and new output? We could also take the chance to consistently change to usage of f-strings?
> 


It does not look particularly nice now but it can be further refined.

I think for the console logger we should suppress the INFO:worker and INFO:games prompts. I would keep the WARNING: 
and ERROR: prompts though.

For the file logger (not implemented now) we could add time stamps.
```
vdbergh@pl:~/fishtest/worker$ python worker.py

INFO:worker:Worker started in /home/vdbergh/fishtest/worker with PID=676571
INFO:worker:Found g++ version 13.3.0
INFO:worker:No clang++ or clang++ is not executable
INFO:worker:machine_id 38370896c30440658170914419b6756c obtained from /etc/machine-id
INFO:worker:Default uuid_prefix: 3b14e644
INFO:worker:Worker version 276 connecting to https://tests.stockfishchess.org:443
INFO:worker:Confirming cached credentials with https://tests.stockfishchess.org:443
INFO:worker:Credentials ok!
INFO:worker:System memory determined to be: 15.308GiB
INFO:worker:Worker constraints: {'concurrency': 1, 'max_memory': 7838, 'min_threads': 1}
INFO:worker:Config file /home/vdbergh/fishtest/worker/fishtest.cfg written
INFO:worker:Writing sri hashes to /home/vdbergh/fishtest/worker/sri.txt
INFO:worker:Verify worker version...
INFO:games:Post request https://tests.stockfishchess.org:443/api/request_version handled in 85.44ms (server: 0.26ms)
INFO:worker:Using g++ 13.3.0
INFO:worker:Toolchain check...
INFO:worker:Obtaining version info for /home/vdbergh/fishtest/worker/testing/fastchess...
INFO:worker:Found fastchess alpha 1.2.0 01010101-f8e58066
INFO:games:Downloading https://raw.githubusercontent.com/official-stockfish/fishtest/master/worker/sri.txt
WARNING:worker:worker.py has been modified!
WARNING:worker:games.py has been modified!
WARNING:worker:This worker is tainted...
INFO:worker:UUID: 3b14e644-bd74-46bb-b05b-86a5f25d8309
INFO:worker:Start heartbeat
INFO:worker:Current time is 2025-05-04 07:38:14.259749+00:00 UTC (local offset: +02:00) 
INFO:worker:Verify worker version...
INFO:games:Post request https://tests.stockfishchess.org:443/api/request_version handled in 93.03ms (server: 0.31ms)
INFO:worker:Remaining number of GitHub api calls = 5000
INFO:worker:Fetching task...
INFO:games:Post request https://tests.stockfishchess.org:443/api/request_task handled in 76.58ms (server: 4.04ms)
INFO:worker:Working on task 723 from https://tests.stockfishchess.org:443/tests/view/68109e2c3629b02d74b166ee
INFO:worker:Running cntst6 vs master
INFO:games:Variable task sizes used. Opening offset = 61542
INFO:games:Downloading https://api.github.com/repos/FauziAkram/Stockfish/zipball/afc3c6080a403f9bb2db469d2149b193c5ab19f7
INFO:games:Build uses default net: nn-1c0000000000.nnue
INFO:games:Build uses default net: nn-37f18f62d772.nnue
INFO:games:Using native target architecture
Existing nn-1c0000000000.nnue validated, skipping download
Existing nn-37f18f62d772.nnue validated, skipping download

Config:
debug: 'no'
sanitize: 'none'
optimize: 'yes'
arch: 'x86_64'
bits: '64'
```

vdbergh:
> 
> By the way, completely unrelated: since [official-stockfish/books#52](https://github.com/official-stockfish/books/pull/52) we have SRI checksums for all the books. So we can check the integrity of local books when deciding whether to download books from the repo. I think you know that part of the worker code best, in case you want to give it a try.

I guess the idea is that the books in the book repo can be updated without changing their name. Correct?

vdbergh:
> > By the way, completely unrelated: since [official-stockfish/books#52](https://github.com/official-stockfish/books/pull/52) we have SRI checksums for all the books. So we can check the integrity of local books when deciding whether to download books from the repo. I think you know that part of the worker code best, in case you want to give it a try.
> 
> I guess the idea is that the books in the book repo can be updated without changing their name. Correct?

BTW is this metadata kept up to date automatically?

robertnurnberg:
Yes, the metadata will be kept up to date from now on. (In the same way that the Sri data for the fishtest worker is kept up to date: manually within each PR).

I think we'll still use new versions for new books. But yes, for things like adding missing move numbers (without changing the exits and their order), we can now use the Sri to trigger a book update on the worker.

The main reason for the new meta data was to ensure the book data does not corrupted locally over time. We take great care to avoid duplicate book exits etc, so it would be a shame if a small edit in a local book copy invalidates that.

vdbergh:
> Yes, the metadata will be kept up to date from now on. (In the same way that the Sri data for the fishtest worker is kept up to date: manually within each PR).

For the worker it is checked by CI. I was wondering if at least something like this was in place for the books.

> 
> I think we'll still use new versions for new books. But yes, for things like adding missing move numbers (without changing the exits and their order), we can now use the Sri to trigger a book update on the worker.
> 
> The main reason for the new meta data was to ensure the book data does not corrupted locally over time. We take great care to avoid duplicate book exits etc, so it would be a shame if a small edit in a local book copy invalidates that.

Ok I understand. Of course people are not supposed to edit files in the testing folder, but who knows what some people might do...

robertnurnberg:
It's not checked within the ci yet. But I can look into setting that up.

vdbergh:
> It's not checked within the ci yet. But I can look into setting that up.

I think it would be a good idea. In my experience people do not keep metadata/documentation up to date unless it is enforced in some way.

robertnurnberg:
> > It's not checked within the ci yet. But I can look into setting that up.
> 
> I think it would be a good idea. In my experience people do not keep metadata/documentation up to date unless it is enforced in some way.

That has now been implemented.

vdbergh:
I am going to close this. It is currently a bit of a distraction.

worker/games.py
```diff
@@ -148,12 +146,7 @@ def update_atime(path):
         mtime = os.stat(path).st_mtime
         os.utime(path, times=(atime, mtime))
     except OSError as e:
-        print(
-            f"Failed to update the atime of {path}:\n",
-            e,
-            sep="",
-            file=sys.stderr,
-        )
+        logger.error(f"Failed to update the atime of {path}: " + str(e))
```
ppigazzini, line 149: I wrote few fishtest utils scripts using `ruff check --select ALL` to iron out all the warnings (eg https://github.com/official-stockfish/fishtest/blob/master/server/utils/delta_update_users.py). I learned than f-string in `logging` is a bad practice, because it is evaluated eagerly, even if that log level is disabled. This is CPU intensive if there are many DEBUG loggings, and it can create weird errors whit code enabled only in DEBUG mode.
`logging` provides %-formatting that defers string interpolation until after the level log check.

```python
import logging

# A class that raises an error when its string representation is needed.
class Broken:
    def __str__(self):
        raise ValueError("Broken __str__ conversion!")

# Create an instance of the problematic class.
broken_instance = Broken()

# Configure the logger: set level to INFO so DEBUG messages are not output.
logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

# Using an f-string: the expression inside {} is evaluated immediately,
# even though DEBUG messages are disabled.
try:
    logger.debug(f"F-string logging: {broken_instance}")
except Exception as e:
    print("F-string raised an exception:", e)

# Using %-formatting: the logger.debug method checks the log level first.
# Since DEBUG is disabled, the string interpolation never happens and
# broken_instance.__str__() is not called.
try:
    logger.debug("%%-formatting logging: %s", broken_instance)
    print("%%-formatting did NOT raise an exception.")
except Exception as e:
    print("%%-formatting raised an exception:", e)
```

vdbergh: Ok that makes sense. I was wondering why the logging frame work has its own string interpolation.

Nowadays it would be natural to use t-strings (lazily evaluated f-strings).

ppigazzini: https://peps.python.org/pep-0750/
template string will be shipped in python 3.14, so it could be used only server side (where not requiring `pyramid` explicit support).
PROD is running with python 3.13, but 3.13 is not officially tested by `pyramid` because it breaks on windows.
https://github.com/Pylons/pyramid/pull/3767

PR #2291, opened 2025-05-03
Remove redundant flag `base_same_as_master`

vdbergh:
This flag is functionally equivalent to

`official_master_sha == resolved_base`


vdbergh:
I am drafting this since apparently `base_same_as_master` means that base is an _ancestor_ or master. It does not have to be master itself.

PR #2293, opened 2025-05-04
Make base is not master warning more precise.

vdbergh:
It really means that base is not an ancestor of master (in other words the value is determined by a three-dot comparison).

Also: fix diff url to compare with the official master and not with the master of the test repo.


vdbergh:
So is it correct that if base is not an ancestor of master then an eventual pull request will not correspond to what has been tested?

If so, that seems pretty dangerous. Are there safeguards in place to prevent this?

vdbergh:
Awaiting comments, I already made the warning stronger

Screenshot from 2025-05-04 08-46-04

vdbergh:
It occurred to me that if the goal is to indicate that a test can be used as the basis for a pull request then it would be more accurate to define `base_same_as_master` as:
```
base is the most recent common ancestor of test and master
```
I checked that this information can be trivially obtained from the compare api.

server/fishtest/schemas.py
```diff
@@ -635,7 +635,7 @@ def flags_must_match(run):
 # about non-validation of runs created with the prior
 # schema.
 
-RUN_VERSION = 13
+RUN_VERSION = 14
```
ppigazzini: I suppose that his must set to 16 after merging #2295

vdbergh: Yep. I will rebase.

vdbergh: Rebased and increased version.

ppigazzini (approved):
Looks good on DEV, PROD updated in minutes, thank you @vdbergh 

image

vondele:
This seems to have the side effect that one can not easily make a sequence on patches, like
commit -am "Take 1"
push and test
commit -am "Take 2"
push and test
which leads to "merge base commit"

Vizvezdenec:
yes, happened to me and nonlinear, pretty inconvenient I must say

vdbergh:
The warning is cosmetic. It should not cause any errors. If it does then that’s a bug.

vondele:
test is not submitted AFAICT.

Vizvezdenec:
> The warning is cosmetic. It should not cause any errors. If it does then that’s a bug.

test isn't submitted to fishtest if warning is there, so it's definitely not cosmetic. And yes, I've seen this happening, it's not a guess.

vdbergh:
It seems to me that for some reason the GitHub api call does not return the merge base commit field on the second test. I have no idea why that could be.

Are the steps you gave a method to reliably reproduce this bug?

vondele:
worked for me first try... so I think so.

vdbergh:
@ppigazzini Have you ever monitored the GitHut rate limit of the server? Is it conceivable that it sometimes runs out of API calls?

vdbergh:
> worked for me first try... so I think so.

Does the error persist if you resubmit the test?

Vizvezdenec:
Even if it doesn't this is probably not the good practice to have random blocking message here and there - I would argue that having this stuff popping up randomly is worse.

vondele:
I tried to submit it twice, failed the first time, passed the second time.

vondele:
so rather random indeed.

vdbergh:
> Even if it doesn't this is probably not the good practice to have random blocking message here and there - I would argue that having this stuff popping up randomly is worse.

The server does many GitHub API calls and each of them will block the submission if they fail.

ppigazzini:
The GitHub Token on PROD should be fine, no end date.

image

Vizvezdenec:
> > Even if it doesn't this is probably not the good practice to have random blocking message here and there - I would argue that having this stuff popping up randomly is worse.
> 
> The server does many GitHub API calls and each of them will block the submission if they fail.

Have never seen this happening with me - literally a single time.  API calls block doesn't show you _patch difference_ - but never prohibits you from submitting a patch.
But what mentioned there blocked me from submitting a patch at least 2 times in span of last 5 days, also happened to other devs.

ppigazzini:
Used 77/5000 core API call right now on PROD.

```yaml
$ curl -n https://api.github.com/rate_limit
{
  "resources": {
    "core": {
      "limit": 5000,
      "used": 77,
      "remaining": 4923,
      "reset": 1749501539
    },
    "search": {
      "limit": 30,
      "used": 0,
      "remaining": 30,
      "reset": 1749501205
    },
    "graphql": {
      "limit": 5000,
      "used": 0,
      "remaining": 5000,
      "reset": 1749504745
    },
    "integration_manifest": {
      "limit": 5000,
      "used": 0,
      "remaining": 5000,
      "reset": 1749504745
    },
    "source_import": {
      "limit": 100,
      "used": 0,
      "remaining": 100,
      "reset": 1749501205
    },
    "code_scanning_upload": {
      "limit": 5000,
      "used": 77,
      "remaining": 4923,
      "reset": 1749501539
    },
    "code_scanning_autofix": {
      "limit": 10,
      "used": 0,
      "remaining": 10,
      "reset": 1749501205
    },
    "actions_runner_registration": {
      "limit": 10000,
      "used": 0,
      "remaining": 10000,
      "reset": 1749504745
    },
    "scim": {
      "limit": 15000,
      "used": 0,
      "remaining": 15000,
      "reset": 1749504745
    },
    "dependency_snapshots": {
      "limit": 100,
      "used": 0,
      "remaining": 100,
      "reset": 1749501205
    },
    "audit_log": {
      "limit": 1750,
      "used": 0,
      "remaining": 1750,
      "reset": 1749504745
    },
    "audit_log_streaming": {
      "limit": 15,
      "used": 0,
      "remaining": 15,
      "reset": 1749504745
    },
    "code_search": {
      "limit": 10,
      "used": 0,
      "remaining": 10,
      "reset": 1749501205
    }
  },
  "rate": {
    "limit": 5000,
    "used": 77,
    "remaining": 4923,
    "reset": 1749501539
  }
}
```

vdbergh:
What is special about this api call is that it compares commits in different repositories. However I neglect to specify the dev repository in the call. Perhaps this intermittently fails.

> > > Even if it doesn't this is probably not the good practice to have random blocking message here and there - I would argue that having this stuff popping up randomly is worse.
> > 
> > 
> > The server does many GitHub API calls and each of them will block the submission if they fail.
> 
> Have never seen this happening with me - literally a single time. API calls block doesn't show you _patch difference_ - but never prohibits you from submitting a patch. But what mentioned there blocked me from submitting a patch at least 2 times in span of last 5 days, also happened to other devs.

Well GitHub API calls do not normally fail. That’s why you don’t observe it. This one should not fail either. I am trying to understand what is happening.

robertnurnberg:
I wonder if we really need such a warning? The approvers will anyway see the git diff with the official master branch, right?

vdbergh:
I don't think so. AFAICS the diff shown is test vs base. Which is correct in IMHO since one may also do tests for research purposes.

A similar warning has always been shown (so there was clearly a justification for it) but I  made more precise what it means (I am trying to sanitize the various adhoc warnings).  

There is currently a bug in the implementation. I am comparing branches in different repositories, without specifying the user names. It seems this usually works (that's why it was not caught during testing) but apparently not always (probably it has something to do with caching).

PR #2295, opened 2025-05-31
Check that a run is not finished before reviving after modifying.

vdbergh:
We also enforce the corresponding condition in the run schema.

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -617,6 +617,21 @@ def flags_must_match(run):
     return True
 
 
+def is_undecided(run):
+    completed_games = (
+        run["results"]["wins"] + run["results"]["losses"] + run["results"]["draws"]
+    )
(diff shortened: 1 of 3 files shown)


server/fishtest/schemas.py
```diff
@@ -635,7 +650,7 @@ def flags_must_match(run):
 # about non-validation of runs created with the prior
 # schema.
 
-RUN_VERSION = 13
+RUN_VERSION = 14
```
ppigazzini: #2297 has already set RUN_VERSION to 14, I wonder if this PR must set 15 or a fast merge is sufficient.

vdbergh: No it should be set to 15. I will do it now.

vdbergh: Fixed and rebased now.

ppigazzini (approved):
Looks good on DEV.

ppigazzini:

PR #2296, opened 2025-05-31
Check book checksums

robertnurnberg:
For some time now the books repo provides checksums for our books in a json file, and these are now available from the server via the `book_sri` key.

This PR makes the worker check the validity of a book before it is being used.

At the moment the logic is as follows:
* if a book is found, and the checksum fails, we delete the book, download it once and re-test the checksum
* if a book was downloaded and the checksum fails, an error message is printed, but the worker continues as normal
* if we fail to obtain the `book_sri` information from the server, the worker prints an error message and continues as normal

We also raise the "expiration" time of local book files in `updater.py` to 365 days since they were last accessed (up from 30 days).

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -35,7 +35,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 276
+WORKER_VERSION = 277
 
 
 @exception_view_config(HTTPException)
--- a/worker/games.py
+++ b/worker/games.py
@@ -1491,7 +1491,8 @@ def run_games(
 
         print(f"Book sri mismatch: {book_sri} != {sri}.", file=sys.stderr)
         if downloaded_book:
-            raise WorkerException(f"Failed to match sri for book {book}")
+            print(f"Failed to match sri for book {book}. Ignoring!", file=sys.stderr)
+            break
 
         try:
             (testing_dir / book).unlink()
--- a/worker/games.py
+++ b/worker/games.py
@@ -1,3 +1,4 @@
+import base64
 import copy
 import ctypes
 import hashlib
@@ -17,7 +18,6 @@ import sys
 import tempfile
 import threading
 import time
-from base64 import b64decode
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
@@ -34,6 +34,13 @@ IS_MACOS = "darwin" in platform.system().lower()
 LOGFILE = "api.log"
 
 LOG_LOCK = threading.Lock()
+
+
+def text_hash(file):
+    # text mode to have newline translation!
+    return base64.b64encode(
+        hashlib.sha384(file.read_text().encode("utf8")).digest()
+    ).decode("utf8")
 
 
 class WorkerException(Exception):
@@ -550,7 +557,9 @@ def download_from_github_api(
     )
     print("Downloading {}".format(item_url))
     git_url = requests_get(item_url, timeout=HTTP_TIMEOUT).json()["git_url"]
-    return b64decode(requests_get(git_url, timeout=HTTP_TIMEOUT).json()["content"])
+    return base64.b64decode(
+        requests_get(git_url, timeout=HTTP_TIMEOUT).json()["content"]
+    )
 
 
 def download_from_github(
@@ -1369,6 +1378,7 @@ def run_games(
 
     book = run["args"]["book"]
     book_depth = run["args"]["book_depth"]
+    book_sri = run["args"].get("book_sri")
     new_options = run["args"]["new_options"]
     base_options = run["args"]["base_options"]
     threads = int(run["args"]["threads"])
@@ -1454,13 +1464,42 @@ def run_games(
 
     os.chdir(testing_dir)
 
-    # Download the opening book if missing in the directory.
-    if not (testing_dir / book).exists() or (testing_dir / book).stat().st_size == 0:
-        zipball = book + ".zip"
-        blob = download_from_github(zipball)
-        unzip(blob, testing_dir)
-    else:
-        update_atime(testing_dir / book)
+    downloaded_book = False
+    while not downloaded_book:
+        # Download the opening book if missing in the directory.
+        if (
+            not (testing_dir / book).exists()
+            or (testing_dir / book).stat().st_size == 0
+        ):
+            zipball = book + ".zip"
+            blob = download_from_github(zipball)
+            unzip(blob, testing_dir)
+            downloaded_book = True
+        else:
+            update_atime(testing_dir / book)
+            print(f"Re-using local book {testing_dir / book}.")
+
+        # very old tests (stopped/restarted) may lack the book_sri key
+        if book_sri is None:
+            print("Failed to obtain book_sri from server.", file=sys.stderr)
+            break
+
+        sri = text_hash(testing_dir / book)
+        if book_sri == sri:
+            print(f"Book sri for {book} matches.")
+            break
+
+        print(f"Book sri mismatch: {book_sri} != {sri}.", file=sys.stderr)
+        if downloaded_book:
+            raise WorkerException(f"Failed to match sri for book {book}")
+
+        try:
+            (testing_dir / book).unlink()
+            print(f"Deleted local book {testing_dir / book}.")
+        except Exception as e:
+            raise WorkerException(
+                f"Failed to remove local book {testing_dir / book}", e=e
+            )
 
     # Clean up the old networks (keeping the num_bkps most recent)
     num_bkps = 10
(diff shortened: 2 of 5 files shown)


vondele:
One comment from my side would be that this would trigger a github download for every test run, this is problematic for the fleet, I think. From this point of view, it would be nicer if the server would pass this to worker as part of the test payload.

robertnurnberg:
Good point. I have no experience with server-side code, so would need help from other devs, I think.

At the end of the day, we would like a field `book_sri` similar to `book_depth` being passed from the server, correct?

vondele:
I think that's probably the code to follow to pass from the server to the client. Contrary to book_depth, it should probably be initialized from the github repo.

vdbergh:
I can make the accompanying server PR.

vdbergh:
See #2297

robertnurnberg:
Very nice, thanks a lot.

Once your pr is merged and dev runs on it, I can align this pr with it.

robertnurnberg:
This is now ready for review.

If I understand the server code correctly, a book not being in `books.json` will now be a critical failure? In the sense that a test scheduled with such a book cannot be run? This places some responsibility on (future) maintainers of the books repo to not commit changes that break the integrity of `books.json`. (For PRs that means only to merge them if all CI checks pass.)

vdbergh:
> This is now ready for review.
> 
> If I understand the server code correctly, a book not being in `books.json` will now be a critical failure? In the sense that a test scheduled with such a book cannot be run? This places some responsibility on (future) maintainers of the books repo to not commit changes that break the integrity of `books.json`. (For PRs that means only to merge them if all CI checks pass.)

Yes that's correct. I will be a critical failure. I don't see a reason why it would not be... Only merging things that pass CI is standard practice and the overhead imposed on the book author is minimal.

vdbergh:
Perhaps it might be useful to extend the README with a section detailing the steps necessary to make a valid book PR. Another possible place for this information might be the wiki.

vdbergh:
I was testing the PR. For a while at least the worker should deal gracefully with the possibility that book_sri might be missing.
```
$ python worker.py

Worker started in /home/vdbergh/fishtest/worker with PID=2652837
Found g++ version 13.3.0
No clang++ or clang++ is not executable
machine_id 38370896c30440658170914419b6756c obtained from /etc/machine-id
Default uuid_prefix: 3b14e644
Worker version 277 connecting to https://tests.stockfishchess.org:443
Confirming cached credentials with https://tests.stockfishchess.org:443
Credentials ok!
System memory determined to be: 15.308GiB
Worker constraints: {'concurrency': 1, 'max_memory': 7838, 'min_threads': 1}
Config file /home/vdbergh/fishtest/worker/fishtest.cfg written
Writing sri hashes to /home/vdbergh/fishtest/worker/sri.txt
Verify worker version...
Post request https://tests.stockfishchess.org:443/api/request_version handled in 41.48ms (server: 0.28ms)
Using g++ 13.3.0
Toolchain check...
Obtaining version info for /home/vdbergh/fishtest/worker/testing/fastchess...
Found fastchess alpha 1.2.0 01010101-f8e58066
Downloading https://raw.githubusercontent.com/official-stockfish/fishtest/master/worker/sri.txt
The master sri file has a different version number. Ignoring!
UUID: 3b14e644-b3e5-4117-a3d3-5b012d2945ac
Start heartbeat
Current time is 2025-06-04 07:26:33.194109+00:00 UTC (local offset: +02:00) 
Verify worker version...
Post request https://tests.stockfishchess.org:443/api/request_version handled in 40.46ms (server: 0.28ms)
Remaining number of GitHub api calls = 5000
Fetching task...
Post request https://tests.stockfishchess.org:443/api/request_task handled in 46.27ms (server: 3.49ms)
Working on task 312 from https://tests.stockfishchess.org:443/tests/view/683ceb8e6ec7634154f9da21
Running blend-next-value-3 vs master

Exception running games:
Exception KeyError at games.py:1381 WorkerVersion: 277 ('book_sri')
Informing the server
Post request https://tests.stockfishchess.org:443/api/failed_task handled in 55.40ms (server: 12.40ms)
Task exited
Releasing the worker lock
Heartbeat stopped
```

robertnurnberg:
Is this with the updated server code?

vdbergh:
Yes. The PR has been merged into master. But there are still tests that were created before the merger.

robertnurnberg:
Pushed a fix.

vdbergh:
Am I correct that the only consequence of an invalid book is a console error message? I think we should throw an exception to stop the task instead.

robertnurnberg:
Yes, for now this PR takes a soft approach.

Stopping the task is an option. We can do this once we tighten the key error handling as well. (If the maintainers and other devs agree.)

vdbergh:
Perhaps only redownload  the book if `book_sri` is present and does not match the computed sri? For now my worker is redownloading the book with almost every task.  Of course this will slowly diminish, but people can pause/stop tests and revive them later. So in theory it can take a very long time before all tests will have `book_sri`.

robertnurnberg:
> Perhaps only redownload the book if `book_sri` is present and does not match the computed sri? For now my worker is redownloading the book with almost every task. Of course this will slowly diminish, but people can pause/stop tests and revive them later. So in theory it can take a very long time before all tests will have `book_sri`.

This is now implemented. I also raise an exception now if the book sri does not match.

robertnurnberg:
CI fails. Let me know if this can be fixed from my side.

vdbergh:
Seems like a fluke related to starting MongoDB.

vdbergh:
I have been running a worker with this PR and it seems to behave as expected.

vondele:
maybe for now there should just be a warning to the even log, just to see what happens?

One potential failure mode is if a test is started with one version of the book, and somebody updates that book leading to a different sri. Just downloading from github can never fix the sri (I don't know if my example is clear).

robertnurnberg:
Yes, that could happen. It would, of course, only affect new workers, that do not have a local copy of the book yet. So in theory the test could be finished by existing workers? Not sure.

In general, the book repo should no longer replace books with the same name (if it ever did that).

What is not so clear to me is what we gain from waiting and seeing how things go. The problem of a book being changed while keeping its name will never go away.

I will wait for some more feedback and/or discussion for now.

vondele:
just to see how many books are involved, but yeah, we could just message this to the event log and replace.

vdbergh:
We could provide an extra safety net by running a scheduled task on the server that downloads `books.json` every 5 minutes and updates `book_sri` in the unfinished runs. That would only be a few lines of code with the current server architecture.

robertnurnberg:
> We could provide an extra safety net by running a scheduled task on the server that downloads `books.json` every 5 minutes and updates `book_sri` in the unfinished runs. That would only be a few lines of code with the current server architecture.

If that's done, then this pr could stay as it is for @vondele ?

I think doing what you suggest every hour should be enough? And updating would only be needed if the books repo has seen a new commit, of course.

vondele:
would probably be fine, however we don't know how much of an issue there is. In all likelihood, all books in use are just fine. (Like generally, I'm obviously in favor of this patch, just want to avoid that e.g. the fleet gets into a mode of repeatedly downloading a book). I'm basically trying to point out a kind of a race.

robertnurnberg:
Btw, something from protecting the fleet from downloading books in general, would be to increase the expiration time for books from 30 days to 180 days, say? Or even longer/infinity?

With regards to this PR, I'd be happy to implement something that both vondele and vdbergh are happy with.

vondele:
not sure what the expiration time exactly is? Anyway, this is not too critical, downloads from github work fine. Just want to avoid that these get into some kind of loop.

vdbergh:
The main issue is if a book is updated without changing its name. This will lead to repeated downloading of the book for tests that were created before the book update (for workers, that don't have the book, e.g. those that are created from scratch each time).

Of course updating a book without changing its name will presumably not happen. However if it would happen then the solution I propose would limit the situation described in the first paragraph to 5 minutes.

robertnurnberg:
I meant this in the update:
https://github.com/official-stockfish/fishtest/blob/fdefcdd1be3617ffa0451b006e3b6f030cf8678d/worker/updater.py#L109

I think if we want to 100% exclude repeated downloads then it is best to replace raising an exception with a simple error message (and continuing regardless). What do you think? @vdbergh

vdbergh:
This is 30 days after the last use (so frequently used books will never be deleted). Of course we can make it longer.

vondele:
I agree we shouldn't update the book without changing the name. Let's start with the simplest solution for now, just delete and download the book if the sri doesn't match.. let's see where this goes.

vdbergh:
If we agree not to update a book without changing its name then there is no problem.

robertnurnberg:
I squashed all previous commits into one, and then added a new one on top that replaces the exception with a warning message. This follows the suggestion by @vondele and is what the PR started out with.

I believe the failed checksum will almost never happen in practice, so it would be a shame to take any risks regarding e.g. the fleet.

If the checksum fails immediately after a new download, the error most likely does not lie with the worker but with our infrastructure instead. So I think it is best to look out for any such occurrences, rather than make the worker fail.

Is printing a warning message to stderr the standard procedure for "logging" such events?

vdbergh:
> 
> I think if we want to 100% exclude repeated downloads then it is best to replace raising an exception with a simple error message (and continuing regardless). What do you think? @vdbergh

Sorry, I missed this. I have no strong feelings either way. If we agree not to update books without changing their  name, then it doesn't matter much since the possible exception would be almost impossible to trigger.

vdbergh:
It seems our posts crossed.

> Is printing a warning message to stderr the standard procedure for "logging" such events?

There is still no way to log messages from the worker to the server without killing the task.

I will try to make such a PR this evening.

vdbergh:
I don't know what @ppigazzini's  schedule is, but I was thinking that probably it is best to keep this PR as it is for now.

Since the PR creating a logging end point on the server (which I plan to do tonight) will need an accompanying worker PR to make the api easy to use. That code is orthogonal to this PR.

ppigazzini (approved):

ppigazzini:
Looks good on windows worker as well.
@vdbergh let me know if I can merge this PR.

vdbergh:
> Looks good on windows worker as well. @vdbergh let me know if I can merge this PR.

Sure it can be merged!

ppigazzini:
PROD updated to trigger the workers upgrade, thank you @robertnurnberg

PR #2297, opened 2025-06-01
Send book_sri from server to worker.

vdbergh:


robertnurnberg:
I am not sure, but should `server/utils/upgrade.py` also be given the new field?

vdbergh:
That script was used once to sanitize the rundb but it is not kept up to date with the relatively frequent changes in the runs schema. Sanitizing the rundb takes a lot of effort so it is not done frequently.

PR #2299, opened 2025-06-05
Create a "worker_log" end point on the server

vdbergh:


vdbergh:
I have drafted this since in retrospect I think it is better to unify the events “log_message” and “worker_log”.

vdbergh:
Undrafted.

ppigazzini (approved):
Looks good on DEV, PROD updated in minutes, thank you @vdbergh

PR #2304, opened 2025-06-07
Mirror `books.json` on the server. 

vdbergh:
Update it every 15 minutes.
    
Instead of including `book_sri` in the run object (where it is redundant), we now add it dynamically during `request task`.


vdbergh:
Further steps I plan (in future PRs).

- Save `rundb.books` in a collection as a backup for when GitHub is down when the server starts.
- get the list of valid books from `rundb.books`, rather than directly from GitHub. Then things can never be out of sync.

ppigazzini (approved):

PR #2305, opened 2025-06-08
Save rundb.books to a collection.

vdbergh:
This serves as backup if GitHub is down during server start.

Also: Get the list of valid books from rundb.books.


ppigazzini (approved):

PR #2309, opened 2025-06-09
Workaround for failing GitHub api call until true cause found.

vdbergh:
See the discussion in #2293.

I have theory why the call fails, but no time to test it now (need to prepare some conference talks). 

In case the call fails, I am logging the result. Hopefully this will make things clearer.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1014,7 +1014,13 @@ def validate_form(request):
     master_diff = requests.get(
         api_url, headers={"Accept": "application/vnd.github+json"}
     )
-    merge_base_commit = master_diff.json()["merge_base_commit"]["sha"]
+    merge_base_commit = data["resolved_base"]
+    json = None
+    try:
+        json = master_diff.json()
+        merge_base_commit = json["merge_base_commit"]["sha"]
+    except Exception as e:
+        print(f"Api call {api_url} failed: {str(e)}. Result = {json}", flush=True)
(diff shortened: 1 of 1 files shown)


ppigazzini (approved):

vdbergh:
@ppigazzini I wonder if the error message already appears in the server log.

ppigazzini:
> @ppigazzini I wonder if the error message already appears in the server log.

I'm away from the keyboard right now.

vdbergh:
> > @ppigazzini I wonder if the error message already appears in the server log.
> 
> I'm away from the keyboard right now.

Sure. It is not urgent. I am just interested in the possible error message at some point.

ppigazzini:
```yaml
sudo journalctl -u fishtest@6543 --since "24 hours ago" | grep "Api call"
Jun 10 05:05:05 ns304014 pserve[40880]: Api call https://api.github.com/repos/official-stockfish/Stockfish/compare/master...399968c57b failed: 'merge_base_commit'. Result = {'message': 'Not Found', 'documentation_url': 'https://docs.github.com/rest/commits/commits#compare-two-commits', 'status': '404'}
Jun 10 05:05:14 ns304014 pserve[40880]: Api call https://api.github.com/repos/official-stockfish/Stockfish/compare/master...399968c57b failed: 'merge_base_commit'. Result = {'message': 'Not Found', 'documentation_url': 'https://docs.github.com/rest/commits/commits#compare-two-commits', 'status': '404'}
```

PR #2320, opened 2025-06-21
bugfix: deal gracefully with corrupt book files

robertnurnberg:
Fixes https://github.com/official-stockfish/fishtest/issues/2318.


ppigazzini (approved):
Looks good with workers on DEV, workers update in minutes, thank you @robertnurnberg

PR #2321, opened 2025-06-22
Eliminate one more bare github url from views.py.

vdbergh:


ppigazzini:
Testing this PR I caught that "Show Diff" was broken by the last commit.
https://github.com/official-stockfish/fishtest/commit/60c25a93c8798ec88197329e9e367588d9e6b68f#commitcomment-160587923

EDIT_000:
fixed by https://github.com/official-stockfish/fishtest/commit/614a87d03d0e0fa328e6bcb4966750e4e8e18295

ppigazzini (approved):

PR #2322, opened 2025-06-22
fix(tests_view.mak): correct ReferenceError

ppigazzini:
The `handleDiff` function in the tests view template was causing a `ReferenceError: request is not defined` on the client-side.

This was because the `baseOfficialMaster` JavaScript variable was being assigned directly from the server-side `request` object, which is not available in the browser's scope.

This commit resolves the error by using Mako's `${...}` syntax with `json.dumps` to properly serialize the `request.rundb.official_master_sha` value from the server and inject it into the JavaScript code. This ensures the value is correctly handled,
including cases where it might be `None`.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -853,7 +853,7 @@
 
     const diffNew  = "${run["args"]["resolved_new"][:10]}";
     const apiOfficialMaster = "https://api.github.com/repos/official-stockfish/Stockfish";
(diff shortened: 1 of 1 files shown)


ppigazzini:
fixes https://github.com/official-stockfish/fishtest/commit/60c25a93c8798ec88197329e9e367588d9e6b68f#commitcomment-160587923

vdbergh:
Sorry… I missed this during testing.

PR #2324, opened 2025-06-22
Temporarily disable some javascript code.

vdbergh:
When base is called master and master in the dev repo has diverged from the official master then an extra button is shown. See here

https://tests.stockfishchess.org/tests/view/6857414fc830e28d5b3937af

I think with the retirement of run["args"]["offcial_master_sha"] this code is no longer correct. The two dot comparison should be replaced by a three dot comparison with the official master. However I have currently no time to study this code. It is quite complicated.

Note that the information shown by the button is already available from the link shown in the warning on the right.

Not tested.


vdbergh:
One difficulty is that the three dot diff is built into the github api but the two dot diff is not. The two dot diff code was written by Peregrine but its signature is not the same as the three dot code. So you cannot just exchange one for the other.

Personally I don't really see the need for two dot diffs in normal SF development (and they are very expensive in api calls).

vdbergh:
Sometimes I wonder if it would not be better to generate the diffs on the server. That would eliminate a lot of tricky javascript code in the browser.

ppigazzini (approved):

ppigazzini:
The "Show Diff" was broken for some days and nobody reported the issue.

PR #2329, opened 2025-06-25
If base is ancestor of test (the common case) use 3 dot diff.

vdbergh:
A 2 dot diff is more expensive in api calls and gives the same result.


ppigazzini:
Other PRs in the topic:
#924
#959
#1918

ppigazzini (approved):
Looks good on DEV, PROD update after the merge, thank you @vdbergh

vdbergh:
> Other PRs in the topic: #924 #959 #1918

There is a reason why GitHub does not provide a two dot diff in its api. IMHO the need for the two dot diff is mostly the result of bad GitHub workflow.

The only legitimate use case I see is to provide the correct diff for a shootout between two different patches. But even then it might be more instructive (and much cheaper in api calls) to show the two (three dot) diffs versus  the common base.

PR #2330, opened 2025-06-26
Add a rate limit item (server and client) to the sidebar.

vdbergh:
The client rate limit takes the personal access token into account.

Screenshot from 2025-06-26 21-59-02

PR #2331, opened 2025-06-27
Cleanup of #2330

vdbergh:
- Remove an incorrect "Accept:" header (which was ignored by GitHub).

- Replace the "token" prefix by "Bearer" which seems slightly preferred.

- Move DOM code from application.js to rate_limits.mak


ppigazzini (approved):
Looks good on DEV. PROD update after merging, thank you @vdbergh

PR #2332, opened 2025-06-28
Introduce convenience function is_master(sha) and use it.

vdbergh:
The result is cached except when a network error occurs.


ppigazzini (approved):

PR #2337, opened 2025-06-30
Fix bug in rateLimit() function.

vdbergh:
It returns incorrect values for unauthenticated users.


vdbergh:
I managed to fix prettier.

ppigazzini (approved):

vdbergh:
> 
> With this PR one can see that the two dot diff cannot be used by unauthenticated users. The number of api calls drops to zero in one go.

I guess the correct story is that the number of api calls needed to create a 2-dot diff is the number of files changed + 1. So for small patches it is not expensive.

PR #2339, opened 2025-07-02
refactor: update the GitHub PAT link

ppigazzini:


ppigazzini:
`.mak` formatted with `prettier` (extension changed to `.html`)

PR #2340, opened 2025-07-02
More refactorings of github_api calls

vdbergh:
- Always pass exceptions to the caller. There are too many things that can go wrong, to silently suppress such exceptions.

- More idiomatic javascript in rate_limits.mak.

- Home grown LRUCache that better suits our needs. The implementation is a simple wrapper around OrderedDict.

- The cacheing of GitHub api calls is now persistent across server restarts.

- Recent users cannot have DEV repo that is not forked from official-stockfish.

### DIFF
--- a/server/fishtest/github_api.py
+++ b/server/fishtest/github_api.py
@@ -1,21 +1,71 @@
 import base64
-from functools import lru_cache
 from pathlib import Path
 from urllib.parse import urlparse
 
 import requests
+from fishtest.lru_cache import LRUCache
 from fishtest.schemas import sha as sha_schema
 from vtjson import validate
 
+"""
+We treat this module as a singleton.
+"""
+"""
+Note: we generally don't suppress exceptions since too many things can
+go wrong. The caller should gracefully handle whatever comes their
+way.
+"""
+GITHUB_API_VERSION = 1
 TIMEOUT = 3
+INITIAL_RATELIMIT = 5000
+LRU_CACHE_SIZE = 6000
 
+_github_rate_limit = None
+_lru_cache = None
+_kvstore = None
+
+# This one is set externally
 _official_master_sha = None
-_github_rate_limit = -1
 
 
-def call(url, *args, **kwargs):
+def init(kvstore):
+    global _github_rate_limit, _kvstore, _lru_cache
+    _kvstore = kvstore
+    _lru_cache = LRUCache(LRU_CACHE_SIZE)
+    github_api_cache = {"version": GITHUB_API_VERSION, "lru_cache": []}
+    try:
+        if "github_api_cache" in _kvstore:
+            github_api_cache = _kvstore["github_api_cache"]
+        else:
+            print("Initializing github_api_cache", flush=True)
+        if github_api_cache["version"] != GITHUB_API_VERSION:
+            raise Exception("Stored github_api_cache has different version")
+        for k, v in github_api_cache["lru_cache"]:
+            _lru_cache[tuple(k)] = v
+    except Exception as e:
+        print(f"Unable to restore github_api_cache from kvstore: {str(e)}", flush=True)
+    try:
+        _github_rate_limit = rate_limit()["remaining"]
+    except Exception as e:
+        print(
+            f"Unable to initialize github rate limit :{str(e)}. Assuming {INITIAL_RATELIMIT}."
+        )
+
+
+def save():
+    global _kvstore
+    _kvstore["github_api_cache"] = {
+        "version": GITHUB_API_VERSION,
+        "lru_cache": [(k, v) for k, v in _lru_cache.items()],
+    }
+
+
+def call(url, *args, _method="GET", _ignore_rate_limit=False, **kwargs):
     global _github_rate_limit
-    r = requests.get(url, *args, **kwargs)
+    if not _ignore_rate_limit and _github_rate_limit < INITIAL_RATELIMIT / 2:
+        raise Exception(r"Rate limit more than 50% consumed.")
+
+    r = requests.request(_method, url, *args, **kwargs)
     resource = r.headers.get("X-RateLimit-Resource", "")
     if resource == "core":
         _github_rate_limit = int(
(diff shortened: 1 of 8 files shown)


vdbergh:
Undrafted

vdbergh:
Drafted because of bug.

vondele:
btw, since a while I have 
image

I guess there is an issue with a access token or config, but the message (No diff available) is not very helpful to figure that out. Anything more explicit would be great.

vdbergh:
I agree. The javascript code on the client spews out various messages if things go wrong but I have not studied it yet. This PR is about the server code however.

I assume this particular case is a rate limit issue. Did you check here?
image

vdbergh:
Yes I am sure it is rate limits. While the javascript code takes great care to generate appropriate error messages for two dots diffs, it doesn't do so for three dot diffs and these are used in the great majority of cases (i.e. when they give the same result as a two dot diff).

vdbergh:
It seems sufficient to remove the `catch` part in the function below. Then the exception will be caught at a higher level
and this will produce a more appropriate error message.
```javascript
  const fetchDiffThreeDots = async (diffApiUrl) => {
    try {
      const token = localStorage.getItem("github_token");
      const options = token
        ? {
            headers: {
              Authorization: "Bearer " + token,
              Accept: "application/vnd.github.diff",
            },
          }
        : {
            headers: {
              Accept: "application/vnd.github.diff",
            },
          };
      const text = await fetchText(diffApiUrl, options);
      return {text: text, count: text?.split("\n")?.length || 0};
    } catch(e) {
      console.log("Error fetching diff: " + e);
      return {text: "", count: 0};
    }
  };
```

vondele:
> I agree. The javascript code on the client spews out various messages if things go wrong but I have not studied it yet. This PR is about the server code however.
> 
> I assume this particular case is a rate limit issue. Did you check here? image

image

vdbergh:
I assume the rate limit has meanwhile been reset. Do you still get "No diff"? For me the diff shows fine.

image

vondele:
ok, now the diff is showing. So probably was a rate limit. Maybe that error message could include something about the rate limit, and could link to that https://tests.stockfishchess.org/rate_limits page.

vdbergh:
I have dealt with the copilot comments.

vondele:
Probably something else on my side, I have:
image
and also
image

(in the same browser tab)

vdbergh:
That's an honest bug! I'll fix it.

vondele:
now back to 
image

yet, with approx 5000 requests remaining (client side)

dav1312:
Saying "_most probably_ API limit exceeded" doesn't make much sense considering you can check how many requests you have left.

vdbergh:
Hmm so it seems the token is not properly sent to GitHub. My javascript is terrible.

vdbergh:
> Saying "_most probably_ API limit exceeded" doesn't make much sense considering you can check how many requests you have left.

Yes this should be cleaned up, but first the bugs must be fixed.

vdbergh:
@dav1312 could you maybe have a look what is wrong with this function?
```
  const fetchDiffThreeDots = async (diffApiUrl) => {
    const token = localStorage.getItem("github_token");
    const options = {
      headers: {
        Accept: "application/vnd.github.diff",
      },
    };
    if (token) {
      options.headers.Authorization = "Bearer " + token;
    }
    const text = await fetchText(diffApiUrl, options);
    return {text: text, count: text?.split("\n")?.length || 0};
  };
```

ppigazzini:
> yet, with approx 5000 requests remaining (client side)

I have tested on DEV with and without (in private mode) PAT, it worked for me with PAT after using 100 calls.

vdbergh:
Yes I also checked that it works. And both ChatGPT and prettier are happy with the code.

@vondele Does your token allow viewing public repositories?

vondele:
image

vondele:
with the help of @dav1312 ... regenerated a token with 364 days lifetime, all works fine.

vdbergh:
Would it be ok to remove that restriction on the Stockfish side?

dav1312:
Just showing the `message` to the user would already help a lot

vdbergh:
> Just showing the `message` to the user would already help a lot

Sure but that requires some rearchitecting for which I have currently no time. tests_view.mak is quite messy and it should basically be completely refactored.

vondele:
> Would it be ok to remove that restriction on the Stockfish side?

366 days is the maximum for finite lifetime. I think it is OK. In principle it is a good thing to have a finite lifetime, and it goes against the recommendations / policies to have very long lifetimes.

vdbergh:
> > Would it be ok to remove that restriction on the Stockfish side?
> 
> 366 days is the maximum for finite lifetime. I think it is OK. In principle it is a good thing to have a finite lifetime, and it goes against the recommendations / policies to have very long lifetimes.

Well this is for a token that allows viewing public repositories. It is not clear to me why that should be restricted in time. 

But anyway something needs to be done to avoid major user confusion (I never had this problem since I am simply using a classic PAT instead of a fine grained one).

dav1312:
> > Just showing the `message` to the user would already help a lot
> 
> Sure but that requires some rearchitecting for which I have currently no time. tests_view.mak is quite messy and it should basically be completely refactored.

Yes, tests_view.mak is a mess...
A "simple" change could be this
```js
async function raiseForStatus(response) {
  if (response.ok) {
    return;
  }
  let errorMsg = `request failed with status code ${response.status}`;
  try {
    const errorBody = await response.json();
    if (errorBody && errorBody.message) {
      errorMsg += `: ${errorBody.message}`;
    }
  } catch (e) {
    // Could not parse response body as JSON.
  }
  throw errorMsg;
}
```
```js
async function fetchText(url, options) {
  const response = await fetch(url, options);
  await raiseForStatus(response); // await
  return response.text();
}
```

vdbergh:
I guess that would take care of `fetchDiffThreeDots` but there is also `fetchDiffTwoDots` which is much more complicated.

vdbergh:
> image

Strangely I cannot duplicate this. I have a fine grained PAT with no expiration date and it appears to work fine on Fishtest.

PR #2341, opened 2025-07-02
feat: set bench timeout to 600 seconds

ppigazzini:
Raise worker version to 285 (also server side)


ppigazzini:
https://tests.stockfishchess.org/actions?action=stop_run&user=&text=timed+out+after

With the previous timeout, 90 seconds, this worker stopped a good test because a temporary unknown slowdown (eg file system).

PR #2342, opened 2025-07-03
Remove superfluous try/catch block.

vdbergh:
The exception is already caught by the caller.  See https://github.com/official-stockfish/fishtest/pull/2340#issuecomment-3031034201.


ppigazzini:
PROD updated

PR #2343, opened 2025-07-03
Fix missing reference

vdbergh:
Hopefully fixes https://github.com/official-stockfish/fishtest/pull/2340#issuecomment-3033007744


ppigazzini:
PROD updated.

vdbergh:
Sorry for the extra work...

PR #2346, opened 2025-07-09
Show reset time in rate_limits page.

vdbergh:


vdbergh:
I drafted this since i want to do it a little differenly.

vdbergh:
It seems I did not draft it, but in any case the PR is now ready for review.

vdbergh:
It looks as follows
Screenshot from 2025-07-09 21-18-11

vdbergh:
There is a subtle bug. If time passes the reset time then the number remaining api calls changes silently and the cached value is no longer valid.

EDIT: I will fix it later.

vdbergh:
Hopefully fixed.

vdbergh:
I fixed the first Copilot comment. The other one is just the suggestion to use the Python logging framework... That's something for another time.

PR #2347, opened 2025-07-11
Better behaviour of PR2346 in case of network failure.

vdbergh:
The fields that cannot be refreshed will color red.


vdbergh:
I implemented the copilot suggestions. I checked that the fallback for `AbortSignal.timeout()` indeed works (but I checked it only on a modern browser, where the fallback is of course not necessary).

PR #2349, opened 2025-07-13
Various cleanups

vdbergh:
- Install signal handlers also in secondary instances (bugfix). It seems this code was accidentally disabled in the past. The result of this was that stopping secondary instances was no longer recorded in the event log.

- Make sure that `gitub_api.py` is properly initialized before we use it.

- Initialize `github_api.py` in `__init__.py` rather than in the constructor of `rundb`. We do this only in the primary instance.

- Instead of directly importing methods from `github_api.py` (recall that we think of it as a singleton object) we now do `import fishtest.github_api as gh` and then use `gh.<method>`.

- Create a private method `_update_rate_limit` which does the heavy lifting for keeping `_github_rate_limit` up to date.

- Fix a tab error in `rate_limits.mak`.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -4,6 +4,7 @@ import sys
 import threading
 import traceback
 
+import fishtest.github_api as gh
 from fishtest.routes import setup_routes
 from fishtest.rundb import RunDb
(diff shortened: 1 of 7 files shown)


ppigazzini (approved):

PR #2350, opened 2025-07-16
Display error message when obtaining diff fails.

vdbergh:
A rudimentary implementation for
https://github.com/official-stockfish/fishtest/pull/2340#issuecomment-3033389107

### DIFF
--- a/server/fishtest/static/js/application.js
+++ b/server/fishtest/static/js/application.js
@@ -167,12 +167,96 @@ function log(message, trace) {
   }
 }
 
+// Convert HTTP status code into status text
+function getStatusText(code) {
+  return (
+    {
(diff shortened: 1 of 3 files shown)


vdbergh:
It looks like this

Screenshot from 2025-07-19 08-01-35-edited2

vdbergh:
CI failed because the server ran out of api calls. Probably the actions script should create a valid .netrc file using the GITHUB_TOKEN.

vdbergh:
Ok these comments were implemented as well.

vdbergh:
CI fails because of rate limit exceeded. This is fixed in #2354 .

ppigazzini (approved):

PR #2352, opened 2025-07-19
Remove a useless whitespace

FauziAkram:
Remove a useless whitespace
Pointed out by salve_sf_fan on Discord

### DIFF
--- a/server/fishtest/templates/rate_limits.mak
+++ b/server/fishtest/templates/rate_limits.mak
@@ -32,7 +32,7 @@
     GitHub personal access token
     </a>
     in your
-    <a href="/user"> profile </a>.
+    <a href="/user"> profile</a>.
   </p>
 </div>
 <script>


server/fishtest/templates/rate_limits.mak
```diff
@@ -32,7 +32,7 @@
     GitHub personal access token
     </a>
     in your
-    <a href="/user"> profile </a>.
+    <a href="/user"> profile</a>.
```
Disservin: what about the one in the front lol?

FauziAkram: If I understand it correctly, we should keep the space before "profile" so that it doesn't get glued to the word before it — without the space, you'd get something like "yourprofile" instead of "your profile".

Disservin: pretty sure you won't..

FauziAkram: Ok, also removed the one in front.

ppigazzini (approved):
Looks good on DEV, PROD updated in seconds, thank you @FauziAkram and salvo_sf_fan :)

PR #2353, opened 2025-07-23
Restrain GitHub api calls by the server

vdbergh:
When a test is being viewed, the server makes some GitHub api calls to potentially show some extra information to the user. Most of the time the results of the api calls are cached.

This system works well under normal use but it seems to break down somewhat when crawlers systematically view all tests. In that case the server becomes just a proxy for executing pointless GitHub api calls.

From a practical side this has not been a real problem. When a crawler visits, the server seems to execute up to 500 GitHub api calls/hour which is far below the limit of 5000.

Still, the server executing pointless GitHub api calls on behalf of random crawlers seems wrong, so in this PR we restrict this as follows:

- Authenticated users are allowed to make api calls (a crawler will be unauthenticated).
- Api calls are allowed for tests < 30 days old. These will be served from the cache most of the time.
- In other cases api calls are not allowed.

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1687,6 +1687,23 @@ def tests_view(request):
     elif run["failed"]:
         # for backward compatibility
         warnings.append("this is a failed test")
+    if run["args"]["tc"] != run["args"]["new_tc"]:
+        warnings.append("this is a test with time odds")
+    book_exits = request.rundb.books.get(run["args"]["book"], {}).get("total", 100000)
+    if book_exits < 100000:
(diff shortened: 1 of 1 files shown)


vdbergh:
This is a screenshot of an occasion where a crawler seems to be visiting.

image

EDIT: It is getting worse (09:47 local time)

image

EDIT2: And now at 10:00

image

vdbergh:
I implemented the first comment. The second one does not apply since the use of UTC aligns with the use in the rest of  the source.

ppigazzini (approved):

PR #2354, opened 2025-07-23
Use GITHUB_TOKEN during CI.

vdbergh:
In this way we can do unit tests for the GitHub api.

We add a few unit tests.

Also: some refactoring. `update_official_master_sha()` is now contained in `github_api.py`.


vdbergh:
I dealt with these comments. They were all more or less to the point.

ppigazzini:
Thank you @vdbergh

PR #2356, opened 2025-07-25
No GitHub api calls if repo is not forked from official-stockfish

vdbergh:
The GitHub api calls which compare the official master repo with the dev repo will fail.

Also: display corresponding warning.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -543,6 +543,7 @@ class RunDb:
         spsa=None,
         username=None,
         tests_repo=None,
+        master_repo=None,
         auto_purge=False,
         throughput=100,
         priority=0,
@@ -579,6 +580,9 @@ class RunDb:
             "priority": priority,
             "adjudication": adjudication,
         }
+
+        if master_repo is not None:
+            run_args["master_repo"] = master_repo
 
         if sprt is not None:
             run_args["sprt"] = sprt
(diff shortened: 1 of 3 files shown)


vdbergh:
The warning looks like this 

image

vdbergh:
I am not particularly enthused by the comments. The value of `value` in the first comment is already validated because it is part of the run object, which is validated during creation.

For the second comment I think that the condition is a bit too trivial now to wrap it in a helper function (which the reader must then look up). This may perhaps change in the future.

ppigazzini:
DEV running with the PR, I don't know how to test the PR.
This test by snicolet (repo forked from mcostalba) shows the usual diff
https://dfts-0.pigazzini.it/tests/view/685a143c09bf761afbfea0a8

In create test the warning message is different.

image

vdbergh:
I will try to demonstrate it, but need to go home now.

vdbergh:
Here is a test with the warning triggered.

https://dfts-0.pigazzini.it/tests/view/688a379ceb16c3531c92df55

I logged in as vdbergh and created a test with a repo https://github.com/vdbergh/Stockfish2 forked from https://github.com/mcostalba/Stockfish

Note that recently created users can no longer use a dev repo that is not forked from official-stockfish/Stockfish. They get the error message you showed above.

ppigazzini:

PR #2357, opened 2025-07-27
Add clarifying note when diff yields status 401 (Unauthorized)

vdbergh:
We do this to help the user, who might otherwise have no idea what is happening. We now get this:

Screenshot from 2025-07-27 07-19-59


vdbergh:
Both comments implemented!

PR #2360, opened 2025-07-31
Log failing API calls to event log.

vdbergh:
Occasionally there are reports of GitHub API calls that fail and then succeed when repeated. E.g.
https://github.com/official-stockfish/fishtest/pull/2293#issuecomment-2956858006

It would be interesting to know how often this occurs. Therefore we now log such events in the event log. We do this by saving certain error replies in the GH API cache but we do not use these entries afterwards. So there is no negative caching. Not at this time anyway.

It is not obvious how to test this PR in real life, but it is tested in `test_github_not_found()` in `test_github_api.py`.

### DIFF
--- a/server/fishtest/__init__.py
+++ b/server/fishtest/__init__.py
@@ -85,7 +85,7 @@ def main(global_config, **settings):
             # - uses the network;
             # - accesses the db;
             # - starts new threads.
-            gh.init(rundb.kvstore)
+            gh.init(rundb.kvstore, rundb.actiondb)
             rundb.update_aggregated_data()
             rundb.schedule_tasks()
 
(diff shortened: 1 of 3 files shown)


vdbergh:
Drafted this since I spotted a bug.

vdbergh:
The first comment is a genuine bug which I fixed.

I prefer to leave the second comment as it is. The reply to failed github api calls is documented and it should always be json. Furthermore it is ok for the methods in `github_api.py` to throw exceptions. The convention is that these are caught and handled by the caller.

ppigazzini:
DEV updated.

vdbergh:
> DEV updated.

Well the code cannot be triggered in real life without control over the GitHub API servers :) It is meant to document some strange behavior of GitHub which has been observed by some people.

But I do check it in the unit test by poisoning the API cache.

ppigazzini:
> Well the code cannot be triggered in real life without control over the GitHub API servers :) 

Please, write the companion PR to control the GitHub API servers :)

ppigazzini:

PR #2361, opened 2025-08-01
Update fastchess

vondele:
update fastchess to latest version, include fixes that might soon be needed.

Take care of additional warnings


ppigazzini (approved):
Looks good on DEV with linux, wsl2 and msys2 workers.
Workers update trigger after merging, thank you @vondele

PR #2364, opened 2025-08-03
Enforce structure of base_options and new_options in schema

vdbergh:


vdbergh:
The pattern is meant to match the output of `sanitize_options`. I wrote a little program to test the pattern
```python
from vtjson import intersect, validate, regex, set_name, ValidationError

ascii = set_name(lambda x: x.isascii(), name="ascii")
set_option = r"[^\s=]+=[^\s=]+"
option_list = intersect(
    str, ascii, regex(rf"(|({set_option} )*{set_option})", "option_list")
)

negative_tests = (
    1,
    "\u1000",
    "\u0000",
    "\u1000=\u1000",
    " ",
    "a",
    "a=",
    "a =5",
    "a= 5",
    "a = 5",
    " a=5",
    "a=5 ",
    "a=5 b =5",
    "a=5  b=5",
)

print("Negative tests")
print("==============")
for t in negative_tests:
    try:
        validate(option_list, t, name="pattern")
        print(f"FAIL: {repr(t)} validates")
    except ValidationError as e:
        print(f"SUCCESS: {repr(t)}:{str(e)}")

positive_tests = ("", "a=1", "a=b c=d", "AAAAAAA=BBBBBBBBB", "a=b c=d e=f")
print("Positive tests")
print("==============")

for t in positive_tests:
    try:
        validate(option_list, t, name="pattern")
        print(f"SUCCESS: {repr(t)} validates")
    except ValidationError as e:
        print(f"FAIL: {repr(t)}:{str(e)}")
```
Output:
```
Negative tests
==============
SUCCESS: 1:pattern (value:1) is not of type 'str'
SUCCESS: 'က':pattern (value:'က') is not of type 'ascii'
SUCCESS: '\x00':pattern (value:'\x00') is not of type 'option_list'
SUCCESS: 'က=က':pattern (value:'က=က') is not of type 'ascii'
SUCCESS: ' ':pattern (value:' ') is not of type 'option_list'
SUCCESS: 'a':pattern (value:'a') is not of type 'option_list'
SUCCESS: 'a=':pattern (value:'a=') is not of type 'option_list'
SUCCESS: 'a =5':pattern (value:'a =5') is not of type 'option_list'
SUCCESS: 'a= 5':pattern (value:'a= 5') is not of type 'option_list'
SUCCESS: 'a = 5':pattern (value:'a = 5') is not of type 'option_list'
SUCCESS: ' a=5':pattern (value:' a=5') is not of type 'option_list'
SUCCESS: 'a=5 ':pattern (value:'a=5 ') is not of type 'option_list'
SUCCESS: 'a=5 b =5':pattern (value:'a=5 b =5') is not of type 'option_list'
SUCCESS: 'a=5  b=5':pattern (value:'a=5  b=5') is not of type 'option_list'
Positive tests
==============
SUCCESS: '' validates
SUCCESS: 'a=1' validates
SUCCESS: 'a=b c=d' validates
SUCCESS: 'AAAAAAA=BBBBBBBBB' validates
SUCCESS: 'a=b c=d e=f' validates
```

vdbergh:
Hmmm. It seems that SF admits spaces in its option names. This means the pattern is wrong (and so is `sanitize_options`, which uses the same pattern). We could for course ignore this until someone protests.
```
option name Debug Log File type string default 
option name Contempt type spin default 24 min -100 max 100
option name Analysis Contempt type combo default Both var Off var White var Black var Both
option name Threads type spin default 1 min 1 max 512
option name Hash type spin default 16 min 1 max 131072
option name Clear Hash type button
option name Ponder type check default false
option name MultiPV type spin default 1 min 1 max 500
option name Skill Level type spin default 20 min 0 max 20
option name Move Overhead type spin default 30 min 0 max 5000
option name Minimum Thinking Time type spin default 20 min 0 max 5000
option name Slow Mover type spin default 84 min 10 max 1000
option name nodestime type spin default 0 min 0 max 10000
option name UCI_Chess960 type check default false
option name UCI_AnalyseMode type check default false
option name UCI_LimitStrength type check default false
option name UCI_Elo type spin default 1350 min 1350 max 2850
option name SyzygyPath type string default <empty>
option name SyzygyProbeDepth type spin default 1 min 1 max 100
option name Syzygy50MoveRule type check default true
option name SyzygyProbeLimit type spin default 7 min 0 max 7
```

ppigazzini:
Someone called for a protest... :)

vdbergh:
Perhaps we should leave it as it is? It is a bit strange to allow spaces in option names if spaces are also a separator. Moreover if there are spaces in values then the parsing is even ambiguous.

Spaces have been disallowed for a while now and nobody has complained. It means that if someone wants to tune `Move Overhead` then they have to make a custom Stockfish version to do it. Not really a difficult task.

PR #2368, opened 2025-08-07
Add a runtime error for crashing bench

vondele:


vondele:
not urgent, but avoids the generic "Unable to parse bench output"

PR #2370, opened 2025-08-08
Add a runtime error for crashing bench

ppigazzini:
Opened for #2368


ppigazzini:
Looks good on DEV with linux/wsl2/linux workers, PROD updated, thank you @vondele

PR #2371, opened 2025-08-10
Disable CORS for the public get_run api

vdbergh:
This only affects browsers. See

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS

CORS is already disabled for the actions api.


vdbergh:
I am locally developing a pull request application. It runs on the client but it needs to download run objects from the server to extract some data.

With CORS enabled the application needs to be served from the same server as the run objects which is very inconvenient for development.

ppigazzini (approved):

PR #2375, opened 2025-08-16
Update fastchess to 66cac47

Disservin:
Update to not run into any issues regarding the behavior before
https://github.com/Disservin/fastchess/pull/903


ppigazzini (approved):
Looks good with linux, wsl2 and msys2 workers on DEV. Workes upgrade triggered just after merging, thank you @Disservin

PR #2384, opened 2025-08-31
fix(tests_view): lazy repo normalization

ppigazzini:
Use the stored repo URL first (no network) and run GitHub checks. If checks raise, normalize once and retry, then recompute the anchor. Flash a warning if the repo URL changed after normalization.

Prevents 500s when github.com is slow/unreachable (ReadTimeout) while keeping the happy path fast and unchanged.

Fix suggested by @robertnurnberg

Closes https://github.com/official-stockfish/fishtest/issues/2381

### DIFF
--- a/server/fishtest/views.py
+++ b/server/fishtest/views.py
@@ -1727,7 +1727,9 @@ def tests_view(request):
             return False
         return True
 
-    user, repo = gh.parse_repo(gh.normalize_repo(tests_repo(run)))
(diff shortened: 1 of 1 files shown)


vdbergh:
I think my solution is conceptually (and practically) simpler....

I haven't studied this PR in detail, but I am afraid that "checking if things work" will lead to the same kind of spamming as we are having by invoking normalize_repo directly.

Also one needs to be careful with "checking if things work" since many github api calls are cached (like is_master).

PR #2386, opened 2025-09-03
Use run_id/task_id in the name of the pgn file

### LINKED ISSUE #2385: pgn file contains batch from wrong test
The pgn from https://tests.stockfishchess.org/tests/view/68b2c1426217b8721dca9d1c contains games from two different tests:

```
> zgrep Event ./25-08-30/68b2c1426217b8721dca9d1c/68b2c1426217b8721dca9d1c.pgn.gz | sed 's/.*://' | sort -u
 moreSpc vs master"]
 test1562 vs master"]
```

Discovered by running the script from https://github.com/official-stockfish/fishtest/issues/2286 on nightly downloads, which flagged duplicate book exits in that pgn file.

vdbergh:
This should fix #2385 (the cause of that issue remains unclear however).

Note: we no longer delete the pgn file. We rely instead on the trim_files() function which is called when starting a new task.

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -36,7 +36,7 @@ Proper configuration of `nginx` is crucial for this, and should be done
 according to the route/URL mapping defined in `__init__.py`.
 """
 
-WORKER_VERSION = 291
+WORKER_VERSION = 292
 
 
 @exception_view_config(HTTPException)
--- a/worker/games.py
+++ b/worker/games.py
@@ -165,7 +165,7 @@ def trim_files(testing_dir, source_dir=None):
         ("stockfish-*-old" + EXE_SUFFIX, 0, -1, True),
         ("stockfish-*" + EXE_SUFFIX, 50, 30, False),
         ("nn-*.nnue", 10, 30, False),
-        ("results-*.pgn", 0, -1, True),
+        ("results-*.pgn", 0, -1, False),
         ("*.epd", 4, 365, False),
         ("*.pgn", 4, 365, False),
     )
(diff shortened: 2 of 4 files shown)


vdbergh:
Undrafted.

robertnurnberg:
I think it's safer to still attempt to delete the pgn file, as in master, even with the new name. 

Once fastchess allows disabling the default append mode, this could then be dropped.

vdbergh:
> I think it's safer to still attempt to delete the pgn file, as in master, even with the new name.
> 
> Once fastchess allows disabling the default append mode, this could then be dropped.

I don't see why it would be safer. Every task will now write to a different pgn file.

I prefer to do cleanup tasks in a single place, rather than sprinkled throughout the source. This makes debugging easier (in current master the pgn file is deleted in _two places_ ...).

robertnurnberg:
> Note: we no longer delete the pgn file. We rely instead on the trim_files() function which is called when starting a new task.

Is there any risk of deleting a pgn file that was not fully uploaded for the previous task yet?

robertnurnberg:
> I don't see why it would be safer. Every task will now write to a different pgn file.

I was thinking about a user who created the pgn file beforehand by chance, but I had missed your change to the trim function.

vdbergh:
> > Note: we no longer delete the pgn file. We rely instead on the trim_files() function which is called when starting a new task.
> 
> Is there any risk of deleting a pgn file that was not fully uploaded for the previous task yet?

The upload happens synchronously.

vdbergh:
> > I don't see why it would be safer. Every task will now write to a different pgn file.
> 
> I was thinking about a user who created the pgn file beforehand by chance, but I had missed your change to the trim function.

I don't think that a user will create a file like `results-68b18eab6217b8721dca9c0c-32.pgn` in the testing directiory by chance (and do it is such a way that `trim_files()` would not delete it).  Such a user would be a malicious user and sadly a malicious user can do anything.

Disservin:
How much longer is the current pgn name vs the new one ? I remember that in the past we had issues with max path length

robertnurnberg:
> I prefer to do cleanup tasks in a single place, rather than sprinkled throughout the source. This makes debugging easier (in current master the pgn file is deleted in _two places_ ...).

It is actually three places, I think. ;) https://github.com/official-stockfish/fishtest/blob/a6afbb4cd13ddac573c169c6b3ff58e7f85213e6/worker/worker.py#L1416

vdbergh:
> How much longer is the current pgn name vs the new one ? I remember that in the past we had issues with max path length

It's about the same length.

ppigazzini (approved):
Looks good with linux, wsl2 and msys2 workers, workers update triggered after merging, thank you @vdbergh

PR #2390, opened 2025-09-11
check integrity of cached binaries

### LINKED ISSUE #2389: Cached Stockfish binaries may improperly stop a test
https://discord.com/channels/435943710472011776/900771437177094156/1415342184525332481

The above happened when the host processor of my Azure VM instance was changed after reboot. (Intel Xeon Platinum 8370C -> Intel Xeon Platinum 8168)

If a host processor is changed (when running on VM/cloud) and architecture is different from which Stockfish was compiled on, bench command will crash and stop the corresponding test prematurely.

Instead of trusting cached executables unconditionally, I think it's preferable to do:
1. Run stockfish
2. If a) worker is using the cached stockfish and b) it fails (statuscode != 0), rebuild stockfish
3. If stockfish compiled in current session crashes, send a stop signal to the server

robertnurnberg:
This PR implements the workflow suggested by @MinetaS in #2389.

Fixes #2389.


MinetaS:
LGTM
Except one trivial thing is that it now runs verify_signature for every cached instance. I was expecting if it can be tested in run_games instead, but maybe it's better not to overhaul the entire function here.

robertnurnberg:
If we wanted, we could save the successful bench run, and then later not repeat. Let's see what the maintainers say.

robertnurnberg:
Implemented the suggestion. Kept both commits for clarity. (And fixed formatting in a third commit. ;) )

worker/games.py
```diff
@@ -823,8 +824,20 @@ def setup_engine(
     engine_path_native = (testing_dir / engine_name).with_suffix(EXE_SUFFIX)
     for path in (engine_path_native, engine_path):
         if path.exists():
-            update_atime(path)
-            return path
+            try:
+                verify_signature(path, signature)
+            except Exception as e:
+                print(f"Cached engine gave exception:\n{e}", file=sys.stderr)
```
MinetaS: Thank you for the work. I tested the PR for a short time and thought it's better to address this:

`run_single_bench` raises three different kinds of exception: `RunException`, `OSError` and `subprocess.SubprocessError`. I'm not sure if we need to catch the last two errors here, but at least `RunException` covers most of the errors, including nonzero exit codes.

If every `Exception` is caught here, it will handle `FatalException` too. which should terminate the worker regardless (especially when user interrupts the worker with SIGINT). I actually had to press Ctrl+C twice to stop the worker when it was verifying the signature of a cached SF instance.

I suggest limiting the scope here to `RunException` at least, or `OSError` and `subprocess.SubprocessError` too if necessary.

robertnurnberg: Thanks for the review. Agreed and done.

worker/games.py
```diff
@@ -825,8 +826,20 @@ def setup_engine(
     engine_path_native = (testing_dir / engine_name).with_suffix(EXE_SUFFIX)
     for path in (engine_path_native, engine_path):
         if path.exists():
-            update_atime(path)
-            return path
+            try:
+                verify_signature(path, signature)
+            except RunException as e:
```
ppigazzini: `run_single_bench()` raises other exception types (eg `OSError`), the bad engine in that case is not deleted.

robertnurnberg: Thanks for doing these changes. I think going to general Exceptions here would also include a user aborting the bench run with ctrl -c? I think that happened to MinetaS, hence my (incomplete) change to Run exceptions.

ppigazzini: A corrupted binary raises other type of errors:
```
Variable task sizes used. Opening offset = 67584.
Computing engine signature...
Signature verification failed for cached engine:
[Errno 8] Exec format error: PosixPath('/home/usr00/_git/__test_folder00/fishtest/worker/testing/stockfish-fc54d8730174cdb5cfc4f7074b90128e706e4040-g++_14_2_0-25b32ad706')
Deleted cached engine /home/usr00/_git/__test_folder00/fishtest/worker/testing/stockfish-fc54d8730174cdb5cfc4f7074b90128e706e404
```

worker/games.py
```diff
@@ -825,10 +826,23 @@ def setup_engine(
     engine_path_native = (testing_dir / engine_name).with_suffix(EXE_SUFFIX)
     for path in (engine_path_native, engine_path):
         if path.exists():
-            update_atime(path)
-            return path
+            try:
+                verify_signature(path, signature)
+            except Exception as e:
```
robertnurnberg: How about checking for KeyboardInterrupt and/or FatalException first, and raising them if encountered?

ppigazzini: It should be processed by our signal code

https://github.com/official-stockfish/fishtest/blob/5745e48ad1604b1aec203da34ca1a8a620b62a4b/worker/worker.py#L1455-L1467

https://github.com/official-stockfish/fishtest/blob/5745e48ad1604b1aec203da34ca1a8a620b62a4b/worker/worker.py#L883-L885

robertnurnberg: Hm, @MinetaS reported having to press ctrl-c twice in an earlier version of the pr with the same exception handling.

robertnurnberg: So I checked the current commit myself. On linux hitting ctrl-c during signature check does indeed not lead to the worker being killed. It needs a second ctrl-c.
```
Computing engine signature...
^CSignature verification failed for cached engine:
Terminated by signal SIGINT.
Deleted cached engine /home/rn/git/fishtest/worker/testing/stockfish-fc54d8730174cdb5cfc4f7074b90128e706e4040-g++_11_4_0-25b32ad706
Downloading https://api.github.com/repos/official-stockfish/Stockfish/zipball/fc54d8730174cdb5cfc4f7074b90128e706e4040...
Heartbeat stopped.
Build uses default net: nn-1c0000000000.nnue
Build uses default net: nn-37f18f62d772.nnue
Using native target architecture.
^C
```

MinetaS: Python delivers the exception in the context of the code that was running at the time when the signal arrived.
https://docs.python.org/3/library/signal.html#handlers-and-exceptions

So anything like `except Exception:` is not desirable anywhere in the worker script as `WorkerException` inherits from `Exception` class.

vdbergh: > So I checked the current commit myself. On linux hitting ctrl-c during signature check does indeed not lead to the worker being killed. It needs a second ctrl-c.
> 
> ```
> Computing engine signature...
> ^CSignature verification failed for cached engine:
> Terminated by signal SIGINT.
> Deleted cached engine /home/rn/git/fishtest/worker/testing/stockfish-fc54d8730174cdb5cfc4f7074b90128e706e4040-g++_11_4_0-25b32ad706
> Downloading https://api.github.com/repos/official-stockfish/Stockfish/zipball/fc54d8730174cdb5cfc4f7074b90128e706e4040...
> Heartbeat stopped.
> Build uses default net: nn-1c0000000000.nnue
> Build uses default net: nn-37f18f62d772.nnue
> Using native target architecture.
> ^C
> ```
Not sure if this is the issue, but long running processes started by the worker should be started in a new session on Linux. 
```
            with subprocess.Popen(
                cmd,
                shell=True,
                env=os.environ,
                start_new_session=False if IS_WINDOWS else True,   <-------
                stdout=None if tests else subprocess.DEVNULL,
                stderr=subprocess.PIPE,
                universal_newlines=True,
                bufsize=1,
                close_fds=not IS_WINDOWS,
            ) as p:
```

vdbergh: This did not solve the problem.

vdbergh:
```
           try:
                verify_signature(path, signature)
            except Exception as e:
                print(
                    f"Signature verification failed for cached engine:\n{e}",
                    file=sys.stderr,
                )
                try:
                    path.unlink()
                    print(f"Deleted cached engine {path}")
                except FatalException:
                    raise
                except Exception as e:
                    raise WorkerException(f"Failed to remove cached engine {path}", e=e)
            else:
                update_atime(path)
                return path
```

I have my reservations about nested exception handling. I would prefer to set flags when exceptions are raised and then act appropriately.

ppigazzini:
TBH I don't like to move the responsibility of the signature check to the setup_engine: nested exceptions, the signature mismatch is boring to test (a miss match triggers the rebuild process)
This night I will test the option to add a fast bench at depth=1 in setup_engine to check the cached engine, keeping the master code for the signature check.

worker/games.py
```diff
@@ -318,8 +318,6 @@ def post_to_worker_log(worker_info, password, remote, message):
     }
     try:
         send_api_post_request(remote + "/api/worker_log", payload)
-    except FatalException:
```
vdbergh: I am kind of curious. Why did you delete this? It is meant to catch Ctrl-C.

MinetaS: iirc there are other places that have nested exceptions and require user to press Ctrl+C twice. I think it would be better to handle this in a separate PR.

vdbergh: This was not a nested exception. Just handling a special case.

vdbergh: Ok I see. The FatalException handling is also not in master. In that case it is indeed best to leave things as they are since AFAICS this has nothing to do with what this PR is about.

robertnurnberg:
Thanks for the changes and cleanup @ppigazzini. I changed the title of the PR and undrafted.

worker/games.py
```diff
@@ -808,6 +808,20 @@ def create_environment():
     return env, env_hash
 
 
+def engine_is_healthy(path: Path, timeout_s: float = 1.0):
+    try:
+        r = subprocess.run(
+            [str(path), "bench", "16", "1", "1", "default", "depth"],
```
vondele: I think it would be worthwhile to test at slightly deeper depth. Like depth 5 or 6, is essentially the same runtime as depth 1:
```
% for i in `seq 1 9`; do time ./stockfish bench 16 1 $i default depth > /dev/null 2>&1; done
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.10s user 0.02s system 97% cpu 0.125 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.09s user 0.01s system 97% cpu 0.113 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.10s user 0.01s system 97% cpu 0.114 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.10s user 0.01s system 97% cpu 0.117 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.10s user 0.01s system 97% cpu 0.120 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.12s user 0.01s system 97% cpu 0.131 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.14s user 0.01s system 97% cpu 0.155 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.17s user 0.02s system 98% cpu 0.185 total
./stockfish bench 16 1 $i default depth > /dev/null 2>&1  0.22s user 0.02s system 98% cpu 0.239 total
```

ppigazzini: Done

PR #2391, opened 2025-09-13
Improve memory estimate

vondele:
recently a few workers had failures due to too small memory estimate. update based on new measurements.


MinetaS:
I watched my worker while it's running a task from a LTC test:

- python3: 63248
- fastchess: 213M
- stockfish: 227-240M

May I ask what `Needed for binaries` means? I think SF binary size is already addressed in `(12 + 139 + 16 * run["args"]["threads"])`

Also without TT size one stockfish process takes 176M which exceeds 12+139+16=167.

vondele:
the binary size is not yet in the 167M, as this is shared between all processes. So the 167 + binaries / (2 * concurrency) should be close to the 176M (without the TT).

ppigazzini (approved):
Looks good on linux/wsl2/msys2 workers. Workers update after merging, thank you @vondele

PR #2394, opened 2025-09-19
Update fastchess

MinetaS:


ppigazzini (approved):
Looks good with all workers, workers update triggered after merging, thank you @MinetaS

PR #2401, opened 2025-10-11
feat(server): switch to python 3.14

ppigazzini:
Also bump some packages.


ppigazzini:
DEV and PROD have been running fine with python 3.14 for few days.

https://docs.python.org/3.14/whatsnew/3.14.html

PR #2408, opened 2025-10-24
Stop task when heartbeat receives an error from server

vdbergh:
Very likely the same error will prevent an update later, resulting in wasted compute.

The main (only?) use case is an external IP address change (e.g. when moving a computer to a different location).


ppigazzini (approved):
Looks good with linux, wsl2, msys2 workers, update just after merging, thank you @vdbergh

vdbergh:
Sorry for not reacting. I was going to, but I am currently very busy.

PR #2410, opened 2025-11-22
update fastchess

robertnurnberg:
This PR serves as a placeholder to keep track of changes in fastchess during its development.

List of changes:
* `--version` becomes `-version`
* fastchess now produces correct PGN output for games ending with illegal moves
* we use the option `-pgnout append=false` as an extra safety measure to avoid corrupt PGN files
* fix crash on older mac workers, see #2412


Disservin:
thanks for the quick update

PR #2411, opened 2025-11-23
worker: build fastchess with build=release for static linking on Windows

maximmasiutin:
Build fastchess with `build=release` to ensure static linking on Windows.

Without this, the binary depends on MSYS64 UCRT DLLs, causing `nanosleep64` entry point errors or `Engine didn't respond to uciok` failures when run outside MSYS64.

Related: https://github.com/Disservin/fastchess/pull/950

### DIFF
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -461,11 +461,11 @@ def setup_fastchess(worker_dir, compiler, concurrency, global_cache, tests=False
         os.chdir(build_dir)
 
         cmds = [
-            f"make -j{concurrency} CXX={compiler} GIT_SHA={FASTCHESS_SHA[:8]} GIT_DATE=01010101",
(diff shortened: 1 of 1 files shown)


Disservin:
i don't understand this.. the environment you compile fastchess in, is also the one supposed to execute it.. there's no need for static linking

ppigazzini:
msys2 ucrt64 builds an application that needs:
  - msys2 ucrt64 dlls, for fastchess they are:
```
  c/msys64/ucrt64/bin/libgcc_s_seh-1.dll
  c/msys64/ucrt64/bin/libwinpthread-1.dll
  c/msys64/ucrt64/bin/libstdc++-6.dll 
```
  - Microsoft UCRT dlls
  
If fastchess.exe does not find the 3 msys2 ucrt64 libraries, neither python3.exe, g++.exe, stockfish.exe should find their required libraries.

PR #2413, opened 2025-12-04
Raise WorkerException when fastchess crashes.

### LINKED ISSUE #2412: fastchess segfaults not shown in events
It seems that currently fastchess has a problem on older mac workers (macOS < 26.0) if the `dir` option is given, fastchess will crash with a SIGSEGV.

Below is the log of the worker output, which shows not really any diagnostics that fastchess ended unexpectedly.

```
Indexing opening suite...
Started game 3 of 160 (New-d7e49e2b20 vs Base-5297ba0a1a)
Started game 1 of 160 (New-d7e49e2b20 vs Base-5297ba0a1a)
Started game 2 of 160 (Base-5297ba0a1a vs New-d7e49e2b20)

Waiting for fastchess to finish... done.
Task exited
```

I will push an update later today which will fix this

https://github.com/Disservin/fastchess/issues/957

vdbergh:
Should fix #2412.

I sent a signal to fastchess while the worker was running and the task stopped gracefully.

https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=&before=1764873583.097603&run_id=

(sorry for forgetting to use DEV).


Disservin:
nice thanks :D

ppigazzini (approved):
Looks good with linux, msys2, wsl2 workers. Update after merging.

Tested with a faster cleaner fastchess:
```c
#include <stdio.h>
#include <string.h>

int main(int argc, char *argv[]) {
    // Check if there is exactly one argument and it equals "-version"
    if (argc == 2 && strcmp(argv[1], "-version") == 0) {
        printf("fastchess alpha 1.5.0 01010101-e892ad92\n");
        return 0; // success
    } else {
        return 1; // error
    }
}
```

vdbergh:
Just saw one in real life  https://tests.stockfishchess.org/actions?max_actions=1&action=&user=&text=fastchess&before=1765303442.779962&run_id=

vdbergh:
I wonder if we should collect, say, the last 10 lines of FC output and include them in the error message.

Disservin:
sometimes could be useful, like seeing the segfault message for example, the fastchess output itself isn't that useful..
note that on discord, maximmasiutin had issues in the past which were related to his environment.. my guess is that those issues haven't been completely fixed

vdbergh:
This worker seems terminally broken. https://tests.stockfishchess.org/actions?text=%22maximmasiutin-1cores-fe4ac214%22 I blocked it.

vdbergh:
It’s rather puzzling that the errors alternate between two workers with different long names but the same short name. My first reaction would be that that is impossible unless the worker is hacked.

EDIT. It is possible if the two workers run in identical but different virtual machines and the user has copied the same fishtest.cfg file (with a [private] section) to both of them (contrary to the instructions in the wiki).

If neither of the workers is able to run a task then they would connect alternatingly, as we are seeing here.

This is the relevant section in the wiki: https://github.com/official-stockfish/fishtest/wiki/Running-the-worker#cloud-and-container-considerations

ppigazzini:
I pinged the user on Discord linking this thread.

PR #2415, opened 2025-12-18
Censor ip addresses in failed_task messages.

vdbergh:
We want to avoid this:

https://tests.stockfishchess.org/actions?max_actions=1&before=1766088916.498543


vdbergh:
#2417 removes the only use case. So I am closing this.

PR #2416, opened 2025-12-18
Fail task when heartbeat returns server error.

vdbergh:
Currently the task is stopped normally.

This PR should be merged after #2415 since it will expose IP addresses.

### DIFF
--- a/worker/games.py
+++ b/worker/games.py
@@ -1089,7 +1089,12 @@ def parse_fastchess_output(
         if current_state["task_id"] is None:
             # This task is no longer necessary.
             # Error message has already been printed.
-            return False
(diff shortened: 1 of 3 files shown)


vdbergh:
It seems the changing IP addresses are all okrout. Perhaps this is what explains the many "stale active task" messages for okrout.

If that's the case then we should probably remove the ip address check. This requires a bit of thought.

vdbergh:
#2417 removes the single known use case for this PR. In addition there is a flaw. A remote error returned by the heartbeat will very likely prevent the resulting failed task event from reaching the server. That makes this PR non-functional.

PR #2418, opened 2025-12-22
Revert pull request commits

vdbergh:
I feel no longer comfortable with the pull request feature. It is a large monolithic block of code and I would like to remove it. Currently removal is still quite easy but it may become harder as fishtest evolves.

Some parts of the pull request code are useful (e.g. error handling) and I will probably resubmit them as separate patches.


vondele:
I think that's fair, wasn't really used, it went a bit further than the original idea to have easy copy&paste of the results. Thanks for the effort though!

PR #2419, opened 2025-12-23
Remove GUI option to priority >0 for non-approvers

anematode:
Recently a new contributor got understandably confused about the global nature of priority and set their test's priority to be >0. This should prevent that from accidentally happening again.

### DIFF
--- a/server/fishtest/templates/tests_view.mak
+++ b/server/fishtest/templates/tests_view.mak
@@ -364,6 +364,9 @@
               class="form-control"
               name="priority"
               id="modify-priority"
+            % if not approver:
+              max="0"
+            % endif
               value="${run['args']['priority']}"
             >
           </div>


vdbergh:
In some sense this is already true. If a non approver raises the priority >0 then the test has to be reapproved (so an approver needs to be involved anyway). In addition a warning will be shown on the test page that the priority is higher than usual.

anematode:
What happened here then...

https://tests.stockfishchess.org/actions?run_id=694716a13c8768ca4507235e

vondele:
well was approved, but might be easy to overlook. I would kind of agree that this is a fair restriction.

vdbergh:
> well was approved, but might be easy to overlook.

This means that the approver chose to ignore the warning...

vdbergh:
Anyway for consistency there should be a similar restriction for throughput > 100.

vdbergh:
For the record, this was the warning ignored by the approver

image

In addition the approve button (not shown here) would be yellow.

ppigazzini:
A user can set high priority and high throughput on the new test page.
We must define the approver’s role:
- approve-only model: approver verifies high values -> no code changes.
- setter‑approver model: approver must assign and approve high values -> new test page must be updated as well dropping the priority and throughput fields

vondele:
I'm slightly less concerned about throughput, it has a more gradual effect. 

Prio 1 on the other hand is rather clear.

vdbergh:
> I'm slightly less concerned about throughput, it has a more gradual effect.
> 
> Prio 1 on the other hand is rather clear.

I prefer consistency... (the current Fishtest also gives a warning if throughput is too high).

vondele:
fair. No problems with that.

PR #2425, opened 2026-01-02
fix(worker): stop run only for a bench mismatch

ppigazzini:
Bench validation previously raised RunException for bench crashes, timeouts, SIGKILL/OOM, and other non-signature-related failures. The worker maps RunException to /api/stop_run, which incorrectly stops the whole test on the server.

Change bench execution and parsing failures to raise WorkerException, so only the current task fails (/api/failed_task). Keep RunException only for true bench signature mismatches, preserving the intended "stop run on mismatch" behavior.

Raise worker version to 307 (also server side).


worker/games.py
```diff
@@ -466,21 +466,22 @@ def run_single_bench(engine, hash_size, threads, depth, timeout=600):
             except subprocess.TimeoutExpired as e:
                 p.kill()
                 message = f"Bench of {engine.name} timed out after {timeout} seconds."
-                raise RunException(message) from e
+                raise WorkerException(message, e=e) from e
```
robertnurnberg, line 469: This should probably still be RunException? We had such rogue patches on the past iirc

ppigazzini: This is the issue about the timeout case:
https://github.com/official-stockfish/fishtest/issues/2325

@vondele let me know if for this case you prefer WorkerException (the worker stops for 15 seconds and then requests a task) or RunException (a qualified worker stops the run).

vondele: Not 100% sure what is best. Probably a WorkerException. In case a test causes hangs on all workers, it will just not start, and get canceled by the developer.

ppigazzini: Ok for WorkerException

PR #2427, opened 2026-01-05
Show worker arch in machines list and task list.

vdbergh:
image

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -420,7 +420,7 @@ worker_info_schema_api = {
     "compiler": union("clang++", "g++"),
     "unique_key": uuid,
     "modified": bool,
-    "worker_arch?": str,  # can become mandatory after worker upgrade
+    "worker_arch": str,
     "ARCH": str,
     "nps": unumber,
     "near_github_api_limit": bool,
--- a/server/fishtest/templates/machines.mak
+++ b/server/fishtest/templates/machines.mak
@@ -17,6 +17,7 @@
       <th>MNps</th>
       <th>RAM</th>
       <th>System</th>
+      <th>Arch</th>
       <th>Compiler</th>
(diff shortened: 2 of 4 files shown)


Disservin:
wait.. what if the patch modifies the arch detection script and adds a new arch ? this won't be shown here.. and a test creator can't limit it to the new arch.. ? we'd need to take the arch from the bash script of the running test and not master.. ?
or should we ignore this case for now?

vdbergh:
> wait.. what if the patch modifies the arch detection script and adds a new arch ? this won't be shown here.. and a test creator can't limit it to the new arch.. ? we'd need to take the arch from the bash script of the running test and not master.. ? or should we ignore this case for now?

I think this is too complicated (and probably not desirable). I think if SF learns to handle a new arch, it should go through master before Fishtest picks it up.

vondele:
I think it is OK for the arch to be shown in the worker list to be the one of master, but building the test/new version should be done with the test version? Only in that way can changes to the arch be tested on fishtest.

Disservin:
> but building the test/new version should be done with the test version?

I think this is anyway the case and won't be touched here

> I think this is too complicated (and probably not desirable). I think if SF learns to handle a new arch, it should go through master before Fishtest picks it up.

I agree on the first bit, but a new arch has to show it's worth it and fishtest is one way to prove that :D

vdbergh:
> I think it is OK for the arch to be shown in the worker list to be the one of master, but building the test/new version should be done with the test version? Only in that way can changes to the arch be tested on fishtest.


Of course nothing changes in the build process. Showing the arch as detected by master (as added by this PR) is purely cosmetic.

However I am now working on a feature where one can specify a filter for the arch (as detected by master). In this way one can for example restrict a test to arm machines.

vondele:
sounds good, that's definitely useful

PR #2436, opened 2026-01-10
Make `arch_filter` an optional field in the run object.

vdbergh:
`arch_filter` is almost always empty. In such a case the convention is that the field should be optional.

Also: Improve layout of check boxes on mobile.


ppigazzini (approved):
Looks good on DEV and mobile (browser emulated), PROD update after merging, thank you @vdbergh

PR #2438, opened 2026-01-11
Now also use a lru_cache decorator for the github_api cache.

vdbergh:
Note that we deleted the ugly error detection code. It has served its purpose:

https://tests.stockfishchess.org/actions?max_actions=1&before=1767274733.476365

This means that the current github_api cache contains invalid entries. So we raised the version. This essentially means that the cache will be cleared on server start. This will slow down test viewing a bit until the cache fills up again.

### DIFF
--- a/server/fishtest/github_api.py
+++ b/server/fishtest/github_api.py
@@ -6,19 +6,14 @@ from urllib.parse import urlparse
 import requests
 from fishtest.lru_cache import LRUCache, lru_cache
 from fishtest.schemas import sha as sha_schema
-from fishtest.schemas import str_int, uint
-from vtjson import ValidationError, union, validate
-from vtjson import url as url_schema
+from vtjson import validate
 
-"""
-We treat this module as a singleton.
-"""
 """
 Note: we generally don't suppress exceptions since too many things can
 go wrong. The caller should gracefully handle whatever comes their
 way.
 """
-GITHUB_API_VERSION = 1
+GITHUB_API_VERSION = 2
 TIMEOUT = 3
 INITIAL_RATELIMIT = 5000
 LRU_CACHE_SIZE = 6000
@@ -33,7 +28,7 @@ _github_rate_limit = {
     "resource": "core",
     "_uninitialized": True,
 }
-_lru_cache = None
+_lru_cache = LRUCache(LRU_CACHE_SIZE)
 _kvstore = None
 
 _dummy_sha = 40 * "f"
@@ -41,10 +36,9 @@ official_master_sha = _dummy_sha
 
 
 def init(kvstore, actiondb):
-    global _actiondb, _github_rate_limit, _kvstore, _lru_cache, _api_initialized
+    global _actiondb, _kvstore, _api_initialized
     _kvstore = kvstore
     _actiondb = actiondb
-    _lru_cache = LRUCache(LRU_CACHE_SIZE)
     try:
         if "github_api_cache" in _kvstore:
             github_api_cache = _kvstore["github_api_cache"]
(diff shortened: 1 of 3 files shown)


server/fishtest/github_api.py
```diff
@@ -336,33 +279,39 @@ def is_ancestor(
...
+def _is_master(sha, ignore_rate_limit=False):
     try:
         merge_base_commit = get_merge_base_commit(
             sha1=sha, sha2=official_master_sha, ignore_rate_limit=ignore_rate_limit
         )
     except requests.HTTPError as e:
         if e.response.status_code == 404:
             # sha has been deleted so it can never become master
-            _lru_cache[inputs] = False
             return False
```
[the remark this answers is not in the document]
vdbergh: True... Fixed.

server/fishtest/github_api.py
```diff
@@ -6,9 +6,7 @@
 import requests
 from fishtest.lru_cache import LRUCache, lru_cache
 from fishtest.schemas import sha as sha_schema
-from fishtest.schemas import str_int, uint
-from vtjson import ValidationError, union, validate
-from vtjson import url as url_schema
+from vtjson import validate
 
 """
 We treat this module as a singleton.
```
ppigazzini: > We treat this module as a singleton.

This comment is not clear to me.

In Python, importing a module always returns the same module object.

If we need a single shared instance of a class, the idiomatic approach is to create it at module level and import it wherever it’s needed.

vdbergh: Well this is just an informal comment. What I want to say is that there can be only one `github_api` object in the whole application (usually called `gh`). Which follows from the fact that it is a module (as you say).

I toyed with the idea of having a `GithubApi` class and let the `gh` object be an instance of that (which would be more pythonic). But then we have  either to make it global or else pass it as a parameter to all methods/functions that need it. Neither option seemed very attractive.

EDIT: I think of `gh` as an object because it has private mutable data (the cache `_lru_cache`).

ppigazzini: Yes, a module is an object in python, a unique object.

https://docs.python.org/3/library/sys.html#sys.modules

```python
import sys
import math

print(math)                     # <module 'math' (built-in)>
print(type(math))               # <class 'module'>
print(id(math))                 # memory identity of the module object

# Importing again returns the *same* object
import math as m2
print(math is m2)               # True

# Modules live in sys.modules (the module registry)
print(sys.modules['math'] is math)   # True
```

vdbergh: Yes that's what I wrote...

vdbergh: If you really want to, you can import a module multiple times by deleting it from `sys.modules.`.. I assume the resulting modules have distinct module level global data. That would be a method to construct multiple `github_api` objects...

vdbergh: Perhaps a more pythonic method would be

- Have a GithubApi class inside github_api.py
- Intantiate a module level global object github_api
- When importing do 
```
import github_api.github_api as gh
```
In that way we can easily instantiate isolated GithubApi object which can be useful for testing.

EDIT: I will try to rewrite it in this way, but not in this PR.

ppigazzini: Deleting a module to have different copies is simply fighting against python itself.

In python you can destroy the interpreter deleting the builtins, replacing the `__import__` function etc.

The original comment suggests that python modules are not unique, and only this one is a special case. IMO this is clearer "github_api must be unique in the application, so we implement it as a module and import it wherever needed."

ppigazzini: > Perhaps a more pythonic method would be
> 
> * Have a GithubApi class inside github_api.py
> * Intantiate a module level global object github_api
> * When importing do
> 
> ```
> import github_api.github_api as gh
> ```
> 
> In that way we can easily instantiate isolated GithubApi object which can be useful for testing.

I agree that this is another pythonic way.

To be clearer, the code is already good, only that comment is confusing.

vdbergh: I will delete it.

Edit: done!

PR #2440, opened 2026-01-14
fix(userdb): clear cached user lookups after add_user_group

ppigazzini:


ppigazzini:
@vdbergh I think that this was a leftover.

vdbergh:
> @vdbergh I think that this was leftover.

Yes sorry about this.

ppigazzini:
Thank you for the fast review.
I hit the bug loading a fresh DB, the script adds the approver group to the user00.

ppigazzini:
The script now finishes correctly.

PR #2441, opened 2026-01-15
Do not start worker if determining arch fails

vdbergh:
Currently the arch is set to "unknown" if determining it fails. However after the tightening of the worker api schema in #2439 this is rejected by the server. We could of course fix this server side by relaxing the worker api schema but this seems overkill for something that happens at most very rarely (I have never seen it).

Besides: if determining the arch fail then it can be for two reasons: network failure, or being unable to run the script "get_native_properties.sh" from SF master.
In both cases the worker will not work properly anyway.


vdbergh:
I am going to close this, since I think there may be issues with a fleet starting up. This will cause the script get_native_properties.sh to be hit many times, likely triggering some rate limit at github.

PR #2442, opened 2026-01-15
Accept unknown arch in worker api schema.

vdbergh:
Also delete the (commented out) arch "x86-64-modern" from util.supported_arches since it is not a true arch but an alias.

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -419,7 +419,7 @@ worker_info_schema_api = {
     "compiler": union("clang++", "g++"),
     "unique_key": uuid,
     "modified": bool,
-    "worker_arch": worker_arch,
+    "worker_arch": union(worker_arch, "unknown"),
     "ARCH": str,
     "nps": unumber,
     "near_github_api_limit": bool,
--- a/server/fishtest/util.py
+++ b/server/fishtest/util.py
@@ -659,7 +659,6 @@ supported_arches = [
     "x86-64-avx512",
     "x86-64-avxvnni",
     "x86-64-bmi2",
-    #    "x86-64-modern",
     #    "x86-64-sse3-popcnt",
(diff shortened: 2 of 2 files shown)


ppigazzini (approved):
Looks good patching the worker, PROD update after merging, thank you @vdbergh 

```diff
diff --git a/worker/worker.py b/worker/worker.py
index 207acc80..2f0e008d 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1192,6 +1192,7 @@ def get_worker_arch(worker_dir):
         arch = "unknown"
     finally:
         os.chdir(working_dir)
+    arch="unknown"
     return arch
```


```diff
diff --git a/worker/worker.py b/worker/worker.py
index 207acc80..2f0e008d 100644
--- a/worker/worker.py
+++ b/worker/worker.py
@@ -1192,6 +1192,7 @@ def get_worker_arch(worker_dir):
         arch = "unknown"
     finally:
         os.chdir(working_dir)
+    arch="AAAAAA"
     return arch
```

vondele:
was wondering if an 'unknown' arch worker joining could be logged to events?

ppigazzini:
In the next days I will open a SF PR to merge the improved script that deals with all the possible ARCHs of Makefile.
An all-new CPU arch "blablabla", with gcc/clang support, will build SF as "general-32" or "general-64"
`unknown` will be still helpful to highlight a bug in the script.

vdbergh:
> was wondering if an 'unknown' arch worker joining could be logged to events?

It can be done since there is a `/api/worker_log` end point (which is currently used when book validation fails) which creates a log event on the server.

Up to now I have kept to the convention that log events should only be created for internal errors in fishtest (other strange conditions can be logged to the server log). So only script failures should be logged. Not network errors. Hence this requires a bit of thought.

PR #2451, opened 2026-01-24
Enforce in schema that user group list can contain groups only once

vdbergh:
Requires vtjson >= 2.2.7

There is a unit test `test_add_user_group` that checks that it works.

Also: make `base_nets` and `new_nets` (the net names used by SF) similarly have unique entries.

### DIFF
--- a/server/fishtest/schemas.py
+++ b/server/fishtest/schemas.py
@@ -38,6 +38,7 @@ from vtjson import (
     set_name,
     size,
     union,
+    unique,
 )
 
 run_id = intersect(str, set_name(ObjectId.is_valid, "valid_object_id"))
@@ -108,7 +109,7 @@ user_schema = {
     "pending": bool,
     "blocked": bool,
     "email": email,
-    "groups": [str, ...],
+    "groups": intersect([str, ...], unique),
     "tests_repo": union(github_repo, ""),
     "machine_limit": uint,
 }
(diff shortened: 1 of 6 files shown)


vdbergh:
Is it true that `base_nets` and `new_nets` should also not contain repeated entries?

vondele:
> Is it true that `base_nets` and `new_nets` should also not contain repeated entries?

not guaranteed.

vdbergh:
> > Is it true that `base_nets` and `new_nets` should also not contain repeated entries?
> 
> not guaranteed.

Well what these fields seem to be saying is that "Stockfish uses these nets". More semantics does not seem to be enforced currently (e.g. which is the big net, small net etc...). So if you view it that way, there should not be repeated entries.

ppigazzini:
> Well what these fields seem to be saying is that "Stockfish uses these nets". More semantics does not seem to be enforced currently (e.g. which is the big net, small net etc...). So if you view it that way, there should not be repeated entries.

I must wait to merge this PR?

vondele:
my comment was just that SF should be able to specify twice the same net (e.g. small net and big net could be the same in tests looking at the strength impact), and that should not be blocked at the fishtest level. However, technically, it could be handled in fishtest showing just one net, I don't know the details enough.

vdbergh:
> > Well what these fields seem to be saying is that "Stockfish uses these nets". More semantics does not seem to be enforced currently (e.g. which is the big net, small net etc...). So if you view it that way, there should not be repeated entries.
> 
> I must wait to merge this PR?

Well if you think it is ok to disallow repeated entries in `base_nets` and `new_nets` then I can also enforce it (that would be my preference). 

I think you changed the type of these fields from a string to a list of strings, so perhaps you know?

EDIT: I am going to enforce uniqueness. I am drafting this PR.

vdbergh:
> my comment was just that SF should be able to specify twice the same net (e.g. small net and big net could be the same in tests looking at the strength impact), and that should not be blocked at the fishtest level. However, technically, it could be handled in fishtest showing just one net, I don't know the details enough.

Ok then I will enforce uniqueness for now. If more semantics is needed then we can implement this.

vdbergh:
Undrafted.

PR #2462, opened 2026-02-05
fix(mako): small fixes in templates

ppigazzini:
- Fix incorrect HTML tag nesting and syntax errors (extra closing tags, wrong tag types, extra braces)
- Replace non-ASCII unicode symbols with HTML entities in button labels and output text
- Replace unicode arrows with ASCII arrows in JavaScript comments

### DIFF
--- a/server/fishtest/templates/elo_results.mak
+++ b/server/fishtest/templates/elo_results.mak
@@ -66,7 +66,7 @@
   % endfor
   % if elo_ptnml_run:
     <br>
-    ${f"nElo: {nelo5:.2f} ± {nelo5_delta:.1f} (95%) PairsRatio: {results5_pairs_ratio:.2f}"}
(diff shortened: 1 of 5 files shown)


dav1312:
I'm guessing this was meant to be replaced too?

https://github.com/official-stockfish/fishtest/blob/864cc9023b60436ddadc0c3b3eb75bc1fee88407/server/fishtest/util.py#L301

ppigazzini:
> I'm guessing this was meant to be replaced too?
> 
> https://github.com/official-stockfish/fishtest/blob/864cc9023b60436ddadc0c3b3eb75bc1fee88407/server/fishtest/util.py#L301

This commit served for some parity scripts used during the conversion from Mako to Jinja2.
I suspect that that is dead code in the new framework, after the merge I will triple check for dead code to be cleaned up in all the files that I kept unchanged to be able to rebase upstream during the development (it took 2 months).

Thank you for the reminder :)

PR #2463, opened 2026-02-05
Fix small error in default key for lru_cache decorator

vdbergh:
The default key incorporates keyword arguments (this was a feature request by copilot). However currently it is sensitive to the order of the keywords. This PR fixes this and adds the appropriate unittest.

Also: fix the names of some unittests.

Also: this fix exposed a bug with an unreleased lock in one of the tests. We fix that also.

### DIFF
--- a/server/fishtest/lru_cache.py
+++ b/server/fishtest/lru_cache.py
@@ -221,7 +221,7 @@ class lru_cache:
         expiration=None,
         refresh=None,
         cache=None,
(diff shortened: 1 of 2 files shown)


ppigazzini:
Looks good on DEV, journal clean, dumps of kvstore are consistent after few server/mongo restarts.
I'm waiting copilot before approving and merging.

PR #2469, opened 2026-02-11
Add button link to tracker (https://stockfish.dev/chart/) on the test page

DanSamek:
Makes it easier to access the test tracker directly from the test page.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -278,7 +278,7 @@ legend {
 .mainnavbar a::before,
 #diff-num-comments + a::before,
 #copy-diff::before,
-#statistics-charts a::before {
+#statistical-charts a::before {
   display: inline-block;
   font-style: normal;
   font-variant: normal;
@@ -454,7 +454,7 @@ legend {
   content: "\f328";
 }
 
-#statistics-charts a::before {
+#statistical-charts a::before {
   content: "\f1fe";
 }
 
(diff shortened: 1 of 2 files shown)


ppigazzini:
image
image


I have few doubts:
- I don't know (my ignorance) who maintains the tracker, when it started collecting data (it seems mid-November 2025), and what is the retention policy of the tracked data (keep forever? only last N months?)
- The chart of tracked data is mostly a random walk for a SPRT test, a line for a NumGames test. I have the impression that a button with that label overstates the meaning provided by the chart (we have a page with proper statistics for any test). I wonder if simply adding a link in the Live Elo page would be a better solution, but this add a click of course

I call for opinions on this PR :)

DanSamek:
> I have the impression that a button with that label overstates the meaning provided by the chart 

I agree, maybe better text would be "Patch tracker",...

But I'll wait on another opinions about this PR.

vondele:
I have also a bit my reservations on adding too many external dependencies. However, I think it would be interesting if fishtest started to collect some 'time based metrics' (e.g. number of cores/workers/nps over time), or per test some data.

vdbergh:
The thing is that all information about a test is contained in its current state. The trajectory that lead to this state contains no information.

One says that the pentanomial frequencies are a _sufficient statistic_ for the pentanomial probabilities. 

Formally: the probability of a certain trajectory before time t, conditioned on the state at time t, is independent of the pentanomial probablities.

PR #2478, opened 2026-02-23
fix(templates): harden Jinja templates, JS bindings, and docs

ppigazzini:
Templates
- Fix base template markup (duplicate attributes and extra closing tag)
- Add rel="noopener noreferrer" to every target="_blank" link
- Escape template values before inserting controlled HTML
- Use |tojson for JS interpolation; stringify ObjectId first
- Replace notification <div> with accessible <button>

JavaScript
- Make protectForms() skip forms that already contain csrf_token
- Remove inline onclick/onsubmit; use delegated listeners instead
- Add defensive id parsing; avoid Array.prototype.at for compat
- Use void for fire-and-forget async in delegated handlers

Python (views + template_helpers)
- Stop pre-escaping run_args with html.escape (rely on autoescape)
- Return Markup from results_pre_attrs, nelo_pentanomial_summary, info_html
- Wrap HTML-containing warnings with Markup so they render under autoescape
- Validate CSS color values before emitting them into attributes

Docs
- Annotate Markup return types in the global functions table
- Add authoring rules: escaping boundary, |safe policy, rel="noopener noreferrer", no inline event handlers

### DIFF
--- a/docs/5-templates.md
+++ b/docs/5-templates.md
@@ -36,8 +36,8 @@ and `http/jinja.py`.
 | `display_residual(r)` | `util.py` | Chi-squared residual display |
 | `is_active_sprt_ltc(run)` | `util.py` | Checks if run is active SPRT LTC |
 | `is_elo_pentanomial_run(run)` | `template_helpers.py` | Checks pentanomial display |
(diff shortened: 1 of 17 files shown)


ppigazzini:
edge, chrome, firefox, opera:
image

mobile:
image

PR #2530, opened 2026-04-08
feat(ui): add Open Graph previews for shared actions links

ppigazzini:
Add route-specific Open Graph metadata to the canonical /actions page so shared events-log URLs preview the first visible action instead of the generic site description.

Preserve the current query string in og:url because filters, time cursors, and explicit max_count values define the shared log slice, while keeping htmx fragments body-only.

Add helper and route coverage for the new metadata contract and sync the UI and HTMX reference docs with the full-page ownership model.


ppigazzini:
image

PR #2540, opened 2026-05-03
[bugfix] Correct engine name output in repeated event log warnings

robertnurnberg:
Master leads to these entries in the Event log:
```
26‑05‑03 13:03:11    worker_log    maximmasiutin-8cores-4066bd50-5bf8    history_prune_threshold-997fe99/423    fastchess has so far said 2 times: 'Warning; No bestmove found in the last line from' for None
26‑05‑03 12:49:45    worker_log    maximmasiutin-8cores-4066bd50-5bf8    history_prune_threshold-997fe99/423    fastchess says: 'Warning; No bestmove found in the last line from Base-1a882efc7f because: Element 'bestmove' not found'
```

The reason is that we did not apply the sha shortening that is applied to the whole `line` also to the two engine names.


ppigazzini:
Wow! Copilot speechless! Instamerge!

ppigazzini (approved):
Looks good with linux, wsl2 and msys2 workers.

ppigazzini:
PROD updated, thank you @robertnurnberg

PR #2543, opened 2026-05-06
Exempt official releases from runExceptions and Event log warnings

robertnurnberg:
We discussed on discord that the warnings from official releases during PTs should be avoided. This PR exempts official releases from raising RunExceptions and from posting warning to the Event log. This also means that we can move some warnings to errors, so that bugged test binaries are flagged immediately.

We also update the fastchess version so that some additional output for `bestmove` parsing failures are posted to the Event log. The relevant fastchess PR by @Disservin is https://github.com/Disservin/fastchess/pull/1003.


robertnurnberg:
The SHAs for the (three most recent) official Stockfish releases are hardcoded in `games.py` for now. I have added a `TODO` in case that @vdbergh would like to deal with that server side in some future improvement.

ppigazzini:
Testing on DEV.

ppigazzini (approved):
Looks good on DEV with linux, wsl2 and msys2 workers, PROD update after merging, thank you @robertnurnberg

PR #2545, opened 2026-05-07
Update fastchess

robertnurnberg:
Master may emit spurious bestmove extraction warnings to the Event log when the engine has in fact timed out.

This was today fixed in fastchess, so we update the worker to the latest fastchess version.

Also needs a small edit to one of the warning texts.


ppigazzini (approved):
Looks good on DEV with linux, msys2, wsl2 workers, PROD update after merging, thank you @robertnurnberg (push the red Rob.... push the red)

PR #2549, opened 2026-05-29
Let fish.finish gracefully stop the worker

robertnurnberg:
Ever since #2163 there is no longer a way to truly gracefully stop the worker, as evidenced by the [Event log](https://tests.stockfishchess.org/actions?text=%22fish.exit%22). This also contradicts our [docs](https://github.com/official-stockfish/fishtest/wiki/Running-the-Worker#stop-the-worker).

Recall that before #2163 the presence of `fish.exit` in the worker directory meant the worker would finish the assigned task and then stop. Since #2163 the worker will wait for the next game pair to finish, and then send the stats and pgn of the games finished so far to the server and stop, irrespective of how many games are currently still running on the machine, and how many games are left to complete for the task.

This PR proposes to let the file `fish.finish` play the old role of `fish.exit`. I am happy to update the docs to reflect the new situation if and when this gets merged. Notice that also our [docker](https://github.com/official-stockfish/docker-fishtest/blob/f766dd26caa6f56ade8a5ba669b1a9e2112378b8/worker/updater.sh#L23-L29) appears to rely on the old implementation. (At least docs and code comments suggest that.) I would propose to migrate these to `fish.finish` if this PR gets merged.

With this PR there are at least four ways to stop the worker:
* CTRL-C (immediate stop, kills all running games, does not send stats update to server)
* `kill -15 pid` (as above)
* `touch fish.exit` (near immediate stop, sends update to server, kills nearly all running games, raises exception)
* `touch fish.finish` (graceful stop, may take time to take effect)

### DIFF
--- a/server/fishtest/api.py
+++ b/server/fishtest/api.py
@@ -18,7 +18,7 @@ from fishtest.schemas import api_access_schema, api_schema, gzip_data
 from fishtest.stats.stat_util import SPRT_elo, get_elo
 from fishtest.util import strip_run, worker_name
 
-WORKER_VERSION = 322
+WORKER_VERSION = 323
 
 WORKER_API_PATHS = {
     "/api/request_version",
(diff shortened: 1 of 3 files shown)


vdbergh:
I think that neither `Crtl-C` nor `touch fish.exit` is "unclean". The raised exception is just an artifact of the implementation of the graceful stop  (by contrast `kill -9 pid` would be truly unclean). But perhaps there should not be a `failed-task` event since the stop is intentional.... On the other hand we want communicate somehow - through the event log - that the task was stopped before completion.

robertnurnberg:
Unclean or not graceful, doesn't really matter what we call it. At the moment there's very little difference between ctrl-c and fish.exit. Both kill several running stockfish instances, that users may want to rather let finish.

A prominent example is our docker implementation. Why waste the compute there just to update the compiler or some other package?

vdbergh:
Well calling it unclean (or not graceful) suggests that there is something wrong with it...

I agree that for large core workers there is not much difference between Ctrl-C and `touch fish.exit`. This is due to the fact that for technical reasons `batch_size` cannot scale with the number of cores for SPRT tests.

For example a 256 core worker after Ctrl-C would potentially waste 31completed +256 unfinished games wheres after `touch fish.exit` it would waste 255 unfinished games.

For low core workers the difference is much more significant however. There are still plenty of workers with say <= 4 cores.

vdbergh:
BTW I am of course not opposed to this or an equivalent PR.

robertnurnberg:
Thanks. We had several users ask on discord for such a feature, hence why I opened this PR.

I updated the PR description to remove the wording "unclean". Let me know if the description is fine now.

@ppigazzini @vondele Given that the author of #2163 is not opposed to this PR, any chance to get this merged? I could then upate our wiki, so that wiki and code agree again.

PR #2550, opened 2026-05-30
docs: add mermaid runtime diagrams to core guides

ppigazzini:
Add GitHub-safe Mermaid diagrams to the architecture, threading, worker API, worker, and deployment guides where the docs already describe ordered runtime flows. Refresh stale worker version and fastchess constants so the touched pages match the current implementation.


ppigazzini:
image


image

image

PR #2557, opened 2026-06-18
itp: focus on latency more than llr

dubslow:
The LLR focus was my idea but I've come back around on it. In the NNUE era, search patches are rarer and fishtest is as much about data collection as finding passers (in contrast to e.g. the HCE era). The recent dip in core count also degrades latency.

So focus instead on a consistent latency for all tests, not the luck-based LLR-path-history of each SPRT.


ppigazzini (approved):
Looks good on DEV, waiting the approval of the SF maintainers for the change of the algorithm.

ppigazzini:
PROD updated, thank you @dubslow

Issue #2140, opened 2024-11-09
Type annotations in Fishtest


vdbergh:
To stay up to date I adapted vtjson to work well with type annotations. This would be the new runs schema: 

```python

import copy
import math
from datetime import datetime, timezone
from typing import Annotated, Literal, NotRequired, TypedDict

from bson.objectid import ObjectId

from vtjson import (
    Apply,
    at_most_one_of,
    div,
    fields,
    ge,
    glob,
    gt,
    ifthen,
    intersect,
    ip_address,
    keys,
    lax,
    one_of,
    quote,
    regex,
    skip_first,
    url,
)

username = Annotated[str, regex(r"[!-~][ -~]{0,30}[!-~]", name="username"), skip_first]
net_name = Annotated[str, regex("nn-[a-f0-9]{12}.nnue", name="net_name"), skip_first]
tc = Annotated[
    str, regex(r"([1-9]\d*/)?\d+(\.\d+)?(\+\d+(\.\d+)?)?", name="tc"), skip_first
]
str_int = Annotated[str, regex(r"[1-9]\d*", name="str_int"), skip_first]
sha = Annotated[str, regex(r"[a-f0-9]{40}", name="sha"), skip_first]
country_code = Annotated[str, regex(r"[A-Z][A-Z]", name="country_code"), skip_first]
run_id = Annotated[str, ObjectId.is_valid]
uuid = Annotated[
    str,
    regex(r"[0-9a-zA-Z]{2,}(-[a-f0-9]{4}){3}-[a-f0-9]{12}", name="uuid"),
    skip_first,
]
epd_file = Annotated[str, glob("*.epd", name="epd_file"), skip_first]
pgn_file = Annotated[str, glob("*.pgn", name="pgn_file"), skip_first]
even = Annotated[int, div(2, name="even"), skip_first]
datetime_utc = Annotated[datetime, fields({"tzinfo": timezone.utc})]

uint = Annotated[int, ge(0)]
suint = Annotated[int, gt(0)]
ufloat = Annotated[float, ge(0)]
sufloat = Annotated[float, gt(0)]


class results_type(TypedDict):
    wins: uint
    losses: uint
    draws: uint
    crashes: uint
    time_losses: uint
    pentanomial: Annotated[list[int], [uint, uint, uint, uint, uint], skip_first]


def valid_results(R: results_type) -> bool:
    l, d, w = R["losses"], R["draws"], R["wins"]
    Rp = R["pentanomial"]
    return (
[...]

PR #2562, opened 2026-06-27
chore: type annotations for workerdb module

TumCucTom:
This is a first incremental installment toward #2140 (type annotations in Fishtest), covering the `server/fishtest/workerdb.py` module.

## What changed
- Added complete type hints to the `WorkerDb` class:
  - `__init__` now annotates the `db` parameter as `pymongo.database.Database[dict[str, Any]]` and the `self.db` / `self.workers` attributes (the latter as `pymongo.collection.Collection[dict[str, Any]]`).
  - `get_worker` is annotated `(worker_name: str) -> dict[str, Any]`.
  - `update_worker` is annotated `(worker_name: str, blocked: bool | None = None, message: str | None = None) -> None`.
  - `get_blocked_workers` is annotated `() -> list[dict[str, Any]]`.

## Notes
- Behavior is unchanged; this is annotations only.
- `ty check` passes cleanly on the file, as do `ruff check` and `ruff format --check`.
- Kept the change small and self-contained to one module, matching the existing typing style used elsewhere (e.g. `userdb.py`, modern PEP 604 unions, no `from __future__` needed under the py3.14 target).

Refs #2140

### DIFF
--- a/server/fishtest/workerdb.py
+++ b/server/fishtest/workerdb.py
@@ -1,19 +1,22 @@
 from datetime import UTC, datetime
+from typing import Any
 
+from pymongo.collection import Collection
+from pymongo.database import Database
 from vtjson import validate
 
 from fishtest.schemas import worker_schema
 
 
 class WorkerDb:
-    def __init__(self, db):
-        self.db = db
-        self.workers = self.db["workers"]
+    def __init__(self, db: Database[dict[str, Any]]) -> None:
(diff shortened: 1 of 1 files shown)


TumCucTom:
closing this - #2140 is aiming at proper TypedDict/vtjson schemas, not dict[str, Any] annotations on individual modules, so this isn't really the right approach. happy to help on the actual schema work if that's useful.

Issue #1679, opened 2023-05-09
The pagination widget for finished runs no longer shows the last page


vdbergh:
Note that pagination for the finished runs always worked well (in contrast to the event log).

![Screenshot from 2023-05-09 18-42-38](https://github.com/glinscott/fishtest/assets/6613103/cc117ced-3f7e-454d-9684-ea104a87c0a3)

dav1312:
https://github.com/glinscott/fishtest/pull/1631

vdbergh:
> #1631

Yes but this was meant to be for the event log. Not for the finished runs (for which there was never any problem).

dav1312:
I'm just linking the PR bcs u didn't in the issue description.
Also the PR doesn't seem to indicate that it was meant only for the events log, it just says "pagination" in general.
The last page (currently 5550) of finished tests has a TTFB of 300ms and the same 5550 page in events has a TTFB of 4 seconds, maybe there are some underlying issues in the events.

vdbergh:
Yes we tried several times to fix this with appropriate indexes, but without success.

vondele:
finished runs had the same problem:
```
$ time wget -q -O - https://tests.stockfishchess.org/tests/user/Vizvezdenec?page=817 > /dev/null

real	0m19.507s
user	0m0.009s
sys	0m0.003s

```

vdbergh:
Yes of course for the user pages.

But there was never a problem for the generic page.

vdbergh:
I always found it nice that with a single click one could observe the humble beginnings of Fishtest. It is a bit like a museum.

It would be trivial to fix it using an optional argument that restores the old behavior when that is desired.

ppigazzini:
fixed in https://github.com/official-stockfish/fishtest/pull/2453

<img width="443" height="209" alt="Image" src="https://github.com/user-attachments/assets/2188d970-14d6-4864-9f79-0c1232d6e0e2" />

PR #2565, opened 2026-06-27
Fix Python 2-style except clauses for Python 3 compatibility

TumCucTom:
Several server modules use Python 2-style `except A, B:` clauses, which are a `SyntaxError` under Python 3 and prevent the affected modules from importing. This converts all of them to the parenthesized `except (A, B):` form across the server tree (16 clauses in 11 files).

Notably this unblocks `server/fishtest/views_helpers.py`, whose import failure broke the finished-runs pagination helper referenced in #1679 — the pagination logic itself already renders the last-page link correctly once the module imports.

Relates to #1679.

### DIFF
--- a/server/fishtest/http/boundary.py
+++ b/server/fishtest/http/boundary.py
@@ -120,7 +120,7 @@ async def get_json_body(request: Request) -> JsonBodyResult:
     """Parse JSON body, preserving legacy error behavior."""
     try:
         body = await request.json()
-    except JSONDecodeError, TypeError, ValueError:
+    except (JSONDecodeError, TypeError, ValueError):
         return JsonBodyResult(body=None, error=True)
     return JsonBodyResult(body=body, error=False)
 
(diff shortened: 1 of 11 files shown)


ppigazzini:
Closing this PR because wrong, that is the new python 3.14 syntax from https://peps.python.org/pep-0758/
`ruff` accepts only the new syntax, enable it with `pre-commit install`
Vibe coding should always account that LLMs have an old cutoff date and some suggestions are outdated.

ppigazzini:
> Notably this unblocks server/fishtest/views_helpers.py, whose import failure broke the finished-runs pagination helper referenced in https://github.com/official-stockfish/fishtest/issues/1679 — the pagination logic itself already renders the last-page link correctly once the module imports.

This is hallucinated; we have already fixed the pagination with https://github.com/official-stockfish/fishtest/pull/2453

image

PR #2566, opened 2026-06-28
itp: higher latency bonus, additive llr bonus

dubslow:
Upon further thought, +20% tp for an average-duration test hardly affects the latency at all
    
This time actually flatten latency a little bit, and de-stack the LLR bonus from the latency bonus

(The first commit is completely optional, idc if it goes in or not or squashed or whatever)


ppigazzini (approved):
Looks good on DEV, PROD update after merging, thank you @dubslow

PR #2567, opened 2026-06-28
User test count: implement hard cap

dubslow:
Requires testing.

Implement a hard cap on the number of tests each user may run at a given time. If the hard cap is exceeded, offending tests are paused. This operates only on priority=0 tests. The second commit also makes the hard cap variable depending on the size of the backlog, but I'm not attached to that aspect, a fixed hard cap is also fine.

----

I think the current situation on fishtest is... less than ideal.

Currently there is a soft cap on tests per user, which is at 6, by punishing itp. This has some drawbacks, especially in the current low-cores environment:

1) Users sometimes go over the soft cap for no particular reason, defeating the purpose of having a cap, which leaves the main page littered with active tests that won't progress

2) In fact, since the priority code guarantees one worker per test regardless of itp, depending on the exact core count, it's possible to get more workers than the soft cap would otherwise permit by just submitting 12(ish) tests. In other words, the soft cap can be bypassed.

3) It greatly strengthens the bias between worker size and task distribution. We cannot be rid of this bias of course, but it's a lot stronger than in a high core count environment with a hard cap. 

I also find the soft cap annoying on submitter-experience grounds, but that is very much subjective.

I do not believe the bias is a particular problem, i.e. likely not altering test results. But it's pretty easy to weaken it by either reducing the tests per user and/or reducing the itp punishment (i.e. removing the soft cap). Aside from bias, I think this will also improve the fishtest viewer's experience, having a shorter list of faster moving tests (nevermind the submitter's experience).

This PR implements both of these paths to weakening the bias, namely reducing the cap and changing it from soft to hard. I am however open to alternative improvements, such as:

1) just reducing the existing soft cap

2) having both a hard cap and soft cap

3) any other ideas?

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -8,6 +8,7 @@ import re
 import textwrap
 import threading
 import time
+from collections import defaultdict
 from datetime import UTC, datetime
 
 import regex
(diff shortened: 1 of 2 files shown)


dubslow:
Should we perhaps indicate the current cap on user test pages?

If the hard cap is accepted, should the pauses generate an eventlog entry?

anematode:
would there be interest in just getting rid of the cap altogether? All tests are created equal (except priority>0, which are more equal than others)...

dubslow:
I would be okay with that, and that was certainly the former state of affairs before the soft cap was implemented (I've never liked the itp punishment of the soft cap)

dubslow:
Today marks the second time this week that someone has busted the soft cap. At the very least the cap should be reduced, whether or not the hard cap is accepted or whatever of the various options

vondele:
IMO, it just works as intended in the current implementation.

dubslow:
Amended into a more logical commit sequence (and tweaked suggested variable cap)

dubslow:
Viz on discord preferred to keep the (soft) cap no lower than 6. He suggested soft 6 hard 10

dubslow:
Per some further discussion on Discord, perhaps the key problem with lowering the cap (soft or hard or otherwise) is that many people don't have the ability to unpause a batch of pre-submitted tests, so running them all at once is the only way to maintain thruput for them.

If auto-unpausing of unstarted tests was implemented, that would close the only thing "worse" about improving test serialization. I think that such unpausing code should be fairly short within the existing codebase...

PR #2570, opened 2026-07-07
Two cleanups

dubslow:
This contains two commits with changes originally found/made while working on 2567. I've separated them out here as being nice improvements independent of the substance of opinions within 2567. (And perhaps merging these separately will reduce the diff over there, helping focus discussion...?)

The first commit fixes a minor thing spotted by Copilot, namely that global stats are only used on the homepage. It is a straightforward improvement to the codebase imo.

The second commit changes `pending_hours`, i.e. "Time remaining" on the homepage, to exclude paused tests, since they on average never resume. This commit is optional.

### DIFF
--- a/server/fishtest/rundb.py
+++ b/server/fishtest/rundb.py
@@ -936,6 +936,7 @@ class RunDb:
         return machines
 
     def aggregate_unfinished_runs(self, username=None):
+        # The global stats are computed for the homepage, i.e. when username is Falsy, and are 0 otherwise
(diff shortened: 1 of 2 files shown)


dubslow:
Per discussion on Discord, it might be nice if we had automatic unpausing of unstarted tests in the near future. In such a case, the second commit of this PR should be amended (slightly) so as to count paused and unstarted tests, while excluding started and paused tests.

PR #2572, opened 2026-07-08
fix: make the SPRT Elo confidence bound model-invariant (issue 2571)

ppigazzini:
The live Elo page and the raw statistics page reported different Elo bounds for the same test: one showed a bare 1000, the other a finite value. Both pages call the same estimator, sprt.sprt.analytics, but built the SPRT object in different Elo models, and lower_cb inverted outcome_prob with a search window whose step size was the bound spread in those model units. outcome_prob maps Elo through the logistic score law, so the window unit had no statistical meaning. Past the SPRT range the LLR is clamped and outcome_prob stops being monotone, so whether brentq bracketed a finite root or fell through to the plus/minus 1000 clamp depended on the model-unit window schedule.

Search the confidence bound on the logistic Elo axis over a fixed window, independent of the SPRT bounds, and take the leftmost descending crossing so the meaningful bound is selected and the spurious rising branch past the SPRT range is ignored. A single shared grid scan brackets the point estimate and both confidence bounds, then brentq refines each; targets that are never reached inside the clamp window pin to the clamp and set the clamped flag. Both pages now report the same, well-defined estimate.

The estimator is display and reporting only and is not on the pass/fail path, so test outcomes are unaffected. Well-posed output is preserved to brentq tolerance. Add server/tests/test_stats_sprt.py covering well-posed parity, model invariance, the divergence reproduction, ordering, and clamp behavior.

### DIFF
--- a/server/fishtest/stats/sprt.py
+++ b/server/fishtest/stats/sprt.py
@@ -10,6 +10,15 @@ from fishtest.stats.brownian import Brownian
 
 
 class sprt:
+    # Elo axis clamp for the confidence-bound search. outcome_prob maps Elo
+    # through the logistic score law, so the search runs in logistic Elo over
+    # [-CB_CLAMP, CB_CLAMP] regardless of the elo_model.
+    CB_CLAMP = 1000.0
+    # Grid step (Elo) for the coarse scan. The descending region of outcome_prob
+    # is tens to hundreds of Elo wide for real pentanomial data, so a 10 Elo grid
+    # cannot skip over the leftmost crossing; brentq then refines each bracket.
+    CB_STEP = 10.0
+
     def __init__(self, alpha=0.05, beta=0.05, elo0=0, elo1=5, elo_model="logistic"):
         assert elo_model in ("logistic", "normalized")
         self.elo_model = elo_model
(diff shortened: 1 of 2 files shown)


dubslow:
Two small suggested tweaks:

1) `_cb_clamp` should perhaps be `400.0` instead. I can recall at least one time recently where the excessive size of the bound `1000` was commented on. Logistic elo loses some meaning by +200, and is practically meaningless beyond +400. Best reflect that in the clamp.

2) No reason for the test suite to double-define the clamp value, it should just use the existing constant:

```
from fishtest.stats.sprt import sprt

CLAMP = sprt._cb_clamp
```

vdbergh:
I am still studying this issue. The observation by the ai that the clamp `+-1000` is model dependent is correct. This is a genuine bug, but I am surprised that this would be a problem in the example in the issue (the right confidence bound is much less than `1000` even in nelo). 

I also think that the claim of the ai that `outcome_prob` is not monotonous, is not correct.

I need to write a little program for the example. I have finally upgraded fishtest. 

It is interesting that the comment in the code says:
```
       # Various error conditions must be handled better here!                                                                                                                         
 ```

**Comment by ppigazzini** [comment]
Thank you for the review, take always with a grain of salt anything written by an ai agent: logic and math sometimes are hallucinated. Feel free to take everything decent from this PR and open a better one.

File: server/fishtest/stats/sprt.py [hunk 1]
```diff
@@ -80,45 +89,80 @@ def outcome_prob(self, elo):
...
+
+        The root is sought on the logistic Elo axis that outcome_prob uses
+        (s = L_(elo)), over the fixed window [-CB_CLAMP, CB_CLAMP]. The search
+        does not depend on the elo_model or on the units of the SPRT bounds, so
+        every caller and every elo_model agrees on the result.
+
+        outcome_prob is monotone decreasing in Elo only while the small-Elo GSPRT
+        approximation holds. Once the observed Elo lies far outside [elo0, elo1]
+        the approximation breaks down: outcome_prob dips to a minimum and rises
+        again toward the clamp. The confidence bound is the leftmost (descending)
```
vdbergh, line 106: The AI appears to be contradicting itself here. The leftmost descending crossing is not the maximal Elo such that " For each probability p in `ps`, return the maximal Elo such that the observed outcome of the test has probability less than p, i.e. the Elo where outcome_prob(elo) == 1 - p" (see first paragraph above).

It is probably true that one should only look at the descending crossings, but the motivation for this is unclear.

ppigazzini:
superseded by #2581

PR #2574, opened 2026-07-16
Make highlighted active contributor row permanently green

FauziAkram:
Make highlighted active contributor row permanently green

Currently, when using the "Jump to my rank" feature on the Contributors page, the row highlights green for 3 seconds before fading out to standard grey. The resulting grey background is very subtle and can be difficult to distinguish from standard alternating table row colors.

This PR changes the behavior to keep the active contributor's row highlighted green permanently.
The fade-out animations and associated keyframes have been removed to keep the styling simple and consistent.

### DIFF
--- a/server/fishtest/static/css/application.css
+++ b/server/fishtest/static/css/application.css
@@ -815,17 +815,9 @@ td {
   opacity: 1;
 }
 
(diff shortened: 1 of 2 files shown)


FauziAkram:
Current:
image

Proposed:
image

PR #2575, opened 2026-07-16
Profile page rework

Disservin:
extracted some profile page improvements from https://github.com/official-stockfish/fishtest/pull/2421

Updated the profile page to make it a bit more UX friendly.. also previously the github token was sent to the server which is wrong since we store it in the users localstorage..

image


Disservin:
@ppigazzini friendly ping

PR #2576, opened 2026-07-20
Fix for issue 2571

vdbergh:
The algorithms in LLRcalc.py and sprt.py for computing a median unbiased Elo estimate from SPRT tests, as well as the corresponding confidence intervals, were never intended to work correctly for large Elo values. So issue 2571 is mainly cosmetic. Nonetheless one would prefer that the algorithms return something reasonable, even for large Elo values.
    
The root cause for issue 2571 is that outcome_prob (the probability of observing a result worse than the actual one,
for a given drift) is not monotonous with the current code. This was correctly observed by the AI although the given reason was wrong.
    
The actual reason is that in the code for outcome_prob the infinitesimal variance of the Brownian approximation is estimated conditionally on the given value of the drift. While this is in principle correct, the resulting distribution cannot be used naively and in fact I don't know how to do it (we are not in the setting of Wilks' theorem for example). So in this PR we estimate the infinitesimal variance unconditionally and hope for the best. For small ELO values this makes no difference as the drift and the variance are orthogonal and it seems that with this change outcome_prob becomes monotonous in all cases.
    
The AI also observes that the clamp +- 1000 depends on the elo model (normalized or logistic). This is correct but harmless as long as we do not hit the clamp. I will fix this is in a separate PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,12 +9,12 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.22
+    rev: v0.16.1
     hooks:
       - id: ruff-check
         args: [--fix, --exit-zero]
       - id: ruff-format
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.11.31
+    rev: 0.12.1
     hooks:
       - id: uv-lock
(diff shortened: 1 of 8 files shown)


vdbergh:
Made this draft temporarily. I am going to reflect a bit more about why exactly outcome_prob is non-monotonous.

vdbergh:
Still working on this. There seem to be two issues.

(1) My method for obtaining an Elo confidence interval is asymptotically incorrect (it is of course ok for small Elo differences). This was pointed out to me by Google AI who also explained the correct method. I still need to implement this.

(2) There is another technical issue. Two methods for determining drift and the variance of the likelihood ratio (in order to apply the Brownian approximation) give different results. I still need to understand this.

Note that none of this has any effect on the correct functioning of Fishtest. The Elo confidence intervals in the context of Fishtest were extensively validated by simulation.

PR #2577, opened 2026-07-22
chore: deps bump July 2026

ppigazzini:
**chore(deps): bump packages, pre-commit hooks, and CI actions**
Raise server floors: awscli 1.45.53, fastapi 0.139.2, numpy 2.5.1,
python-multipart 0.0.32, regex 2026.7.19, scipy 1.18.0, uvicorn 0.51.0,
httpx2 2.7.0, hatchling 1.31.0. Raise dev tools: pre-commit 4.6.1,
ruff 0.15.22, ty 0.0.62. Regenerate the root, server, and worker
lockfiles.

Keep the worker pins requests 2.32.4 and hatchling 1.27.0, the last
releases supporting Python 3.8.

Update pre-commit hooks (ruff v0.15.22, uv 0.11.31) and workflows:
checkout v7, setup-python v7, setup-node v7, ruff-action v4.1.0,
setup-uv v9.0.0. Move Node.js from the end-of-life 25.x line to 26.x.

**chore(worker): re-vendor certifi, charset_normalizer, and idna**
Update the vendored packages to their newest releases still compatible
with Python 3.8: certifi 2026.7.22, charset_normalizer 3.4.9, and
idna 3.15 (idna 3.16 requires Python >= 3.9). Keep requests 2.32.4 and
urllib3 2.2.3, the last releases supporting Python 3.8.

Vendor charset_normalizer from the pure py3-none-any wheel and drop the
stale cp38 mypyc speedups: upstream publishes binary wheels only for
cp39+ so the pure wheel is what pip serves a Python 3.8 client. The
certifi wheel now ships a small certifi/tests subpackage, kept to
mirror the upstream wheel content.

**fix: adapt to ty 0.0.62 and ruff-pre-commit hook rename**
Use a falsiness check in the required-options loop of views_run.py.
ty 0.0.62 rejects len() there because the flow-insensitive value union
of the form dict includes int, list, and dict keys that are assigned
only after the loop; every value present at the check is a string, so
the falsiness test is equivalent and type-safe.

Rename the deprecated ruff hook id to ruff-check; the old id is
reported as a legacy alias by current ruff-pre-commit.

**chore(web): bump Font Awesome to 7.3.1**
Update the jsDelivr URL and the SHA-256 SRI hash in the base template.
The hash is verified against the fetched asset and the jsDelivr package
metadata. The other CDN assets (bootstrap 5.3.8, htmx 2.0.10,
highlight.js 11.11.1, diff 9.0.0) are already at their latest
releases.


ppigazzini:
Looks good on DEV with linux, wsl2, msys2 workers and python 3.8.20 - 3.14.x

ppigazzini:
worker version bump 7b62ef48a1914fd203ac8c827b3fd70cd894dc04 to push the refreshed vendored packages to the workers

PR #2578, opened 2026-08-01
Add % c column to SPSA parameter table

FauziAkram:
This PR adds a % c column to the SPSA parameter table in the test details view (/tests/view/<id>).

Currently, the relative movement of an SPSA parameter (% c) is visualized in the SPSA kernel smoother chart, but it is not directly visible in the parameter table above it. Adding this column allows developers to quickly inspect the relative parameter shift alongside its raw value and initial start value without needing to navigate the graph.


FauziAkram:
@ppigazzini Hi, can you please merge when possible? I would really love to see the % c column in the table in my current tuning run, and would help me when I copy the whole table into excel.
https://tests.stockfishchess.org/tests/view/6aaea105dbd128aac8fd5854

PR #2588, opened 2026-09-05
Update Progression tests base to Stockfish 19.

vondele:


ppigazzini (approved):
Looks good on DEV with linux, msys2, wsl2 workers, PROD update after merging, thank you @vondele

PR #2590, opened 2026-09-07
update fastchess

robertnurnberg:
Includes the recent fix https://github.com/Disservin/fastchess/pull/1066. Without it, fastchess reported spurious "`Illegal move <none>`" warnings for development versions of stockfish.

Other changes in the new fastchess version:
* faster compilation
* better exception messages for wrong option values


ppigazzini (approved):
Looks good on DEV with workers (linux, wsl2, msys2), PROD update after merging, thank you @robertnurnberg

PR #2599, opened 2026-09-22
feat(ui): switch the client bundle from htmx 2 to htmx 4

ppigazzini:


ppigazzini:
Improved and tested for few months with htmx4 beta4-beta6, htmx4 final version ran fine on PROD.

